PR #11912, opened 2020-12-07
Update Addon API min ver

### LINKED ISSUE #12426: many addons have been disabled after updating to snapshot Version: alpha-22782,7e6561a0 U
### Steps to reproduce:
  After updating
  ### Actual behavior:
  addon(s) do not work as well as their settings have been omitted from NVDA preferences.
  ### Expected behavior:
  I can use addon like instantTranslate-4.4.1 or quickDictionary-2.1.2-dev
  ### System configuration
  NVDA installed
  #### NVDA version:
  alpha-22782,7e6561a0 U
  #### Windows version:
  Windows 10 64-bit the last version. (up to date)
  
  Laptop 8GB of RAM, processor: coreI7 64-bit
  ### Other questions
  yes it does.
  earlier snapshot versions before around May 14th 2021 worked well.
  N/A
  N/A

feerrenrut:
2021.1 will have several API breaking changes.
Addons need to be re-tested and have their manifest updated.

This does not need to be merged until 2021.1 branches off for beta. Doing so earlier would cause add-ons to be marked as incompatible for alpha users. However by the time the beta comes, addon authors should have been able to test and provide pre-release versions of their addons.


seanbudd (dismissed):

leonardder:
Given there are more and more add-ons breaking, I'd like to propose to merge this ASAP. See also #12426

source/addonAPIVersion.py
```diff
@@ -13,7 +13,7 @@
 """
 
 CURRENT = (buildVersion.version_year, buildVersion.version_major, buildVersion.version_minor)
-BACK_COMPAT_TO = (2019, 3, 0)
+BACK_COMPAT_TO = (2021, 1, 0)
 """
 As BACK_COMPAT_TO is incremented, the changed / removed parts / or reasoning should be added below.
 EG: (x, y, z): Large changes to speech.py
```
leonardder, line 19: I think there should be some reasoning below this line about the version updating.

feerrenrut:
We intentionally haven't merged this because:
- Merging this signifies the commitment to the API. 
- Once we merge this, all addon's must update their manifest before they can even test for compatibility. Delaying this process allows authors and adventurous addon users to test and report problems with the upcoming API without a new addon release being made.
- The downside is that users are not warned that add-ons may not work. However remember, this is alpha software, there will be rough edges.

source/addonAPIVersion.py
feerrenrut: Hmm, yes you are right. Though given the size of the list, it might be better to refer to the changelog.

josephsl:
Hi, I think a representative sample of about two or three major things might be useful to add – I can think of wxPython 4.1.1, say all handler refactor, to name a few. Thanks.

LeonarddeR (approved):
Looks good to me now.

leonardder:
> We intentionally haven't merged this because:

Makes sense. Having said that, i think this might yet be the time to go with this change:

1. Alpha started with the 2021.2 release cycle, so I assume the API of 2021.1 can be considered stable?
2. It will urge add-on authors once more to fix their add-ons.

feerrenrut:
I believe we are very close to doing this.

>Alpha started with the 2021.2 release cycle, so I assume the API of 2021.1 can be considered stable?

Note quite, we allowed 2021.2 to start so that changes intended specifically for that release could be merged without effecting 2021.1. 

> It will urge add-on authors once more to fix their add-ons.

Yes, that would be good.

PR #10973, opened 2020-04-09
Add a command to check formatting for the text under the caret.

lukaszgo1:
### Link to issue number:
Closes #9505 
First step towards #7314
### Summary of the issue:
It is not possible to check formatting for the text under the caret without moving review cursor to its position. This is problematic for people working with review cursor not following caret. In addition this can be done with a braille display, so it should also be possible for speech only users.
### Description of how this pull request fixes the issue:
I've created a separate script assigned to NVDA+f to check formatting at the caret position. The current command for checking it at the review cursor has been reassigned to NVDA+shift+f because for most users checking formatting happens  more often at the caret not at the review cursor. As by default review follows caret there is no functional difference for people who haven't changed this.
While at it I've also split up these scripts into a separate ones as per #7314 - they are obviously not assigned.
### Testing performed:
1. Tested that formatting can still be checked and displayed in browse mode with NVDA+shift+f at the review position,
2. Tested that with the new script formatting at caret is properly announced/displayed in browse mode.
3.  Assigned a keyboard command to each of the separate scripts and confirmed that they are functioning as expected.
4. Compiled user guide - made sure that the changes to the keyboard commands are displayed as they should be.
### Known issues with pull request:
- I haven't documented the new separate scripts - I don't think this should be done at this early stage.  When #7314 would be fully implemented we should add a general note to one of the UG sections though.
### Change log entry:

New features:
It is now possible to check formatting at the caret with NVDA+f
Changes:
Command used to check formatting at review position has been reassigned to NVDA+shift+f

Issue #6479, opened 2016-10-18
No formatting information in edit mode in Excel


Qchristensen:
In any cell, with or without text in it, press F2 to go into edit mode, then press NVDA+f to read formatting information.  NVDA reads "No formatting information" instead of advising the current font, etc.

Tested in Excel 2016.

In Excel 2007, if the cell is blank, NVDA reads "No formatting information".
If the cell has text, and the focus is at the end of the text, NVDA reads "space" when NVDA+f is pressed.
If the cell has text, and the focus is anywhere else in the text, NVDA reads "space" and then the formatting information.

Possibly connected to #6478 ?

feerrenrut:
I tested this in Excel 2013 and got the same behaviour.

feerrenrut:
We are unable to get access to the formatting information while in edit mode as it currently stands, we will need an update to excel for this.

feerrenrut:
P3 since its perhaps less common to individually format words within cells than format the whole cell.

Adriani90:
I agree here. Formating single parts of content in a cell is not common practice.

Qchristensen:
Trying this now with NVDA 2020.3 and Office 365 (64-bit) Version: 16.0.13426.20270, I find:

- On a cell (not in edit mode), NVDA reads: "No caret".
- Press F2 to go into edit mode (the focus is after the last character in the cell).  NVDA reads "No formatting information".
- In edit mode, move to within the text in the cell: NVDA reads the correct formatting information.

The comment above that it is less common to format individual words within a cell is, I believe, still correct, but that only makes the lack of formatting information when not in edit mode more pronounced.

Meanwhile, the work around is, go into edit mode, and left arrow once into the text.

Adriani90:
Did you try with nvda+shift+f?

cc: @lukaszgo1, @CyrilleB79

CyrilleB79:
@Qchristensen wrote:
> * In edit mode, move to within the text in the cell: NVDA reads the correct formatting information.

On my side, it is somewhat different (Excel 2016, NVDA 2020.3).
In a cell in edit mode and after having moved the caret in the middle of the text, NVDA reports "baseline", which is correct, but not bold and italic, either with NVDA+F or with NVDA+shift+F.

If the cell is not in edit mode, NVDA+F indicates only "No caret", but NVDA+shift+F indicates the correct formatting information for the whole cell, assuming it is uniform.





[...]

PR #11914, opened 2020-12-07
Fix-up of #10973: Fallback to POSITION_FIRST when reporting formatting info at caret to report in MS Excel

lukaszgo1:
### Link to issue number:
Discussion in #6479

### Summary of the issue:
It is currently impossible to request formatting info for an Excel cell with NVDA+F.

NVDA+F now falls back to POSITION_FIRST if there is no caret.
### Testing performed:
- Ensured that NVDA+F reports formatting info for an Excel cell.
- Ensured that 'No formatting information' is reported for cases in which there is none such as a 'browse' button in the run dialog.
None known
### Change log entry:

Section: Bug fixes
It is now possible to report formatting information for the focused Excel cell using NVDA+F.

### DIFF
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -1552,7 +1552,7 @@ class GlobalCommands(ScriptableObject):
 		category=SCRCAT_SYSTEMCARET,
 	)
 	def script_reportFormattingAtCaret(self, gesture):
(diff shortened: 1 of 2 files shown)


michaelDCurran (approved):

feerrenrut:
Is it important that this is in 2020.4? As it currently stands it will not be, and the changelog needs adjustment.

Issue #11861, opened 2020-11-24
Update OCR documentation


Qchristensen:
### Is your feature request related to a problem? Please describe.

The information on OCR is broken up into three sections in the user guide:

Section 10. Content Recognition, which describes what OCR is, explains how it follows the navigator object, and describes the virtual window the results are presented in, and how to navigate them and close the window

Section 10.1. Windows 10 OCR immediately follows, which discusses the languages and points to the User Guide section to set  the OCR language, and then provides instructions for starting OCR (NVDA+r).

Section 12.1.16. Windows 10 OCR Settings explains the settings in the OCR section of the NVDA settings dialog.

The two "parts" of section 10 are disjointed, with the specific instructions being separated (how to navigate and close the results comes before how to actually start OCR for instance).  Additionally, users are often confused about how much information NVDA's OCR feature provides.  A common scenario is to open a PDF file, run OCR on it, and wonder why only a small amount of text is recognised.

### Describe the solution you'd like

I suggest for the information we have in section 10, it could be done all in one block.  If we want to have that Section 10.1 to discuss our solution (perhaps with the idea of multiple possible solutions or discussion of specific add-ons at some point in the future?) then I suggest having only the following text prior to subsection 10.1:

"When authors don't provide sufficient information for a screen reader user to determine the content of something, various tools can be used to attempt to recognize the content from an image."

I would then take this paragraph and move it to the end of the section (after NVDA+r is mentioned):

"Once recognition is complete, the result will be presented in a document similar to browse mode, allowing you to read the information with cursor keys, etc. Pressing enter or space will activate (normally click) the text at the cursor if possible. Pressing escape dismisses the recognition result."

I would also move the discussion about languages above the paragraph about recognition being from the current navigator object.  That would bring all discussion about "how to" together, and immediately before the specific instructions of what to press.

Finally, I would add the following paragraph after the current navigator information and before the "press NVDA+r" instructions:

OCR is only performed on text which is visible on screen.  Text which is too small to be read is also not able to be processed.  Where possible, setting the view to show at "100%" or to show "one page", or "one page width" should provide good results.

My proposed full reworded section 10 below (note I just copied the text, so links in the original text are lost):

`10. Content Recognition

When authors don't provide sufficient information for a screen reader user to determine the content of something, various tools can be used to attempt to recognize the content from an image.

10.1. Windows 10 OCR

[...]

PR #11921, opened 2020-12-09
Prevent running Windows 10 OCR if screen curtain is enabled.

### LINKED ISSUE #11911: Warning when using OCR when screen curtain is enabled
For a long time I thought that the OCR function (Windows 10 OCR) just didn't work properly, because I got no results every time. Today I found out that I had turned on the screen curtain feature of NVDA, so the OCR only saw a black screen. After I turned off the screen curtain, the OCR worked fine.
If that's technically possible, I'd like it when NVDA warns you when you try to use the OCR feature while Screen Curtain is on.
I don't know any alternatives, this seems to me the easiest solution.
### Additional context
I am using Windows 10 1909 (64-bit) build 18363.1198 and NVDA 2020.3.

CyrilleB79:
### Link to issue number:

Fixes #11911

### Summary of the issue:

User are not informed that OCR cannot be achieved when the screen curtain is enabled. They may then think that Windows 10 OCR is a non-working / broken feature.

* Do not perform OCR if screen curtain is active. Instead, issue a message to inform the user that he needs to disable the screen curtain to perform OCR.
* Add some information in the user guide (screen curtain section and OCR section) to document this incompatibility.
* Also document some other related limitations:
  * screenshots cannot be taken if OCR is active
  * other visual aids may cause incorrect results in recognition. I did not mention explicitly them.

For reference, even if I did not include these details in the guide, here is the status of incompatibility between OCR and visual aids:
* screen curtain: fully incompatible
* Windows Magnifier: partially incompatible: the OCRed object need to be fully visible on the screen to have a correct result
* Visual Highlight (ex: Focus Highlight): incompatible in some rare cases, e.g. when the browse mode cursor highlighter ioverlaps the OCRed zone
* Zoomtext remains fully compatible with OCR even if the OCRed zone is not in the zoomed view

### Testing performed:

* Tested that OCR command (NVDA+R) still works if screen curtain is disabled
* Tested that no OCR is performed and that the warning message is heard when trying to OCR when screen curtain is enabled. The 3 way to enable screen curtain have been tested:
* temporary activation via shortcut
* full activation via shortcut
* activation from the vision setting window (before pressing OK or cancel)

This PR is limited to warn and document the incompatibility between OCR and screen curtain. It does not offer a solution.
The goal is to have this PR merged quite quickly so that people using screen curtain do not think erronously that OCR is just a broken feature.

### Change log entry:

Section: Bug fixes
`NVDA now prevents to perform Windows 10 OCR if screen curtain is enabled. (#11911)`


CyrilleB79:
This PR is ready to review.

Since the original issue may let some think that OCR feature does not work and since the modification is quite small, I am wondering if this could still be integrated in 2020.4.

I know there is another issue regarding the doc organization concerning the OCR (#11861. However, I expect the current PR to be integrated quite quickly if possible (2020.4 or 2021.1); on the contrary, I expect #11861 to generate more comments and discussion regarding the doc organization, not only regarding OCR.

Cc @Qchristensen for the doc review, @feerrenrut for the rest and the question about integration in 2020.4.

michaelDCurran (approved):

PR #11924, opened 2020-12-10
Making KeyboardInputGesture.fromName to use en_us locale by default.

mltony:
### Link to issue number:
N/A
### Summary of the issue:
KeyboardInputGesture.fromName() function by default tries to resolve key names in current system locale. A user of my BrowserNav add-ons complained to me that the add-on wouldn't start. After investigating I found that statements like:
```
KeyboardInputGesture.fromName("Control+C")
```
wouldn't work when the default locale on computer is set to non-latin, e.g. Russian in this case. This happens, because the letter C cannot be resolved in Russian locale. This change doesn't seem to affect NVDA core at all, since it doesn't seem to be resolving keystrokes including any Latin letters, however it affects my BrowserNav add-on and possibly others.
I propose to resolve keyboard shortcuts in English locale, en_us for example. I think it is unlikely someone will ever need to create keyboard shortcuts that include non-Latin letters.
### Testing performed:
Tested using NVDA Python console that shortcuts can now be successfully created regardless of current system locale.
N/A
### Change log entry:
Making KeyboardInputGesture.fromName to use en_us locale by default.
Section: Bug fixes

### DIFF
--- a/source/keyboardHandler.py
+++ b/source/keyboardHandler.py
@@ -548,6 +548,7 @@ class KeyboardInputGesture(inputCore.InputGesture):
 				time.sleep(0.01)
 				wx.Yield()
 
+	en_us_input_Hkl = 1033 + (1033 << 16)
 	@classmethod
 	def fromName(cls, name):
 		"""Create an instance given a key name.
@@ -569,7 +570,7 @@ class KeyboardInputGesture(inputCore.InputGesture):
 				vk, ext = getNVDAModifierKeys()[0]
 			elif len(keyName) == 1:
 				ext = False
-				requiredMods, vk = winUser.VkKeyScanEx(keyName, getInputHkl())
+				requiredMods, vk = winUser.VkKeyScanEx(keyName, en_us_input_Hkl)
 				if requiredMods & 1:
 					keys.append((winUser.VK_SHIFT, False))
 				if requiredMods & 2:
--- a/source/keyboardHandler.py
+++ b/source/keyboardHandler.py
@@ -570,7 +570,7 @@ class KeyboardInputGesture(inputCore.InputGesture):
 				vk, ext = getNVDAModifierKeys()[0]
 			elif len(keyName) == 1:
 				ext = False
-				requiredMods, vk = winUser.VkKeyScanEx(keyName, en_us_input_Hkl)
+				requiredMods, vk = winUser.VkKeyScanEx(keyName, cls.en_us_input_Hkl)
 				if requiredMods & 1:
 					keys.append((winUser.VK_SHIFT, False))
 				if requiredMods & 2:


JulienCochuyt:
@mltony: For the record, would you please have a log of the error this PR aims at preventing?

CyrilleB79:
> I propose to resolve keyboard shortcuts in English locale, en_us for example. I think it is unlikely someone will ever need to create keyboard shortcuts that include non-Latin letters.

Yes it may happen, with non-latin letters or with symbols that are not at the same position on non-US keyboard, e.g. French one.
As an example, I am using the dynamic keystrokes feature of your own add-on Tony's enhancements. If I define control+$ in the dynamic keystrokes table, it is failing to report the current line with the test build of this PR since the dollar key is not located at the same place on the French keyboard and the US keyboard (and also because issuing $ on US keyboard needs additional shift modifier on contrary to French keyboard).

Thus, in order not to break compatibility with existing add-ons, I suggest to add an additional optional parameter to the `fromName`method.
`def fromName(cls, name, hkl=None):`
Then if hkl is is not None, i.e. is passed as parameter, use it to call `winUser.VkKeyScanEx`, else use `getInputHkl()`.

JulienCochuyt:
To complement on @CyrilleB79's https://github.com/nvaccess/nvda/pull/11924#issuecomment-742380355: This is the reason why NVDA ships `locale/??/gestures.ini`

mltony:
Here is the traceback reported to me by a user:
```
  File "C:\Users\User\AppData\Roaming\nvda\addons\browsernav\globalPlugins\browserNav.py", line 193, in <module>
    kbdControlC = keyboardHandler.KeyboardInputGesture.fromName("Control+c")
  File "keyboardHandler.pyc", line 572, in fromName
  File "winUser.pyc", line 608, in VkKeyScanEx
LookupError
```
My reasoning as to why en_us should be the default, or any other English locale, is that by keeping things as they are, this is a time bomb ticking - it is only a matter of time when someone else would need to use Control+C or another shortcut involving a letter, and then it'll fail on the computers with non-English locale. I admit this is probably not a serious time bomb by any mean, but it's a time bomb nevertheless. And in the future it is likely someone would have to spend time investigating this issue again. That's why I proposed to fix it now.
@CyrilleB79, I understand this change would break your current config. But would it be possible to rename gestures for the  dynamic keystrokes to use English locale?

CyrilleB79:
> My reasoning as to why en_us should be the default, or any other English locale, is that by keeping things as they are, this is a time bomb ticking - it is only a matter of time when someone else would need to use Control+C or another shortcut involving a letter, and then it'll fail on the computers with non-English locale. 

In BrowserNav, It should not fail if you follow the proposal in my previous comment to add an optional parameter to `fromName` method, provided you use:
```
from keyboardHandler import KeyboardInputGesture
en_us_input_Hkl = 1033 + (1033 << 16
KeyboardInputGesture.fromName('kb:control+C', hkl=en_us_input_Hkl)
```
On the contrary, in Tony's enhancement, you should keep this line unchanged:
```
kb = keyboardHandler.KeyboardInputGesture.fromName(tokens[1]).identifiers[0]
```

> @CyrilleB79, I understand this change would break your current config. But would it be possible to rename gestures for the dynamic keystrokes to use English locale?

If I understand well, it will not only break my current config. It will also force me (or any user of Tony's enhancements add-on) to investigate to know which key on the en_US keyboard corresponds to the $ key on the French keyboard. And reading the dynamic key table won't be explicit at all.
Let me know how I should proceed if I am mistaken.

mltony:
@CyrilleB79, I think there are two separate questions here:
1. Which way should be the default in NVDA? I would argue that English should be default, because otherwise, keystrokes like Control+C should not be allowed to be created, since they will be impossible to create depending on the system locale. If someone argues this is desired behavior - then it should be documented.
2. How should I parse dynamic keystrokes in Tony's enhancements. And this one is even more tricky. I understand your point. However, if there are any Russian users of Tony's enhancements add-on, they probably won't be able to specify Control+C shortcut at all, since C is not a letter of Russian alphabet. They might be able to specify Russian letter that is on the same key as letter C in English layout, but I'm not sure. I can only say for Russian users that they would prefer having keystrokes parsed in English layout. So no matter what either French or Russian users would suffer. I hate to put it this way but I don't see a good way out, I mean I guess it is possible to create a checkbox defining in which locale to parse all the keystrokes.

CyrilleB79:
> 1. Which way should be the default in NVDA? I would argue that English should be default, because otherwise, keystrokes like Control+C should not be allowed to be created, since they will be impossible to create depending on the system locale. 

@mltony did you read my proposal in previous comments. I actually give an example on how control+C could be issued with whatever type of keyboard.

> 2. How should I parse dynamic keystrokes in Tony's enhancements. And this one is even more tricky. I understand your point. However, if there are any Russian users of Tony's enhancements add-on, they probably won't be able to specify Control+C shortcut at all, since C is not a letter of Russian alphabet. They might be able to specify Russian letter that is on the same key as letter C in English layout, but I'm not sure. I can only say for Russian users that they would prefer having keystrokes parsed in English layout. So no matter what either French or Russian users would suffer. I hate to put it this way but I don't see a good way out, I mean I guess it is possible to create a checkbox defining in which locale to parse all the keystrokes.

I am not sure of the solution to use in Tony's enhancement among these three:
* use English layout
* use local layout
* use one or the other according to how the key combination is made, i.e. contains ascii letter as main key, contain non-ascii letter as main key, contain non-letter as main key (e.g. punctuation), contains modifiers or not.
An input from a Russian user would help. Maybe a Russian user is just using English keyboard layout for coding (which is a main use case of dynamic keystrokes) since all the keywords of a coding language are written with the latin alphabet...
Anyway, all these questions are implementation details of this add-on and it will finally be your choice.

But the most important point to not in my proposal regarding this PR's code is that the developer using `fromName` can choose to use whatever keyboard layout he wants (local, English or other), whereas your proposal forces the developer to use English layout, thus creating limitations.
A second point to note in my proposal is that the behaviour of this method would not change and thus not break existing code.

I hope all is clear now.

mltony:
Closing as this PR doesn't make sense anymore. I had a mistaken belief that resolving keystrokes in English locale would work for all users.

PR #11936, opened 2020-12-13
Python Console: Fix handling of the tab key in the input pane (#11532)

### LINKED ISSUE #11532: Tab behaves strangely when not being used for completion in Python Console
I am not sure what the tab is meant to do in the Python Console.  However, since I am used to using it for indentation, I do, and it normally works fine for this purpose.
However, I recently had occasion to need to indent some code I had written further than it previously was.
I can just use space, but by default, I am used to using tab.

### Steps to reproduce:

1.  Press keys: NVDA+n, t, p, to open the python console.
2.  Type some code such as this:
    ```python
    print("Some test code.")
    ```
3.  That code runs fine. But now, let's say you need to put that code in a function. Obviously this example is easy to just retype, but imagine if it was 10 lines of very complex code, and you now wanted to encapsulate it in a function. First, enter:
    ```python
    def testFunc(parameter):
    ```
4.  So far, so good. But now, up arrow and try to include your code from before. You can't just accept the print statement as entered, because it isn't indented. You can press home to get to the start of the line, and then press tab to indent it.

### Actual behavior:

When you press tab, the cursor jumps back to the end of the line, and **then** the tab is inserted.
So the tab appears at the end of the line, instead of at the beginning where it was expected.

### Expected behavior:

If a tab is going to show up anywhere in this situation, I would expect it to be at the beginning of the line (I.E. where the cursor is.).
A tab on a blank line, produces a tab character at the start of that line. I would expect a tab at the start of any line to do the same, since there is no completion context for tab to operate on.

### System configuration

NVDA portable alpha-20793,9a4074bc.
Windows 10 2004.

JulienCochuyt:
### Link to issue number:

Fixes #11532

### Summary of the issue:

Currently, when editing a non-empty input line, NVDA's Python Console does not support inserting a tab at the beginning of the line for indentation.
It does not support either performing tab-completion anywhere but at the end of an input line: This is fine when typing a new line, but limiting when editing an existing one.

### Change log entry:

Section: Bug fixes

In the Python Console, inserting a tab for indentation at the beginning of a non-empty input line and performing tab-completion in the middle of an input line are now supported.

### DIFF
--- a/source/pythonConsole.py
+++ b/source/pythonConsole.py
@@ -308,7 +308,7 @@ class ConsoleUI(
 	RE_COMPLETE_UNIT = re.compile(r"[\w.]*$")
 
 	def complete(self):
-		textBeforeCursor = self.inputCtrl.GetRange(0, self.inputCtrl.GetInsertionPoint())
+		textBeforeCursor = self.inputCtrl.GetRange(0, self.inputCtrl.GetSelection()[0])
 		try:
 			original = self.RE_COMPLETE_UNIT.search(textBeforeCursor).group(0)
 		except AttributeError:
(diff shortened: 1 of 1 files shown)


JulienCochuyt:
@XLTechie, could you please check if the try-build of this PR meets your expectations?

michaelDCurran (approved):

PR #11943, opened 2020-12-16
Update CLDR to version 38.1

OzancanKaratas:
Please merge the beta branch.
@leonardder @feerrenrut


OzancanKaratas:
Sorry for forgetting some changes. Please don't close this. We can merge it using patch.

leonardder:
We're not going for beta with CLDR 38.1. It will have to be 2021.1 instead.

josephsl:
Hi, in addition, one important consideration: although things are not done with beta, we have to focus on stability. Although CLDR 38.1 may introduce new emoji characters, we have to give it a chance to be tested by testers, and beta is not a good branch at this time. This is more so now that 2020.4 beta 1 was just declared, which to me means stability takes priority (personally, I’m against applying dependency updates to beta branch when a beta cycle is active unless the dependency fixes an issue introduced in a beta release). Therefore, I’m in agreement with Leonard, and move to target 2021.1. Thanks.

OzancanKaratas:
Ok, now it's ready for review. Thanks for the branch change @leonardder.

michaelDCurran (approved):

PR #11597, opened 2020-09-12
Fix regression introduced by #11478

michaelDCurran:
### Link to issue number:
Fixes #11587 

### Summary of the issue:
Pr #11478 fixed issue #9463 by moving the sending of VK_NONE (for modifiers that would normally perform an action) from the key hook into NvDA's main thread, after executing a script.
However, it caused issue #11587  
Specifically: if a script is bound to a gesture that uses the alt key, and that script takes a long time to run, and the user releases the alt key before the script is finished, and therefore before the VK_NONE is sent, then the alt key will perform its normal action of activating the menu bar.

### Description of how this pull request fixes the issue:
Rather than sending the VK_NONE after the script finished executing only if the gesture had not had its 'send' method called, now the VK_NONE is sent just before the script is executed (still in the main thread of course). 
The tracking of how many times 'send' has been called has now also been removed as it is no longer needed.

 
#### Testing for #11587 
* added a time.sleep of 0.5 seconds to script_review_nextLine and bound that script to alt+rightArrow. Witout this fix, releasing the alt key before the script finishes causes the menu bar to activate. With the changes in this pr the menu bar is no longer activated when pressing alt+rightArrow.
* Provided a try build on the issue, and the reporter confirmed the issue is fixed.
#### Testing for issue #9463 
* Switched to an input language with multiple keyboards installed. Then selected text on google.com in Mozilla Firefox using control+shift+downArrow. Confirmed that NVDA announced the text selection in browse mode, and that the input language did not change (due to control+shift) being pressed.
* Followed the steps in issue #9463 
* Provided a try build on the issue, and a user confirmed that the issue is still fixed.

### Known issues with pull request:
None

### Change log entry:
None required.

PR #11478, opened 2020-08-10
No longer freeze or restart when pressing control+shift+downArrow in MS Word

michaelDCurran:
An alternative of pr #11477 
This time, send the VK_NONE from the main thread, if the script does not send the gesture. This completely avoids deliberate reentrancy in the keyboard hook, and also avoids a time.sleep and the call to wx.yield in the winInputHook thread (which causes a wx assertion). Users report that no only does this fix the issue but this also massively speeds up holding down of control+shift+downArrow in MS Word, no doubt due to no longer sleeping or yielding in the keyboard hook.

### Link to issue number:
Fixes #9463 

#### background
As NVDA detects key presses in its key hook, it may block a certain key from continuing through to the Operating System if that key press is bound to a particular NvDA script. The script may or may not then choose to send the key onto the Operating system later, depending on the action it is taking. 
However, certain key combinations such as control+shift, may cause an action in Windows such as changing the input language, if a main key is not also included with that key combination. E.g. control+shift by itself changes input languages, but control+shift+downArrow does not. But, if NVDA blocks control+shift+downArrow, and does not send any key, then the input language is still changed. To get around this, in its key hook, NVDA sends a special VK_NONE (0xff) when modifiers like control+shift are pressed, (whether or not NVDA scripts later send a key) to ensure that the Operating System does not change input languages.
#### Actual problem
To send the VK_NONE key press in its key hook, currently NVDA uses KeyboardInputGesture.send, which among other things, enters a lock and sets a global variable which flags that NVDA should ignore any injected keys at the moment (I.e. this one).
The assumption was that during the call to KeyboardInputGesture.send (or keybd_event more specifically) only the key press NVDA sent  would be executed in NVDA's key hook. And because the OS would have marked it as injected, and because ignoreInjected was true, then we would exit from the reentered key hook early.
But, on some Operating systems (E.g. users in #9463 ) other key presses currently pending, may also execute in the key hook inside keybd_event. In this case though, they are not injected, and therefore the entire key hook runs, and as these key presses are most likely repeats of modifier keys that require us to send the VK_NONE, then we do this, but cause a deadlock in the winInputHook thread as our ignoreInjected lock is not reentrant. Further to this, NVDA's main thread may then actually itself call KeyboardInputGesture.send in a script (perhaps for control+shift+downArrow) which also tries to take the lock, and waits for ever as well.

See the following Python stacks:
```
Python stack for thread 10376 (winInputHook):
  File "threading.pyc", line 890, in _bootstrap
[...]

PR #11944, opened 2020-12-16
Allow QWERTY input from Papenmeier braille displays to work again

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
Some Papenmeier braille displays have the ability to allow QWERTY input. However, it seems an equality check that needed to be corrected with the conversion to Python 3 was missed which disabled input completely, plus some reorganisation of keyboard input support in NVDA 2020.3 caused NVDA to freeze at a random point after typing for a while with a Papenmeier braille display.

The Papenmeier braille display driver has been updated to correctly detect and allow QWERTY input, and several new gesture bindings have also been added. The update to this driver was provided to me by Papenmeier.

keyboardHandler.injectRawKeyboardInput has been simplified to just call the keybd_event win32 API directly, and let the OS handle the input from there. Previously, NVDA would send the input manually to its keyboard input hook, and then if the hook didn't handle the input, calling keybd_event. This avoids assumptions about how NVDA's input hook worked, which are no longer true since #11478 and  #11597  were merged.

keyboardHandler.injectRawKeyboardInput is currently only used by the Papenmeier braille display driver, thus why it probably broke in NVDA 2020.3. 

Although I could not get a good log from papenmeier showing a useful stack, my guess is that their input thread was somehow getting stuck on the 'ignoreInjected' lock in keyboardHandler as NVDA did not expect its input hook (internal_keyDownEvent) to be called directly from that thread.
 
### Testing performed:
* Provided a try build to Papenmeier and had them test the driver including QWERTY input for multiple hours. They reported that the freeze they used to experience was now gone and they could type accurately.

None

### Change log entry:
Bug fixes:
* QWERTY input on Papenmeier braille displays that support it again works and no longer causes NVDA to randomly freeze.


feerrenrut (dismissed):
Looks good thanks @michaelDCurran (and Papenmeier).

feerrenrut (approved):

PR #11958, opened 2020-12-21
Fix cyclic import context help

feerrenrut:
### Link to issue number:
#11950

### Summary of the issue:
When installing add-ons which imported gui in their InstallTasks module NVDA failed to restart properly.

This was caused by the fact that when gui was imported it also imported contextHelp.ContextHelpMixin and contextHelp in turn also imports gui causing circular import.

Alternative fix to #11951
Breaks the circular import in contextHelp by making imports of gui, ui, and queueHandler dynamic imports.

This exposes another problem, the order of imports mattered. Several modules imported by gui, depended on ContextHelpMixin (imported name) in the gui module. This is fixed by referring directly to gui.contextHelp.ContextHelpMixin rather than gui.ContextHelpMixin in all those files. To prevent this from happening again (or being used by add-ons) it has been imported as _ContextHelpMixin.

### Testing performed:
Tested as per #11950

Dependencies and layout of code in gui package needs attention. This would be an API breaking change, as such it can happen in 2021.1

### Change log entry:
None


source/cursorManager.py
```diff
@@ -15,6 +15,7 @@
 import documentBase
 import gui
 from gui import guiHelper
+import gui.contextHelp
```
lukaszgo1, line 18: Same as above

source/gui/__init__.py
```diff
@@ -1,9 +1,15 @@
 # -*- coding: UTF-8 -*-
-#gui/__init__.py
-#A part of NonVisual Desktop Access (NVDA)
-#Copyright (C) 2006-2018 NV Access Limited, Peter Vágner, Aleksey Sadovoy, Mesar Hameed, Joseph Lee, Thomas Stivers, Babbage B.V.
-#This file is covered by the GNU General Public License.
-#See the file COPYING for more details.
+# gui/__init__.py
```
lukaszgo1, line 2: Please remove this line as per Copyright headers in the wiki

source/updateCheck.py
```diff
@@ -20,7 +20,7 @@
 	raise RuntimeError("No update version type, update checking not supported")
 import addonAPIVersion
 # Avoid a E402 'module level import not at top of file' warning, because several checks are performed above.
-from gui.contextHelp import ContextHelpMixin  # noqa: E402
+import gui.contextHelp  # noqa: E402
```
lukaszgo1, line 23: Another redundant import

lukaszgo1 (dismissed):
Aside from failing Linter checks and some cosmetic changes which I've requested below this looks fine to me and resolves this issue on my side. Nice work!

feerrenrut:
I don't think these imports should be considered redundant. The `gui` package now no longer explicitly exposes that name, if no other module imported `ContextHelp` it wouldn't be found in the `gui` package. Depending on this is depending on an implementation detail. Instead I think it is better to be explicit. 

For more (and better) descriptions of this see:
- [the submodules are added to the package namespace trap](http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-submodules-are-added-to-the-package-namespace-trap)
- this [stack overflow answer](https://stackoverflow.com/a/49641638/346188)

PR #8165, opened 2018-04-11
Try to use window.edit.Edit when an editable window supports it

LeonarddeR:
### Link to issue number:
Closes #8102
Closes #636

### Summary of the issue:
NVDA uses the display model for editable text cases where there is a caret in the window and there doesn't seem to be another overlay class that implements editing. However, NVDA is a bit conservative in this, only using window.edit.Edit for objects with a particular window class or for windowless richEdit.

### Description of how this pull request fixes the issue:
When finding an overlay class for an object and there is a caret in the window, rather than directly choosing DisplayModelEditableText, this pr will send the EM_LINELENGTH message to the window. This window message always returns 1 or higher for windows that support the edit API, also for single line edit controls. This means that, when the window responds with >=1 to EM_LINELENGTH, the edit API is supported. In this case, a new UnidentifiedEdit class is used that makes no assumptions about the edit API version. For these controls, this pr will try to construct an ITextDocumentTextInfo object to retrieve text information. If that fails, the implementation will fall back to EditTextInfo.

### Testing performed:
Tested the notetab text editor. Instead of using DisplayModelEditableText for the edit control, UnidentifiedEdit is now used, and the control even implements ITextDocument support. Also tested a corporate Delphi application that has edit controls with custom window classes.

### Known issues with pull request:
None

### Change log entry:
* Bug fixes
    - In some text controls, particularly in Delphi applications, the information provided about editing and navigating is now much more reliable. (#636, #8102)

PR #11966, opened 2020-12-22
Don't use UnidentifiedEdit for windows with empty windowText

lukaszgo1:
### Link to issue number:
None, regression from PR #8165
### Summary of the issue:
PR #8165 started using `UnidentifiedEdit` for windows which contain a caret and for which `EM_LINELENGTH` returns 1 or more. Unfortunately  I've seen cases in the wild in which `EM_LINELENGTH` returns count of lines in the edit field yet it's `windowText` is empty and `UnidentifiedEdit` relies on `windowText` being correct to retrieve content.

If the given window responds to `EM_LINELENGTH` with a positive value before adding `UnidentifiedEdit` to its MRO we're now checking if it windowText is not empty.
### Testing performed:
Ensured that main editor in BASCOM_AVR from MCS Electronics is once again readable.

It's theoretically possible that getting windowtext before adding `UnidentifiedEdit` can cause performance problems. I'm reluctant to use an appModule based approach here because using class which requires windowText to be correct for windows for which windowText is empty makes no sense.
### Change log entry:

Section: Bug fixes

Fixed access to edit fields in MCS Electronics IDE's


michaelDCurran:
@lukaszgo1  Can you please give a bit more explanation as to the exact issue with edit fields in MCS Electronics IDE? If we suppress these windows from being classed as an unidentified edit, what does the user then experience? How is it "once again readable"?

lukaszgo1:
@michaelDCurran wrote:
> Can you please give a bit more explanation as to the exact issue with edit fields in MCS Electronics IDE? If we suppress these windows from being classed as an unidentified edit, what does the user then experience? How is it "once again readable"?



When we're not using `unidentified edit` to access these edit fields NVDA falls back to the `DisplayModelEditableText`  which makes text of these controls reported when navigating with the system caret. Without this fix these edit fields are reported as empty.

michaelDCurran (approved):

PR #10703, opened 2020-01-20
Use enum for speech reason

feerrenrut:
### Link to issue number:
Stemmed from comment: https://github.com/nvaccess/nvda/pull/10593#pullrequestreview-329358023

### Summary of the issue:
Speech `reason` was a string, this made it hard to know what the "allowed" values are and restrict values to domains that make sense.

### Description of how this pull request fixes the issue:
Replace speech `reason` constants with an enum.
Note there was one `reason` found defined outside of `controlTypes`, `REASON_QUICKNAV = "quickNav"` in `BrowseMode.py`.  It's possible, but unlikely that there are others.
Should not break compatibility, as long as addons are using the constants for reason defined in `controlTypes`.

### Testing performed:
Ran from source, attempting to exercise as many speech "reason" paths as possible. 

### Known issues with pull request:
If addons or other code are using string literals for the reason argument, they will break. This is considered low risk, since it's expected that they are using the constants in `controlTypes`

### Change log entry:

Changes for developers:
`- speech 'reason' has been converted to an Enum, see controlTypes.OutputReason class. `

PR #11782, opened 2020-10-23
Fix misleading log message for aria-current="true"

feerrenrut:
### Link to issue number:
None

### Summary of the issue:
There is a misleading log message for aria-current="true". It was a non-handled value, an exception was raised, a message added to the log, and the value `True` (bool rather than str) was used.

When trying to follow this code it is hard to know where the values are strings, None, or bool. In some cases the naming is overly specific.

### Description of how this pull request fixes the issue:

Types and naming for isCurrent were confusing, so normalize the types
using an enum and only name as ariaCurrent when the data is specific to aria-current,
otherwise use 'isCurrent'. Include comments to link aria-current as the
potential (likely) source for 'isCurrent'

### Testing performed:
Used Chrome, Firefox, IE, and Edge with https://a11ysupport.io/tests/html/aria/aria-current.html

### Known issues with pull request:
This is technically an API breaking change. If there are any concerns about this, we may have to wait for the 2021.1 release. 

### Change log entry:

```
changes for developers:
- 'NVDAObject' (and derivatives) property 'isCurrent' now strictly returns 'controlTypes.IsCurrent'. It is no longer Optional, and thus will not return None. When an object is not current 'controlTypes.IsCurrent.NO' is returned.
- The 'controlTypes.isCurrentLabels' has been removed, instead use the 'displayString' property on a 'controlTypes.IsCurrent' enum value. EG 'controlTypes.IsCurrent.YES.displayString'
```

PR #11969, opened 2020-12-25
Convert NVDA's code base to use output reasons from enum in controlTypes and remove deprecated REASON_* module level constants

lukaszgo1:
### Link to issue number:
Discussion in #10703 and #10732
### Summary of the issue:
PR #10703 introduced new `OutputReason` class which should be used in place of previously used `REASON_*` constants from the `controlTypes` module. To provide backward compatibility old module level constants were kept in place, but marked as deprecated. Since 2021.1 intents to break backwards compatibility anyway it makes sense to use output reasons from the new enum throughout the code and remove these deprecated constants.
This pr converts all usages of `REASON_*` constants to the new enum (git grep was used to look for all occurrences of them) and removes old constants from the `controlTypes` and `browseMode` modules.
### Testing performed:
- Ensured that unit tests still passes.
- Used NVDA for around a day with these changes without experiencing any regressions in functionality.
- With git grep ensured that only mentions of `REASON_` in the code is in the changelog file
I'd be great to merge this soon not only to give it a good amount of testing on Alpha, but it touches many files and therefore is quite  prone to conflicts.
### Change log entry:
Changes for developers:
- Module level  `REASON_*` constants are removed from `controlTypes` - please use `controlTypes.OutputReason` instead.
- `REASON_QUICKNAV` has been removed from `browseMode` - use `controlTypes.OutputReason.QUICKNAV` instead


lukaszgo1:
@feerrenrut As I've predicted in the PR description due to the amount of modified files this had conflicts already. Would it be possible to review this to avoid multiple back and forth of fixing merge conflicts here?

feerrenrut:
Ok, I'll get #10732 delivered first , then I'll come and review this. To speed up the review, please go over the approach in #10732 to ensure the approach here is consistent.

feerrenrut:
Sorry, I meant PR #11782, I'll edit my last comment.

lukaszgo1:
@feerrenrut This PR merely starts using enum added in #10703 so I believe consistency should not be an issue. In deed `OutputReason` matches style used in #11782

michaelDCurran (approved):

PR #11973, opened 2020-12-26
Sphinx: remove 2.2.2 restriction and update readme regarding building source code docs locally

josephsl:
Hi,
Possible oversight when Sphinx was introduced:

### Link to issue number:
None

### Summary of the issue:
When Sphinx was introduced, readme wasn't updated to point out that source code documentation can be built in Python 3 version of NVDA. Also, used this opportunity to remove Sphinx version restriction (latest release is 3.4.1, which succeeds in building source code docs).

Updates:

* Readme: add instructions on building source code documentation, and remove dev docs limitation statement.
* Replace Epydoc with Sphinx.

Sphinx requirement: remove version restriction.

### Testing performed:
Built NVDA dev docs with Sphinx 2.2.2 and 3.4.1 with both succeeding.

None

### Change log entry:
None

Thanks.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -223,7 +223,7 @@ To generate the HTML-based source code documentation, type:
 scons devDocs
 ```
 
-The documentation will be placed in the `devDocs` folder in the output directory.
+The documentation will be placed in the `NVDA` folder in the output directory.
 
 To generate developer documentation for nvdaHelper (not included in the devDocs target):
 
--- a/readme.md
+++ b/readme.md
@@ -216,7 +216,14 @@ scons developerGuide
 ```
 
 The developer guide will be placed in the `devDocs` folder in the output directory.
-Note that the Python 3 sources of NVDA currently do not support building NVDA developer documentation using the `scons devDocs` command.
+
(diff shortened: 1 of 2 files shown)


readme.md
```diff
@@ -122,6 +121,7 @@ Although this [must be run manually](#linting-your-changes), developers may wish
 
 The following dependencies aren't needed by most people, and are not included in Git submodules:
 
+* To generate developer documentation: [Sphinx](http://sphinx-doc.org/), version 3.4.1
```
lukaszgo1, line 124: This is inconsistent with the requirements file. I can imagine that someone reports an issue stating that readme asks to install specific version of Sphinx whereas requirements file installs whatever is the latest version at the time of installation. It'd be great to either remove specifif version number from the readme, or enforce version 3.4.1 in the requirements file. I personally prefer the latter as we cannot guarantee that dev docs would always be buildable with the most recent version of Sphinx.

josephsl:
Hi, although not ideal, specifying a specific release can clarify things for now. Thanks for pointing this out.

michaelDCurran (approved):

PR #11975, opened 2020-12-27
Synth drivers/eSpeak NG internal: pass in a NULL pointer (path string) when obtaining eSpeak NG version string

josephsl:
Hi,

Needed for Python 3.8, although it is equally applicable to Python 3.7 and makes NVDA code more compliant with eSpeak NG API:

### Link to issue number:
None

### Summary of the issue:
eSpeak NG fails to load on Python 3.8 version of NVDA, traced to espeak.dll::espeak_Info function not receiving path string.

Pass in NULL (None) when eSepak NG asks for path string in sepak_Info function because what NVDA is interested in is eSpeak NG version string.

### Testing performed:
Tested under source code and via a private binary copy:

* Success in Python 3.7 version
* Success in Python 3.8 version

None

### Change log entry:
None

### Impact:
This is one of the major changes required to move NVDA to Python 3.8, especially given that eSpeak NG will be set as default speech synthesizer on Windows releases other than 10.

Thanks.

### DIFF
--- a/source/synthDrivers/_espeak.py
+++ b/source/synthDrivers/_espeak.py
@@ -1,5 +1,4 @@
 # -*- coding: UTF-8 -*-
-# synthDrivers/_espeak.py
 # A part of NonVisual Desktop Access (NVDA)
 # Copyright (C) 2007-2020 NV Access Limited, Peter Vágner
 # This file is covered by the GNU General Public License.
@@ -370,7 +369,7 @@ def terminate():
 
 def info():
 	# Python 3.8: a path string must be specified, a NULL is fine when what we need is version string.
-	return espeakDLL.espeak_Info(None)
+	return espeakDLL.espeak_Info(None).decode()
 
 def getVariantDict():
 	dir = os.path.join(globalVars.appDir, "synthDrivers", "espeak-ng-data", "voices", "!v")
--- a/source/synthDrivers/_espeak.py
+++ b/source/synthDrivers/_espeak.py
@@ -369,7 +369,8 @@ def terminate():
 	onIndexReached = None
 
 def info():
-	return espeakDLL.espeak_Info()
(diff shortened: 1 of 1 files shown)


source/synthDrivers/_espeak.py
```diff
@@ -1,9 +1,9 @@
 # -*- coding: UTF-8 -*-
-#synthDrivers/_espeak.py
-#A part of NonVisual Desktop Access (NVDA)
-#Copyright (C) 2007-2017 NV Access Limited, Peter Vágner
-#This file is covered by the GNU General Public License.
-#See the file COPYING for more details.
+# synthDrivers/_espeak.py
```
lukaszgo1, line 2: Please remove this line to comply with the standard for copyright headers.

source/synthDrivers/_espeak.py
```diff
@@ -369,7 +369,8 @@ def terminate():
 	onIndexReached = None
 
 def info():
-	return espeakDLL.espeak_Info()
+	# Python 3.8: a path string must be specified, a NULL is fine when what we need is version string.
+	return espeakDLL.espeak_Info(None)
```
lukaszgo1, line 373: Since this function returns bytes, and the only place in which it is used is the message in the log stating which version of eSpeak is in use have you considered decoding this to avoid leading "b" being shown in the log viewer i.e.:
`return espeakDLL.espeak_Info(None).decode()`

lukaszgo1 (changes requested):

josephsl:
Hi, should we also remove encoding statement at the top of the file? If yes, I’ll do that while addressing comments. As for encoding from bytes to str, I’ll do that once Grep tells me it is safe to do so. Thanks.

michaelDCurran (approved):

PR #11977, opened 2020-12-27
Correctly report accented letters after cap

### LINKED ISSUE #11948: nvda does not announce the accentuation of the letter if the check box is Say capitalization before capitalization
### Steps to reproduce:
Open a text editor
In the nvda speech settings select the checkbox
Say capitalization before capitalization
return to the text editor
turn on capslook
type letters with accents
examples:
Á
À
Ã
É
Ê
Í
Ó
Ô
Ú
read the accented letters typed with the left and right arrows

### Actual behavior:
nvda informs
the character name is spoken
character accent is not spoken
is spoken capitalized

### Expected behavior:
nvda informs:
speak the name of the character
speaks the accentuation of the character
speaks  capital letter
instaled
#### NVDA version:
nvda.exe, NVDA alpha-21524,7b92e442

#### Windows version:
windows 10 19042.662
any text editor
if the checkbox
Say capitalization before capitalization
is unchecked and the user chooses another way to receive the capitalization status of the letters, everything works fine.
In this case I want the nvda to announce the capitalization of the letters, and the accentuation of the letters as well.
tested with several synthesizers and the problem remains

yes
yes
yes
no

CyrilleB79:
### Link to issue number:

Fixes #11948

### Summary of the issue:

As described in #11948, some letter with diacritic is not spelt correctly with some synth if "Say cap before capitale" is checked, e. g. 'À' is spelt 'cap A' instead of 'cap A grave'.
Looking at `getSpellingSpeech` function in `speech\__init__.py`, it appears that this is due to the fact that the string 'cap À' is passed to the synth in the speech sequence in such a case. But the synth is only asked to use character mode for the strings whose length is 1. This is wrong in this case, because 'cap' should be spoken with character mode off, but 'À' should be spoken with character mode on.

In `getSpellingSpeech`, in the string sequence, I have separated the 'cap' message string from the character to be spelt. If the character to be spelt is of length 1, i.e. not replaced by symbol description or modified pronunciation, character mode is activated. If a string in the speech sequence has a length > 1, because it is a capitalization message or because it corresponds to a character description or symbol replacement, character mode is disabled.

Note that the 'cap' message can be of two types:
* with 'cap' indication before as in English: 'cap %s'
* with 'cap' indication after as in French: '%s majuscule'
Thus this message is split in 2 parts in the speech sequence.

I have also taken advantage of the rework I had to do in `getSpellingSpeech` to split this function that was too complex. I have thus removed the corresponding flake8 exception (noqa comment).

### Testing performed:

* Tested with 'Œ' character in French that it was pronounced as a character when spelt with eSpeak and with OneCore, what was not the case before this PR.
* Checked the speech sequence in the log: 1-character long string are correctly preceded by a CharacterMode(True) command if not already in character mode.

* This PR does not fix all character spelling issues. E.g. OneCore seems not to spell 'Á' correctly even if set in char mode, but the 'Say cap before capitale' option has not impact on this issue.
* This PR assumes that the translated capitalization message has 0 or at least 2 characters before the '%s'; and similarly, the translated capitalization message should have 0 or at least 2 characters after the '%s'. Should it not be the case, some parts of this message would be spelt by the synth instead of being normally read. I have checked with a [python script](https://github.com/nvaccess/nvda/files/5745402/testTranslations.py.txt)
 that this assumption is verified in all the existing translations of NVDA.
If this does not seems robust enough, I may add a unit test such as the check_pot to check in all the .po files that this condition is met. Just let me know.

### Change log entry:

Section: Bug fixes

`NVDA should no longer spell incorrectly some letters with accents or other diacritic if 'Say cap before capitale' option is checked.  (#11948)`


source/speech/__init__.py
```diff
@@ -230,10 +230,64 @@ def speakSpelling(
...
+	if capMsgBefore:
+		yield capMsgBefore
+	yield speakCharAs
+	if capMsgAfter:
+		yield capMsgAfter
+	if uppercase and isPitchSupported and synthConfig["capPitchChange"]:
+		yield PitchCommand()
+
+
+def getSpellingSpeechWithoutCharMode(
```
feerrenrut, line 290: Please make this a private method for now:
```suggestion
def _getSpellingSpeechWithoutCharMode(

```

source/speech/__init__.py
```diff
@@ -230,10 +230,64 @@ def speakSpelling(
...
+				if not charMode:
+					yield CharacterModeCommand(True)
+					charMode = True
+			elif charMode:
+				yield CharacterModeCommand(False)
+				charMode = False
+		yield item
+
+
+def getSpellingCharWithCaseInfo(
```
feerrenrut, line 255: Please make this a private method for now:
```suggestion
def _getSpellingCharWithCaseInfo(

```

source/speech/__init__.py
```diff
@@ -230,10 +230,64 @@ def speakSpelling(
 	speak(seq, priority=priority)
 
 
-# C901 'getSpellingSpeech' is too complex
-# Note: when working on getSpellingSpeech, look for opportunities to simplify
-# and move logic out into smaller helper functions.
-def getSpellingSpeech(  # noqa: C901
+def getSpellingSpeechAddCharMode(
```
feerrenrut, line 233: Please make this a private method for now:
```suggestion
def _getSpellingSpeechAddCharMode(

```

feerrenrut (changes requested):
It would be really helpful if this added unit tests for these modified functions.

CyrilleB79:
> It would be really helpful if this added unit tests for these modified functions.

OK, I will have a look at it.
Since `getSpellingSpeech` has been split in various functions, do you expect unit tests only on `getSpellingSpeech` with various use cases to test all the branches (use cases) of the old `getSpellingSpeech`?
Or do you also expect unit tests on the newly created helper functions called by `getsSpellingSpeech` (`_getSpellingSpeechAddCharMode`, `_getSpellingCharWithCaseInfo`, `_getSpellingSpeechWithoutCharMode`)?

Passing this PR as draft while developing unit tests.

feerrenrut:
>Since `getSpellingSpeech `has been split in various functions, do you expect unit tests only on `getSpellingSpeech `with various use cases to test all the branches (use cases) of the old `getSpellingSpeech`?
Or do you also expect unit tests on the newly created helper functions called by `getsSpellingSpeech` (`_getSpellingSpeechAddCharMode`, `_getSpellingCharWithCaseInfo`, `_getSpellingSpeechWithoutCharMode`)?

It's probably clearer to test the three internal functions separately.

CyrilleB79:
Hello @feerrenrut 

I am trying to set up unit test for the new functions. I am not used to unit test development and have not ever done it for NVDA. Thus, I need some guidance.

I have created a `test_speech.py` file containing a `Test_getSpellingSpeechAddCharMode` class.
This class contains a `test_symbolNameAtTheEnd` function as follow:

> ```
> 	def test_symbolNameAtTheEnd(self):
> 		seq = makeGen(['h', EndUtteranceCommand(), 'e', EndUtteranceCommand(), 'l', EndUtteranceCommand(), 'l', EndUtteranceCommand(), 'o', EndUtteranceCommand(), 'bang', EndUtteranceCommand()])
> 		expected = repr([CharacterModeCommand(True), 'h', EndUtteranceCommand(), 'e', EndUtteranceCommand(), 'l', EndUtteranceCommand(), 'l', EndUtteranceCommand(), 'o', EndUtteranceCommand(), CharacterModeCommand(False), 'bang', EndUtteranceCommand()])
> 		output = _getSpellingSpeechAddCharMode(seq)
> 		self.assertEqual(repr(list(output)), expected)
> ```

When running the unit test, I get the following error:

> ```
> File "D:\Cyrille\Dev\GIT\nvda\tests\unit\test_speech.py", line 29, in test_symbolNameAtTheEnd
> self.assertEqual(repr(list(output)), expected)
> File "D:\Cyrille\Dev\GIT\nvda\source\speech\__init__.py", line 222, in _getSpellingSpeechAddCharMode
> synthConfig = config.conf["speech"][synth.name]
> AttributeError: 'NoneType' object has no attribute 'name'
> ```

There is probably some sort of stub in unit tests that allow to avoid having a real synth driver.
 Where and how should I create a synth stub? The same for config that will probably be missing.

Also, any feedback is welcome. All this is quite new for me.

Thanks for your help.

feerrenrut:
Ok, I see the issue, several of these newly introduced methods are coupled to `config`. Rather than look up the synth driver, and config from directly in the method, pass them in as arguments. If this is done for all of these functions then it is only the "glue" function `getSpellingSpeech` that is coupled to config. This will also make it easier to be sure you have tested all the combinations.

The test would become like:
```python
def test_symbolNameAtTheEnd(self):
	seq = makeGen([
		'h', EndUtteranceCommand(),
		'e', EndUtteranceCommand(),
		'l', EndUtteranceCommand(),
		'l', EndUtteranceCommand(),
		'o', EndUtteranceCommand(),
		'bang', EndUtteranceCommand()
	])
	expected = repr([
		CharacterModeCommand(True),
		'h', EndUtteranceCommand(),
		'e', EndUtteranceCommand(),
		'l', EndUtteranceCommand(),
		'l', EndUtteranceCommand(),
		'o', EndUtteranceCommand(),
		CharacterModeCommand(False),
		'bang', EndUtteranceCommand()
	])
	output = _getSpellingSpeechAddCharMode(seq, useSpellingFunctionality = True)
	self.assertEqual(repr(list(output)), expected)
```
You would also add a test for `useSpellingFunctionality = False`

CyrilleB79:
@feerrenrut, I have added unit tests for `_getSpellingSpeechAddCharMode` and `_getSpellingCharAddCapNotification`.

Regarding `_getSpellingCharAddCapNotification`, it would be interesting to add another test case for languages such as French where the 'cap %s' message is translated with `%s`  before ('%s majuscule' in French). I guess I should override the `_` function in some way. How should I do?

Regarding `_getSpellingSpeechWithoutCharMode` it contains the majority of the code that was in `getSpellingSpeech` and that I did not modified. Would you recommend to add unit tests for this function also? Passing synth config parameters as additional arguments do not seem to simplify the code. What do you think?
Any other idea of refactor?

Thanks!

feerrenrut:
> it would be interesting to add another test case for languages such as French where the 'cap %s' message is translated with %s before ('%s majuscule' in French). I guess I should override the _ function in some way. How should I do?

Yes. Since you know that there are variations that must be handled, you could do it like that. Rather than call it a French test, which implies that locale is being set, I'd say "test variations on placeholder position (which occurs in other languages like French)". It would be good the code didn't have to deconstruct those strings for these variations. I don't fully understand the need for it and was hoping that the unit tests will demonstrate it to me.

> Regarding _getSpellingSpeechWithoutCharMode it contains the majority of the code that was in getSpellingSpeech and that I did not modified. Would you recommend to add unit tests for this function also?

Yes, please do. Since you have taken the time to understand it, and what the expected output is, it would be great to document that via unit tests. 

>Passing synth config parameters as additional arguments do not seem to simplify the code. What do you think?
Any other idea of refactor?

While it might not immediately seem like it, I think it is a good thing to expose the underlying complexity. The fact is that this code depends on many variables. It is better for this to be explicit. One additional advantage is that it is then clear that it doesn't also modify config.

CyrilleB79:
> > it would be interesting to add another test case for languages such as French where the 'cap %s' message is translated with %s before ('%s majuscule' in French). I guess I should override the _ function in some way. How should I do?
> 
> Yes. Since you know that there are variations that must be handled, you could do it like that. Rather than call it a French test, which implies that locale is being set, I'd say "test variations on placeholder position (which occurs in other languages like French)". It would be good the code didn't have to deconstruct those strings for these variations. I don't fully understand the need for it and was hoping that the unit tests will demonstrate it to me.

The 'cap %s' string has to be deconstructed in some way because:
* the word 'cap' has to be pronounced as word by the synth (i.e. has to follow a `CharacterModeCommand(False)` command) 
* '%s' has to be pronounced as character by the synth (i.e. has to follow `CharacterModeCommand(True)` command).
    I agree this seems ugly code but I did not find a more elegant way to do this.

> > Regarding _getSpellingSpeechWithoutCharMode it contains the majority of the code that was in getSpellingSpeech and that I did not modified. Would you recommend to add unit tests for this function also?
> 
> Yes, please do. Since you have taken the time to understand it, and what the expected output is, it would be great to document that via unit tests.

OK, will do.

CyrilleB79:
As requested, I have refactored a bit `_getSpellingSpeechWithoutCharMode` and implemented some unit tests for it.

For now I have skipped `test_capNotificationsWithPlaceHolderAfter` since, as explained before, I do not know how to implement it correctly. The way it is implemented however gives the idea of the goal of this test.
If you have a suggestion to implement it correctly, I will be happy to include it. If not, I may remove this test.

tests/unit/test_speech.py
```diff
@@ -0,0 +1,325 @@
...
+		output = _getSpellingCharAddCapNotification(
+			speakCharAs='A',
+			sayCapForCapitals=True,
+			capPitchChange=0,
+			beepForCapitals=False,
+		)
+		self.assertEqual(repr(list(output)), expected)
+
+	@unittest.skip("Need to implement patching of _ function.")
+	def test_capNotificationsWithPlaceHolderAfter(self):
```
feerrenrut, line 148: I would use a [setup / teardown,](https://docs.python.org/3/library/unittest.html#class-and-module-fixtures) there is nicer syntax for this in pyTest, but lets stick with what we've got:
``` python
class Test_getSpellingCharAddCapNotification(unittest.TestCase):
	_translationResults: typing.Dict[str, str] = {}
	
	@classmethod
	def translationFunctionFake(cls, s: str)
		if str in cls._translationResults:
			return cls._translationResults[str]
		return cls.originalTranslationFunction(s)

	@classmethod
	def setUpClass(cls):
		global _
		cls.originalTranslationFunction = _
		_ = cls.translationFunctionFake

	@classmethod
	def tearDownClass(cls):
		global _
		_ = cls.originalTranslationFunction 

	...
	Some code later
	...
		def test_capNotificationsWithPlaceHolderAfter(self):
			self._translationResults["cap %"] = "%s cap"
			... etc
```

feerrenrut (commented):
I have given some suggestions, let me know if you'd like more guidance.

source/speech/__init__.py
```diff
@@ -211,19 +211,17 @@ def speakSpelling(
...
 	"""Inserts CharacterMode commands in a speech sequence generator to ensure any single character
 	is spelt by the synthesizer.
 	@param seq: The speech sequence to be spelt.
 	"""
-	synth = getSynth()
-	synthConfig = config.conf["speech"][synth.name]
 	charMode = False
 	for item in seq:
 		if isinstance(item, str):
-			if len(item) == 1 and synthConfig["useSpellingFunctionality"]:
```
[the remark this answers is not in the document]
CyrilleB79: I have looked at it again and I did not find anything letting me think that it would change the behaviour.
If you make a comparison between the initial code (master branch) and this branch, you will have to consider the following key point:
The variables `charMode` and `synthConfig["useSpellingFunctionality"]` are both used only to add `CharacterModeCommand` commands in the speech sequence. The addition of `CharacterModeCommand` commands has just been extracted in a separate function in the new code.

Please let me know if I missed something or if something more specific is unclear to you.

tests/unit/test_speech.py
CyrilleB79: I have modified the unit test implementation according to your suggestion.
But the `test_capNotificationsWithPlaceHolderBefore` test (renamed) still doesn't work. For now I have skipped it to show that the other tests work.

The issue is that the patching of `_` function (gettext translation function) is not working as I would expect. Probably a Python issue rather than a unittest lib issue.

@feerrenrut, could you please help me on this point? Thanks.
Note: the skip decorator has to be removed.

feerrenrut: I see the issue, you have in several places I typed `str` when I meant `s`. I have a more tidy solution for you now, I have pushed this to your branch, I hope you don't mind.

CyrilleB79: > I see the issue, you have in several places I typed `str` when I meant `s`.

Oh yes, my bad.
Just for curiosity, I spent 5 minutes trying to make it work with 's' instead of 'str' but did not had any success. I finally gave up due to the other solution you implemented.

> I have a more tidy solution for you now, I have pushed this to your branch, I hope you don't mind.

No problem.
I would not have found the gettext specific class trick; I am not very familiar with gettext.
Specifically, after reviewing your commit, the only thing I am not able to explain is why installing an instance of `gettext.NullTranslations' is equivalent to `gettext.install('nvda')`.

If you can clarify it for my curiosity and so that I may be able to explain what stands in my own PR, that would be nice.

In any case, I close this conversation since the point is resolved. Thanks!

CyrilleB79:
@feerrenrut, thanks for your last commit.
I have reviewed your changes that seems OK to me. I have closed the 2 last open conversations with my comments.

If you do not have any other point to comment, this PR is ready from my side.

feerrenrut (approved):
Thanks @CyrilleB79

Issue #11968, opened 2020-12-24
NVDA stops playing audio when audio output device is changed using latest betas


fisher729:
Hi.

I use a USB sound device with my computer. While using the beta, the following has been consistently reproduceable:
1. Use NVDA when the sound card is plugged in.
2. Unplug the USB sound device.
3. Try to invoke progress bar sounds, browse/focus mode sounds etc.
4. Plug back in the USB sound device.5.

Expected behaviour: NVDA should play the sounds using the device that is in use. Both synthesizer and voice should come through the same audio output device.
Actual behaviour: NVDA's synthesizer output comes through the right device in steps 3 and 4, but the minute the audio device changes, NVDA stops playing its sounds until it is restarted.

NVDA versions where this happens: NVDA 2020.4beta1, NVDA 2020.4beta2, but not NVDA 2019.3 or older
Default audio output device: Microsoft Sound Mapper
Installed or portable: Installed
Windows version: Windows 10 20H2 (64-bit) build 19042.685

Brian1Gaff:
For the record, the exact opposite happens here in windows 7. Ie the windows 
sounds still work but nvda won't until its rebooted.

 This sort of thing has always been a little flaky in Windows to be honest.
 Brian

bglists@blueyonder.co.uk
Sent via blueyonder.
Please address personal E-mail to:-
briang1@blueyonder.co.uk, putting 'Brian Gaff'
in the display name field.
Newsgroup monitored: alt.comp.blind-users
----- Original Message -----

kvark128:
I confirm the problem on Windows 10 1809 with NVDA 2020.4beta2. When disconnecting the USB headphones, any speech output causes the following exception:
ERROR - queueHandler.flushQueue (10:22:41.502) - MainThread (9388):
Error in func cancelSpeech
Traceback (most recent call last):
  File "queueHandler.pyc", line 55, in flushQueue
  File "speech\__init__.pyc", line 146, in cancelSpeech
  File "speech\manager.pyc", line 767, in cancel
  File "synthDrivers\espeak.pyc", line 146, in cancel
  File "synthDrivers\_espeak.pyc", line 245, in stop
  File "nvwave.pyc", line 501, in stop
  File "nvwave.pyc", line 112, in _winmm_errcheck
OSError: [Errno 6] There is no driver installed on your system.

fisher729:
Hi.

I get the same errors, but I use a different synth. In fact, the
synthesizer plays no role in whether this occurs or not.

[...]

PR #11978, opened 2020-12-29
Fix no speech after removed USB audio device

feerrenrut:
### Link to issue number:
#11968

### Summary of the issue:
An error during a call to `nvWave.stop` does not reset the device state, and allows the error to propagate all the way up the stack, leaving `nvWave` in an unknown state.

Ensure all calls to `winmm` that take the `_waveout` param are wrapped in try blocks. On error, close the device and log (nvwave category). This should force nvWave to re-open on the next feed, and will ensure that `windll.kernel32.SetEvent(self._waveout_event)` gets called to prevent `sync` from blocking on `waitForSingleObject`.

### Testing performed:
I don't have a USB audio device for testing, I'll ask reporters to test a try build.

Will require extensive testing.

### Change log entry:
None

### DIFF
--- a/source/nvwave.py
+++ b/source/nvwave.py
@@ -536,9 +536,9 @@ class WavePlayer(garbageHandler.TrackedObject):
 		self._close()
 
 	def _safe_winmm_call(
-		self,
-		winmmCall: Callable[[Optional[int]], None],
-		messageOnFailure: str
+			self,
+			winmmCall: Callable[[Optional[int]], None],
+			messageOnFailure: str
 	) -> bool:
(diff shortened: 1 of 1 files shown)


feerrenrut:
The system test failure is very likely unrelated, I have kicked off another build.

I have confirmation from three testers of the the try build that this resolves their issue. 

I believe this is ready for review.

LeonarddeR (approved):
Thanks, looks good @feerrenrut

leonardder:
Could we have this in master as soon as possible?

PR #11974, opened 2020-12-26
Global commands: replace gestures map with script decorator

josephsl:
Hi,

Targetting 2021.1 and to encourage usage of script decorator throughout nVDA Core and add-ons:

### Link to issue number:
Closes #11964 

### Summary of the issue:
NVDA 2018.3 introduced scriptHandler.script decorator to ease definition of scripts. one of the places where it is used is global commands where a mixture of script decorator and gestures map is used. To encourage usage of script decorator, standardize around it, beginning with global commands.

### Description of how this pull request fixes the issue:
Replace gestures map with script decorator. One exception is sleep mode command where the script uses an attribute that isn't accepted by script decorator yet.

### Testing performed:
Tested in source code form (with script decorator applied in global commands).

### Known issues with pull request:
None

### Change log entry:
None (not really needed)

### Considerations:
If this approach is successful, I expect people will file pull requests to convert other parts of NVDA to use script decorator, which will result in more add-ons adopting script decorator approach. This change is significant enough to target 2021.1.

Thanks.

PR #11979, opened 2020-12-29
Script decorator: add ability to specify that given script should be active in sleep mode.

lukaszgo1:
### Link to issue number:
None - discovered  by @josephsl during his work on #11974

### Summary of the issue:
It is impossible to specify that given script should be active in sleep mode for a script decorated with script decorator.

Script decorator now accepts additional parameter `allowInSleepMode` which just sets it on the script.
While at it I've also added type hints to the list of its arguments and extended unit test to ensure that `allowInSleepMode` works.
### Testing performed:
I've decorated script which enables/disables sleep mode for a current app and set `allowInSlepMode` to true ensuring that this script indeed works when in sleep mode. This code isn't in the PR as IMHO it makes sense to merge tis PR first and then perform the actual change as part of #11974.

None known
### Change log entry:
Changes for developers:
- It is now possible to make given script work in sleep mode when using script decorator by specifying `allowInSleeepMode` in the list of script decorator's 
-arguments.


josephsl (dismissed):
Thanks. Blocks PR #11974

michaelDCurran (approved):

PR #11639, opened 2020-09-19
V2 of refactor LiveText to use diff-match-patch

codeofdusk:
### Link to issue number:
Supersedes #11500. Closes #3200. Closes #11498.

### Summary of the issue:
`LiveText` objects use [difflib](https://docs.python.org/3/library/difflib.html) to calculate text changes:

* Difflib is slow, making 10,000-line objects completely unusable and slowing down or locking up NVDA for other large objects.
* Comparisons are made at the line level, so edits in the middle of lines cause characters to the right of the caret to be read out.

### Description of how this pull request fixes the issue:
This PR changes diffing functions to operate at the string rather than line level and adds [diff-match-patch](https://github.com/leutloff/diff-match-patch-cpp-stl) (DMP) as an optional diffing algorithm for `LiveText` objects. It is anticipated that DMP will become the default in a future NVDA release pending positive user testing.

Unlike #11500, this PR does not import or dynamically link to DMP due to licensing issues. Instead, [a Python application](https://github.com/codeofdusk/nvda_dmp) is run in another process that calls [the DMP extension](https://github.com/JoshData/diff_match_patch-python), and communicates over standard IO.

### Testing performed:
* Tested this approach, difflib, and a dynamically linked DMP Python extension on very large (millions of characters) blocks of text. Observed that while DMP was able to return diffs in about 10 seconds (even with insignificant IPC overhead), Difflib ran for several minutes (I aborted the diff before it could complete).
* Tested deleting and inserting characters in the middle of a line both with difflib and DMP. With difflib the characters to the right of the caret were read out (as stated in #3200). With DMP only the changed characters were read.

### Known issues with pull request:
None.

### Change log entry:
== Bug Fixes ==
- In terminal programs on Windows 10 version 1607 and later, when inserting or deleting characters in the middle of a line, the characters to the right of the caret are no longer read out. (#3200)
    - This experimental fix must be manually enabled in NVDA's advanced settings panel by changing the diff algorithm to Diff Match Patch.

== Changes For Developers ==
- `LiveText._getTextLines` has been removed. Instead, override `_getText` which returns a string of all text in the object. (#11639)
- `LiveText` objects can now calculate diffs by character. To forcibly modify this for a particular object, override the `diffAlgo` property (see the docstring for details). (#11639)

PR #11981, opened 2020-12-31
Fix error in user guide due to DMP retargetting

codeofdusk:
Follow-up of #11639.

Just a small edit to the DMP section of the user guide, since the PR was targeted for 2021.1 (not 2020.4).

The referenced version for "previous behaviour" was changed from 2020.3 to 2020.4.


feerrenrut (approved):
Thanks @codeofdusk

Issue #11992, opened 2021-01-09
Errors in new diffing algorithm breaks automatic reading of consoles till NVDA restarts


Neurrone:
### Steps to reproduce:

Unsure exactly what caused this, I think it happened right as I switched from another application into the windows 10 console (conhost, am not using windows terminal).

### Actual behavior:

The following error is logged. Subsequent text written to the console is no longer read until NVDA restarts.

```
ERROR - diffHandler.DiffMatchPatch.diff (13:48:59.482) - Dynamic_WinConsoleUIAEditableTextWithAutoSelectDetectionUIA._monitorThread (1516):
Exception in DMP, falling back to difflib
Traceback (most recent call last):
  File "diffHandler.pyc", line 94, in diff
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 322: unexpected end of data
ERROR - diffHandler.DiffMatchPatch.diff (13:48:59.545) - Dynamic_WinConsoleUIAEditableTextWithAutoSelectDetectionUIA._monitorThread (1516):
Exception in DMP, falling back to difflib
Traceback (most recent call last):
  File "diffHandler.pyc", line 91, in diff
OverflowError: cannot fit 'int' into an index-sized integer
```

### Expected behavior:

Reading of new text does not break

#### NVDA installed/portable/running from source:
Installed

#### NVDA version:
alpha-21599,d3aaca53

#### Windows version:
Windows 10 2004

#### Name and version of other software in use when reproducing the issue:
Conhost

Neurrone:
cc @codeofdusk

codeofdusk:
Yes, this is a known issue.

feerrenrut:
It seems this could fail in a more graceful way. @codeofdusk are you investigating this?

codeofdusk:
I'm not really sure how to fix this, although I can reliably repro on my system. As I said over in the PR, flushing stdin/stdout doesn't fix it. I've added logging to nvda_dmp which appears to be receiving everything correctly. I've also contacted the author of the original DMP library to see about obtaining a dual-licensed version for NVDA (obviating the need for nvda_dmp) but haven't received a response.

codeofdusk:
@Neurrone Can you still repro this on latest master?

Neurrone:
I'll need a few days to test it with this setting enabled. Will report back soon.

Thanks @codeofdusk for investigating this 🙏

codeofdusk:
[...]

PR #11998, opened 2021-01-11
Update nvda_dmp to possibly address an autoread hang

codeofdusk:
### Link to issue number:
Possible resolution for #11992.

### Summary of the issue:
Currently, nvda_dmp sends the number of Unicode characters in the response header, then sends utf-8 encoded data. NVDA expects the header to contain the number of bytes, resulting in the diff hanging when the number of characters is fewer than the number of bytes.

nvda_dmp now sends the number of bytes (not characters) in the response header.

### Testing performed:
I can no longer reproduce the hang described in #11992 and https://github.com/nvaccess/nvda/pull/11639#issuecomment-753325754 (I previously thought it was an issue with locking until further testing revealed this bug).

* This PR definitely improves behaviour for me, but should be merged ASAP to allow for testing on master.

### Change log entry:
None needed.

### DIFF
--- a/source/diffHandler.py
+++ b/source/diffHandler.py
@@ -89,6 +89,8 @@ class DiffMatchPatch(DiffAlgo):
 				(size,) = struct.unpack("=I", sizeb)
 				while len(buf) < size:
 					buf += DiffMatchPatch._proc.stdout.read(size - len(buf))
+				DiffMatchPatch._proc.stdin.flush()
+				DiffMatchPatch._proc.stdout.flush()
 				return [
 					line
 					for line in buf.decode("utf-8").splitlines()


codeofdusk:
Cc @feerrenrut, @neurrone.

michaelDCurran (approved):

feerrenrut:
From https://stackoverflow.com/a/54063928 I expected that at each "synchronization boundary" between NVDA and the DMP process there should be a call to flush, in both processes. 

Eg if this code is sending some data and waiting for DMP to read it and respond, there should be a flush of std-out before the read. Otherwise you could end up with the DMP process waiting on extra data from NVDA, and NVDA waiting on data from DMP, therefore a deadlock.

feerrenrut:
So for instance after line 80 `DiffMatchPatch._proc.stdin.write(new)`

PR #12005, opened 2021-01-18
Remove blank lines in browseable message

### LINKED ISSUE #12004: Unexpected empty lines in browseable messages (e.g. formatting information)
### Steps to reproduce:

* set the cursor or virtual cursor on a text having formatting information such as text in a web page
* Press two times quickly NVDA+F to get formatting info in a browseable window
* Examine the formatting information line by line by pressing down arrow many times

Extra:
* Open the Python console and type:
  `import ui;ui.browseableMessage("Hello\nworld")
`

### Actual behavior:

NVDA reads "blank" one line in two in formatting info as well as in "Hello world" message.

### Expected behavior:

NVDA should not read "blank" between formatting information line nor in "Hello world" message since:
* As a user, we do not need blank lines to examine a text such as formatting information
* Visually, there is no blank line
* The source string of the "Hello world" message does not contain any blank line.

all
#### NVDA version:
2020.4beta4
#### Windows version:
10
N/A
N/A
Yes
Yes, 2020.3: same issue.
It seems to me that this issue has always been there.
Yes
No

CyrilleB79:
### Link to issue number:

Fixes #12004

### Summary of the issue:
When reading a text in a browseable message (such as formatting information) line by line pressing down arrow, there is one blank line between each line of text.

* Modified the HTML message template to use only `innerHTML` attribute and not `innerText` anymore.
* The raw text formatting is done in Python code by adding `<pre>` tag and escaping `<` character.

### Testing performed:

* Tested with formatting info message, with "Hello world" message in raw text.
* Tested HTML message with [Character information](https://addons.nvda-project.org/addons/charInfo.en.html) add-on.
* Tested long raw text (1000000 character) with [Clip Contents Designer](https://addons.nvda-project.org/addons/clipContentsDesigner.en.html) from @nvdaes and @abdel792.
Incidentally, this PR also fixes a visual issue in raw text browseable message containing lines longer than the window: moving the cursor on blank lines would pane the whole text to its right-most position, making the beginning of the lines disappear. This was visually strange. This is not the case anymore with this PR since the unexpected blank lines have been removed.

None.

### Change log entry:

Bug fixes!
`Formatting information and other browseable messages do not present unexpected blank lines anymore. (#12004)`

### DIFF
--- a/source/ui.py
+++ b/source/ui.py
@@ -57,7 +57,7 @@ def browseableMessage(message,title=None,isHtml=False):
 		title = _("NVDA Message")
 	if not isHtml:
 		message = f"<pre>{message.replace('<', '&lt;')}</pre>"
-	dialogString = u"true;{title};{message}".format(title=title, message=message)
+	dialogString = f"{title};{message}"
 	dialogArguements = automation.VARIANT( dialogString )
 	gui.mainFrame.prePopup() 
 	windll.mshtml.ShowHTMLDialogEx( 
--- a/source/ui.py
+++ b/source/ui.py
@@ -55,8 +55,9 @@ def browseableMessage(message,title=None,isHtml=False):
 	if not title:
 		# Translators: The title for the dialog used to present general NVDA messages in browse mode.
 		title = _("NVDA Message")
-	isHtmlArgument = "true" if isHtml else "false"
-	dialogString = u"{isHtml};{title};{message}".format( isHtml = isHtmlArgument , title=title , message=message ) 
+	if not isHtml:
+		message = f"<pre>{message.replace('<', '&lt;')}</pre>"
+	dialogString = u"true;{title};{message}".format(title=title, message=message)
 	dialogArguements = automation.VARIANT( dialogString )
 	gui.mainFrame.prePopup() 
 	windll.mshtml.ShowHTMLDialogEx( 
--- a/source/ui.py
+++ b/source/ui.py
@@ -15,6 +15,7 @@ import sys
 from ctypes import windll, byref, POINTER, addressof
 from comtypes import IUnknown
 from comtypes import automation 
+from html import escape
 from logHandler import log
 import gui
 import speech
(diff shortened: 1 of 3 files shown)


source/ui.py
```diff
@@ -55,8 +55,9 @@ def browseableMessage(message,title=None,isHtml=False):
 	if not title:
 		# Translators: The title for the dialog used to present general NVDA messages in browse mode.
 		title = _("NVDA Message")
-	isHtmlArgument = "true" if isHtml else "false"
-	dialogString = u"{isHtml};{title};{message}".format( isHtml = isHtmlArgument , title=title , message=message ) 
+	if not isHtml:
+		message = f"<pre>{message.replace('<', '&lt;')}</pre>"
```
michaelDCurran, line 59: Please use something like html.escape() to ensure that the text is totally safe to include within html. &lt; is not enough.

michaelDCurran (changes requested):

source/ui.py
CyrilleB79: Thanks for this suggestion.

According to the [French MDN page](https://developer.mozilla.org/fr/docs/Web/HTML/Element/pre):

> L'élément HTML <pre> représente du texte préformaté, généralement écrit avec une police à chasse fixe. Le texte est affiché tel quel, les espaces utilisés dans le document HTML seront retranscrits.
> Note : Il est nécessaire d'échapper les caractères '<' en '&lt;' afin de s'assurer que le code écrit entre les éléments ne soit pas interprété de façon involontaire.

Roughly translated (Google Translate) as follows:

> The HTML <pre> element represents preformatted text, typically written in a fixed-width font. The text is displayed as is, the spaces used in the HTML document will be transcribed.
> Note: It is necessary to escape the characters '<' in '<' to ensure that the code written between the elements is not unintentionally interpreted.

Thus escaping < character seems to be enough (confirmed by the tests).

However, the [English MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre) does not indicate this note.

Thus, for more safety, I will use `html.escape`anyway, as suggested by you.

CyrilleB79:
@michaelDCurran, this PR is ready. I have addressed your comments and performed again the teste of the initial description of this PR.

michaelDCurran (dismissed):

michaelDCurran (approved):

PR #12020, opened 2021-01-31
Python 3.8/liblouis helper: add NVDA executable path by calling os.add_dll_directory

### LINKED ISSUE #10416: Python 3.8: add NVDA source directory when attempting to import liblouis DLL
Hi,

A variation for #10385 

Python 3.8 adds os.add_dll_directory function that takes path string. This is done for security reasons, as Python's DLL path lookup routine has changed. This affects all sorts of DLL's, including NVDA Helper (see #10385) and for this one, liblouis.dll.

Use os.add_dll_directory function right before importing liblouis.dll in lous helper module.

Leave it as is, although it breaks Python 3.8 compatibility.

### Additional context
See #10384 for details.

josephsl:
Hi,

Required in Python 3.8:

### Link to issue number:
Fixes #10416

### Summary of the issue:
Python 3.8 adds os.add_dll_directory function to add a specified path to known DLL folders, used to load DLL's more securely. Without this call, liblouis.dll cannot be imported - instead, Python says "ModuleNotFoundError".

Calls os.add_dll_directory function (context manager form) from louisHelper.py to add NVDA executable path to known DLL's directory in order to import liblouis.dll.

### Testing performed:
Tested from source code version of Python 3.8 NVDA with no problems - braille input and output works as expected.

None

### Change log entry:
None

Thanks.

### DIFF
--- a/source/louisHelper.py
+++ b/source/louisHelper.py
@@ -9,7 +9,7 @@
 # Thus manually add NVDA executable path to DLL lookup path for loading liblouis.dll.
 import os
 import globalVars
-with os.add_dll_directory(os.path.dirname(globalVars.appDir)):
+with os.add_dll_directory(globalVars.appDir):
 	import louis
 from logHandler import log
 import config
--- a/source/louisHelper.py
+++ b/source/louisHelper.py
@@ -1,4 +1,3 @@
-# louisHelper.py
 # A part of NonVisual Desktop Access (NVDA)
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
@@ -9,7 +8,8 @@
 # Python 3.8 changes the way DLL's are loaded due to security.
 # Thus manually add NVDA executable path to DLL lookup path for loading liblouis.dll.
 import os
-with os.add_dll_directory(os.path.dirname(__file__)):
+import globalVars
+with os.add_dll_directory(os.path.dirname(globalVars.appDir)):
 	import louis
 from logHandler import log
 import config
--- a/source/louisHelper.py
+++ b/source/louisHelper.py
@@ -6,7 +6,11 @@
 
 """Helper module to ease communication to and from liblouis."""
 
-import louis
+# Python 3.8 changes the way DLL's are loaded due to security.
+# Thus manually add NVDA executable path to DLL lookup path for loading liblouis.dll.
+import os
+with os.add_dll_directory(os.path.dirname(__file__)):
+	import louis
 from logHandler import log
 import config
 
--- a/source/louisHelper.py
+++ b/source/louisHelper.py
@@ -1,8 +1,8 @@
-#louisHelper.py
-#A part of NonVisual Desktop Access (NVDA)
-#This file is covered by the GNU General Public License.
-#See the file COPYING for more details.
-#Copyright (C) 2018 NV Access Limited, Babbage B.V.
+# louisHelper.py
+# A part of NonVisual Desktop Access (NVDA)
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+# Copyright (C) 2018-2021 NV Access Limited, Babbage B.V., Joseph Lee
 
 """Helper module to ease communication to and from liblouis."""
 


josephsl:
Hi,

No need to worry about Nose test failure.

Thanks.

source/louisHelper.py
```diff
@@ -1,12 +1,16 @@
-#louisHelper.py
-#A part of NonVisual Desktop Access (NVDA)
-#This file is covered by the GNU General Public License.
-#See the file COPYING for more details.
-#Copyright (C) 2018 NV Access Limited, Babbage B.V.
+# louisHelper.py
```
lukaszgo1, line 1: Please remove this line

source/louisHelper.py
```diff
@@ -1,12 +1,16 @@
...
+# See the file COPYING for more details.
+# Copyright (C) 2018-2021 NV Access Limited, Babbage B.V., Joseph Lee
 
 """Helper module to ease communication to and from liblouis."""
 
-import louis
+# Python 3.8 changes the way DLL's are loaded due to security.
+# Thus manually add NVDA executable path to DLL lookup path for loading liblouis.dll.
+import os
+with os.add_dll_directory(os.path.dirname(__file__)):
```
lukaszgo1, line 12: `os.path.dirname(__file__)` is not gonna work for binary copies since it would point to nvda_install_dir\library.zip. Please use `globalVars.appDir` instead.

lukaszgo1 (changes requested):

josephsl:
Hi, forgot about globalVars.appDir. Thanks for pointing this out.

josephsl:
Hi,

Upon further testing, globalVars.appDir will not work on source code version but __file__ will. I'll find out if the binary copy is affected by this and come up with something if needed.

Thanks.

josephsl:
Hi,

Traceback from source version while using globalVars.appDir:

CRITICAL - __main__ (10:56:05.868) - MainThread (13492):
core failure
  File "nvda.pyw", line 247, in <module>
    core.main()
  File "core.py", line 268, in main
    import appModuleHandler
  File "appModuleHandler.py", line 27, in <module>
    import NVDAHelper
  File "NVDAHelper.py", line 26, in <module>
    import eventHandler
  File "eventHandler.py", line 12, in <module>
    import api
    import review
    from displayModel import DisplayModelTextInfo
  File "displayModel.py", line 17, in <module>
    import mouseHandler
  File "mouseHandler.py", line 8, in <module>
    import gui
  File "gui\__init__.py", line 25, in <module>
    import ui
  File "ui.py", line 22, in <module>
    import braille
  File "braille.py", line 20, in <module>
    import louisHelper
  File "louisHelper.py", line 13, in <module>
    import louis
  File "louis\__init__.py", line 61, in <module>
    liblouis = _loader["liblouis.dll"]
  File "C:\Python38\lib\ctypes\__init__.py", line 448, in __getitem__
    return getattr(self, name)
  File "C:\Python38\lib\ctypes\__init__.py", line 443, in __getattr__
    dll = self._dlltype(name)
  File "C:\Python38\lib\ctypes\__init__.py", line 373, in __init__
    self._handle = _dlopen(self._name, mode)
FileNotFoundError: Could not find module 'liblouis.dll' (or one of its dependencies). Try using the full path with constructor syntax.

This was after compiling NVDA from source (Python 3.8/SCons).

Thanks.

lukaszgo1:
That's quite strange. Have you compared `os.path.dirname(__file__)` with `globalVars.appDir` logically they should point to the same location?

josephsl:
Hi, nope it doesn’t. As you pointed out, using __file__ results in library.zip whereas globalVars.appDir results in the parent folder of the executable file e.g. if code is located in nvdapy38/source, globalVars.appDir will point to just nvdapy38 folder. Although ugly, one possible workaround is detecting source vs binary and add “source” to source code version. What’s more interesting is that liblouis.dll is loaded as part of braille input/output initialization later (and one can then work with the DLL from that point on), as the issue with louisHelper/add_dll_directory is encountered early in core.main. Thanks.

source/louisHelper.py
```diff
@@ -1,12 +1,16 @@
...
+# Copyright (C) 2018-2021 NV Access Limited, Babbage B.V., Joseph Lee
 
 """Helper module to ease communication to and from liblouis."""
 
-import louis
+# Python 3.8 changes the way DLL's are loaded due to security.
+# Thus manually add NVDA executable path to DLL lookup path for loading liblouis.dll.
+import os
+import globalVars
+with os.add_dll_directory(os.path.dirname(globalVars.appDir)):
```
lukaszgo1, line 12: I believe I've found the problem. There is no point in using  `(os.path.dirname` on `globalVars.appDir` as it represents path to the directory already. On my machine using `globalVars.appDir` alone allows source copy of NVDA to import liblouis and start correctly whereas when `(os.path.dirname` is used I'm getting the traceback you've posted yesterday.

michaelDCurran (approved):

PR #12023, opened 2021-02-01
System tests: run Chrome in kiosk mode to stop treeGrid test failing

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
Around 50% of the time, NVDA builds (including prs) fail on the ARIA treegrid system test. This seems to be because the test is needlessly pressing f6 when already on the document, most likely because of a delay in focus events caused by Chrome loading the content of an iframe (the treeGrid system test embeds a standard ARIA accessibility test in our normal test page using an iFrame).

The test framework now always starts Chrome in kiosk mode (meaning that the address bar and other parts of the chrome are hidden, leaving only the document content visible). The test framework no longer has to worry about pressing f6 a number of times to get to the document, rather it only needs to wait until NVDA lands on the "Before test" marker itself.

### Testing performed:
* Produced a try build twice with no system test failures.
* PR has been built multiple times now with no system test failures.

Before this pr, the system test never failed on my machine, but only on appveyor. This change so far seems to run fine on both appveyor and my machine, but only time will tell with appveyor. We have tried to fix this in the past. Let's hope this one works.
 
### Change log entry:
None.

### DIFF
--- a/tests/system/libraries/ChromeLib.py
+++ b/tests/system/libraries/ChromeLib.py
@@ -94,7 +94,7 @@ class ChromeLib:
 		return marker in speech and documentIndex < speech.index(marker)
 
 	def _waitForStartMarker(self, spy, lastSpeechIndex):
-		""" Wait until the page loads and NVDA reads the start marker. 
+		""" Wait until the page loads and NVDA reads the start marker.
 		@param spy:
 		@type spy: SystemTestSpy.speechSpyGlobalPlugin.NVDASpyLib
 		@return: None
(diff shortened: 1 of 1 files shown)


LeonarddeR (approved):

feerrenrut (approved):
I'm worried that this might just be fixing the symptom rather than the cause. It would be good to have a solid understanding of why the test doesn't detect the correct focus location. I wonder if we could more reliably replicate the failure in a local VM. 

I think it is important that we resolve these false failures now, but in the longer run this might just be hiding the problem. When / if we get to testing window chrome in browsers while need to face this.

PR #11079, opened 2020-05-01
Add advanced setting to allow UIA in chromium based browsers

LeonarddeR:
### Link to issue number:
Related to #10555 

### Summary of the issue:
UIA is now forcefully disabled in chrome based browsers. While this makes perfect sense, it could be beneficial for developers and/or development purpose to circumvent this without hacking.

### Description of how this pull request fixes the issue:
1. Adds an advanced setting that allows using UIA in Chromium based browsers, particularly in Edge. This option defaults to only when necessary, which is when UIA is exposed by the browser and we are unable to inject. If yes, UIA is used, but still only when UIA is exposed by the browser (not by default in Chrome). If no, there's just no browse mode when we can't inject.
2. Added NVDAObjects.UIA.web and NVDAObjects.UIA.chromium. The chromium module is meant to contain (Edge) Chromium specific code. The web module is meant to contain code shared between Edge Chromium and Edge Legacy. The Edge module contains Edge Legacy specific code. The code I moved to the web module was moved mainly by guessing what we'd need and what not, so some code in there can still be considered Edge legacy specific. It is very hard to assess this, though. I haven't considered renaming edge to edgeLegacy.

### Known issues (Todo)
- [ ] Update documentation
- [ ] Headings announcements are duplicated, probably because Edge Chromium both exposes them as control and format field. We'd have to decide what to use and whether the EdgeHeadingQuicknavIterator should be moved to the Web module.
- [ ] The makeTextInfo implementation of the Edge tree interceptor has been moved to the web tree interceptor, so is also used for Chromium. Without this, I"m getting lots of errors indicating that a range couldn't be cloned. This is either a bug in Edgium, or we need another workaround, as navigation in browse mode is still pretty buggy and bouncy.
- [ ] It looks like navigation by line is also broken, as I've seen several cases (e.g. at the top of babbage.com) where lines are broken into words.

### Testing performed:
* Tested in Edge that the config flag works.

### Change log entry:
t.b.d.

Issue #10555, opened 2019-11-28
Support UI Automation in Microsoft Edge Chromium and Google Chrome


codeofdusk:
### Is your feature request related to a problem? Please describe.
When starting Chrome with the `--enable-experimental-ui-automation` flag, browse mode is not functional.

### Describe the solution you'd like
Support for UIA Chrome, including browse mode and full functionality of all widgets.

### Describe alternatives you've considered
Continue to use the IA2 implementation and forcibly ignore UIA (similar to #7497) if UIA becomes no longer behind a feature flag. However, UIA would likely improve performance on large pages.

josephsl:
Hi, I think there is an issue on Chromium Edge/UIA, but I think we should use this one. CC @LeonardDer

leonardder:
I closed the other issue.

gdata1:
The native UI Automation is enabled by default in microsoft edge starting with version 80.0.355.0 (Official Build) canary.

leonardder:
Even in the stable build of edge, UIA now turns out to be the default.

gdata1:
I reported it to Microsoft. they said there should be UIA support on the nvda side. they also offer to turn it off manually, but not everyone knows how to do it. I don't think a lot of people will be able to turn it off in edge.

leonardder:
The problem is that NVDA doesn't create a tree interceptor to activate 
browse mode. This is not something on Microsofts end.

Adriani90:
In Edge Cromium 81 which is the last stable version, UIA is enabled by default and browse mode seems to work as expected.
@codeofdusk can you confirm?

codeofdusk:
On my machine (Chrome 83) with the `--enable-experimental-ui-automation` flag, NVDA now seems to force IAccessible 2.

josephsl:
Hi, this is because UIA will be ignored from document window itself. Other controls will be recognized as UIA if they are UIA elements. Thanks.

codeofdusk:
Makes sense, so this issue is still not resolved (we still need to implement support for UIA in Chrome).

Adriani90:
CC: @feerrenrut after our private discussion I think this is maybe nice to keep an eye on for the future.

leonardder:
Personally, I think a first pr should address two things:

1. Ad a flag to NVDA's advanced settings to enforce IA2 in Edge and Chrome, which is enabled by default
[...]

PR #12025, opened 2021-02-02
Add support for UIA with chromium based browsers 

feerrenrut:
This is based off of PR "Add advanced setting to allow UIA in chromium based browsers #11079"

However, due to the large amount of moved code, NV Access has decided not to squash this PR. Instead the commits have been cleaned up to make following the history as easy as possible. The changes from #11079 are as follows:
- Code is moved between files verbatim, in commits independent from any other changes. It should be noted that these commits will not run, but serve to make diffs / blame easier to follow.
- Linting is done in independent commits. This allows logic changes to be differentiated from "clean up"
- UIA/edge becomes UIA/spartan_edge and a placeholder is added for UIA/anaheim_edge

This PR supersedes and thus closes #11079, however the history there is still relevant and useful.
 
### Link to issue number:
Related to #10555

### Summary of the issue:
Chromium based browsers now contain a UIA implementation which may be exposed (E.G. in Edge Anaheim). Allowing this to be used will allow developers to work to stabilize and mature this technology. For users on systems where process injection is not possible, access to UIA in Edge can act as a fallback option.

Provides an advanced setting to use UIA when available. Moves significant portions of the UIA code for reuse:
- Web
- Legacy Edge (Spartan)
- Chromium based browsers such as Edge Anaheim

### Testing performed:
Manual testing.

None

### Change log entry:

New features
```
- Early support for UIA with Chromium based browsers (such as Edge).
```

### DIFF
--- a/source/NVDAObjects/__init__.py
+++ b/source/NVDAObjects/__init__.py
@@ -417,10 +417,12 @@ class NVDAObject(documentBase.TextContainerObject, baseObject.ScriptableObject, 
 		"""
 		return ""
 
-	def _get_role(self):
+	#: Type definition for auto prop '_get_role'
+	role: int
+
+	def _get_role(self) -> int:
 		"""The role or type of control this object represents (example: button, list, dialog).
 		@return: a ROLE_* constant from L{controlTypes}
-		@rtype: int
 		"""  
 		return controlTypes.ROLE_UNKNOWN
 
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -903,6 +903,17 @@ class UIA(Window):
 				clsList.append(spartanEdge.EdgeList)
 			else:
 				clsList.append(spartanEdge.EdgeNode)
+		elif self.windowClassName == "Chrome_RenderWidgetHostHWND":
+			from . import chromium
+			if (
+				self.UIATextPattern
+				and self.role == controlTypes.ROLE_DOCUMENT
+				and self.parent
+				and self.parent.role == controlTypes.ROLE_PANE
+			):
+				clsList.append(chromium.ChromiumUIADocument)
+			else:
+				clsList.append(chromium.ChromiumUIA)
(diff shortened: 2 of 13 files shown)


leonardder:
I'm slightly worried about introducing underscores in the names here, as it is not following the naming conventions we usually use.

michaelDCurran:
True we have not used underscores before, however, I think this is a 
suitable way to very clearly denote very different implementations of 
Edge. Especially when they are living in the same directory.

michaelDCurran (dismissed):
I have done the same manual testing I did on the original pr and all was good for modern Edge.
I hit an issue in classic edge, but this was actually introduced by pr #11969 so I have separately opened pr #12026 to address that. But other than that things seemed to be working well.

feerrenrut:
> very clearly denote very different implementations of Edge. Especially when they are living in the same directory.

Hmm perhaps there should be more structure here eg "namespaces". Consider if we moved the files:
```
- UIA
  |- New Package: Web
  |   |- New Package: Chromium
  |   |   |- EdgeAnaheim.py
  |   |- EdgeSpartan.py
```

michaelDCurran:
I'm okay with either way. Don't think the further directory structure is 
necessary work, but fine if it gets done.

source/NVDAObjects/__init__.py
```diff
@@ -417,6 +417,9 @@ def _get_name(self):
 		"""
 		return ""
 
+	#: Type definition for auto prop '_get_role'
+	role: str
```
lukaszgo1, line 421: I don't think this annotation is correct since roles are represented by `ROLE_` constants in `controlTypes` and they are integers.

feerrenrut: Yes you are right @lukaszgo1, good catch.

feerrenrut:
I've been looking at the layout of this code and reconsidering whether a nested structure is better. I think it is better for this code to stick with a flatter structure, otherwise there are dependencies going in the wrong direction:
- chromium depends on Web
- Web depends on UIA
- spartan_edge depends on UIA
Nesting these as packages requires requires multiple level relative imports or absolute imports.

I will however rename spartan_edge and anaheim _edge to spartanEdge and anaheimEdge respectively.

PR #12026, opened 2021-02-02
Fix up of #11969: correct logic for heading levels in getFormatfieldSpeech

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
In pr #11969 REASON_* constants were replaced with OutputReason.* enum constants.
However, a logic error was introduced in speech.getFormatFieldSpeech which caused NVDA to no longer  be able to read content in edgeHTML (Edge classic) documents. Causing the following exception:
```
File "speech\__init__.py", line 2033, in getFormatFieldSpeech
    text=_("heading level %d")%headingLevel
TypeError: %d format: a number is required, not NoneType
```

This is due to the 
```
headingLevel != oldHeadingLevel
  ```
being moved out of its original brackets and into a new set of outer brackets.

Moves the code back into the original place, maintaining linting rules.

### Testing performed:
Did a search in the start screen, and arrowed down into the extra content (which is an Edge classic document). Previusly this caused an exception. Now all the content reads again.

None.

### Change log entry:
None.

### DIFF
--- a/source/speech/__init__.py
+++ b/source/speech/__init__.py
@@ -2026,8 +2026,8 @@ def getFormatFieldSpeech(  # noqa: C901
 			and (
 				initialFormat
 				and (reason == OutputReason.FOCUS or unit in (textInfos.UNIT_LINE, textInfos.UNIT_PARAGRAPH))
+				or headingLevel != oldHeadingLevel
 			)
-			or headingLevel != oldHeadingLevel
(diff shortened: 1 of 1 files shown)


michaelDCurran:
@lukaszgo1

lukaszgo1 (approved):
Looks fine to me.

PR #12027, opened 2021-02-03
add styles for userguide and changes

feerrenrut:
### Link to issue number:
None

### Summary of the issue:
The user guide has long since been missing any basic styling. While very utilitarian, the default appearance will likely be perceived by many as ugly and potentially hard to read. 

Add simple style rules for the user guide, changes file, and key commands

### Testing performed:
Tested opening these from NVDA.

None

### Change log entry:
Section: Changes
```
- The user guide, changes file, and key commands listing now have a refreshed appearance.
```


feerrenrut:
Note: I haven't updated the appveyor file to package the css with the docs. For reviewing a build I don't think this is necessary. @michaelDCurran could you comment on what we would need to do to get the css file to accompanying the userguide / changes file to the website?

@Qchristensen, here is a demo:
[styledUserGuide.zip](https://github.com/nvaccess/nvda/files/5915681/styledUserGuide.zip)

michaelDCurran:
Although styles.css was being copied to each language directory, it also needs to be copied to the output directory. Thus I just pushed a commit that ensures that styles.css is copied to the output directory if the userGuide or changes targets are requested to be build with scons.
As for getting styles.css included when we actually publish a snapshot or release: this will be a two-line change to our appveyorHook internally. But can't be done until alpha/beta snapshots start producing these with this pr.
Either this pr must be held back until NVDA 2020.4 is fully released, or this should go into beta.

feerrenrut:
I'm not sure I follow, can we merge this and just delay updating the appveyorHook until after 2020.4 is released?

michaelDCurran:
You're right, there is no disadvantage to merging this to master at any 
time. I thought the snapshots page linked to a changes.html file, but 
these days it just links to a github git commit log, so that is fine.

But once this does ever hit the beta branch, the appVeyorHook must be 
updated before any beta, rc or final release is made.

feerrenrut:
Ok, in that case I'll wait for @Qchristensen to review / approve the style choices and we can proceed.

Qchristensen (dismissed):
Great work.  I agree, no urgency with merging it.

Qchristensen:
I didn't comment on the style, but I like the use of the purple.  My only thoughts were

- For tables if we could shade every second row might help reading visually.
- Is it worth purring a HR or something to more clearly denote the change between topics?

I thought maybe changes like that which weren't critical could be done later down the track.

feerrenrut:
Thanks for taking a look @Qchristensen, yes those changes could be made fairly easily after this is merged. The other thing I was hoping we could do was to delineate shortcuts clearly.

I'll merge this in now. It should be visible when docs are launched from NVDA, but won't become available on our website until 2021.1 reaches beta.

PR #12034, opened 2021-02-04
Fixes somes context help issues in advanced settings panel.

CyrilleB79:
### Link to issue number:
None

### Summary of the issue:

In #12025 (superseding #11079) and in #11639, new options were added to the advanced settings panel and a dedicated paragraph was added to the user doc corresponding to each of them.
However, context help does not work for these options. I.e. when the focus is on one of this option, pressing F1 reports "No help available here".
This is due to 2 issues:
* No help event was bound to these controls. Thus F1 does not opens the doc on the dedicated paragraph
* No helpId was assigned to AdvancedPanelControls subpanel. Thus the default behaviour when the controls do not have associated help events (i.e. opening the doc on the advanced settings paragraph) was not executed either.

1. Bound both controls to the corresponding help events
2. Added helpId to AdvancedPanelControls subpanel.
3. Rename the tag "ChromiumUIA" to "AdvancedSettingsChromiumUIA" to conform to the majority of other help tags formatting, I.e. combining panel name and option name. This is still possible because this option has not been released yet. On the contrary it is not recommended to change tag names once the doc was included in a stable release of NVDA.

### Testing performed:

* Tested point 2 during implementation when point 1 and 3 were not yet implemented to check that pressing F1 on these 2 options jumps to the main "Advanced settings" paragraph.
* Once the three points were implemented, tested that F1 opens the user guide at the appropriate paragraph.

None

### Change log entry:

None: this PR only fixes issues that have not yet been released.

### DIFF
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -2584,7 +2584,7 @@ class AdvancedPanelControls(
 			pgettext("advanced.uiaWithChromium", "No"),
 		)
 		self.UIAInChromiumCombo = UIAGroup.addLabeledControl(label, wx.Choice, choices=chromiumChoices)
-		self.bindHelpEvent("AdvancedSettingsChromiumUIA", self.UIAInChromiumCombo)
+		self.bindHelpEvent("ChromiumUIA", self.UIAInChromiumCombo)
 		self.UIAInChromiumCombo.SetSelection(config.conf["UIA"]["allowInChromium"])
 		self.UIAInChromiumCombo.defaultValue = self._getDefaultValue(["UIA", "allowInChromium"])
 
@@ -2632,7 +2632,7 @@ class AdvancedPanelControls(
 			"difflib"
 		)
 		self.diffAlgoCombo = terminalsGroup.addLabeledControl(diffAlgoComboText, wx.Choice, choices=diffAlgoChoices)
-		self.bindHelpEvent("AdvancedSettingsDiffAlgo", self.diffAlgoCombo)
+		self.bindHelpEvent("DiffAlgo", self.diffAlgoCombo)
 		curChoice = self.diffAlgoVals.index(
 			config.conf['terminals']['diffAlgo']
 		)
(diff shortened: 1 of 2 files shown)


CyrilleB79:
Cc @feerrenrut, @leonardder and @codeofdusk as authors of initial PR's and/or potential reviewers of future PRs.

In the future, please do not forget context help when new controls are added in the GUI.

feerrenrut:
Thanks @CyrilleB79, I have created a code review checklist to remind us of this and other similar forgotten tasks: https://github.com/nvaccess/nvda/pull/12037

feerrenrut:
>Rename the tag "ChromiumUIA" to "AdvancedSettingsChromiumUIA" to conform to the majority of other help tags formatting, I.e. combining panel name and option name. This is still possible because this option has not been released yet. On the contrary it is not recommended to change tag names once the doc was included in a stable release of NVDA.

Hmm, while it's a loss of consistency I think it is better not to include the category in the tag. Since we can't / don't want to change the name of these tags, then moving settings to new sections will result in a very confusing tag. It would be more sensible if the tag does not include the section name.  The main reason I can see for wanting to include the section name is to ensure that the tag it is unique. Instead of a tag naming scheme, I'd propose an automated test. There isn't much we can do about the ones already in place, however I would like to also note that some of the "advanced settings" are quite likely to eventually graduate to another settings panel.

feerrenrut:
Otherwise the changes look good.

CyrilleB79:
> Hmm, while it's a loss of consistency I think it is better not to include the category in the tag. Since we can't / don't want to change the name of these tags, then moving settings to new sections will result in a very confusing tag. It would be more sensible if the tag does not include the section name. The main reason I can see for wanting to include the section name is to ensure that the tag it is unique. Instead of a tag naming scheme, I'd propose an automated test. There isn't much we can do about the ones already in place, however I would like to also note that some of the "advanced settings" are quite likely to eventually graduate to another settings panel.

Yes, this makes sense. I have removed the category name from the tags of these 2 options that have not yet been included in any release.
Also re-tested the help link (F1) on the updated branch.

feerrenrut (approved):
Thanks @CyrilleB79

PR #12036, opened 2021-02-04
Fix usages of some remaining Python 2 names

lukaszgo1:
### Link to issue number:
None

### Summary of the issue:
During migration from Python 2 to 3 two usages of Python 2 only names in the source code were missed.
- In NVDACOntroller client Python example replaced `xrange` with `range`
- in `locationHelper` removed check for `long` as it no longer exists in Python 3.

In addition since some methods in `locationHelper` used to create location classes from various types aren't used in the code at all I've expanded unit tests to ensure that they raise expected exceptions when incompatible arguments are given to them.

### Testing performed:
Newly added unit tests for locationHelper are passing now whereas previously they weren't.
None
### Change log entry:
None needed.


feerrenrut (approved):
Thanks @lukaszgo1

PR #12072, opened 2021-02-14
Py3.8: allow NVDA scripts to again run when in a wx popup menu or message box.

### LINKED ISSUE #12058: Task: Fix commands don't work in dialogs or context menu's with wxPython 4.1.1
As reported on issue Upgrade wxPython to 4.1.1 #11056. After updating to wxPython 4.1.1 NVDA commands no longer work when the NVDA menu (a context menu) or a dialog is open. It appears that the main thread can no longer be woken by the background thread using wx. This can be confirmed by executing some commands, then pressing a key that is just passed through eg left arrow, the commands will now run.

Instead we will need to implement an alternative mechanism to wake the main thread. One potential solution would be to look for an appropriate Windows message to send to signal that there is work to be done.

michaelDCurran:
### Link to issue number:
Fixes #12058 

### Summary of the issue:
If NVDA is run with wxPython 4.1.1, NVDA-specific  key commands (scripts) fail to execute while in a WX popup menu (such as the NVDA menu) or wx message box (such as the NVDA About dialog). Yet, key commands that are not trapped by NVDA (E.g. pressing downArrow in a menu) still work okay.
  In recent versions of wxWidgets, if in a menu or message box, calling wxWidget's CallAfter function does not execute the given callback until the menu or message box is dismissed (I.e. the wx main loop is again running). There are some exceptions to this, such as if a key or other input is sent to the application. But of course in the case of trying to execute an NVDA-specific command, NVDA blocks the key from getting to the application and therefore the wx event loop never wakes up.
In wxWidgets, it looks like CallAfter calls QueueEvent, which calls WakeupIdle.
Further investigation shows that queuing any kind of wx event (command, idle etc) does not wake up the wx event loop while in popup menus or message boxes.

Monkeypatch wx.CallAfter to post a WM_NULL message to our top-level window after calling the original CallAfter, which causes wx's event loop to wake up enough to execute the callback.

### Testing strategy:
Note: this pr (nor the py3.8 branch at time of this comment) does not yet specifically use wxPython 4.1.1, thus it is necessary for testing that you first:
* cd to include/wx
* Delete everything in this directory
* Fetch wxPython 4.1.1 into this directory with the command: py -m pip install -t . wxPython==4.1.1
This pr has so far been tested manually by:
* Starting NVDA
* Executing the ReportTime NVDA script by pressing NVDA+f12 and noting that NVDA speaks the time.
* Opening the NVDA menu with NVDA+n
* Again executing reportTime script with NVDA+f12 and noting that NVDA speaks the time. previous to this pr, if using wxPython 4.1.1, the command would fail to execute at all.
* Open NVDA menu -> Help -> About...
* Again execute reportTime script with NVDA+f12 and note that NVDA speaks the time. 
Unit tests do not apply to this pr as the issue is within wxPython.
System testing is certainly possible here. But only once wxPython 4.1.1 is in NVDA.

None.

### Change log entry:
None needed.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### DIFF
--- a/source/gui/__init__.py
+++ b/source/gui/__init__.py
@@ -597,6 +597,21 @@ def initialize():
 		raise RuntimeError("GUI already initialized")
 	mainFrame = MainFrame()
 	wx.GetApp().SetTopWindow(mainFrame)
+	# In wxPython >= 4.1,
+	# wx.CallAfter no longer executes callbacks while NVDA's main thread is within apopup menu or message box.
+	# To work around this,
+	# Monkeypatch wx.CallAfter to
+	# post a WM_NULL message to our top-level window after calling the original CallAfter,
+	# which causes wx's event loop to wake up enough to execute the callback.
+	old_wx_CallAfter = wx.CallAfter
+
+	def wx_CallAfter_wrapper(func, *args, **kwargs):
+		old_wx_CallAfter(func, *args, **kwargs)
+		# mainFrame may be None as NVDA could be terminating.
+		topHandle = mainFrame.Handle if mainFrame else None
+		if topHandle:
+			winUser.PostMessage(topHandle, winUser.WM_NULL, 0, 0)
+	wx.CallAfter = wx_CallAfter_wrapper
 
 def terminate():
 	import brailleViewer


josephsl (approved):
Thanks- tested manually myself and works, too.

leonardder:
Is this change in WX actually a bug? How did you find out this workaround?

michaelDCurran:
I just found it out through trial and error.

I guess it could be seen as a wx bug, though I doubt it was designed 
specifically to handle the way we use it in NVDA, even if it did work 
previously. I.e. running arbitrary code (including other wx events) 
within a popup menu or message box (which usually would be classed as 
blocking) is not something that most apps do -- though very common in a 
screen reader.

feerrenrut (approved):
Looks fine to me. 

I've been through the checklist, they can all be ticked. 

Could you please raise an issue on the wxPython issue tracker? This will help us to remove the workaround if / when the issue is fixed.

PR #12075, opened 2021-02-15
Py3.8: Transparently use a Python virtual environment under the hood

### LINKED ISSUE #11056: Upgrade wxPython to 4.1.1
Hi,

Update: originally 4.1.0, but wxPython 4.1.1 was released on November 25, 2020, hence the change of title and content. Unless requested, subsequent comments below refer to 4.1.0.

This issue serves as both a meta issue and a concrete one.

wxPython 4.1.0 was released on April 24, 2020, with 4.1.1 released about seven months later (November 2020). Besides internal changes to use wxWidgets 3.1.x (master branch), wxPython 4.1.1 includes the following:

* More useful notification message handling.
* Deprecated code paths.
* Python 3.8 compatibility.
* Others to be discovered.

As updating to Python 3.8 and future 3.x releases require dependency checks, wxPython 4.1.1 fulfills the first dependency check (the other major ones include SCons and Py2exe unless a much better packaging tool is found; update: Py2exe supports Python 3.8 and 3.9 as of November 2020).

Upgrade to wxPython 4.1.1.

None.

#### Blockers:
NVDA's own commands cannot be performed if NVDA menu is open. Also. NVDA may not appear to respond when a message dialog opens.

#### Past or mitigation possible:
There exists an issue in wxPython 4.1.0 where if one calls wx.Locale().Init, it changes locale and encoder seen by Python, which causes text processing to fail. The issue was traced to two areas: wxPython calling wx.App.InitLocale when an app is instantiated (can be mitigated by defining wx.App.OnPreInit), and InitLocale itself which tells wxWidgets to set the CPython locale to be in sync with wxWidgets locale.

Thanks.

### LINKED ISSUE #11794: Update to Py2exe 0.10.0.2

### LINKED ISSUE #12059: Task:  Add docs for manual virtual env process
To ensure we understand the process, and to provide information in case any aspect fails for some developer provide docs explaining the process.
Docs should include:
- Creating virtual env.
- How to populate with dependencies.
- What to do when deps change.
- Where packages for deps are cached.

In order to write these docs, the approach will need to be proved and tested locally.

### LINKED ISSUE #12060: Task: Update sconsscript to exit early when not in virtual environment
We don't want developers accidentally building NVDA in an unknown environment. Get the build system to check and exit with an error which points them to the docs.

### LINKED ISSUE #12061: Task: Create scripts to automate venv creation
Several python scripts will need to be made to automate the manual process of creating a virtual env with a known location, installing the dependencies, checking dependencies installed match the requirements file, and removing the virtual environment.

This allows us to have full programmatic control over the virtual environment.
Because pip is quite slow, we can detect that dependencies have change simply by caching / copying the requirements file during install. If the files don't match, the deps have changed since the last installation of dependencies. Developers will need to be able to downgrade dependences (Eg when checking out beta after having master), pip doesn't support this well. To support this in a simple and reliable way, we can just delete the virtual environment, and re-install the dependencies from the (updated) requirements file. Because the packages are cached, this should be a fairly quick operation.

For now these can be left for developers to run manually. Point to these from the docs.

### LINKED ISSUE #12062: Task: Create a "runnvda" script to activate venv, run nvda, then deactivate venv
If the virtual environment is already activated (the developer may actively be using it) then don't try to activate or deactivate.

This will likely be served by several python scripts. These should be documented.

### LINKED ISSUE #12063: Task: Update scons.bat to run venv creation scripts before build
After completing:
- [Task: Create scripts to automate venv creation #12061](https://github.com/nvaccess/nvda/issues/12061)
- [Task: Create a "runnvda" script to activate venv, run nvda, then deactivate venv #12062](https://github.com/nvaccess/nvda/issues/12062)

Update scons.bat to detect if a new virtual environment needs to be created (missing, or dependencies out of date), then create the virtual environment and install dependencies. It should also enable the virtual environment before building.

### LINKED ISSUE #12065: Task: update readme and any other docs
Readme can likely have the extra dependencies section removed or replaced with a note to see requirements.txt. If referring to requirements.txt make it clear this is automatically setup by the build system.

Ensure that the docs for the virtual env / dependencies are up to date. In both readme and other docs explicitly mention that cmd.exe is the only officially supported way to build.

### LINKED ISSUE #12116: wxPython 4.1.1: Binary versions of NVDA crashes on Windows 8.1 and earlier.
### Steps to reproduce:
1. Download a try build from PR #12075 
2. Try to execute downloaded launcher on Windows 8.1, Windows Server 2012 or Windows 7 SP1.

### Actual behavior:
NVDA startup sound plays and the Windows display a dialog informing that NVDA Aplication has stopped working.
[Dump file from NVDA](https://drive.google.com/file/d/1crN7drWPOojmZPrN_nFnSY0dluFygPhc/view?usp=sharing)
[nvda-old.log](https://github.com/nvaccess/nvda/files/6083188/nvda-old.log)

### Expected behavior:
NVDA should work normally.

Running from launcher and portable - no problems when running from source on the same machines.

#### NVDA version:
nvda_snapshot_try-py3.8_venv-21812,b834a4e6

#### Windows version:
Windows 7 SP1, Windows Server 2012, Windows 8.1
None
Yes
Versions using WxPython 4.0.3 works okay.

Yes -- tested with a clean config

N/A

# #Additional info:
I've finally tracked it down to two things:
1. WXPython 4.1.1 - if I build the launcher with wxPython 4.1.0 there is no crash.
2. Version of Universal CRT bundled in the launcher - if I don't bundle these libraries or bundle different versions of them (see below) NVDA starts properly. Note that it does not matter if the machine in question has Universal CRT installed or not since according to the [Microsoft help page about dll search order](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-desktop-applications) directory of the current application always has precedence over dll's bundled with Windows.

## Possible solutions:
Ideally we should get to the bottom of this problem and understand why wxPython 4.1.1 causes a crash with the version of the Universal CRT that we bundle. I don't know if the dump above is sufficient to debug further - my  experience with using a debugger is almost non existent it'd be great if someone more experienced could take a look at it. If this proves not to be easily fixable I see a few possible solutions:
### Use different version of the Universal CRT
I've discovered that using Universal CRt libraries from Windows 10 SDK 18362 or later works around this issue. This is in contrast which what Microsoft recommends for applications which need to bundle these libraries but it seems to work. Since I do not really understand why it works it is hard to recommend it as a solution.
### Don't bundle Universal CRt with NVDA.
This also solves the problem but causes NVDA not to work on machines without Universal CRt i.e. clean installs of Windows so it also isn't very appealing.
### Use wxPython  4.1.0
As described above this crash does not occur with 4.1.0 which is compatible with Python 3.8.

michaelDCurran:
### Link to issue number:
Fixes #11056 
Fixes #11794 
Fixes #12059 
Fixes #12060 
Fixes #12061 
Fixes #12062 
Fixes #12063 
Fixes #12065 
Fixes #12116 

### Summary of the issue:
NVDA and its build system have has many Python dependencies, the majority of which are fetched as git submodules to official git repositories, some are git submodules for our own git repositories containing pre-built dependencies (E.g. `wxPython`) and some are fetched via `pip`. And even the ones via `pip` are fetched differ as to where they are fetched into our build system.
 
The current approach to Python dependencies as the following issues:
* There is no standard way of tracking / fetching Python dependencies
* For certain dependencies such as `wxPython`, there is a great deal of extra work to pre-build and store in a git repository each time we want to upgrade to a new version of that dependency
* Python dependencies installed within our build system may conflict with Python packages already installed globally on the developer's system
* Python dependencies for NVDA may be inappropriately installed globally on the developer's system, which could affect things outside of NVDA.

As we move to `Python 3.8`, some of our Python dependencies (namely `wxPython`, `pySerial` and `py2exe`) need to be upgraded. We should take this opportunity to standardize on a cleaner way of fetching Python dependencies that address all the above issues. 

A Python virtual environment is now used transparently by the NVDA build system, and all Python dependencies are installed into this environment using pip.
From a developer's perspective using the NVDA build system, there is no need to worry about the Python virtual environment behind the scenes, nor should you create and or activate one manually. All NVDA build system commands will handle this transparently.
* To build NVDA, SCons should continue to be used in the usual way. E.g. executing scons.bat in the root of the repository. Running py -m SCons is no longer supported, and scons.py has also been removed.
* To run NVDA from source, rather than executing source/nvda.pyw directly, the developer should now use runnvda.bat in the root of the repository. If you do try to execute source/nva.pyw, a message box will alert you this is no longer supported.
* To perform unit tests, execute rununittests.bat [<extra unittest discover options>]
* To perform system tests: execute runsystemtests.bat [<extra robot options>]
* To perform linting, execute runlint.bat <base branch>
Behind the scenes, the above batch files (scons, runnvda, rununittests, runsystemtests and runlint) ensure that the Python virtual environment is created and up to date, activates the environment, runs the command and then deactivates. All transparently. A developer should not have to know about the Python virtual environment at all.
The first time one of these commands are run, the virtual environment will be created, and all required Python dependencies will be installed with pip. You can see the entire list of packages and their exact versions that pip will use, in requirements.txt in the root of the repository.
venvUtils/ensureVenv.py contains the logic to check for, create and update the virtual environment.
If a previous virtual environment exists but has a miss-matching Python version or pip package requirements have changed, The virtual environment is recreated with the updated version of Python and packages.
If a virtual environment is found but does not seem to be ours, the user is asked if it should be overwritten or not.
This script also detects if it is being run from an existing 3rd party Python virtual environment and aborts if so. thus, it is impossible to execute SCons or NVDA from source within another Python virtual environment.
venvUtils/ensureAndActivate.bat can be used to ensure the virtual environment exists and is up to date, and then activates it in the current shell, ready for other commands to be executed in the context of NVDA's build system Python virtual environment. this would never normally be executed by itself, though appVeyor uses it at the beginning of its execution and leaves the environment active for the remainder of its execution.
venvUtils/venvCmd.bat is a script that runs a single command within the context of the NVDA build system Python virtual environment. It ensures and activates the environment, executes the command, and then deactivates the environment. this script is what all the high-level batch files use internally.
SConstruct, and source/nvda.pyw both contain logic that detects the NVDA build system Python virtual environment, and abort if it is not active.
As this PR is specific for Python 3.8, it also upgrades the following Python dependencies:
* wxPython to 4.1.1
* pySerial to 3.5
* py2exe to 0.10.1.0

### Testing strategy:
* Compiled NVDA locally with `scons source`, `scons dist` and `scons launcher`
* Run unit tests and system tests locally with no errors.
* Successfully generated try builds on appVeyor.

None so far.

### Change log entry:
Changes for developers:
```
- NVDA now requires Python 3.8.
- NVDA's build system now fetches all Python dependencies with pip and stores them in a Python virtual environment. This is all done transparently.
* To build NVDA, SCons should continue to be used in the usual way. E.g. executing scons.bat in the root of the repository. Running ``py -m SCons``  is no longer supported, and ``scons.py`` has also been removed. 
* To run NVDA from source, rather than executing ``source/nvda.pyw`` directly, the developer should now use ``runnvda.bat`` in the root of the repository. If you do try to execute ``source/nvda.pyw``, a message box will alert you this is no longer supported.
* To perform unit tests, execute ``rununittests.bat [<extra unittest discover options>]``
* To perform system tests: execute ``runsystemtests.bat [<extra robot options>] ``
* To perform linting, execute ``runlint.bat <base branch>``
```

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.


appveyor.yml
```diff
@@ -59,6 +59,16 @@ init:
 clone_depth: 1
 
 install:
+ # Ensure we have the latest version of pip
+ - py -m pip install --upgrade pip
+ # Create a Python virtual environment
+ - py -3.8-32 -m venv .venv
```
lukaszgo1, line 65: Since `PY_PYTHON` is being set to 3.8 x32 above you can just use `py` here.

requirements.txt
```diff
@@ -0,0 +1,26 @@
+# NVDA's build system is SCons
+SCons==4.1.0.post1
+
+# NVDA's runtime dependencies
+comtypes==1.1.8
+pyserial==3.5
+wxPython==4.1.1
```
lukaszgo1, line 7: Installing wx like that will also pull NumPy from PyPI. As mentioned by @josephsl in #11967 if NumPy is installed Py2exe bundles parts of in in NVDA's binary builds because ComTypes has an optional dependency on NumPy arrays. NumPy is also a dependency of screencaplibrary. Perhaps we need to be able to specify `--no-deps` for a particular package?

lukaszgo1 (commented):
- Please add .venv to the .gitignore.
- Why exactly you've deleted `scons.bat` and `sconns.py` when we would look into automating this process they probably would need to be reintroduced in some form.

readme.md
```diff
@@ -119,14 +112,30 @@ Although this [must be run manually](#linting-your-changes), developers may wish
...
+```
+However, it is strongly recommended that you create a Python virtual environment for this repository, and activate it before installing required python packages and or building or running NVDA. This avoids any Python package version conflicts with packages you may have installed system-wide.
+To create a Python virtual environment in the .venv directory:
+```
+py -m venv .venv
+```
+To activate the virtual environment:
+```
+.venv\scripts\activate.bat
+```
```
MarcoZehe, line 135: Actually, I would switch the paragraphs around and simply instruct users to first create a virtual environment and then run the pip command. Since this is what AppVeyor will also do, and it is good practice anyway, why bother with howevers and maybes? Just my 2 Cents.

requirements.txt
feerrenrut: NumPy will be excluded from the Py2exe bundle explicitly, see `source/setup.py`. I think NumPy is required indirectly by robotframework-screencaplibrary anyway.

readme.md
feerrenrut: I agree. I think we should move all the "doing a local build" stuff into a dedicated file or at least section. It should be ordered so it can be followed step by step.

appveyor.yml
```diff
@@ -59,6 +59,12 @@ init:
 clone_depth: 1
 
 install:
+ # Ensure we have the latest version of pip
+ - py -m pip install --upgrade pip
```
lukaszgo1, line 63: Upgrading `pip` here does not prevent warnings about it being outdated during packages installation because when creating an venv a predefined version of `pip` is used rather than one currently installed. It makes sense to update it after our venv is activated.

lukaszgo1:
Also for  whatever reason NVDA fails to exit properly after install on AppVeyor with the following log:

```

CRITICAL - watchdog._crashHandler (10:38:49.715) - MainThread (4732):
NVDA crashed! Minidump written to C:\projects\nvda\testOutput\nvda_install.log\..\nvda_crash.dmp
INFO - watchdog._crashHandler (10:38:49.715) - MainThread (4732):
Python stack for thread 5020 (watchdog):
  File "threading.pyc", line 932, in _bootstrap_inner
  File "winKernel.pyc", line 228, in waitForSingleObject

Python stack for thread 5056 (winInputHook):
  File "threading.pyc", line 932, in _bootstrap_inner

Python stack for thread 3044 (_UIAHandler.UIAHandler.MTAThread):
  File "threading.pyc", line 932, in _bootstrap_inner
  File "_UIAHandler.pyc", line 329, in MTAThreadFunc
  File "threading.pyc", line 302, in wait

Python stack for thread 1068 (braille._BgThread):
  File "threading.pyc", line 932, in _bootstrap_inner
  File "braille.pyc", line 2193, in func

Python stack for thread 4732 (MainThread):
  File "nvda.pyw", line 261, in <module>
  File "core.pyc", line 576, in main

INFO - watchdog._crashHandler (10:38:49.715) - MainThread (4732):
Restarting due to crash

```

source/core.py
```diff
@@ -560,6 +560,15 @@ def run(self):
 	else:
 		log.debug("initializing updateCheck")
 		updateCheck.initialize()
+	# If running from source, try to disconnect from the console we may have been executed in.
+	# NVDA may reconnect to read it later,
+	# but it is better to assume we are not connected to anything at the start.
+	try:
+		ctypes.windll.kernel32.FreeConsole()
```
lukaszgo1, line 567: This code assumes that `FreeConsole` would raise a `WindowsError` if  NVDA wasn't connected to the console but it seems not to be the case. According to the documentation this function returns true if the console has been fried and False if `FreeConsole` fails or there was no connected console in the first place.

tests/system/libraries/NvdaLib.py
```diff
@@ -51,7 +51,7 @@ def __init__(self):
 		self.whichNVDA = builtIn.get_variable_value("${whichNVDA}", "source")
 		if self.whichNVDA == "source":
 			self._runNVDAFilePath = _pJoin(self.repoRoot, "source/nvda.pyw")
-			self.baseNVDACommandline = f"pyw -3.7-32 {self._runNVDAFilePath}"
+			self.baseNVDACommandline = f"pyw {self._runNVDAFilePath}"
```
lukaszgo1, line 54: Can't you just use runnvda.bat here?

venvUtils/ensureVenv.py
```diff
@@ -0,0 +1,129 @@
...
+
+def createVenvAndPopulate():
+	"""
+	Creates the NVDA build system's Python virtual environment and installs all required packages.
+	this function will overwrite any existing virtual environment found at c{venv_path}.
+	"""
+	print("Creating virtual environment...", flush=True)
+	subprocess.run(
+		[
+			"py", "-3.8-32",
```
lukaszgo1, line 59: Specifying Python version explicitly in more places creates more places in which we would need to update it in the future. You can just use `sys.executable` here.

venvUtils/ensureVenv.py
```diff
@@ -0,0 +1,129 @@
...
+		check=True,
+		shell=True,
+	)
+	shutil.copy(requirements_path, venv_orig_requirements_path)
+
+def ensureVenvAndRequirements():
+	"""
+	Ensures that the NVDA build system's Python virtual environment is created and up to date.
+	If a previous virtual environment exists but has a miss-matching Python version or pip package requirements have changed,
+	The virtual environement is recreated with the updated version of Python and packages.
```
lukaszgo1, line 91: environement -> environment

venvUtils/ensureVenv.py
```diff
@@ -0,0 +1,129 @@
...
+	If a virtual environement is found but does not seem to be ours,
```
lukaszgo1, line 92: Also here...

venvUtils/ensureVenv.py
```diff
@@ -0,0 +1,129 @@
...
+	"""
+	if not os.path.exists(venv_path):
+		print("Virtual environment does not exist.")
+		return createVenvAndPopulate()
+	if (
+		not os.path.exists(venv_python_version_path)
+		or not os.path.exists(venv_orig_requirements_path)
+	):
+		if askYesNoQuestion(
+			f"Virtual environement at {venv_path} probably not created by NVDA. This directory must be removed before continuing. Should it be removed?"
```
lukaszgo1, line 103: And here - probably bad copy/paste or autocomplete.

lukaszgo1 (commented):
Please make sure that readme in tests/system reflects the new situation.

runnvda.bat
```diff
@@ -0,0 +1,2 @@
+@echo off
+call "%~dp0\venvUtils\venvCmd.bat" py "%~dp0\source\nvda.pyw" %*
```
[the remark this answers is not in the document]
michaelDCurran: If called from a batch file, pythonw will actually block. Calls in batch files always wait on any process executed. And since it blocks, and I added the freeConsole call in core.py so that the console is still usable by NVDA, we may as well use python rather than pythonw so that critical errors are not missed.
However, I have now also added runnvdaw.bat which executes NvDA in the virtual environment with start pyw, which again provides the behaviour developers were used to by using pyw or executing nvda.pyw directly from Explorer etc. I.e. it does not block, and no console window stays around for the life time of its execution.
Of course this being a draft pr, all this is still up for debate -- I certainly welcome other ideas on the best way we can allow developers to run NVDA.

appveyor.yml
```diff
@@ -59,6 +59,10 @@ init:
 clone_depth: 1
 
 install:
+ # Activate the NvDA build system Python virtual environment, creating / updating it if necessary.
+ - .\venvUtils\ensureAndActivate.bat
+ # Save an exact list of the actual Python packages and their versions that got installed, to aide in reproducing a build
+ - py -m pip freeze >output\pip_packages.txt
```
lukaszgo1, line 65: At this point there is no directory called 'output' - it is created later on by SCons, so this command fails.

venvUtils/ensureVenv.py
```diff
@@ -0,0 +1,129 @@
...
+	oldRequirements = fetchRequirementsSet(venv_orig_requirements_path)
+	newRequirements = fetchRequirementsSet(requirements_path)
+	addedRequirements = newRequirements - oldRequirements
+	if addedRequirements:
+		print(f"Added or changed package requirements. {addedRequirements}")
+		return createVenvAndPopulate()
+
+if __name__ == '__main__':
+	# Ensure we are not inside an already active Python virtual environment. 
+	virtualEnv = os.getenv("VIRTUAL_ENV")
```
lukaszgo1, line 122: Ideally this should not warn if the active venv is the NVDA venv.

dingpengyu:
Using Windows version 10.0.19042 workstation testing NVDA version try-py3.8_venv-21803



Successfully tested in Simplified Chinese
[nvda.log](https://github.com/nvaccess/nvda/files/6066209/nvda.log)

michaelDCurran:
I believe I have addressed all review comments up to now.
I have now updated the pr description, and this pr is no longer a draft.
I'm looking for final reviews before this can possibly be merged.

requirements.txt
```diff
@@ -0,0 +1,26 @@
+# NVDA's build system is SCons
+SCons==4.1.0.post1
+
+# NVDA's runtime dependencies
```
codeofdusk, line 4: Please also add `diff_match_patch_python` so that DMP will work again on Python 3.8 and newer Python versions going forward (currently the extension in the submodule is only compatible with 3.7). Once we switch to 3.8 I'll remove the extension from the nvda_dmp repo.

codeofdusk: I've now opened PR #12113 (to be merged into this branch and fix DMP).

requirements.txt
```diff
@@ -0,0 +1,29 @@
+# NVDA's build system is SCons
+SCons==4.1.0.post1
+
+# NVDA's runtime dependencies
+comtypes==1.1.7
+pyserial==3.5
+wxPython==4.1.1
```
leonardder, line 7: Please consider adding configobj from git:

```suggestion
wxPython==4.1.1
git+git://github.com/DiffSK/configobj@3e2f4cc#egg=configobj
```

Note, this is the last commit from current master, therefore this would require a readme update. No major changes

appveyor.yml
```diff
@@ -95,12 +100,12 @@ build_script:
      Set-AppveyorBuildVariable "sconsOutTargets" $sconsOutTargets
      Set-AppveyorBuildVariable "sconsArgs" $sconsArgs
  - 'echo scons args: %sconsArgs%'
- - py scons.py source %sconsArgs%
+ - scons source %sconsArgs%
  # We don't need launcher to run checkPot, so run the checkPot before launcher.
- - py scons.py checkPot %sconsArgs%
+ - scons checkPot %sconsArgs%
```
feerrenrut, line 105: It might be better to run the same commands that developers are expected to run here. The same thought applies to manually activating the virtual environment and installing dependencies.

scons.bat
```diff
@@ -1,13 +1,3 @@
...
-if "%ERRORLEVEL%" == "0" (
-	rem Python launcher is present in the PATH
-	rem Call python 3.8 32-bit
-	py -3.8-32 "%~dp0\scons.py" %*
-) else (
-	rem Python registers itself with the .py extension, so call scons.py.
-	"%~dp0\scons.py" %*
-)
+rem Executes SScons within the NVDA build system's Python virtual environment.
+"%~dp0\venvUtils\venvCmd.bat" py -m SCons %*
```
feerrenrut, line 3: Please add a new line to the end of this file.

sconstruct
```diff
@@ -7,6 +7,16 @@ import sys
 import os
 import platform
 
+# Ensure we are inside the NVDA build system's Python virtual environment.
+nvdaVenv = os.getenv("NVDA_VENV")
+virtualEnv = os.getenv("VIRTUAL_ENV")
+if not nvdaVenv or nvdaVenv != virtualEnv:
+	print(
+		"Error: SCons cannot detect  the NVDA build system Python virtual environment.\n"
```
feerrenrut, line 15: Double space after 'detect'

source/core.py
```diff
@@ -560,6 +560,11 @@ def run(self):
 	else:
 		log.debug("initializing updateCheck")
 		updateCheck.initialize()
+	# If running from source, try to disconnect from the console we may have been executed in.
+	# NVDA may reconnect to read it later,
+	# but it is better to assume we are not connected to anything at the start.
+	if ctypes.windll.kernel32.FreeConsole() != 0:
+		log.debug("Freed from parent console process.")
```
feerrenrut, line 567: Why is this change necessary or related to the rest of the PR?

source/garbageHandler.py
```diff
@@ -20,7 +20,9 @@ class TrackedObject:
 	"""
 
 	def __del__(self):
-		notifyObjectDeletion(self)
+		# notifyObjectDeletion could be None if Python is shutting down.
+		if notifyObjectDeletion:
+			notifyObjectDeletion(self)
```
feerrenrut, line 25: Could you elaborate on this some more? Has this behavior changed with Python 3.8?

tests/system/libraries/NvdaLib.py
feerrenrut: Yes, running the bat file might be better. It would be good to centralize the method used to invoke python. I've run into nasty issues before where a different version of python was used to execute the system tests. Hence the `-3.7-32` argument.

venvUtils/ensureAndActivate.bat
```diff
@@ -0,0 +1,9 @@
+@echo off
+rem this script ensures the NVDA build system Python virtual environment is created and up to date,
+rem and then activates it.
+rem this script should be used only in the case where many commands will be executed within the environment and the shell will be eventually thrown away. 
+rem E.g. an Appveyor build.
+py -3.8-32 "%~dp0\ensureVenv.py"
+if ERRORLEVEL 1 exit /b %ERRORLEVEL%
+call "%~dp0\..\.venv\scripts\activate.bat"
+set NVDA_VENV=%VIRTUAL_ENV%
```
feerrenrut, line 9: Please add a new line to the end of this file.

venvUtils/ensureVenv.py
```diff
@@ -0,0 +1,137 @@
...
+			# Activate virtual environment
+			os.path.join(venv_path, "scripts", "activate.bat"),
+			"&&",
+			# Ensure we have the latest version of pip
+			"py", "-m", "pip",
+			"install", "--upgrade", "pip",
+			"&&",
+			# Install required packages with pip
+			"py", "-m", "pip",
+			"install", "-r", requirements_path,
```
feerrenrut, line 84: This should probably check that each or the operations were successful, and perhaps the output should be logged somewhere?

venvUtils/venvCmd.bat
```diff
@@ -0,0 +1,11 @@
...
+rem this script executes the single given command and arguments inside the NVDA build system's Python virtual environment.
+rem It activates the environment, creating / updating it first if necessary,
+rem then executes the command,
+rem and then finally deactivates the environment.
+
+setlocal
+call "%~dp0\ensureAndActivate.bat"
+call %*
+call "%~dp0\..\.venv\scripts\deactivate.bat"
+endlocal
```
feerrenrut, line 11: Please add a newline to the end of this file.

source/nvda.pyw
```diff
@@ -18,14 +19,27 @@ if getattr(sys, "frozen", None):
 	sys.path.append(sys.prefix)
 	appDir = sys.prefix
 else:
+	# we are running from source
+	# Ensure we are inside the NVDA build system's Python virtual environment.
+	nvdaVenv = os.getenv("NVDA_VENV")
+	virtualEnv = os.getenv("VIRTUAL_ENV")
+	if not nvdaVenv or nvdaVenv != virtualEnv:
```
feerrenrut, line 26: This will make it difficult to do the virtual env setup manually, though perhaps this should be considered beyond scope. I was thinking that setting up or modifying the virtual env manually might be helpful when experimenting with new libraries, avoiding recreating the virtual env each time. On the other hand, recreating the virtual env each time from the requirements file will prevent silly mistakes that lead to drawing incorrect conclusions while experimenting.

feerrenrut (changes requested):
Overall this looks really good!!

I'd also like to see this documented in the devDocs folder, I commented on the docs issue: https://github.com/nvaccess/nvda/issues/12059#issuecomment-789513379

source/core.py
michaelDCurran: If running NvDA from source in a console with runnvda.bat (in other words python) (not runnvdaw.bat or pythonw) NVDA is associated with the controle, and standard out / standard error are printed there. However, as a screen reader, NVDA itself connects to the console with focus in order to access its contents (at least the older implementation that doesn't use UIA). However, as NVDA is already connected to the console when it starts, then the winConsoleHandler code fails to connect again and or manage connecting / disconnecting from the console with focus. Thus, once NVDA is fully started up, and any critical errors would have been written to the console, as a screen reader, it is better to forcefully disconnect, and it may or may not reconnect later in order to read its content. However, I would be happy to remove this commit and address it in a future pr as it is not specifically to do with Python 3.8. However developers will still hit this if executing runnvda.bat.

source/garbageHandler.py
michaelDCurran: this has not changed with Python 3.8. In fact you reported this bug to me many months ago. I only really noticed it properly as I was testing the rununittests.bat file. I can remove this commit and address it in a future pr if you like? Though now making unittests much easier and faster to run with the new command, that traceback is going to become more noticeable.

source/nvda.pyw
```diff
@@ -10,6 +10,7 @@
 import sys
 import os
 import globalVars
+import ctypes
```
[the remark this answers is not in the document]
michaelDCurran: The only way we could allow this would be that if nvda.pyw is executed outside of the virtual environment that we then execute a subprocess of runnvda or runnvdaw.bat depending on whether python or pythonw was used to originally execute it. But I think that starts getting into some possible recursion danger. I feel it is probably better and safer not to allow executing like this from source anymore, without using runnvda or runnvdaw.

source/nvda.pyw
michaelDCurran: The only way we could allow this would be that if nvda.pyw is executed outside of the virtual environment that we then execute a subprocess of runnvda or runnvdaw.bat depending on whether python or pythonw was used to originally execute it. But I think that starts getting into some possible recursion danger. I feel it is probably better and safer not to allow executing like this from source anymore, without using runnvda or runnvdaw.

venvUtils/ensureVenv.py
michaelDCurran: Because of the "&&"s and the check=True, if any of those commands fail with an error code, the whole command will fail I believe.
I don't feel it is really that necessary to save that output to a file -- we don't tend to do that for scons or other commands (yet). Perhaps that is something that we can look at in future?

appveyor.yml
```diff
@@ -61,9 +62,12 @@ clone_depth: 1
 install:
  # Activate the NvDA build system Python virtual environment, creating / updating it if necessary.
  - .\venvUtils\ensureAndActivate.bat
+ - if ERRORLEVEL 1 exit %ERRORLEVEL%
  # Save an exact list of the actual Python packages and their versions that got installed, to aide in reproducing a build
  - py -m pip freeze >installed_python_packages.txt
+ - if ERRORLEVEL 1 exit %ERRORLEVEL%
  - appveyor PushArtifact  installed_python_packages.txt
+ - if ERRORLEVEL 1 exit %ERRORLEVEL%
```
feerrenrut: Rather than do this here, it could be done in the `on_finish` section. Docs:
>Finalize both successful and failed builds:

It would need to first call `.\venvUtils\ensureAndActivate.bat`

Then, this will be testing the dev workflow more closely. EG calling `scons source` sets up the venv and activates.

devDocs/buildSystemNotes.md
```diff
@@ -0,0 +1,23 @@
+A Python virtual environment is now used transparently by the NVDA build system, and all Python dependencies are installed into this environment using `pip`.
+
+From a developer's perspective using the NVDA build system, there is no need to worry about the Python virtual environment behind the scenes, nor should you create and or activate one manually. All NVDA build system commands will handle this transparently.
+
+* To build NVDA, SCons should continue to be used in the usual way. E.g. executing scons.bat in the root of the repository. Running `py -m SCons`  is no longer supported, and scons.py has also been removed.
```
feerrenrut, line 5: I don't think we need to go over how to build, run, or execute tests here. Those should be documented in a single place, currently that is the main readme. I'd make this doc more of a how / why explanation.

I'd certainly references to unsupported "old ways" of doing this. The change log / PR can used to see what has changed. The python file is no longer available, so it would be pretty hard to make that mistake.

I think we should add an explicit line that builds / running should be done from a cmd.exe window, but that probably belongs in the main readme.md file.

After the first little introduction about how "the build system uses virtual environments transparently", I'd just have a couple of headings.
- How the build system works
  This should highlight the names of the main entry point scripts, list all scripts with an overview of their purpose, and certainly point out they all rely on `ensureVenvAndActivate`.
- Motivation for using virtual environments
  Eg In order to ensure the build environment is clean, no conflicts with other installed packages

devDocs/buildSystemNotes.md
```diff
@@ -0,0 +1,23 @@
+A Python virtual environment is now used transparently by the NVDA build system, and all Python dependencies are installed into this environment using `pip`.
```
feerrenrut, line 1: Rather than "now" I'd make this a statement about the current behavior, it is more simple to read the concrete current truth rather than interpreting a series of changes.

```suggestion
A Python virtual environment is used transparently by the NVDA build system, and all Python dependencies are installed into this environment using `pip`.

```

devDocs/buildSystemNotes.md
```diff
@@ -0,0 +1,23 @@
+A Python virtual environment is now used transparently by the NVDA build system, and all Python dependencies are installed into this environment using `pip`.
+
+From a developer's perspective using the NVDA build system, there is no need to worry about the Python virtual environment behind the scenes, nor should you create and or activate one manually. All NVDA build system commands will handle this transparently.
```
feerrenrut, line 3:
```suggestion
NVDA's build system commands will handle all aspects of the virtual environment.
Developers should not create or activate the virtual environment manually, unless working on the build system itself.
```

appveyor.yml
```diff
@@ -193,20 +203,15 @@ test_script:
...
+    
+     .\runsystemtests.bat --variable whichNVDA:installed
      if($LastExitCode -ne 0) {
       $errorCode=$LastExitCode
       Add-AppveyorMessage "System test failure"
      }
-     $wc = New-Object 'System.Net.WebClient'
+     Compress-Archive -Path "$systemTestOutput\*" -DestinationPath "$testOutput\systemTestResult.zip"
+     Push-AppveyorArtifact "$testOutput\systemTestResult.zip"
+          $wc = New-Object 'System.Net.WebClient'
```
feerrenrut, line 214: This indentation also seems wrong.
```suggestion
     $wc = New-Object 'System.Net.WebClient'
```

readme.md
```diff
@@ -160,9 +159,7 @@ Please see the `release` keyword argument for what compiler optimizations it wil
 
 ## Running the Source Code
 It is possible to run NVDA directly from source without having to build the full binary package and launcher.
-Two batch files (in the root of this repository) are provided for this purpose:
-* runnvda.bat runs NVDA and blocks until NVDA is exited. Any messages written to standard out or standard error during its initialization will be written directly to the console.
-* runnvdaw.bat runs NVDA but does not block. This is the behaviour that developers are previously used to by using pythonw manually.
+To launch NVDA from source, execute runnvda.bat in the root of the repository.
```
feerrenrut, line 162:
```suggestion
To launch NVDA from source, using `cmd.exe`, execute `runnvda.bat` in the root of the repository.
```

feerrenrut (changes requested):
This is looking good! I'm happy to see that the readme for the system tests has been updated too. It actually is now easier than before to run them, that's great. 

I'm guessing that it wouldn't be so simple to make the `checkPot` stuff an external command, and even if it is, it is not really necessary for this PR. I'd be curious to hear your thoughts though, is that a direction we should head in?

devDocs/buildSystemNotes.md
michaelDCurran: Perhaps you can have a go at re-wording / re-structuring this doc if you have some good ideas in your head already. Though I have taken all your other suggestions.

tests/system/libraries/NvdaLib.py
lukaszgo1: I don't think you've addressed this one.

codeofdusk:
PR #12133 now contains a later nvda_dmp commit than this PR. Might there be a merge conflict? Should I base that PR on this branch instead?

appveyor.yml
feerrenrut: @michaelDCurran What do you think of this suggestion?

devDocs/buildSystemNotes.md
feerrenrut: Yep, I'll take a look at this.

feerrenrut (approved):
Looks good to me. Let's rebase this on master and merge straight to master.

feerrenrut:
On second thoughts, no rebase is necessary. Just make sure this is up to date with master and change the target (base) branch to master.

leonardder:
Let's rebase this on master and merge straight to master.

Note that #11912 is not yet part of master. Also this should be up to date with the py3.8 branch before merged into master, I assume?

PR #9679, opened 2019-06-06
Add --copy-portable-config command line parameter to automatically copy the portable configuration with a silent installation

LeonarddeR:
### Link to issue number:
Closes #9676 

### Summary of the issue:
There is currently no automated way to install NVDA silently, thereby also copying a configuration profile folder to the user account. This is currently only possible when manually installing from a portable copy.

### Description of how this pull request fixes the issue:
Added a new --copy-portable-config command line argument, which is stored as a boolean. If the core module initializes and discovers that copying the portable config is hazardous (particularly when running the currently installed config), this variable is set to None. This evaluates to False in most situations. IN the gui however, we explicitly check for None in order to disable the checkbox to enable this feature.

Note that commit 7eaf961c8584efe09d30add contains changes to the formatting of several huge lines in nvda.pyw. It has no functional changes.

### Testing performed:
1. Started a portable copy, started the installation gui. Made sure that "copy the current user config" check box was unchecked.
2. Started a portable copy with --copy-portable-config parameter, started the installation gui. Made sure that "copy the current user config" check box was checked.
3. Started a portable copy with `--copy-portable-config -c %appdata%\nvda`, started the installation gui. Made sure that "copy the current user config" check box was disabled.
4. Started a snapshot from the launcher with `--copy-portable-config -c somepath`. Made sure that the checkbox was enabled and checked.
5. Started a portable copy with `--copy-portable-config --install`. Made sure that the configuration was successfully kopied after the installation.

### Known issues with pull request:
There is no warning when --copy-portable-config is silently discarded for silent installations. I guess we really want to stay silent, and this will hardly happen.

### Change log entry:
* New features
    + Added the `--copy-portable-config` command line parameter that allows you to silently install NVDA and automatically copy the provided configuration to the user account. (#9676)

PR #12076, opened 2021-02-15
Fix-up of pr #9679: Once again allow to copy configuration when installing from a portable NVDA.

### LINKED ISSUE #12071: Important error in NVDA2020.4RC1: The "Copy portable configuration to current user account" checkbox is missing when installing a portable copy on a computer that has never installed NVDA
### Steps to reproduce:
1. Run the installed NVDA and create a portable version.
2. Check the "Copy current user configuration" option.
3. Uninstall the installed NVDA and make sure there is no nvda directory in %appdata% after uninstalling.
4. Run the portable copy created in (1) and select "Install NVDA" under the "Tools" menu.
PS: You can also omit (3) and install the created portable copy in Windows Sandbox.

**Special attention is needed to ensure that there is no NVDA directory in %appdata%**

### Actual behavior:
"Copy portable configuration to current user account" is missing.

### Expected behavior:
"Copy portable configuration to current user account" is displayed normally and can be checked.

only portable version

#### NVDA version:
NVDA2020.4RC1, alpha-21753

#### Windows version:
Windows 10 20H2 (64-bit) build 19042.746

none

yes

alpha-21753

yes

none

### LINKED ISSUE #12205: Portable copy: an install dialog NVDA does not show "copy portable config to this account" or something like that at the first time
Hello,

when I roled back from latest alpha - 22012... to alpha - 21810... via a portable copy I found next bug:
after I ran a portable copy and went to nvda menu, tools, install nvda on this computer, in the first dialog, which has been shown, there was missing a check box to copy portable config to this account. after installing nvda, I ran portable copy again and went to menu and again pressed install nvda to this computer and now checkbox appeared.

### Steps to reproduce:

assuming, that portable copy with user configuration is already built, here are my steps:
- uninstall nvda on system
- reboot a system
- delete nvda folder in my account /appdata/roming
- run portable copy, which was previously created
- on nvda menu, tools press install nvda on a system or something like that

### Actual behavior:

at a first time checkbox to copy portable user config to this account is missing. it appears after nvda is installed on a system and portable copy is running again.
also found, that at the first time folder "system config" is missing in program files (x86)/nvda. is appears after second install too.

### Expected behavior:

this checkbox should be presented at first time, when dialog shows.

portable and then installed (need both of them to replicate this problem)

#### NVDA version:

tested with 22012 and 21810. but probably this issue exist longer.

#### Windows version:

w10 20h2 

n/a

yes

reported in description

n/a (part of user config)

n/a

lukaszgo1:
### Link to issue number:
Fixes #12071
Fixes #12205

### Summary of the issue:
PR #9679 introduced additional command line parameter which allows to copy config from a portable NVDA during installation. Unfortunately it also made it impossible to copy config from a portable copy when installing manually on an user account  without NVDA config in appdata (the corresponding option was disabled in the GUI).
Rather than disabling  this option when result of `config.getUserDefaultConfigPath`  is the same as the current config path (which was also true for portable copies) I've added additional method which disables this option only when NVDA is started from the launcher and configPath is not overridden by the user, or in case of portable copies when config currently in use is placed in appdata, and therefore NVDA would copy it into itself.
Please note that first two commits are merely updating copyright header (1f8a0fa), and cleaning up unused imports from the installer GUI (741c20b) there are no functional changes in them.
### Testing strategy:
Manual testing:
1. With nvda started from the launcher and NVDA config in appdata ensured that "copy config to the current user account" was disabled in the install dialog.
2. With NVDA started from the launcher with the custom config path ensured that the checkbox in question was enabled.
3. With NVDA started from the portable copy and no config in appdata ensured that this option was shown.
4. with portable NVDA started with the `-c %appdata%\nvda` ensured that "copy portable config to the current user account" was disabled.
5. With portable NVDA and config in appdata ensured that the option was enabled.

It should be possible to create system test for this PR but it would require a big refactoring so that first test is executed in a portable copy and ensures that right options are shown in the installer and then subsequent tests are being run for an installed NVDA. I don't thing small change like this justifies the required effort.

I've also repeated tests from #9679:
1. Started a portable copy, started the installation gui. Made sure that "copy the current user config" check box was unchecked.
2. Started a portable copy with --copy-portable-config parameter, started the installation gui. Made sure that "copy the current user config" check box was checked.
3. Started a portable copy with --copy-portable-config -c %appdata%\nvda, started the installation gui. Made sure that "copy the current user config" check box was disabled.
4. Started a snapshot from the launcher with --copy-portable-config -c somepath. Made sure that the checkbox was enabled and checked.
5. Started a portable copy with --copy-portable-config --install. Made sure that the configuration was successfully kopied after the installation.

None known
### Change log entry:

Section:  Bug fixes
- It is once again possible to copy config from a portable NVDA when installing it under user account without previous NVDA config.
### Code Review Checklist:

- [X] Unit tests.
- [X] Manual tests.
- [X] User Documentation.
- [X] Change log entry.

### DIFF
--- a/source/gui/installerGui.py
+++ b/source/gui/installerGui.py
@@ -20,6 +20,26 @@ from gui.dpiScalingHelper import DpiScalingHelperMixinWithoutInit
 import systemUtils
 
 
+def _canPortableConfigBeCopied() -> bool:
+	# In some cases even though user requested to copy config from the portable copy during installation 
+	# it should not be done.
+	if globalVars.appArgs.launcher:
+		# Normally when running from the launcher
+		# and configPath is not overridden by the user copying config during installation is rather pointless
+		# as we would  copy it into itself.
+		# However, if a user wants to run the launcher with a custom configPath,
+		# it is likely that he wants to copy that configuration when installing.
+		return globalVars.appArgs.configPath != config.getUserDefaultConfigPath(useInstalledPathIfExists=True)
+	else:
+		# For portable copies we want to avoid copying the configuration to itself,
+		# so return True only if the configPath
+		# does not point to the config of the installed copy in appdata.
+		confPath = config.getInstalledUserConfigPath()
+		if confPath and confPath == globalVars.appArgs.configPath:
+			return False
+		return True
+
+
 def doInstall(
 		createDesktopShortcut=True,
 		startOnLogon=True,
@@ -48,7 +68,8 @@ def doInstall(
 		if copyPortableConfig:
 			installedUserConfigPath=config.getInstalledUserConfigPath()
 			if installedUserConfigPath:
-				gui.ExecAndPump(installer.copyUserConfig,installedUserConfigPath)
+				if _canPortableConfigBeCopied():
+					gui.ExecAndPump(installer.copyUserConfig, installedUserConfigPath)
 	except Exception as e:
 		res=e
 		log.error("Failed to execute installer",exc_info=True)
@@ -208,9 +229,9 @@ class InstallerDialog(
 		self.copyPortableConfigCheckbox = optionsHelper.addItem(createPortableBox)
 		self.bindHelpEvent("CopyPortableConfigurationToCurrentUserAccount", self.copyPortableConfigCheckbox)
 		self.copyPortableConfigCheckbox.Value = (
-			bool(globalVars.appArgs.copyPortableConfig) and self._shouldEnableCopyConfig()
+			bool(globalVars.appArgs.copyPortableConfig) and self._canPortableConfigBeCopied()
 		)
-		self.copyPortableConfigCheckbox.Enable(self._shouldEnableCopyConfig())
+		self.copyPortableConfigCheckbox.Enable(self._canPortableConfigBeCopied())
 
 		bHelper = sHelper.addDialogDismissButtons(guiHelper.ButtonHelper(wx.HORIZONTAL))
 		if shouldAskAboutAddons:
@@ -259,25 +280,6 @@ class InstallerDialog(
 			# the defaults from the installer are fine. We are testing against the running version.
 		)
 		incompatibleAddons.ShowModal()
-
-	@staticmethod
-	def _shouldEnableCopyConfig() -> bool:
-		# In some cases option to copy configuration from the portable copy
-		# should be disabled at all costs.
-		if globalVars.appArgs.launcher:
-			# Normally when running from the launcher
-			# and configPath is not overridden by the user this checkbox should be hidden.
-			# However, if a user wants to run the launcher with a custom configPath,
-			# it is likely that he wants to copy that configuration when installing.
-			return globalVars.appArgs.configPath != config.getUserDefaultConfigPath(useInstalledPathIfExists=True)
-		else:
-			# For portable copies we want to avoid copying the configuration to itself,
-			# so show this option only if the configPath
-			# does not point to the config of the installed copy in appdata.
-			confPath = config.getInstalledUserConfigPath()
-			if confPath and confPath == globalVars.appArgs.configPath:
-				return False
-			return True
 
 
 class InstallingOverNewerVersionDialog(
--- a/source/gui/installerGui.py
+++ b/source/gui/installerGui.py
@@ -207,7 +207,9 @@ class InstallerDialog(
 		createPortableBox = wx.CheckBox(optionsBox, label=createPortableText)
 		self.copyPortableConfigCheckbox = optionsHelper.addItem(createPortableBox)
 		self.bindHelpEvent("CopyPortableConfigurationToCurrentUserAccount", self.copyPortableConfigCheckbox)
-		self.copyPortableConfigCheckbox.Value = bool(globalVars.appArgs.copyPortableConfig)
+		self.copyPortableConfigCheckbox.Value = (
+			bool(globalVars.appArgs.copyPortableConfig) and self._shouldEnableCopyConfig()
+		)
 		self.copyPortableConfigCheckbox.Enable(self._shouldEnableCopyConfig())
 
 		bHelper = sHelper.addDialogDismissButtons(guiHelper.ButtonHelper(wx.HORIZONTAL))
--- a/source/gui/installerGui.py
+++ b/source/gui/installerGui.py
@@ -208,10 +208,7 @@ class InstallerDialog(
 		self.copyPortableConfigCheckbox = optionsHelper.addItem(createPortableBox)
 		self.bindHelpEvent("CopyPortableConfigurationToCurrentUserAccount", self.copyPortableConfigCheckbox)
 		self.copyPortableConfigCheckbox.Value = bool(globalVars.appArgs.copyPortableConfig)
-		if globalVars.appArgs.copyPortableConfig is None:
-			# copyPortableConfig is set to C{None} in the main loop,
-			# when copying the portable configuration should be disabled at all costs.
-			self.copyPortableConfigCheckbox.Disable()
+		self.copyPortableConfigCheckbox.Enable(self._shouldEnableCopyConfig())
 
 		bHelper = sHelper.addDialogDismissButtons(guiHelper.ButtonHelper(wx.HORIZONTAL))
 		if shouldAskAboutAddons:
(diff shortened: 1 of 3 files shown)


cary-rowen:
Is it possible to postpone the release of 2020.4 to incorporate this fix PR?

dingpengyu:
hello @lukaszgo1 
Recommend to change the merge master to beta
thanks

LeonarddeR (dismissed):
2020.4 has been released. This fix will be in 2021.1.

feerrenrut (changes requested):
Generally this looks good @lukaszgo1

Could you please run through the testing for #9679 and add the procedure and results to the testing section of this PR? I'm a little concerned this may break cases for that. Does this change now allow config to be copied to itself with the `--copy-portable-config` option?

I think the help for `--copy-portable-config` is now misleading, to me it sounds like you must also specify `--config-path` for it to work? Could you confirm or clarify the help please?

lukaszgo1:
@feerrenrut  wrote:

> Could you please run through the testing for #9679 and add the procedure and results to the testing section of this PR? I'm a little concerned this may break cases for that.

Done. As expected all cases from #9679 were working correctly with this PR as it only affects the checkbox being enabled and disabled and not the actual install procedure.

> Does this change now allow config to be copied to itself with the `--copy-portable-config` option?

I'm afraid I don't understand what you're asking here. You cannot copy the config directory to itself obviously.

> 
> I think the help for `--copy-portable-config` is now misleading, to me it sounds like you must also specify `--config-path` for it to work? Could you confirm or clarify the help please?



No you don't. For me current help message is pretty clear - that is if you specify `--copy-portable-config` NVDA would copy current config to the appdata of the current user. Fell free to suggest better wording for the help message.

feerrenrut:
Perhaps you can explain the full flow for processes, rather than just the changes you are making. This would help anyone reading this to be on the same page with how the underlying system works.

Since the failsafe in `core.py` has been removed, looking at just the diff of this PR it seems like it would allow copying from user directory to itself if `--copy-portable-config` command line argument was provided.
Could you highlight where this is prevented?

The help for `--copy-portable-config` currently says:
```
When installing, copy the portable configuration from the provided path (--config-path, -c) to the current user account
```
It doesn't say what happens when there is no provided path. It should probably mention where the config is copied from when run as a portable copy explicitly. IE don't assume the reader know where a portable config is kept.

lukaszgo1:
@feerrenrut wrote:

> Perhaps you can explain the full flow for processes, rather than just the changes you are making. This would help anyone reading this to be on the same page with how the underlying system works.
> 
> Since the failsafe in `core.py` has been removed, looking at just the diff of this PR it seems like it would allow copying from user directory to itself if `--copy-portable-config` command line argument was provided.
> Could you highlight where this is prevented?

In the `gui.installerGui.doInstall` 

> 
> The help for `--copy-portable-config` currently says:
> 
> ```
> When installing, copy the portable configuration from the provided path (--config-path, -c) to the current user account
> ```
> 
> It doesn't say what happens when there is no provided path. It should probably mention where the config is copied from when run as a portable copy explicitly. IE don't assume the reader know where a portable config is kept.



I've reworded the help message in a way which avoids mentioning `--configPath` at all - is it better now?

lukaszgo1:
@feerrenrut Would you be able to review latest changes here? It is pretty annoying to copy config from the portable version after installation each time.

source/gui/installerGui.py
```diff
@@ -1,29 +1,45 @@
...
 
+def _canPortableConfigBeCopied() -> bool:
+	# In some cases even though user requested to copy config from the portable copy during installation
+	# it should not be done.
+	if globalVars.appArgs.launcher:
+		# Normally when running from the launcher
+		# and configPath is not overridden by the user copying config during installation is rather pointless
+		# as we would  copy it into itself.
+		# However, if a user wants to run the launcher with a custom configPath,
+		# it is likely that he wants to copy that configuration when installing.
```
feerrenrut, line 31: `configPath` here sounds like the command line argument `-c CONFIGPATH | --config-path=CONFIGPATH`. Given this function is all about copying the configuration, I'd suggest using the terms 'source' or 'origin' configuration and 'destination' config to clarify. This also applies to variable names.

lukaszgo1:
> 
> 
> I think the docs for these commands could still be clarified, but first lets make sure we are on the same page about the behavior and intentions here.
> 
> My understanding of this is:
> 
>     * A portable copy has its config stored with it in a subdirectory (`userConfig`)
> 
>     * An installed copy has its config stored in `%APPDATA%/nvda`

In both these cases location of the config can be overwritten by providing `--configpath` from the command line when starting NVDA.

> 
>     * The main use case for `--copy-portable-config`, is when running a portable copy, and wanting to install and have the newly installed NVDA version configured in the same way as the portable copy. This is done via the "Install NVDA" option in the tools submenu of the NVDA menu, when the 'Copy portable configuration to current user account ' option is checked.
> 
>     * "Install NVDA" is not shown for installed copies. But it is shown when running from portable or from the launcher.
> 
>     * When "Install NVDA" is selected, the installation happens from within the current running process. The launcher is not started, and the config option `--copy-portable-config` is not used.

That's not true - if `--copy-portable-config` is provided  from the cli the checkbox in the GUI is automatically checked

> 
>     * However, for some admins or advanced users, it is useful to be able to install in NVDA in a silent manner from a portable copy. In this case the `--copy-portable-config` option is useful.
> 
>     * `--copy-portable-config` should also protect against copying over the currently in use config.
> 
>     * Using the `--copy-portable-config` option should also honor the `--config-path` argument which typically specifies the configuration that NVDA should use to run, and is the source configuration to copy.
> 
> 
> The copy portable config code seems to consider running with the launcher, but what happens when this argument is used on an installed version of NVDA. I think it should probably give an error.


That is an interesting point. Currently nothing happens but it is no different to other command line options which are not applicable in the current situation. Consider the following two cases:

- `nvda --install-silent` from the run dialog if NVDA is already installed it tries to install itself again and fails since it cannot overwrite its own files.
- `nvda -aaaaaa` - there is no indication that invalid option has been given from the command line. 

I agree we should clean this up and present an error for options which are not applicable in the given state, however I don't think this should be done as part of tis PR.

feerrenrut:
> That's not true - if --copy-portable-config is provided from the cli the checkbox in the GUI is automatically checked

Can you clarify this. Running which executable from the CLI (launcher, portable NVDA, installed NVDA)?

I seem to have confused two use cases here:
- Using the GUI to install from portable.
- Using the CLI to install from portable NVDA.exe passing the the `--copy-portable-config` option.

It might be helpful to explicitly mention (and give examples) of some of the CLI use cases in the docs. There are a lot permutations of arguments, ways to run them, and NVDA state. It would be easy for one of these situations to be broken. Anything we can do to reduce this would be good.

It seems to me to only make sense to use the `--copy-portable-config` command line argument when running a portable copy, when also specifying '--install' or `--install-silent`. It should also be acceptable to optionally use `--portable-path` and `--config-path` to specify the locations for config files for source and destination respectively.

lukaszgo1:
I feel most of the confusion here comes from naming of the argument since `--copy-portable-config` suggests that the option works only for portable copies which is not true. What this option in fact does is it just copies a  configuration used during NVDA install for silent installations unless location of the config equals location to which config would be copied, and for GUI installs from the portable copies it affects state of the checkbox used to decide if portable config should be copied or not. It probably be better to name this option `--copy-current-config-during-install` or similar.
It is unfortunate these concerns weren't brought up when reviewing #9679 since now it is too late to rename the command line parameter.

CyrilleB79:
> It is unfortunate these concerns weren't brought up when reviewing #9679 since now it is too late to rename the command line parameter.

IMO, the flag may be renamed if it adds some clarity; also the flag documentation should be updated to cover all use cases.
The old flag could still be supported during a certain amount of time in order not to break existing scripts and then eventually be deprecated.

feerrenrut:
I'm in support of adding a new alias for this argument. But it would be good to name it carefully. I don't like the word "current", I think it is ambiguous between the running copy which is potentially very temporary. What about `--copy-config-to-installation` or `--copy-config-during-installation`. Then in the help also add:
> optionally use --portable-path and --config-path to specify the locations for config files for source and destination respectively. If not specified the default locations will be used.

Then in the help for `--portable-path` and `--config-path` explicitly state the default locations. EG:
- `userConfig` directory relative to the nvda.exe
- `%APPDATA%/nvda`

feerrenrut:
Note for those who consume this via email, I had to edit my last comment to finish the sentence about ambiguity: "ambiguous between the running copy which is potentially very temporary and the already installed or about to be installed copy. "

lukaszgo1:
`--portable-path` is irrelevant here as it is just the folder in which portable copy of NVDA is created - it has no effect when installing.

source/gui/installerGui.py
lukaszgo1: Using `configPath` in this comment is appropriate because this function decides if config can be copied or not by looking at the value of `globalVars.appArgs.configPath`.

lukaszgo1:
@feerrenrut While I do share your concerns about better naming of this option, documentation not being extensive enough to cover all possible use cases and giving warnings when options not applicable in the current situation are provided from the CLI I feel it is way more important not to break GUI installations from the portable copies than improving command line arguments. Since it would be great to move this PR forward and include it in 2021.1 would you agree  to proceed as follows:
- Merge this as is reverting eb5dd4f 
- - create an issue to discuss alternative name for `--copy-portable-config`, possible improvements to  its documentation etc separately.

I'd more than happy to work on this subsequently given I needed to  familiarize myself with this code during this PR. Doing it as a follow up ensures that more common use case i.e installing from the GUI is not negatively affected by discussion about way less commonly used CLI options which it seems can take a while.

feerrenrut:
Ok, I'm fine with that plan. Could you create an new issue to track improving the docs around these CLI arguments?

feerrenrut (approved):
Thanks @lukaszgo1

lukaszgo1:
@feerrenrut  As a result of our discussion on this PR I've created:
- #12795 for  warning about unknown CLI params.
- #12796 warning aabout parameters not applicable in the current context.
- #12797 for improvements to the documentation of `--copy-portable-config`


I'm pretty sure all discussed points are captured in these issues but  feel free to add to them if you feel something has been missed.

Issue #10036, opened 2019-08-05
UIA in Windows consoles: spaces are stripped from end of line, blank reported when arrowing through them


leonardder:
cc @codeofdusk 

### Steps to reproduce:
With UIA enabled in Windows consoles, type multiple spaces and arrow back.

### Actual behavior:
NVDA reports blank for every press of left arrow

### Expected behavior:
NVDA reports space for every press of left arrow

#### NVDA installed/portable/running from source:
Installed

#### NVDA version:
Alpha alpha-18290

#### Windows version:
Windows 10 1903 build 18362.239

#### Name and version of other software in use when reproducing the issue:
Windows command prompt (cmd)

#### Does the issue still occur after restarting your PC?
Yes

#### Have you tried any other versions of NVDA? If so, please report their behaviors.
No

Andre9642:
Any news on this?

michaelDCurran:
Please note that Windows consoles do not store space characters at all. If space is typed and or programmatically written, the character cell at that position is just left blank.
We should possibly return a space character for UIA text ranges that are blank in TextInfo.text. This at least will say 'space' for spaces in the middle of a line like it used to. However, spaces at the end of a line will be impossible as the UIA text ranges for a line don't cover the full terminal width, rather they only go up to the last non-space character on the line. 
If the in-line space issue was fixed, would you consider  allowing this issue to be closed as fixed, even though spaces at the end of the line (which again don't really exist anyway) was not fixed?

codeofdusk:
> Please note that Windows consoles do not store space characters at all. If space is typed and or programmatically written, the character cell at that position is just left blank.
> We should possibly return a space character for UIA text ranges that are blank in TextInfo.text. This at least will say 'space' for spaces in the middle of a line like it used to. However, spaces at the end of a line will be impossible as the UIA text ranges for a line don't cover the full terminal width, rather they only go up to the last non-space character on the line.
> If the in-line space issue was fixed, would you consider allowing this issue to be closed as fixed, even though spaces at the end of the line (which again don't really exist anyway) was not fixed?

I'd be happy with this solution – it also seems to have the added bonus of more accurately reflecting the console's contents.

PR #12094, opened 2021-02-24
UI Automation in Windows Console: rename is21H1Plus to isImprovedTextRangeAvailable

codeofdusk:
### Link to issue number:
microsoft/terminal#9239

### Summary of the issue:
Windows 10 21H1 has been released to insiders with significantly reduced scope (i.e. excluding the UIA changes it was originally set to include), making our internal names misleading. Additionally, there are plans to "undock" conhost from Windows, removing the guarantee that a given Windows version will run a particular version of the console (it's technically already possible to run newer console on older Windows, I do it for testing).

Renames:

* `NVDAObjects.UIA.winConsoleUIA.is21H1Plus` -> `NVDAObjects.UIA.winConsoleUIA.isImprovedTextRangeAvailable`
* `NVDAObjects.UIA.winConsoleUIA.consoleUIATextInfo` -> `NVDAObjects.UIA.winConsoleUIA.ConsoleUIATextInfo` (Start class name with upper case)
* `NVDAObjects.UIA.winConsoleUIA.consoleUIATextInfoPre21H1` -> `NVDAObjects.UIA.winConsoleUIA.ConsoleUIATextInfoWorkaroundEndInclusive` (because much of the implementation deals with the fact that before microsoft/terminal#4018 text ranges had both end points inclusive, necessitating workarounds for `expand`, `collapse`, `compareEndPoints`, `setEndPoint`, etc)

Also moves the `text` override to the workaround `textInfo` as #10036 is no longer reproducible in new console.
### Testing performed:
Verified that console still works as intended. These names were used internally only.

None.

### Change log entry:
Changes for developers:
```
- ``NVDAObjects.UIA.winConsoleUIA.is21H1Plus`` renamed ``NVDAObjects.UIA.winConsoleUIA.isImprovedTextRangeAvailable``. (#12094)
- ``NVDAObjects.UIA.winConsoleUIA.consoleUIATextInfo`` renamed to start class name with upper case. (#12094)
- ``NVDAObjects.UIA.winConsoleUIA.consoleUIATextInfoPre21H1`` renamed ``NVDAObjects.UIA.winConsoleUIA.ConsoleUIATextInfoWorkaroundEndInclusive``. (#12094)
  - The implementation works around both end points being inclusive (in text ranges) before microsoft/terminal#4018
  - Workarounds for ``expand``, ``collapse``, ``compareEndPoints``, ``setEndPoint``, etc
```

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### DIFF
--- a/source/NVDAObjects/UIA/winConsoleUIA.py
+++ b/source/NVDAObjects/UIA/winConsoleUIA.py
@@ -1,7 +1,7 @@
 # A part of NonVisual Desktop Access (NVDA)
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
-# Copyright (C) 2019-2020 Bill Dengler
+# Copyright (C) 2019-2021 Bill Dengler
 
 import ctypes
 import NVDAHelper
@@ -16,7 +16,7 @@ from ..behaviors import EnhancedTermTypedCharSupport, KeyboardHandlerBasedTypedC
 from ..window import Window
 
 
-class consoleUIATextInfo(UIATextInfo):
+class ConsoleUIATextInfo(UIATextInfo):
 	def __init__(self, obj, position, _rangeObj=None):
 		collapseToEnd = None
 		# We want to limit  textInfos to just the visible part of the console.
@@ -33,7 +33,7 @@ class consoleUIATextInfo(UIATextInfo):
 				log.warning("Couldn't get bounding range for console", exc_info=True)
 				# Fall back to presenting the entire buffer.
 				_rangeObj, collapseToEnd = None, None
-		super(consoleUIATextInfo, self).__init__(obj, position, _rangeObj)
+		super(ConsoleUIATextInfo, self).__init__(obj, position, _rangeObj)
 		if collapseToEnd is not None:
 			self.collapse(end=collapseToEnd)
 
@@ -90,26 +90,17 @@ class consoleUIATextInfo(UIATextInfo):
 
 	def _move(self, unit, direction, endPoint=None):
 		"Perform a move without respect to bounding."
-		return super(consoleUIATextInfo, self).move(unit, direction, endPoint)
+		return super(ConsoleUIATextInfo, self).move(unit, direction, endPoint)
 
 	def __ne__(self, other):
 		"""Support more accurate caret move detection."""
 		return not self == other
 
-	def _get_text(self):
-		# #10036: return a space if the text range is empty.
-		# Consoles don't actually store spaces, the character is merely left blank.
-		res = super(consoleUIATextInfo, self)._get_text()
-		if not res:
-			return ' '
-		else:
-			return res
 
-
-class consoleUIATextInfoPre21H1(consoleUIATextInfo):
-	"""Fixes expand/collapse on end inclusive UIA text ranges, uses rangeFromPoint
-	instead of broken GetVisibleRanges for bounding, and implements word
-	movement support."""
+class ConsoleUIATextInfoWorkaroundEndInclusive(ConsoleUIATextInfo):
+	"""Implementation of various workarounds for pre-microsoft/terminal#4018
+	conhost: fixes expand/collapse, uses rangeFromPoint instead of broken
+	GetVisibleRanges for bounding, and implements word movement support."""
 	def _getBoundingRange(self, obj, position):
 		# We could use IUIAutomationTextRange::getVisibleRanges, but it seems very broken in consoles
 		# once more than a few screens worth of content has been written to the console.
@@ -139,12 +130,12 @@ class consoleUIATextInfoPre21H1(consoleUIATextInfo):
 		return (_rangeObj, None)
 
 	def collapse(self, end=False):
-		"""Works around a UIA bug on Windows 10 versions before 21H1.
+		"""Works around a UIA bug on conhost versions before microsoft/terminal#4018.
 		When collapsing, consoles seem to incorrectly push the start of the
 		textRange back one character.
 		Correct this by bringing the start back up to where the end is."""
 		oldInfo = self.copy()
-		super(consoleUIATextInfo, self).collapse(end=end)
+		super(ConsoleUIATextInfo, self).collapse(end=end)
 		if not end:
 			self._rangeObj.MoveEndpointByRange(
 				UIAHandler.TextPatternRangeEndpoint_Start,
@@ -153,7 +144,7 @@ class consoleUIATextInfoPre21H1(consoleUIATextInfo):
 			)
 
 	def compareEndPoints(self, other, which):
-		"""Works around a UIA bug on Windows 10 versions before 21H1.
+		"""Works around a UIA bug on conhost versions before microsoft/terminal#4018.
 		Even when a console textRange's start and end have been moved to the
 		same position, the console incorrectly reports the end as being
 		past the start.
@@ -168,7 +159,7 @@ class consoleUIATextInfoPre21H1(consoleUIATextInfo):
 
 	def setEndPoint(self, other, which):
 		"""Override of L{textInfos.TextInfo.setEndPoint}.
-		Works around a UIA bug on Windows 10 versions before 21H1 that means we can
+		Works around a UIA bug on conhost versions before microsoft/terminal#4018 that means we can
 		not trust the "end" endpoint of a collapsed (empty) text range
 		for comparisons.
 		"""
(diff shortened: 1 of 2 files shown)


codeofdusk:
Cc @feerrenrut, @michaelDCurran, @carlos-zamora, @DHowett.

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -348,7 +348,7 @@ def _get_windowThreadID(self):
 			threadID = super().windowThreadID
 		return threadID
 
-	def _get_is21H1Plus(self):
+	def _get_isPostUIARefactor(self):
```
feerrenrut, line 351: This isn't very descriptive. If the main feature that consoleUIATextInfo relies on is GetVisibleRanges then perhaps it should be called 'isGetVisibleRangeAvailable`

codeofdusk: This, too, is somewhat misleading I think. `GetVisibleRanges` was available in the bad UIA implementation as well, but it returned incorrect results in some cases.

codeofdusk:
Essentially, `GetVisibleRanges` is a convenient way to check whether microsoft/terminal#4018 is available, as that work fixed many bugs in the UIA implementation.

feerrenrut:
Is there any other way to check for that? If not, then please just change the names and docs here to be really explicit about what you are trying to determine and why.

Perhaps a name like `IsImprovedUIATextRangeAvailable` with a docstring including all the notes you have given me in comments here, and also pointing to that PR.

codeofdusk:
> Is there any other way to check for that?

Sadly I don't think so, CC @miniksa and @carlos-zamora.

josephsl:
Hi, one possible idea might be the ability to check for conhost.exe file metadata, as Bill pointed out that it is possible to test newer versions of this binary. I tried this approach but couldn’t get it to work (I used looking up app version for conhost.exe using file metadata method defined in app module class to no avail). Thanks.

miniksa:
> > Is there any other way to check for that?
> 
> Sadly I don't think so, CC @miniksa and @carlos-zamora.

We are strongly advised by the Windows compatibility experts that detecting functionality is always supreme to detecting version, so we would have to say that we prefer the existing method of testing for the functionality to know you're on an appropriate version.

However, if you must go beyond that, you might be able to use [VerifyVersionInfo](https://docs.microsoft.com/windows/win32/api/winbase/nf-winbase-verifyversioninfow) of which I have a sample here https://github.com/microsoft/terminal/blob/22fd06e19b3e564d448d60ebf34ffd63764f8080/src/cascadia/TerminalSettingsModel/CascadiaSettings.cpp#L1158-L1168. I'm obviously guilty of ignoring their strong recommendation and checking the version there... but `VerifyVersionInfo` is preferred by the compat team to getting the number and doing the comparison yourself with one of the `GetVersion*` APIs.

Do note that generally speaking, the Windows compatibility team has installed a number of measures to make it very difficult for applications to get the version of the operating system and force them to feature detect instead. So you may get a "version lie" out of that API or any of the others that is crafted to the maximum version that NVDA is [pre-manifested](https://docs.microsoft.com/windows/win32/sbscs/application-manifests#supportedos) to support.

codeofdusk:
I've updated comments to express intent more clearly, but haven't yet changed any names yet.

Something like `isImprovedTextRangeAvailable` as a replacement for `is21H1Plus` sounds good, but what should the new name for `ConsoleUIATextInfoPre21H1` be?

codeofdusk:
@feerrenrut I settled on `TextInfoWorkaroundEndInclusive` (see description).

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -99,17 +99,17 @@ def __ne__(self, other):
...
 
 
-class consoleUIATextInfoPreRefactor(consoleUIATextInfo):
-	"""Fixes expand/collapse on end inclusive UIA text ranges, uses rangeFromPoint
-	instead of broken GetVisibleRanges for bounding, and implements word
-	movement support."""
+class TextInfoWorkaroundEndInclusive(TextInfo):
+	"""Implementation of various workarounds for pre-microsoft/terminal#4018
+	conhost: fixes expand/collapse, uses rangeFromPoint instead of broken
+	GetVisibleRanges for bounding, and implements word movement support."""
```
feerrenrut, line 112: This looks good.

feerrenrut (commented):
Let me know if you will change the TextInfo. Either way I think this can be merged.

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -16,7 +16,7 @@
 from ..window import Window
 
 
-class consoleUIATextInfo(UIATextInfo):
+class TextInfo(UIATextInfo):
```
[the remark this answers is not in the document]
codeofdusk: Yes, but then we'll have `ConsoleUIATextInfo` and `TextInfoWorkaroundEndInclusive`: one name mentions console UIA and one doesn't.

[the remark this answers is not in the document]
feerrenrut: I'm not sure I understand why that is a problem. Ideally `ConsoleUIATextInfo` would tell you something about how it is different from a `UIATextInfo` or a `TextInfo`. But at least giving it a unique name means that text tools can be used on it. 

But like I said, lots of other classes are called `TextInfo` so it's not introducing something new. It just means that smarter tools that understand namespaces must be used.

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, obj, position, _rangeObj=None):
 				log.warning("Couldn't get bounding range for console", exc_info=True)
 				# Fall back to presenting the entire buffer.
 				_rangeObj, collapseToEnd = None, None
-		super(consoleUIATextInfo, self).__init__(obj, position, _rangeObj)
+		super(TextInfo, self).__init__(obj, position, _rangeObj)
```
lukaszgo1, line 36: Can't you just use Python 3 `super` here? Same goes for all other modified calls to `super` below.

source/NVDAObjects/UIA/winConsoleUIA.py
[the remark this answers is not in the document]
codeofdusk: Module-level internal consistency (and, as you point out, consistency with the rest of NVDA).

codeofdusk:
Ready for a merge whenever you are.

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -90,26 +90,17 @@ def move(self, unit, direction, endPoint=None):
...
-			return ' '
-		else:
-			return res
-
 
-class consoleUIATextInfoPre21H1(consoleUIATextInfo):
-	"""Fixes expand/collapse on end inclusive UIA text ranges, uses rangeFromPoint
-	instead of broken GetVisibleRanges for bounding, and implements word
-	movement support."""
+class TextInfoWorkaroundEndInclusive(TextInfo):
```
feerrenrut, line 100: The problem with using `TextInfo` as a name is that code like this becomes quite ambiguous for a reader. I think it is unnecessarily ambiguous. Consistency is worthwhile if the convention scales, I'm not sure this convention really makes anything better, other than allowing the developer a way out of coming up with a descriptive name.

Code is read many more times than it is written. Better to optimize for the reader IMO.

feerrenrut: When I say ambiguous, I mean that you have to be aware that there is a class within this module called TextInfo, and check that the import doesn't do `from textInfos import TextInfo`.

codeofdusk: OK, the name is now `ConsoleUIATextInfo` (with capitalization consistent with the rest of NVDA).

In the long term (i.e. once microsoft/terminal#6986 has been resolved), I hope to transition to just using the standard `UIATextInfo` implementation for consoles where `isImprovedTextRangeAvailable`... but that's in a follow-up PR.

feerrenrut (dismissed):
I'll look to merge this on Monday.

feerrenrut (dismissed):
Thanks @codeofdusk. This is being merged into Beta, it will be merged back into alpha/master shortly.

feerrenrut:
A note @codeofdusk, this requires a change log. Code is only "internal use" if the class / function is prefixed with an underscore. Otherwise it is considered an API change and should be documented.

codeofdusk:
Something like this in dev changes:

* `WinConsoleUIA.is21H1Plus` has been renamed to `WinConsoleUIA.isImprovedTextRangeAvailable` and `ConsoleUIATextInfoPre21H1` has been renamed to `ConsoleUIATextInfoWorkaroundEndInclusive` due to Microsoft changes
* `consoleUIATextInfo` has been renamed to `ConsoleUIATextInfo`

feerrenrut:
I have updated the description for you.

codeofdusk:
LGTM.

feerrenrut (approved):

Issue #12064, opened 2021-02-10
Task: Update to latest (most appropriate) version of Python


feerrenrut:
- Update appveyor.yml
- Update to latest version of dependencies:
  - Scons
  - comtypes
  - wxPython
  - pySerial
  - py2exe
  - brltty

Initially Python 3.8 was being considered. 3.8 was chosen because 3.9 drops support for Windows 7 (see discussion: https://github.com/nvaccess/nvda/issues/10933#issuecomment-737709194). The plan was to upgrade to windows 3.8, give Windows 7 users early notice, then upgrade again a year later.

Instead, we ran into a bug in python which prevented updating: (see https://bugs.python.org/issue38748 and python/cpython#26204)

When this is resolved (at the next compat breaking release) we will update to the latest (most appropriate version of Python).
If this is not fixed, we may have to consider moving NVDA to 64 bit, (the bug does not affect 64 bit python). This would be a large and risky change.

feerrenrut:
Fixed with PR Py3.8: Transparently use a Python virtual environment under the hood #12075

lukaszgo1:
I'm not sure I fully understood your last comment @zstanecic but in short what I've proposed above was to stay on Python 3.7 but use the latest releases by building them from sources since Python developers are no longer releasing binary versions of Py 3.7. Also it seems to me you're  not realizing how serious the issue in libffi is - it affects much more than support for SAPI4 -  there is no way to learn in advance in which cases NVDA would crash as a result of this bug.

mazen428:
Hello.
I think upgrading to python 3.8 in 2023 wouldn't make sense, because by then python 3.8 and even 3.9 would be considered old especially because they won't be getting any bug fixes, only security ones, so then we're back to the same issue.

michaelDCurran:
The point is that we canot move off Python 3.7 before NVDA 2022.1 due to 
a stack corruption bug in Python.

Once this is addressed, we would move to what ever Python version seemed 
appropriate at the time.

zstanecic:
Hi,

How it will be to tr to update to latest version and try it firstly in the try build, to see was this fixed?

Best,

Zvonimir

feerrenrut:
@zstanecic We have added automated tests to python (see  https://bugs.python.org/issue38748 and python/cpython#26204) which show the issue, the problem is still not fixed.

I'll update the title of this issue to clarify.

josephsl:
Hi,

[...]

PR #12141, opened 2021-03-09
Python 3.8: Mention additional system requirements for Windows 7

lukaszgo1:
### Link to issue number:
None, related to #12064

### Summary of the issue:
Python 3.8 introduced [changes to the way in which dll's are loaded on Windows](https://bugs.python.org/issue36085). This however does not work on clean installs of Windows 7 - a specific update needs to be installed.

User guide has been updated to mention that on Windows 7 SP1 KB3063858 has to be installed. I've mentioned KB3063858 specifically rather than KB2533623 which is mentioned in the Python issue above as KB3063858 replaces it and KB2533623 can no longer be downloaded from Microsoft servers anyway.

### Testing strategy:
Manual testing:
- On my main fully uptodate Windows 7 box ensured that  latest build from #12075 starts correctly.
- On a clean install of Windows 7 SP1 tried the same launcher, ensured that it initially fails, installed KB3063858 and ensured that after it is installed NVDA can start.

Changes like these cannot be unit tested and we don't have system tests running under Windows 7.

I've also compiled user guide and ensured that it displays correctly.
I haven't tested  specifically on Server 2008 R2 but since it is based on the same sources as Windows 7 I don't expect any troubles here.
### Change log entry:
None needed - Python 3.8 is not in a release yet.

### Code Review Checklist:

- [X] Unit tests.
- [X] Manual tests.
- [X] User Documentation.
- [X] Change log entry.

### DIFF
--- a/user_docs/en/userGuide.t2t
+++ b/user_docs/en/userGuide.t2t
@@ -68,7 +68,7 @@ For details regarding exceptions, access the license document from the NVDA menu
 + System Requirements +[SystemRequirements]
 - Operating Systems: all 32-bit and 64-bit editions of Windows 7, Windows 8, Windows 8.1, Windows 10, and all Server Operating Systems starting from Windows Server 2008 R2.
  - For Windows 7 and Windows Server 2008 R2 NVDA requires Service Pack 1 and the [KB3063858 https://support.microsoft.com/en-us/kb/3063858] update.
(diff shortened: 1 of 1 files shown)


josephsl:
Apart from spelling and grammar, it looks fantastic - thanks for pointing these out.

michaelDCurran (approved):

PR #12148, opened 2021-03-10
Stop printing some exceptions in __del__ to standard error

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
At the end of a unit test run, or when NVDA exits and is connected to a console, the following is sometimes printed to standard error:
```
Exception ignored in: <function TrackedObject.__del__ at 0x03DA8D68>
  File "C:\Users\mick\programming\git\nvda2\source\garbageHandler.py", line 23, in __del__
TypeError: 'NoneType' object is not callable
```
and also:
```
Exception ignored in: <function newCpbDel at 0x10909588>
  File "C:\Users\mick\programming\git\nvda2\source\comtypesMonkeyPatches.py", line 104, in newCpbDel
TypeError: 'NoneType' object is not callable
```

As Python exits, it finalizes (deletes) objects and modules in random order. Any symbol (dictionary entry) directly referencing a deleted object is set to None. Thus, code that executes while Python exits may see that some global variables or modules have a value of None. This is especially noticeable from within an object's `__del__` method.
 
```garbageHandler.TrackableObject.__del__``` now checks to make sure that Python is not exiting before it executes code to track deleting objects.
comtypesMonkeyPatches override of ```__del__``` on compointer_base also now checks to make sure that Python is not exiting before trying to reference or execute any globals, such as calling garbageHandler. The original ```__del__``` that is being overridden has also been stored on the class now, rather than as a global variable, to further decrease the possibility of it becoming None when the overriding ```__del__``` is called.
 
### Testing strategy:
* Ran the unit tests and confirmed the errors are no longer printed.
* Ran nvda.pyw with python.exe in a console and confirmed errors were no longer being printed on exit.
* Temporarily placed a print call in both ```__del__``` methods where Python was not exiting, to ensure that that code was in deed normally being executed still.

There may be more ```__del__``` methods which should specifically check to see if Python is exiting. But these are the two that show up the most when running unit tests.

### Change log entry:
None needed.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### CODE REVIEW

File: source/comtypesMonkeyPatches.py [hunk 1]
```diff
@@ -95,14 +96,21 @@ def new__call__(self,*args,**kwargs):
...
 from comtypes import _compointer_base
 
-_cpbDel = _compointer_base.__del__
+_compointer_base._oldCpbDel = _compointer_base.__del__
 def newCpbDel(self):
+	# __del__ may be called while Python is exiting.
+	# In this state, global symbols may be set to None
+	# Therefore avoid calling into garbageHandler.
+	# Using local variables or calling other methods on this class is still okay.
+	isFinalizing = getattr(sys, 'is_finalizing', lambda: True)()
```
feerrenrut, line 105: This seems wrong, but it probably works by accident, a function type is truthy, but isn't actually being called.
```suggestion
	isFinalizing = getattr(sys, 'is_finalizing', True)()

```

source/garbageHandler.py
```diff
@@ -20,7 +21,11 @@ class TrackedObject:
 	"""
 
 	def __del__(self):
-		notifyObjectDeletion(self)
+		# __del__ may still be called while Python is exiting.
+		# And therefore some symbols may be set to None.
+		isFinalizing = getattr(sys, 'is_finalizing', lambda: True)()
```
feerrenrut, line 26:
```suggestion
		isFinalizing = getattr(sys, 'is_finalizing', True)()

```

source/comtypesMonkeyPatches.py
```diff
@@ -95,14 +96,21 @@ def new__call__(self,*args,**kwargs):
 # This causes Release() to be called more than it should, which is very nasty and will eventually cause us to access pointers which have been freed.
 from comtypes import _compointer_base
 
-_cpbDel = _compointer_base.__del__
+_compointer_base._oldCpbDel = _compointer_base.__del__
 def newCpbDel(self):
+	# __del__ may be called while Python is exiting.
+	# In this state, global symbols may be set to None
+	# Therefore avoid calling into garbageHandler.
```
feerrenrut, line 103: Might be worth adding "or `log`, unless `isFinalizing` is checked first to ensure they are still available"

feerrenrut (changes requested):

source/comtypesMonkeyPatches.py
CyrilleB79: I did not review this whole PR. But @feerrenrut, your suggestion does not seem correct to me. @michaelDCurran's initial code seems correct.
Did you notice the parenthesis pair at the end of the line?
If is_finalizing is not an attribute of sys, True will be called as a function, what does not seem intended.

michaelDCurran: Note that:
* sys.is_finalizer is a function.
*  sys is a module, except for on exit it may be None.
* ```is_finalizer = getattr(sys, 'is_finalizer', lambda: True)()``` works as follows:
 * if sys is the valid sys module, getattr will return the is_finalizer function
 * If sys is None, getattr will return a lambda that if executed will return True
 * We then execute the result from getattr (executing either the real is_finalizer function, or the lambda).
Note the reason why we can't just check if sys is None by itself is that it may become None after that check. Thus, we use getattr as it is atomic.

CyrilleB79: If you confirm it is atomic, maybe you could use a try ... except AtributeError ... statement instead. This would be more pythonic and self-explaining IMO. Just a suggestion.

feerrenrut: > Did you notice the parenthesis pair at the end of the line?

My mistake, I missed the parenthesis. I didn't realize that `sys.is_finalizer` was a function. I think this could be made easier to read. Eg as @CyrilleB79 suggests in a try block, or like:
```python
isFinalizingFunc = getattr(sys, 'is_finalizing', lambda: True)
isFinalizing = isFinalizingFunc()
```

michaelDCurran:
@feerrenrut  I have split that on to two lines as you suggested.

feerrenrut (approved):

PR #11421, opened 2020-07-25
Support auto complete IntelliSense in SSMS and Visual Studio 2017

LeonarddeR:
### Link to issue number:
Related to #7504 

### Summary of the issue:
In Microsoft SQL Server Management Studio, IntelliSense does not work.

### Description of how this pull request fixes the issue:
I implemented support for IntelliSense using the UIA Element Selected event. In the past this was very slow, but with selective UIA event registration enabled, this works pretty nice. Having said that, even without selective registration enabled it at least reads results, though with somewhat more delay.

### Testing performed:
Tested auto complete IntelliSense in SSMS 18. SSMS 18 has the same Accessibility backend as Visual Studio 2017, so it should also work there. I'm not sure whether VS2017 is still used somewhere though, probably in some corporate environments.
I can also provide an implementation based on an app module.

### Known issues with pull request:
Not tested with Visual Studio 2017. May be @Luke-Robinett is interested? Otherwise we can also consider making this SSMS specific.

### Change log entry:
* New features
    + Support for auto complete suggestions (IntelliSense) Microsoft SQL Server Management Studio and Visual Studio 2017. (#7504)

Issue #12146, opened 2021-03-10
NVDA causes the VS 2019 .NET 5.0 Forms Designer to lock up


MarcoZehe:
### Steps to reproduce:

1. Run Visual Studio Community Edition 2019 and update to the latest release v16.9.1.
2. Create a new WinForms App (.NET framework). make sure that it is the second one in the default list of templates. The first one is for .NET 4.7.2, but we want the second one which is for .NET 5.0.
3. Go through the rest of the wizard and let it generate itself. It will first complain about .NET 5.0 not yet being installed, and install it in the background.
4. Once the form designer gains focus, press CTRL+Alt+X to go to the toolbox, tap L a b for Label, and press Enter to add a label to the blank form.
5. This is when things start to get bad. NVDA will slow down with VS 2019, keystrokes take long to process. To test, open and close the View menu, for example.
6. Now, while still in the Form designer, press Tab.
7. You may still be OK, or you may already be frozen. You can try to press F7. If it opens the editor with your forms1.cs file, you're still good. Press Shift+F7 to go back to the form designer, and maybe press F4 to go to properties view.
8. Repeat steps 6 and 7 a few times, trying to tab around and editing properties.

### Actual behavior:

NVDA is very sluggish with VS 2019. And eventually it will cause VS 2019 to become unresponsive (the title bar gets the "Not Responding" text, and it won't recover from it, even after an NVDA restart. Observations:

1. It is worse with Braille active than without, but even without, you'll eventually manage to hang VS.
2. The WinTen Apps add-on has a bearing, too, although with the 20210310 version, this is less of a problem. @josephsl has the details on what he did to fix it, it had something to do with a COM exception being raised.
3. All of this does not happen with JAWS 2021 latest update. JAWS, too, does not speak anything when tabbing through the .NET 5.0 WinForms designer, but at least it doesn't cause VS 2019 to freeze and eventually crash.

### Expected behavior:

NVDA should be smooth and not cause hangs or crashes in VS 2019.

#### NVDA installed/portable/running from source:
 Installed And Portable.
#### Name and version of other software in use when reproducing the issue: 

Microsoft Visual Studio Community 2019
Version 16.9.1
VisualStudio.16.Release/16.9.1+31105.61
Microsoft .NET Framework
Version 4.8.04084

Installierte Version: Community

Visual C++ 2019   00435-60000-00000-AA291
Microsoft Visual C++ 2019

Allgemeine Azure-Tools   1.10
[...]

PR #12170, opened 2021-03-14
Performance increase interacting with Visual Studio

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
Users are reporting bad performance in NVDA 2020.4 when interacting with Visual Studio.
Performance seemed to decrease after NVDA 2020.3 due to PR #11421.
In NVDAObjects.UIA.visualStudio.findExtraOverlayClasses, parent is fetched for any object given. This causes a great deal of recursion as many ancestors of a given object in Visual Studio will also run through this findExtraOverlayClasses.

In NVDAObjects.UIA.visualStudio.findExtraOverlayClasses: make sure to only fetch parent when absolutely necessary. 
Specifically, reverse a check that would look at parent's type and then check a UIA property on self.

### Testing strategy:
* Moved up and down Solution explorer, opening and closing branches of the solution. Confirming that there is much less of a delay.
* Tabbing into a property page in Visual Studio options. NVDA starts speaking sometimes up to 4 times quicker.

Other users should try and confirm that performance has improved for them.
Users should confirm that support for auto complete / intellisense s is still read correctly by NVDA.

None known.

### Change log entry:
Bug fixes:
* Performance when interacting with Visual Studio has been improved.

### Code Review Checklist:

- [ ] Unit tests.
- [ ] Manual tests.
- [ ] User Documentation.
- [ ] Change log entry.

### DIFF
--- a/source/NVDAObjects/UIA/VisualStudio.py
+++ b/source/NVDAObjects/UIA/VisualStudio.py
@@ -85,7 +85,7 @@ class CompletionToolTip(ToolTip):
 def findExtraOverlayClasses(obj, clsList):
 	if obj.UIAAutomationId in _INTELLISENSE_LIST_AUTOMATION_IDS:
 		clsList.insert(0, IntelliSenseList)
-	elif isinstance(obj.parent, IntelliSenseList) and obj.UIAElement.cachedClassName == "IntellisenseMenuItem":
+	elif obj.UIAElement.cachedClassName == "IntellisenseMenuItem" and isinstance(obj.parent, IntelliSenseList):
 		clsList.insert(0, IntelliSenseItem)
 	elif (
 		obj.UIAElement.cachedClassName == "LiveTextBlock"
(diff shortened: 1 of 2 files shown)


cary-rowen:
It seems that he has indeed improved performance, can it be further optimized?
In addition, NVDA and VS freeze together in the Winform form designer, which can also be reproduced in 2020.4.

thanks.

LeonarddeR (approved):
I'm very sorry for causing this regression

cary-rowen:
Maybe someone else can test this build.

cary-rowen:
I tested the latest build version and it still freezes in the window designer and causes VS to crash. The crash happened after adding a combo box to the window.
The following is the log:
[nvda.log](https://github.com/nvaccess/nvda/files/6145236/nvda.log)

leonardder:
> I tested the latest build version 

TO make sure, did you test latest Alpha or the build for this pull request? Also, did the problem you describe occur in 2020.3 as well? In latter case, this pr won't fix this. It will only fix a regression introduced in NVDA 2020.4.

cary-rowen:
@leonardder 
Yes, I tested the latest Alpha Version and the issue exists, but it should not be a regression from 2020.4. And the issue only appears within the form designer of .net5 WinForm apps.
If this is the case, I will try to open a new Issue.
Thanks!

cary-rowen:
I have confirmed that my situation is exactly the same as #12146.
@leonardder

PR #12174, opened 2021-03-15
Chrome system tests: ensure the page is fully loaded (including all iframes) before testing the page

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
Recently some of our system tests for Chrome started breaking. Specifically the tests that include an iframe.
It looks as if the test logic is starting to run before the content in the iframe is fully loaded.

Place an initial line at the top of the test page, before the start marker. NVDA will initially land on this line as soon as the document gets focus, but possibly before iframes have fully loaded.
The page now has an onload event which forces focus to the start marker. This onload event only fires once all content (including iframes) in the document are fully loaded.
When the start marker finally gets focus, NVDA will detect this and start running the test logic.

### Testing strategy:
Successfully ran all system tests locally.

None known.

### Change log entry:
None needed.

### Code Review Checklist:

- [ ] Unit tests.
- [ ] Manual tests.
- [ ] User Documentation.
- [ ] Change log entry.

### DIFF
--- a/tests/system/libraries/SystemTestSpy/speechSpyGlobalPlugin.py
+++ b/tests/system/libraries/SystemTestSpy/speechSpyGlobalPlugin.py
@@ -68,7 +68,8 @@ class NVDASpyLib:
 
 	# callbacks for extension points
 	def _onNvdaStartupComplete(self):
-		self._isNvdaStartupComplete = True
+		queueHandler.queueFunction(queueHandler.eventQueue, lambda: setattr(self, '_isNvdaStartupComplete', True))
+		#self._isNvdaStartupComplete = True
 
 	def _onNvdaSpeech(self, speechSequence=None):
 		if not speechSequence:
--- a/tests/system/libraries/SystemTestSpy/speechSpyGlobalPlugin.py
+++ b/tests/system/libraries/SystemTestSpy/speechSpyGlobalPlugin.py
@@ -68,9 +68,7 @@ class NVDASpyLib:
 
 	# callbacks for extension points
 	def _onNvdaStartupComplete(self):
-		# Queue the setting of the completion variable,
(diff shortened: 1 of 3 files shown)


tests/system/libraries/SystemTestSpy/speechSpyGlobalPlugin.py
```diff
@@ -68,7 +68,9 @@ def _registerWithExtensionPoints(self):
 
 	# callbacks for extension points
 	def _onNvdaStartupComplete(self):
-		self._isNvdaStartupComplete = True
+		# Queue the setting of the completion variable,
+		# To ensure that NvDA's core loop has started running, and it has processed initial focus.
+		queueHandler.queueFunction(queueHandler.eventQueue, lambda: setattr(self, '_isNvdaStartupComplete', True))
```
feerrenrut, line 73: Perhaps the code at `source/core.py:564: postNvdaStartup.notify()` should be queued instead? If we are saying the loop must have started before NVDA's startup is complete, then the `postNvdaStartup` action is incorrect.

feerrenrut (changes requested):
Overall looks good, couple of minor suggestions.

tests/system/libraries/SystemTestSpy/speechSpyGlobalPlugin.py
michaelDCurran: I think for system tests it makes sense to ensure the loop has started and initial focus is processed. But for other uses I'm not sure. I'm more inclined just to change the system tests for now, rather than change the behaviour of that action, which other 3rd party code *could* be using.

feerrenrut (approved):
Looks good to me, assuming the tests pass.

Issue #10916, opened 2020-03-30
UIA handler (backward compatibility): unify public and private modules under a package


josephsl:
Hi,

The following should wait until work on next incompatible version of NVDA begins:

### Is your feature request related to a problem? Please describe.
For years, UIA handler module was split into two parts: a publicly accessible module (UIAHandler.py) and a private module (_UIAHandler.py). This split was necessary in 2011 in order for NVDA to run without UI Automation client support (namely on Windows XP, Server 2003, and old releases of Windows Vista and Server 2008). Additionally, private UIA handler module wasn't used if UIA was forcefully disabled. Lastly, other accessibility API handlers such as JAB and IAccessible are contained inside one module.

### Describe the solution you'd like
Unify public and private UIA handler modules into one file.

### Describe alternatives you've considered
Leave the UIA handler arrangement as is.

### Additional context
Consider that:

1. NVDA supports Windows 7, Server 2008 R2 and later, which comes with a bit more robust UIA support than old Windows releases, but not up to the level as seen with Windows 8/Server 2012 and later.
2. If UIA is forcefully disabled, an error is raised.
3. Other accessibility API handlers are housed inside one file per API, whereas UIA is split in two. This means majority of UIA support improvements are housed inside the private module, with the public module importing everything.

As this proposal deals with an important accessibility API handler, I propose this change as a backwards incompatible change.

Thanks.

feerrenrut:
I understand that the initial rationale for splitting this module no longer applies, but can you add to this the benefits to this refactor? 

As with any refactor there is risk of breakage, or as already pointed out breaking backwards compatibility. The benefits have to outweigh this.

josephsl:
Hi, the biggest benefit to this move is module content consistency with other accessibility API’s: namely editing only one module would be sufficient rather than thinking about public interface versus implementation. Another benefit is code deduplication: debug flag from private UIA interface is imported by the public after all module contents are imported, so this flag is imported twice. Combined, these benefits will allow ease of future maintenance for people wishing to work with UIA support. At the moment there are no known modules that import the private module directly, as NVDA will assume that public module includes everything from the private module.

feerrenrut:
One benefit from having a private module is that we can change the private interface without breaking compatibility. Generally I wouldn't want to refactor code unless it is already being modified for some other purpose.

josephsl:
[...]

Issue #10833, opened 2020-03-02
Unable to run NVDA alpha


CyrilleB79:
### Steps to reproduce:

Build and run NVDA alpha.

### Actual behavior:

Critical error. Here is the log:

<details> <summary>Log</summary>
INFO - __main__ (18:18:04.681) - MainThread (4004):
Starting NVDA version source-master-1e139d4
INFO - core.main (18:18:04.919) - MainThread (4004):
Config dir: D:\Cyrille\Dev\GIT\nvdaNew\source\userConfig
INFO - config.ConfigManager._loadConfig (18:18:04.919) - MainThread (4004):
Loading config: .\userConfig\nvda.ini
INFO - core.main (18:18:05.036) - MainThread (4004):
Using Windows version 10.0.18362 workstation
INFO - core.main (18:18:05.036) - MainThread (4004):
Using Python version 3.7.5 (tags/v3.7.5:5c02a39a0b, Oct 14 2019, 23:09:19) [MSC v.1916 32 bit (Intel)]
INFO - core.main (18:18:05.036) - MainThread (4004):
Using comtypes version 1.1.7
INFO - core.main (18:18:05.036) - MainThread (4004):
Using configobj version 5.1.0 with validate version 1.0.1
CRITICAL - __main__ (18:18:05.451) - MainThread (4004):
core failure
Traceback (most recent call last):
  File "source\nvda.pyw", line 215, in <module>
  File "core.py", line 256, in main
    import appModuleHandler
  File "appModuleHandler.py", line 27, in <module>
    import NVDAHelper
  File "NVDAHelper.py", line 19, in <module>
    import eventHandler
  File "eventHandler.py", line 9, in <module>
    import api
  File "api.py", line 19, in <module>
    import NVDAObjects.IAccessible
  File "NVDAObjects\IAccessible\__init__.py", line 22, in <module>
    import IAccessibleHandler
  File "IAccessibleHandler.py", line 18, in <module>
    import UIAHandler
  File "UIAHandler.py", line 11, in <module>
    from _UIAHandler import *  # noqa: F403
  File "_UIAHandler.py", line 29, in <module>
    from comtypes.gen import UIAutomationClient as UIA
ImportError: cannot import name 'UIAutomationClient' from 'comtypes.gen' (D:\Cyrille\Dev\GIT\nvdaNew\include\comtypes\comtypes\gen\__init__.py)

</details>

### Expected behavior:

NVDA running.

#### NVDA installed/portable/running from source:
[...]

PR #12201, opened 2021-03-19
Make com interfaces ide friendly

feerrenrut:
### Link to issue number:
None
Related: https://github.com/nvaccess/nvda/issues/10916

### Summary of the issue:
When working on modules that make heavy use of `comInterfaces` support from tools or IDEs would be helpful.
The files in `comInterfaces` are generated by `comtypes.gen`.
There are pairs of files, a module named with the `comInterface` ID, and one with a friendly name that imports symbols from the ID file.
The friendly name files generated by `comtypes.gen` expose symbols using runtime logic which some tools can not follow.
Making these symbols appear as errors/warnings about unknown symbols.

This PR modifies the build scripts to edit the friendly name file to include a fallback import statement which tools are more likely to be able to follow.

Essentially, the following:
```python
from comtypes.gen import _C523F390_9C83_11D3_9094_00104BD0D535_0_3_0
globals().update(_C523F390_9C83_11D3_9094_00104BD0D535_0_3_0.__dict__)
__name__ = 'comtypes.gen.AcrobatAccessLib'
```
becomes:
```python
try:
	from comtypes.gen import _C523F390_9C83_11D3_9094_00104BD0D535_0_3_0
except ModuleNotFoundError:
	import _C523F390_9C83_11D3_9094_00104BD0D535_0_3_0
	from _C523F390_9C83_11D3_9094_00104BD0D535_0_3_0 import *
globals().update(_C523F390_9C83_11D3_9094_00104BD0D535_0_3_0.__dict__)
__name__ = 'comtypes.gen.AcrobatAccessLib'
```

In addition, I noticed that Accessibility.py is generated, despite being commented out by the sconscript.
Presumably some other interface has a dependence on it?
Regardless this is used heavily in NVDA so it is now explicitly included.

The UIAutomationClient interface is not being generated, and instead was manually checked in.
This is because the version provided by Appveyor (and thus included in NVDA official builds) may not be as up to date as it could be. Instead, as new features of UIA are required this file must be manually updated by generating it on a machine that has the latest version available.

### Testing strategy:
Built and ran locally.
Will test the appveyor build.
Ask community to test building and running on different versions of Windows.

None.

### Change log entry:

Changes for devlopers:
```
- comInterfaces are now more easily consumable by developer tools such as IDEs. 
```
### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.


leonardder:
> The UIAutomationClient interface was not being generated, and instead was manually checked in.

This was and is on purpose. The version of the com interface is tied to the operating system. If a newer version of the interface is released with a new version of Windows, it is usually not yet available in the appveyor image.

josephsl:
Hi, for example, dialog detection through UIA was possible by using UIA COM interfaces from Windows SDK 10.0.17763, and I remember having to update UIA COM interface file then (NVDA 2018.3). This becomes more important once Microsoft releases IUIAutomation7 interface (saw this interface in a WIP build in 2019 but didn’t see it since then), and I look closely at changes to UIAutomationCore.dll whenever dev channel builds come out to spot and support UIA interface changes. When we do update the UIA COM interface, we wait until a Windows 10 feature update with COM interface changes to enter release preview channel before sending in a pull request to update the interface (same can be said about Windows 10 feature update/build map), or in some cases, I do my best to detect parts of the changed COM interface early in WinTenApps add-on. Thanks.

lukaszgo1:
Can't we use .idl files from Windows 10 SDK to generate UIAutomationClient COM interface similar to what were doing for IA2? We can require specific version of Windows 10 SDK and it isn't tied to the version of Windows in use.

CyrilleB79:
Could this PR change the behaviour described in #10833 (scons --clean deletes some files that are fetched from the repo)?
Thanks.

michaelDCurran:
@feerrenrut  yes several other typelibraries would have depended on the accessibility typelibrary. IAccessible2, and UI Automation being two of them.

source/comInterfaces_sconscript
```diff
@@ -26,18 +30,62 @@ def new_my_import(fullname):
...
+	globals().update(_C523F390_9C83_11D3_9094_00104BD0D535_0_3_0.__dict__)
+	__name__ = 'comtypes.gen.AcrobatAccessLib'
+	```
+	Become:
+	``` python
+	try:
+		from comtypes.gen import _C523F390_9C83_11D3_9094_00104BD0D535_0_3_0
+	except ModuleNotFoundError:
+		from _C523F390_9C83_11D3_9094_00104BD0D535_0_3_0 import *
+	globals().update(_C523F390_9C83_11D3_9094_00104BD0D535_0_3_0.__dict__)
```
michaelDCurran, line 49: This example is missing an import in the except block, and therefore does not reflect what it really does below.

source/comInterfaces_sconscript
```diff
@@ -26,18 +30,62 @@ def new_my_import(fullname):
...
+	importTemplate = (
+"""try:
+	from comtypes.gen import {libIdentifier}
+except ModuleNotFoundError:
+	import {libIdentifier}
+	from {libIdentifier} import *
+""")
+
+	importPattern = re.compile(r"from comtypes\.gen import ([\w]+)\n")
+	for line in fileinput.input(path, inplace=True):
```
michaelDCurran, line 63: A comment here mentioning that fileinput.Input temporarily replaces sys.stdout would be useful.

michaelDCurran (changes requested):
We don't quite know what version of UIAutomationCore Appveyor has... I think its Windows server 2019 or something. Thus we cannot guarantee that UI Automation client will contain all possible interfaces and constants we need.
We need to either keep our pre-generated version of UIAutomationClient COM interfaces, or perhaps include uiAutomationCore.dll in miscDeps... though I think the latter would be breaking licensing restrictions.

source/comInterfaces_sconscript
feerrenrut: Yep. I think I can remove this example, I don't think it really helps to clarify anything now.

source/comInterfaces_sconscript
feerrenrut: Ah yes, you are right. It is an unintuitive library.

michaelDCurran (approved):

PR #11598, opened 2020-09-12
Report appointment categories in Microsoft Outlook

LeonarddeR:
### Link to issue number:
None

### Summary of the issue:
Outlook appointments could have multiple categories (or colors) assigned to them. However, they are not reported.

### Description of how this pull request fixes the issue:
Fetch the categories from the object model and report them when appropriate.

### Testing performed:
Tested appointment without, with one and with multiple categories.

#### Details
To reproduce / test:
* Open Outlook
* Press control+2 to go to calendar view
* Tab or shift+tab to focus an appointment

Note: using up/down arrows will read the appointment, but does not focus it and will not read the category.

Tested in Outlook 2016 and 365
- with Calendar view 'working week' and day.

If you do not have an appointment with a category, you can add one the following way:
* Open Outlook
* Press control+2 to go to calendar view (working week for me)
* Tab to an appointment
* Press context menu key
* Select "categorize" item
* Select a category in the sub-menu

### Known issues with pull request:
None known

### Change log entry:
* New features
    + NVDA now reports the categories assigned to an appointment in Microsoft Outlook, if any. (#11598)

PR #12210, opened 2021-03-22
Optional experimental support for Microsoft Excel via UI Automation

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
Since 2013, Microsoft has been slowly adding a UI Automation implementation to Microsoft Excel. For quite some time, this implementation was missing too much to be at all useful for NVDA. However, in the last year, Microsoft has increased its effort in this area, starting to fill many of the gaps in the implementation, providing more of a reason for NVDA to start consuming this experimentally, with the goal of one day switching to UI Automation in Excel by default.
Not only would this provide significant performance improvemts for NVDA with Excel, but it would also enable scenarios such as using Excel in Application Guard, or remotely via Windows Virtual Desktop, where access to the object model and in-process injection is no longer possible.
 
NVDA will fall back to relying on UI Automation In Microsoft Excel if available, if NVDA is not able to inject in-process.
A new option has been added to NVDA's Advanced settings, which when enabled, forces NVDA to use UI Automation in Microsoft Excel always when available. 
Although so far navigating, editing and querying cells is all supported, the following features are not yet implemented:
* NvDA Elements List for listing formulas, comments etc. The needed UI Automation extensions are not yet available.
* Browse mode quick nav. Again, same reason as above.
* Setting / consuming screen reader specific column and row headers. This feature probably will not be supported. Instead it is recommended to make these particular sets of data into tables, marking the needed rows and columns as headers in the standrd way for modern Microsoft Excel.

As some Excel-specific information on cells does not quite fit with standard NVDA concepts. A new Cell Appearance script (NVDA+o) has been added, which presents a browse mode document, listing these specific bits of information. This script may be changed or removed in future as it is possible to find this information out by looking in the Cell formatting dialog etc anyway, but for now it is there to demonstrate NVDa's ability to fetch this kind of information from UI Automation.
 
Many of Excel's specific features have been exposed by Microsoft through UI automation via either annotations, or custom properties.
Standard UIA annotations that NvDA consumes for this support are:
* comment: For notes and comments. NVDA exposes the "has comment" state on cells, and reports the comment with a reportComment script (NVDA+alt+c).
* Data validation error: the error text is included in the cell's description.
* Formula error: the error text is included in the cell's description.
* Circular reference error: the error text is included in the cell's description.

Microsoft has made use of UI Automation's custom property feature in order to expose  many more Excel-specific properties: https://docs.microsoft.com/en-us/office/uia/excel/excelcustomproperties
This pr adds an infrastructure in UIAHandler / UIA NVDAObjects which allows registering and accessing these custom property IDs in a standard way.
The properties that Microsoft exposes in Excel are:
* cellFormula: mapped to NvDA's hasFormula state.
* cellNumberFormat: exposed in Cell appearence script (NvDA+o). Note however that Excel currently exposes the raw template value and not the friendly name. So right now this is not as useful as it could be.
* hasDataValidation: Exposed in Cell appearence script (NvDA+o)
* hasDataValidationDropdown: Mapped to NvDA's submenu state.
* dataValidationPrompt: exposed in cell description and Cell appearence script (NvDA+o)
* hasConditionalFormatting: exposed in Cell appearence script (NvDA+o) 
* areGridlinesVisible: exposed in Cell appearence script (NvDA+o)
* commentReplyCount: Included in reportComment script (NvDA+alt+c).

The above annotations and properties have only been available in Microsoft Excel since build 16.0.13522.10000.
Thus turning on this option will only be useful for this build or higher.

### Testing strategy:
Manual testing with [this spreadsheet](https://github.com/nvaccess/nvda/files/6185717/excel.custom.properties.tests.xlsx) which contains examples of each of the properties and annotations.

* No support for Elements List
* No support for browse mode quick nav

### Change log entry:
What's new:
* Optional experimental support for Microsoft Excel via UI Automation

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [ ] Change log entry.


michaelDCurran:
@feerrenrut  There are 5 more commits for you to review which were added after your last rebase of this branch. These comments add further functionality exposed by existing standard UI Automation annotations, and completes all possible support we can do with what Microsoft has currently exposed.

lukaszgo1:
The option to use UIA for Excel should be documented in the User Guide. Don't forget to add context help to it.

codeofdusk:
Nice to see that this is finally PR ready after nearly two years of work!

user_docs/en/userGuide.t2t
```diff
@@ -1850,6 +1850,14 @@ The combo box has the following options:
 - No: Don't use UIA, even if NVDA is unable to inject in process. This may be useful for developers debugging issues with IA2 and want to ensure that NVDA does not fall back to UIA.
 -
 
+==== Use UI automation to access Microsoft  Excel spreadsheet controls when available ====[AdvancedSettingsUseUiaForExcel]
+When this option is enabled, NVDA will try to use the Microsoft UI Automation accessibility API in order to fetch information from Microsoft Excel Spreadsheet controls.
+This is an experimental feature, and some features of Microsoft Excel may not be available in this mode.
+Examples being NVDA's Elements List for listing formulas and comments, and Browse mode quick navigation to jump to form fields on a spreadsheet.
```
feerrenrut, line 1856:
```suggestion
For instance, NVDA's Elements List for listing formulas and comments, and Browse mode quick navigation to jump to form fields on a spreadsheet features are not available.

```

feerrenrut (changes requested):
Some minor changes requested. I think that this work can be squash merged.

source/NVDAObjects/UIA/excel.py
```diff
@@ -0,0 +1,458 @@
...
+	columnHeaderText = None
+
+	#: Typing information for auto-property: _get_areGridlinesVisible
+	areGridlinesVisible: bool
+
+	def _get_areGridlinesVisible(self) -> bool:
+		parent = self.parent
+		# There will be at least one grid element between the cell and the sheet.
+		# There could be multiple as there might be a data table defined on the sheet.
+		while parent.role == controlTypes.ROLE_TABLE:
```
leonardder, line 115: Make sure that this loop covers `parent is None`. I know it is very unlikely, but better safe than sorry,

source/NVDAObjects/UIA/excel.py
```diff
@@ -0,0 +1,458 @@
...
+			# Translators: The height of the cell in points
+			"Cell height: {0.y:.1f} pt"
+		)
+		infoList.append(tmpl.format(self.cellSize))
+
+		if self.rotation is not None:
+			tmpl = pgettext(
+				"excel-UIA",
+				# Translators: The rotation in degrees of an Excel cell
+				"Rotation: {0} degrees"
```
leonardder, line 205: Have you considered npgettext here, degree vs degrees?

source/NVDAObjects/UIA/excel.py
```diff
@@ -0,0 +1,458 @@
...
+		else:
+			# Translators: A message in Excel when there is no note
+			ui.message(_("No note or comment thread on this cell"))
+
+
+class ExcelWorksheet(ExcelObject):
+	role = controlTypes.ROLE_TABLE
+
+	# The grid UIAElement dies each time the sheet is scrolled.
+	# therefore this grid would be announced in the focus ancestory each time which is bad.
```
leonardder, line 439: Typo, start with uppercase

```suggestion
	# Therefore this grid would be announced in the focus ancestory each time which is bad.
```

source/NVDAObjects/UIA/excel.py
```diff
@@ -0,0 +1,458 @@
...
+	# Suppress this.
+	isPresentableFocusAncestor = False
+
+	def _get_parent(self):
+		parent = super().parent
+		# We want to present the parent (sheet) in the focus ancestry
+		# As that is what has the sheet name.
+		parent.isPresentableFocusAncestor = True
```
leonardder, line 447: This looks slightly hacky to me. Couldn't we do this on an overlay?

source/NVDAObjects/UIA/excel.py
```diff
@@ -0,0 +1,458 @@
...
+		parent.states.discard(controlTypes.STATE_SELECTED)
+		return parent
+
+
+class CellEdit(ExcelObject):
+	name = ""
+
+
+class BadExcelFormulaEdit(ExcelObject):
+	shouldAllowUIAFocusEvent = False
```
leonardder, line 458: Could we have more context on this object, may be in its doc string?

source/_UIAConstants.py
```diff
@@ -0,0 +1,62 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2021 NV Access Limited
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+
+import enum
+
+
+class FillType(enum.IntEnum):
+	none = 0
```
leonardder, line 10: Please make the enum constants fully uppercase for consistency

source/_UIAConstants.py
```diff
@@ -0,0 +1,62 @@
...
+	# Translators: a style of fill type (to color the inside of a control or text)
+	FillType.picture: pgettext("UIAHandler.FillType", "picture"),
+	# Translators: a style of fill type (to color the inside of a control or text)
+	FillType.pattern: pgettext("UIAHandler.FillType", "pattern"),
+}
+
+
+# Some newer UIA constants that could be missing
+class UIAutomationType(enum.IntEnum):
+	Int = 1
```
leonardder, line 33: Idem as above, please use upper case

source/_UIAConstants.py
```diff
@@ -0,0 +1,62 @@
...
+	Array = 8
+	Out = 9
+	IntArray = 10
+	BoolArray = 11
+	StringArray = 12
+	DoubleArray = 13
+	PointArray = 14
+	RectArray = 15
+	ElementArray = 16
+	OutInt = 17
```
leonardder, line 49:
```suggestion
	OUT_INT = 17
```

LeonarddeR (changes requested):
I'm very, very happy to see a pr for this

source/NVDAObjects/UIA/__init__.py
```diff
@@ -1219,6 +1219,21 @@ def _get_selectionContainer(self) -> "typing.Optional[UIA]":
 			return obj
 		return None
 
+	def _get_UIAAnnotationObjects(self):
```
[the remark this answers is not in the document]
feerrenrut:
```suggestion
	def _get_UIAAnnotationObjects(self) -> typing.Dict[int, UIAHandler.IUIAutomationElement]:

```

source/NVDAObjects/UIA/excel.py
michaelDCurran: It is pretty hard to identify the sheet element as it contains now className or automationID. We could do an overlay class, but there is a risk it could decrease performance when looking for parents like that. Doing it this way is closest to how it is actually used as presentableFocusAncestor is only relevent when going parent.

leonardder: Could you add that in a comment so when reading back this code, we have a rationale as of why it has been done this way? In that case, I'm ok with it..

source/_UIAConstants.py
michaelDCurran: My goal was to keep the constants as close to what is in UIAutomationCore.idl. They start with an uppercase letter, and contain no underscores. Similar to other constants in UI automation.
I can convert all these to uppercase, but I'm not so sure on the underscore, or what the rule there is. But I want to be really sure this is necessary as it will be a big change which needs to be checked.
@feerrenrut  your thoughts?

source/NVDAObjects/UIA/excel.py
michaelDCurran: I don't believe we have any other code in NVDA that uses this? Do we know this works okay with poedit / the rest of our translation system?
Perhaps something to test outside of this pr.

leonardder: I tested it with #11598 as per @feerrenrut's request after @bramd suggested it

michaelDCurran: Ah good to know. However, in this case rotation is a float not an int, so even for 1 degree it will be 1.0 degrees.

leonardder: I'm convinced, feel free to leave it as is 😊

source/_UIAConstants.py
feerrenrut: I'd like to have a consistent style, and we have previously chosen uppercase for enums. I understand the desire to reflect UI Automation, but I think it is also good to make our code clearly ours. I should be able to update this fairly quickly with an IDE, I'll push a change shortly.

feerrenrut:
@michaelDCurran I have renamed those enum values.
@leonardder I believe all your feedback has been addressed.

LeonarddeR (approved):
🙌

user_docs/en/userGuide.t2t
```diff
@@ -1850,6 +1850,14 @@ The combo box has the following options:
 - No: Don't use UIA, even if NVDA is unable to inject in process. This may be useful for developers debugging issues with IA2 and want to ensure that NVDA does not fall back to UIA.
 -
 
+==== Use UI automation to access Microsoft  Excel spreadsheet controls when available ====[AdvancedSettingsUseUiaForExcel]
```
CyrilleB79, line 1853: In https://github.com/nvaccess/nvda/pull/12034#issuecomment-773847508, @feerrenrut recommanded not to include the category name in the tag, especially for advanced settings that may migrate in another category one day.

```suggestion
==== Use UI automation to access Microsoft  Excel spreadsheet controls when available ====[UseUiaForExcel]

```

michaelDCurran:
@feerrenrut I think you are still requesting changes on this pr...

feerrenrut (approved):

feerrenrut:
I've now marked as approved

source/gui/settingsDialogs.py
```diff
@@ -2566,6 +2566,13 @@ def __init__(self, parent):
 		self.UIAInMSWordCheckBox.SetValue(config.conf["UIA"]["useInMSWordWhenAvailable"])
 		self.UIAInMSWordCheckBox.defaultValue = self._getDefaultValue(["UIA", "useInMSWordWhenAvailable"])
 
+		# Translators: This is the label for a checkbox in the
+		#  Advanced settings panel.
+		label = _("Use UI Automation to access Microsoft &Excel spreadsheet controls when available")
+		self.UIAInMSExcelCheckBox = UIAGroup.addItem(wx.CheckBox(self, label=label))
```
CyrilleB79, line 2572: Please do not forget context help, as mentionned by @lukaszgo1.

```suggestion
		self.UIAInMSExcelCheckBox = UIAGroup.addItem(wx.CheckBox(self, label=label))
		self.bindHelpEvent("UseUiaForExcel", self.UIAInMSExcelCheckBox)

```

PS: The first line did not change, but I do not know how to make a suggestion to insert one line without modifying the others in GitHub.

leonardder:
I will file a follow up pr for support of merged cells.

leonardder:
> 
> 
> I will file a follow up pr for support of merged cells.

Ugh, this looked easy since merged cells are reflected in their row and column span,  but it looks like it's not possible to get address and value that easy. Even when you include a merged cell in a selection, an error is generated:

```txt
  File "NVDAObjects\UIA\excel.py", line 370, in _get_cellCoordsText
    firstSelected = sc.GetCurrentPropertyValue(
AttributeError: 'NoneType' object has no attribute 'QueryInterface'
```

lijianwei2019:
There is a problem with advanced related options.@michaelDCurran

CyrilleB79:
Thanks. Fixed in #12223.

PR #12242, opened 2021-03-28
Microsoft Edge with UIA: support following same-page links 

michaelDCurran:
### Summary of the issue:
When following a same-page link in a web browser, the page is scrolled to the target of the same-page link, and it is expected that the screen reader will move its browse mode caret to that location as well.

Recently a new ActiveTextPositionchanged event was added to UI Automation to allow browsers to communicate this to screen readers. Only in the last week or so, Microsoft Edge Canary has started firing this event when same-page links are activated.
NVDA should make use of this.
 
NVDA now registeres for UI Automation's ActiveTextPositionChanged event globally, and when one is received, fires its own UIA_activetextPositionChanged NvDA event. This NVDA event is implemented on UIABrowseModeDocument, which uses BrowseMode's _handleScrollTo to move the browse mode caret to the new location.
_handleScrollTo had to be changed to accept not just an NvDAObject, but now also a TextInfo, as the ActiveTextPositionChanged event is actually passed a IUIAutomationTextRange.

### Testing strategy:
* Enabled UIA in Chromium browsers in NvDA's advanced settings
* Loaded the NvDA User Guide in Microsoft Edge Canary
* Followed one of the same-page links in the table of contents.
* Confirmed that the browse mode caret moved to the target of the same-page link (above the heading) and that then pressing down arrow or starting a sayAll would read the appropriate content.

None.

### Change log entry:
Same-page links are now supported in Microsoft Edge when using UI automation.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### DIFF
--- a/source/_UIAHandler.py
+++ b/source/_UIAHandler.py
@@ -234,7 +234,8 @@ class UIAHandler(COMObject):
 		UIA.IUIAutomationEventHandler,
 		UIA.IUIAutomationFocusChangedEventHandler,
 		UIA.IUIAutomationPropertyChangedEventHandler,
-		UIA.IUIAutomationNotificationEventHandler
+		UIA.IUIAutomationNotificationEventHandler,
+		UIA.IUIAutomationActiveTextPositionChangedEventHandler,
 	]
 
 	def __init__(self):
@@ -377,6 +378,12 @@ class UIAHandler(COMObject):
 		# #7984: add support for notification event (IUIAutomation5, part of Windows 10 build 16299 and later).
 		if isinstance(self.clientObject, UIA.IUIAutomation5):
 			self.globalEventHandlerGroup.AddNotificationEventHandler(
+				UIA.TreeScope_Subtree,
+				self.baseCacheRequest,
+				self
+			)
+		if isinstance(self.clientObject, UIA.IUIAutomation6):
+			self.globalEventHandlerGroup.AddActiveTextPositionChangedEventHandler(
 				UIA.TreeScope_Subtree,
 				self.baseCacheRequest,
 				self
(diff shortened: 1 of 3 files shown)


source/_UIAHandler.py
```diff
@@ -670,6 +677,34 @@ def IUIAutomationNotificationEventHandler_HandleNotificationEvent(
...
+		if not self.MTAThreadInitEvent.isSet():
+			# UIAHandler hasn't finished initialising yet, so just ignore this event.
+			if _isDebug():
+				log.debug("HandleActiveTextPositionchangedEvent: event received while not fully initialized")
+			return
+		import NVDAObjects.UIA
+		try:
+			obj = NVDAObjects.UIA.UIA(UIAElement=sender)
+		except Exception:
+			if _isDebug():
```
feerrenrut, line 694: I assume this would be a rare scenario? It might be worth having this log message show without specific debugging category enabled.

source/browseMode.py
```diff
@@ -1649,10 +1651,15 @@ def _handleScrollTo(self, obj):
...
-			return False
+		if isinstance(obj, NVDAObject):
+			try:
+				scrollInfo = self.makeTextInfo(obj)
+			except (NotImplementedError, RuntimeError):
+				return False
+		elif isinstance(obj, textInfos.TextInfo):
+			scrollInfo = obj.copy()
+		else:
+			raise ValueError(f"{obj} is not a supported type")
```
feerrenrut, line 1662: Was it already true that `_handleScrollTo` could be called with either a textInfo or an NVDAObject?

source/_UIAHandler.py
michaelDCurran: All the other UIA event handlers also put this message behind _isDebug. In a perfect world this should not happen, but if the window is destroyed before we process the event, or perhaps the UIAElement dies before or as we make the NVDAObject, it could result in None. See ```source/NvDAObjects/__init__.py``` line 86.

source/browseMode.py
michaelDCurran: No, this is new. Previously it could only be passed in an NVDAObject which would then be made into a TextInfo. Now, you can pass in a TextInfo directly.

source/_UIAHandler.py
feerrenrut: I think if it is a rare scenario, and useful debug information, it probably doesn't need to be hidden behind a category.

source/browseMode.py
feerrenrut: In that case, can this method be kept simple and pass in the object? There doesn't seem to be any need to create the Text info outside?

michaelDCurran: The ActiveTextPositionChanged UI Automation event is fired presumably on the document (sender is the document) and it is passed in a IUIAutomationTextRange (textRange), which is the specific position within the document. 
We create a TextInfo from the IUIAutomationTextRange and use that as scrollInfo in _handleScrollTo. If we just used an NVDAObject this would not be enough as the NVDAObject would be for the document, and not for the specific position within the document. 
So _handleScrollTo was changed to allow passing ina TextInfo directly.

feerrenrut: Right, I didn't release that we didn't already have an appropriate NVDA object. It would help if `IUIAutomationActiveTextPositionChangedEventHandler_HandleActiveTextPositionChangedEvent` had type annotations.

feerrenrut: Sorry, I meant `event_UIA_activeTextPositionChanged`, but it would be nice if both had annotations

feerrenrut (approved):
I'll leave it up to you about adding annotations. Otherwise this can go ahead.

PR #12250, opened 2021-03-30
Fix NVDA crash when formatting a timestamp of a log entry under some versions of Universal CRT

### LINKED ISSUE #12160: Crash of the latest alpha NVDA builds on a system with Russian locale

lukaszgo1:
### Link to issue number:
Fixes #12160
Related to https://bugs.python.org/issue36792

### Summary of the issue:

As described in [This Python bug report](https://bugs.python.org/issue36792) (many thanks to @kvark128  for his research) under some versions of Universal CRT Python can crash when formatting time with `time.localtime` and locale is set to a language which contains underscore it its name.
Unfortunately Python developers decided  not to workaround this bug on their side since this crash is fixed in more recent version of the Universal CRT. Sadly the version containing this bug is present in Windows 10 1809 which is a base version for Server 2019 and Windows 10 LTSC 2019 both are supported until 2028 and we should support them too if possible. For users this manifests in a crash of NVDA since default log formatter attempts to format current time with `time.localtime` when writing anything to the log.

1. WX Python no longer tries to set Python locale to an nonexistent one when creating an instance of `wx.app`. While this is not necessary any longer when `time.localtime` is not used we are managing locale ourselves in `languageHandler` so it makes sense to do it just in one place.
2. When formatting time of the log entries `time.localtime` is not used any more. I've provided an alternative implementation which relies on win32 API calls to convert time_t to a string representation.

While at it I've also removed some unused imports from `logHandler` and cleaned up star imports from `winKernel`.
### Testing strategy:
On both Windows 7 and Windows 10 1809 in Polish started NVDA with the following language settings:
1. Language set to User default.
2. Language set to the locale without underscore in the name e.g. English, Polish.
3. Language containing underscore in the name e.g. de_CH.

Confirmed there are no crashes whereas under Windows 10 1809 it was impossible to start NVDA both when set to user default and when set to de_CH before fix from this pr.

In theory it should be possible to create an unit test for this (AppVeyor machines are on Server 2019 after all) annoyingly I am unable to reproduce this crash from the Python interpreter. I'll keep  trying but this should not block this PR.

- It might be more performant to use the new `formatTime` method only for Windows versions on which  UCRT has this bug, but since I am not  sure in which version of Windows 10 it started nor on which it has been fixed it seemed safer to always use this new approach.
- While I've made sure that `time.localtime` is not used anywhere in our code it is theoretically possible it is still used in one of our  dependencies or somewhere in standard Python library. This is very hard to track down so unless someone reports a crash I see no way to detect such cases.

### Change log entry:
Bug fixes:

It is once again possible to use NVDA in a languages containing underscores in the locale name such as de_CH on Windows 10 1803 and 1809.

Please note that this crash occurred also in 2020.4 for these languages hence the change log entry.

### Code Review Checklist:

- [X] Unit tests.
- [X] Manual tests.
- [X] User Documentation.
- [X] Change log entry.

### DIFF
--- a/source/logHandler.py
+++ b/source/logHandler.py
@@ -10,13 +10,13 @@ import os
 import ctypes
 import sys
 import warnings
-from encodings import utf_8
 import logging
 import inspect
 import winsound
 import traceback
-from types import MethodType, FunctionType
+from types import FunctionType
 import globalVars
+import winKernel
 import buildVersion
 from typing import Optional
 
@@ -292,7 +292,6 @@ class Formatter(logging.Formatter):
 		"""Custom implementation of `formatTime` which avoids `time.localtime`
 		since it causes a crash under some versions of Universal CRT ( #12160, Python issue 36792)
 		"""
-		import winKernel
 		timeAsFileTime = winKernel.time_tToFileTime(record.created)
 		timeAsSystemTime = winKernel.SYSTEMTIME()
 		winKernel.FileTimeToSystemTime(timeAsFileTime, timeAsSystemTime)
--- a/source/logHandler.py
+++ b/source/logHandler.py
@@ -288,6 +288,20 @@ class Formatter(logging.Formatter):
 	def formatException(self, ex):
 		return stripBasePathFromTracebackText(super(Formatter, self).formatException(ex))
 
+	def formatTime(self, record, datefmt=None):
+		"""Custom implementation of `formatTime` which avoids `time.localtime`
+		since it causes a crash under some versions of Universal CRT ( #12160, Python issue 36792)
+		"""
+		import winKernel
+		timeAsFileTime = winKernel.time_tToFileTime(record.created)
+		timeAsSystemTime = winKernel.SYSTEMTIME()
+		winKernel.FileTimeToSystemTime(timeAsFileTime, timeAsSystemTime)
+		timeAsLocalTime = winKernel.SYSTEMTIME()
+		winKernel.SystemTimeToTzSpecificLocalTime(None, timeAsSystemTime, timeAsLocalTime)
+		res = f"{timeAsLocalTime.wHour:02d}:{timeAsLocalTime.wMinute:02d}:{timeAsLocalTime.wSecond:02d}"
+		return self.default_msec_format % (res, record.msecs)
+
+
 class StreamRedirector(object):
 	"""Redirects an output stream to a logger.
 	"""
--- a/source/logHandler.py
+++ b/source/logHandler.py
@@ -288,7 +288,7 @@ class Formatter(logging.Formatter):
 	def formatException(self, ex):
 		return stripBasePathFromTracebackText(super(Formatter, self).formatException(ex))
 
-	def formatTime(self, record, datefmt=None):
+	def formatTime(self, record: logging.LogRecord, datefmt: Optional[str] = None) -> str:
 		"""Custom implementation of `formatTime` which avoids `time.localtime`
 		since it causes a crash under some versions of Universal CRT ( #12160, Python issue 36792)
 		"""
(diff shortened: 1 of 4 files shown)


lukaszgo1:
This still needs some work - marking as a draft for now.

seanbudd:
@lukaszgo1 - this fix looks good to me, what more work needs to be done?

I saw wxPython has this fix implemented - but unreleased. Maybe all we need is a comment referencing that this hack can go when wx is updated
https://github.com/wxWidgets/Phoenix/commit/040c59fd991cd08174b5acee7de9418c23c9de33

josephsl:
Critical CC: @RobinD42

RobinD42:
I suggest to make it conditional. The `setlocale` is only needed on Windows and on Python >= 3.8.

lukaszgo1:
> I suggest to make it conditional. The `setlocale` is only needed on Windows and on Python >= 3.8.

NVDA is Windows only, and it checks for the Python version on startup failing to run under any other version than py 3.8.

lukaszgo1:
@seanbudd Sorry for the delay. As I've explained in the updated PR description backporting this method has not been enough to fix this crash. Now this is ready for review.

source/logHandler.py
```diff
@@ -288,6 +288,19 @@ class Formatter(logging.Formatter):
 	def formatException(self, ex):
 		return stripBasePathFromTracebackText(super(Formatter, self).formatException(ex))
 
+	def formatTime(self, record, datefmt=None):
```
seanbudd, line 291:
```suggestion
	def formatTime(self, record: logging.LogRecord, datefmt: Optional[str] = None) -> str:
```

I know that python doesn't even do this nor should this be used directly, but it would be helpful for developers if we have to change this further

source/winKernel.py
```diff
@@ -160,6 +159,40 @@ class SYSTEMTIME(ctypes.Structure):
...
+	timeAsFileTime.dwHighDateTime = res >> 32
+	return timeAsFileTime
+
+
+def FileTimeToSystemTime(lpFileTime: FILETIME, lpSystemTime: SYSTEMTIME) -> None:
+	if kernel32.FileTimeToSystemTime(byref(lpFileTime), byref(lpSystemTime)) == 0:
+		raise WinError()
+
+
+def SystemTimeToTzSpecificLocalTime(lpTimeZoneInformation, lpUniversalTime, lpLocalTime):
```
seanbudd, line 187: could this also get typing information in the function header?

seanbudd (changes requested):
Since I haven't been able to reproduce the error on VMs or my machine I can't verify if the fix works, but the overall strategy appears solid. I've requested some minor changes for type information and highly appreciate the work gone into this from all.

CyrilleB79:
Just to inform also that this fix has been tested successfully by @kvark128 and me in environments where the bug was seen before. See the comments in #12160.

lukaszgo1:
@seanbudd All your changes are now addressed.

seanbudd (approved):
Thank you @lukaszgo1 , and thanks @CyrilleB79 and @kvark128 for testing and investigating

PR #12252, opened 2021-03-31
Fix Issue 12147: new focus target is not always announced

### LINKED ISSUE #12147: New focus target is not always announced
When an element is focused via JS, screen readers will usually announce it. However, this isn't always the case with NVDA, especially when the new focus target replaces the triggering element.  This is a fairly common pattern that is found on the web, especially within dynamic widgets or single page applications.

### Steps to reproduce:

1. Go to https://codepen.io/mfairchild365/pen/qBbrwXv
2. Activate the 'load' button in each example
3. Each example will 'pass' if the new focus target is announced by NVDA
4. Observe that most examples fail in NVDA 2020.4

Update: This bug is also present when focus is sent to an element that already exists in the DOM long before the trigging element is removed. This differs from my first example, where the target element is added to the DOM after the triggering element is removed. I tested this new scenario against JAWS, VO (macOS, and iOS), and NVDA. Only NVDA is silent. For example, see https://codepen.io/mfairchild365/pen/OJWydob

### Actual behavior:

Some examples fail and other pass. Details of what passed and failed can be found at the end of the codepen.

### Expected behavior:

In all examples, I'd expect the new focus target to be announced. Some screen readers do this (such as VoiceOver), and others are inconsistent (such as NVDA).

Installed

#### NVDA version:

2020.4

#### Windows version:

Windows 10 20H2

- Chrome 89
- FF 86

Yes

Yes, I tried older versions of NVDA with similar results.

Yes

No

mfairchild365:
### Link to issue number:
fixes #12147

### Summary of the issue:

The new focus target is not always announced when:

1. The triggering button is activated while in browse mode
2. The triggering button is removed
3. The focus target is adjacent to where the triggering button was

This is because of the overlapping logic for browse mode focus changes that is used to prevent duplicate announcements. That overlapping logic did not account for this scenario.

See the linked codepens in  #12147 for more examples.

I can not specially point to an instance of this in the wild, as all of my use-cases are behind a firewall. However, I know that this behavior is not unusual in SPA-style web apps.

This PR fixes the issue by altering the overlapping detecting logic so that:

1. We look at the state of the previous focus object
2. If the state of the previous focus object is `DEFUNCT` (no longer available) and an overlap was detected, then continue as if there was not an overlap.

### Testing strategy:

- I added a system test for this (I noticed that it times out from time to time, but that is likely an unrelated bug)
- I manually tested against the linked codepens in the latest versions of Chrome, Firefox, and Edge to verify that this resolves the issue.

### Change log entry:

Section: New features, Changes, Bug fixes

### Code Review Checklist:

- [ ] Unit tests.
- [x] Manual tests.
- [ ] User Documentation.
- [ ] Change log entry.


mfairchild365:
I'm feeling pretty confident about this solution, but this is my first PR and dive into NVDA, so please let me know if I can improve it in any way.

michaelDCurran (approved):
Thanks @mfairchild365  this is great.

Issue #10446, opened 2019-10-30
Shorten "out of element" messageShortening "out of element" message


leonardder:
### Is your feature request related to a problem? Please describe.
In browse mode or documents, when leaving a particular element with the caret, "out of element" is announced.

### Describe the solution you'd like
Based on how this is implemented in braille, I'd like to suggest "element end". So, instead of "out of list", announce 'list end". This is somewhat shorter and therefore more intuitive. Furthermore, the change is relatively small.

### Describe alternatives you've considered
Suppress element ends if they are followed by an element of the same role, see #10420

yozaam:
Wow, this is something I would have never thought of, I'll send a pull request right now, but I would like to ask the mods if there is a reason, to syntactically use the "out of" phrase, 
also I'm a little bit of a beginner to this codebase so forgive me if im going in the wrong direction,

[in the src/speech/intit file](https://github.com/nvaccess/nvda/blob/master/source/speech/__init__.py)
Also for eg. in the 		line 2107 & 1699	text=_("link") if link else _("out of %s")%_("link")

I have changed "out of %" to "%s end"

Also in other locations when we call _()
such as 
line 2143          text=_("out of grammar error")     
line 2174          _("out of table")

And other files like  [source/locale/th/LC_MESSAGES/nvda.po](https://github.com/nvaccess/nvda/blob/master/source/locale/th/LC_MESSAGES/nvda.po)    where we have msgid "out of %s" 
on 12 different lines including 2403

should I change these as well or are they relevant since it is better to say "out of grammar error" than "grammar error end" or those in nvda.po are not relevant to this issue

Would be very happy if you or somebody experienced could give me a tip, 
Am i going in the right direction?

Adriani90:
You do not have to change the .po files. They will change automatically since they will get the right strings from the _() function out of the code in speechinit.py through Gettext when compiling.

yozaam:
> You do not have to change the .po files. They will change automatically since they will get the right strings from the _() function out of the code in speechinit.py through Gettext when compiling.

Thank you, you completely solved my doubt,  so the two changes I have made are enough? And what about the other lines where we don't use %s like 
line 2143 text=("out of grammar error")
line 2174 _("out of table")

Should I change these as well,
Again, sorry I am new to this and I am very grateful for your guidance

feerrenrut:
[...]

PR #12260, opened 2021-04-02
Shorten "out of element" message

GdePaulo:
In browse mode or documents, when leaving an element with the caret, announce "element end" instead of "out of element".

Note that there is another outdated pull request (#10714) which tried to tackle the same issue. If this one is merged, then that one can be closed as well.

### Link to issue number:

#10446

See conversation on #10714 before reviewing this PR

### Summary of the issue:

In browse mode or documents, when leaving an element with the caret, NVDA announces "out of element". It would be shorter and perhaps more intuitive to announce "element end" instead. 

In `source/speech/__init__.py`:

* Changed the string when exiting an element (i.e. a list, table, link, comment or text with grammar or spelling errors) to announce "element end" instead of "out of element".

In `source/appModules/kindle.py`:

* Changed the string when exiting frequently highlighted text to announce "popular highlight end" instead of "out of popular highlight".

### Testing strategy:

In Google Chrome, performed the following on a webpage:

* Exit a link element, ensuring that "link end" is announced.
* Exit a table element, ensuring that "table end" is announced.
* Exit a list element, ensuring that "list end" is announced.

In Microsoft Word, performed the following in a document:

* Exit text containing spelling errors, ensuring that "spelling error end" is announced.
* Exit text annotated with a comment, ensuring that "comment end" is announced.

None.

### Change log entry:

Section: Changes

In browse mode or documents, when leaving a particular element with the caret, NVDA now announces "element end" instead of "out of element".

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.


lukaszgo1:
@GdePaulo Would you be able to resolve merge conflicts here?

seanbudd (commented):
I've merged in master and resolved the conflicts. In terms of issues raised on the other PR #10714

https://github.com/nvaccess/nvda/pull/10714#pullrequestreview-584086179
> As I [mentioned on the issue](https://github.com/nvaccess/nvda/issues/10446#issuecomment-773852041) I'm happy let our alpha users try this out and give feedback.
> Though using `git grep -i "_(\"out of" -- "*.py"` I found one more:

I've confirmed that there is no more "out of" usages.

https://github.com/nvaccess/nvda/pull/10714#issuecomment-773883091
>For real containers (tables, lists, etc.), I do not have any better suggestion than the existing "out of". For containers that seem more formatting (e.g. spelling error, link), I guess "no spelling error" / "no link" would be suitable, but maybe it's out of the scope of this PR.

I'm not sure if we are still awaiting feedback from braille users, or if this issue raised is otherwise blocking the PR.

CyrilleB79:
This PR should also be tested as much time as possible with new translations.
Thus let me express the following recommendation:
When alpha is merged into beta, translators should be clearly notified of the foreseen release date of 2021.2beta1. This would allow them to ensure they have their translations included in the first beta.

source/speech/speech.py
```diff
@@ -2376,7 +2377,7 @@ def getFormatFieldSpeech(  # noqa: C901
 		link=attrs.get("link")
 		oldLink=attrsCache.get("link") if attrsCache is not None else None
 		if (link or oldLink is not None) and link!=oldLink:
-			text=_("link") if link else _("out of %s")%_("link")
+			text = _("link") if link else _("%s end") % _("link")
```
feerrenrut, line 2380: - This should have a translation comment
- I think it makes sense to put these all into categories to prevent collisions in translation msgID
- I think it's a bad idea to construct the message like this, especially given other "end" messages are not constructed. EG "comment end". This should either enforce for all that the message is ordered this way, or give translators the ability to order it correctly for their language.

feerrenrut (changes requested):
I'm ok with the UI change but note it might be a controversial change. Let's go ahead, but be aware of feedback for this.

michaelDCurran:
There have been comments on the linked issue which bring up the fact there is a simantic difference between "out of element" and "element end". I agree with this concirn. Note that Speech and braille are very different modalities, and the goal here should not be to bring their messages completely in line with each other.

When moving up and down a document, Speech announces the entry and exit of elements. This is true both when entering an element from the top and exiting at the bottom, or entering at the bottom and exiting from the top. It is conveying relative in and out, rather than start and end. However, in Braille, the user is shown what could be thought of as a static section (or window) of the document content. Thus communicating an element's start and end in physical space, rather than conveying relative in and out.

PR #12276, opened 2021-04-06
Bug issue template: Rewrote COM Reg. Fix Tool question, to make it more action-oriented

### LINKED ISSUE #12275: Proposal: make the COM Registration Fixing Tool question on the NVDA bug template be more provocative

In various Github questions I've observed and commented upon, the solution has turned out to be running the COM Registration Fixing Tool.

The current bug issue template asks whether the user has run the tool.
In every such case, the user has responded "no" to that question.
Later, someone notices that the user said "no" to that, and comments on the issue "Please try running the COM Reg Fix Tool", or words to that effect. Then this solves the problem and the issue can be closed.

I would like the question to become more provocative of the user actually running the tool, not just a passive question that gathers information about whether the user ran the tool.

The add-on question, for example, does not ask "Have you tried running with add-ons disabled?", which would be a similar passive question. Instead, it asks "If add-ons are disabled, is your problem still occurring?", which provokes the user to try running NVDA with add-ons disabled before submitting the issue.

I would like to see the CRFT question be asked in a similar way.

Retain the current format of the question, as a passive gatherer of information.

### Additional context

A recent example issue: #12120

XLTechie:
### Link to issue number:

Closes #12275 

### Summary of the issue:

The current wording of the bug report issue template, makes the question about running the COM Registration Fixing Tool, a passive question that just gathers information.

Make the question more likely to provoke the user to try running the tool before submitting the issue, by phrasing it more like the other non-optional questions (such as the one which asks if the issue still happens when you disable add-ons).

Also, this regularizes the name of the tool in the question, to be consistent with the NVDA Tools menu name for the tool.

### Testing strategy:

N/A, no code changes.

None

### Change log entry:

None needed.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -35,4 +35,4 @@ Please also note that the NVDA project has a Citizen and Contributor Code of Con
 
 #### If add-ons are disabled, is your problem still occurring?
 
-#### Did you try to run the COM registry fixing tool in NVDA menu / tools?
+#### Does the issue still occur after you run the COM Registration Fixing Tool in NVDA's tools menu??


XLTechie:
CC @Qchristensen

Qchristensen (approved):
Looks good, thanks!

Issue #12159, opened 2021-03-13
NVDA doesn't disable assistive technologies after uninstall (java apps can't run and show java.awt.AWTError)


JayDi85:
You can find workaround at the end

### Steps to reproduce:
* Install NVDA in Windows
* Ensure that it works with java's GUI applications
* Uninstall NVDA
* Now you can't run java applications due one of the errors:
  * `java.lang.UnsatisfiedLinkError: D:\SOFT\Programming\ETC\JAVA\openjdk-11.0.2\bin\javaaccessbridge.dll: Can't find dependent libraries`
  * `java.awt.AWTError: Assistive Technology not found: com.sun.java.accessibility.AccessBridge`

### Actual behavior:
* Java apps with GUI are broken.
* Full java error example:
```
java.awt.AWTError: Assistive Technology not found: com.sun.java.accessibility.AccessBridge
        at java.desktop/java.awt.Toolkit.newAWTError(Unknown Source)
        at java.desktop/java.awt.Toolkit.fallbackToLoadClassForAT(Unknown Source)
        ...
        at java.base/java.util.stream.ReferencePipeline.forEach(Unknown Source)
        at java.desktop/java.awt.Toolkit.loadAssistiveTechnologies(Unknown Source)
        at java.desktop/java.awt.Toolkit.getDefaultToolkit(Unknown Source)
        at de.sciss.syntaxpane.DefaultSyntaxKit.<clinit>(DefaultSyntaxKit.java:105)
        at com.xk72.charles.gui.lib.UIUtils.<clinit>(Unknown Source)
        at com.xk72.charles.gui.Main.GSXJ(Unknown Source)
        at com.xk72.charles.Main.FwRs(Unknown Source)
        at com.xk72.charles.gui.Main.main(Unknown Source)
        ...
Caused by: java.lang.ClassNotFoundException: com.sun.java.accessibility.AccessBridge
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(Unknown Source)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(Unknown Source)
        at java.base/java.lang.ClassLoader.loadClass(Unknown Source)
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Unknown Source)
```

### Expected behavior:
* Java apps with GUI must work;
[...]

PR #12279, opened 2021-04-07
nvda now disables the java access bridge on uninstallation, fix for #12159

christopherpross:
### Link to issue number:
#12159 
### Summary of the issue:
NVDA has the feature to automatecly enable the java access bridge if it is not avaiable. Also NVDA delivers the DLL itselfs, the problem is that if NVDA is removed, the enabled settings of Java searches for the DLL of NVDA and don't find it.
This caused an exception in GUI Frameworks, if no other access bridge is installed but the .accessibility.properties is still enabled from NVDA.

I simply write a disableBridge function in JABHandler.py, and define the contents for the file  right under the other variable, that have now the subfix "ENABLED".
In "installer.py" I called the "disableBridge" function, after checking whether the the brigde is enabled.

### Testing strategy:
I called the function "disableBridge" manualy on my system and have confirmed that the access bridge was  disabled.
N/A
### Change log entry:
Section: Changes
`If NVDA is removed from the system, it now disables the Java Access Bridge to prevent errors of no existing DLL-File (#12159 )`

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### DIFF
--- a/source/JABHandler.py
+++ b/source/JABHandler.py
@@ -46,9 +46,14 @@ import globalVars
 #: to enable JAB.
 A11Y_PROPS_PATH = os.path.expanduser(r"~\.accessibility.properties")
 #: The content of ".accessibility.properties" when JAB is enabled.
-A11Y_PROPS_CONTENT = (
+A11Y_PROPS_CONTENT_ENABLED = (
 	"assistive_technologies=com.sun.java.accessibility.AccessBridge\n"
 	"screen_magnifier_present=true\n"
+)
+#: The content of ".accessibility.properties" when JAB is disabled.
+A11Y_PROPS_CONTENT_DISABLED = (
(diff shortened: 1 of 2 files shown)


leonardder:
I don't think we should go this way. This will disable the access bridge even when other screen readers are installed, and that's certainly not what should happen in my opinion.
Even when we register whether NVDA installed the bridge (e.g. with a flag in the registry), we can't be sure that touching accessibility.properties like this is the right thing to do.

akash07k:
I too think that it is a good approach

christopherpross:
Ok, sorry for that. How would you fix this bug. Its my first contribution, so  sorry for the inconvenience

leonardder:
No problem at all, many thanks for your contribution.

It would be best to discuss the best approach in https://github.com/nvaccess/nvda/issues/12159

PR #12051, opened 2021-02-09
Elements list is now available in Microsoft Word straight away without having to enable browse mode first

michaelDCurran:
### Link to issue number:
Fixes #12050

### Summary of the issue:
Documentation for NVDA stated that Elements list (NVDA+f7) was available in focus mode. Although this wa strue for web documents and Excel, it was missed for Microsoft Word.

### Description of how this pull request fixes the issue:
A treeInterceptor is now created for Microsoft Word documents (both with UIA enabled or disabled) but the treeInteceptor is defauled to focus mode. Thus Elements list is available straight away. This is the exact same approach as taken by the treeInterceptor for Microsoft Excel.

### Testing strategy:
Manual testing done:
* Ensured UIA in Microsoft Word document controls was disabled in NVDA's davanced settings.
* Opened Microsoft Word.
* Typed a line of text and formatted it as a heading with shift+alt+rightArrow
* Pressed NVDA+f7 to bring up the Elements list.
* Restarted NvDA.
* Enabled UIA in Microsoft Word document controls in NVDA's advanced settings.
* Typed a line of text and formatted it as a heading with shift+alt+rightArrow
* Pressed NVDA+f7 to bring up the Elements list.

In theory a system test could be created for this, however this would require appveyor vms including Microsoft Word, which is something we would need to look into in the future.
I don't believe unit tests really apply to this kind of infrastructure change.

### Known issues with pull request:
This change means that a treeInterceptor is now always created for Word documents. Although Word document treeInterceptors are not in any way heavy (they don't cache much at all) it is still a change which could affect the lifetime of treeInterceptors or other objects in NVDA.
I think this is very low risk, but either this needs to be tested on another rc for a while by heavy users of Microsoft Word, or we hold off until 2021.1 and make an edit to the 2020.4 documentation excluding Microsoft Word.
 
### Change log entry:
None needed.

### Code Review Checklist:

This checklist is a reminder of things commonly forgotten in a new PR.
Please do a self-review to check these items.
Reviewers will not approve the PR until these are met.

- [x] Pull Request description is up to date.
- [x] Unit tests.
- [x] System (end to end) tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.
- [x] Context sensitive help for GUI changes.

PR #12320, opened 2021-04-21
Windows 10 Mail: ensure browse mode is used in the reading pane.

### LINKED ISSUE #12117: nvda is not reading line by line emails when using the native windows 10 email application
### Steps to reproduce:
open an email in the native email app on windows 10 to view emails
walk with the arrow below to read the email line by line. 
### Actual behavior:
nvda keeps repeating the current line with the focus. 
### Expected behavior:
nvda should read the email line by line 
instaled
#### NVDA version:
nvda.exe, NVDA alpha-21810,db664bef

#### Windows version:
10.0.19043.844
HxOutlook.exe, microsoft.windowscommunicationsapps 16005.13426.20688.0

yes
no
yes
no
log:
[email.txt](https://github.com/nvaccess/nvda/files/6085236/email.txt)

michaelDCurran:
### Link to issue number:
Fixes #12117 

### Summary of the issue:
In Windows 10 Mail, a Microsoft Word document control is used to display content of received emails and emails currently being composed. In NVDA 2020.4, NVDA would use browse mode for reading emails, but not for writing emails.
However, after merging of pr #12051  browse mode is no longer used by default when reading emails. This is because the base Microsoft Word document NVDAObject now creates a TreeInterceptor all the time, but set to focus mode, so that elements list is always available in Microsoft Word.
But as the hxMail implementation assumed browse mode would be available for the TreeInterceptor always, and only created the TreeInterceptor in the reading pane, Windows 10 mail ended up getting no treeInterceptor for writing email (ok) but for reading email it got a treeInterceptor but set to focus mode (not okay).

In the hxMail appModule: rather than overriding shouldCreateTreeInterceptor, expose an isInReadingPane property, and in the treeInterceptor's treeInterceptor_gainFocus event, correctly set browse mode or focus mode based on whether we are now in the reading pane, and if we were or not before.
Aso the overriding of the treeInterceptor's isAlive property has been removed, as the treeInterceptor should now stay around as long as that physical control exists.
Note that Mail uses the same physical instance of the same control for both reading and composing mail.
 
### Testing strategy:
Opened Windows 10 mail. Opened a received email, and verified that the arrow keys could again be used to navigate / read the content of the email message and that quick navigation was available. Then pressed control+n to create a new message, tabbed to the message document, and ensured that 

None known.

### Change log entries:
None needed.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### DIFF
--- a/source/appModules/hxmail.py
+++ b/source/appModules/hxmail.py
@@ -13,12 +13,26 @@ from NVDAObjects.UIA.wordDocument import WordDocument
 class MailWordDocumentTreeInterceptor(WordDocument.treeInterceptorClass):
 
 	def _get_isAlive(self):
-		return super(MailWordDocumentTreeInterceptor,self).isAlive and self.rootNVDAObject.shouldCreateTreeInterceptor
+		return super().isAlive and self.rootNVDAObject.isInReadingPane
+
+	def __init__(self, rootNVDAObject: "MailWordDocument"):
+		super().__init__(rootNVDAObject)
+		if rootNVDAObject.isInReadingPane:
+			# The base WordDocument TreeInterceptorClass forces focus mode by default
+			# As a TreeInterceptor is created for all word documents
+			# so that the NVDA elements list is available.
+			# However, Windows 10 Mail's reading pane should start in browse mode.
+			self.disableAutoPassThrough = False
+			self.passThrough = False
 
 class MailWordDocument(WordDocument):
 
 	treeInterceptorClass=MailWordDocumentTreeInterceptor
-	def _get_shouldCreateTreeInterceptor(self):
+
+	# typing information for isInReadingPane property
+	isInReadingPane: bool
+
+	def _get_isInReadingPane(self) -> bool:
 		# Locate the Reading pane in the ancestors
 		condition=UIAHandler.handler.clientObject.createPropertyCondition(UIAHandler.UIA_ClassNamePropertyId,"ReadingPaneModern")
 		walker=UIAHandler.handler.clientObject.createTreeWalker(condition)
--- a/source/appModules/hxmail.py
+++ b/source/appModules/hxmail.py
@@ -12,18 +12,23 @@ from NVDAObjects.UIA.wordDocument import WordDocument
 
 class MailWordDocumentTreeInterceptor(WordDocument.treeInterceptorClass):
 
-	def _get_isAlive(self):
-		return super().isAlive and self.rootNVDAObject.isInReadingPane
+	_wasInReadingPane: bool = False
 
-	def __init__(self, rootNVDAObject: "MailWordDocument"):
-		super().__init__(rootNVDAObject)
-		if rootNVDAObject.isInReadingPane:
+	def event_treeInterceptor_gainFocus(self):
+		isInReadingPane = self.rootNVDAObject.isInReadingPane
+		if isInReadingPane != self._wasInReadingPane:
+			self._wasInReadingPane = isInReadingPane
 			# The base WordDocument TreeInterceptorClass forces focus mode by default
 			# As a TreeInterceptor is created for all word documents
 			# so that the NVDA elements list is available.
-			# However, Windows 10 Mail's reading pane should start in browse mode.
-			self.disableAutoPassThrough = False
-			self.passThrough = False
+			# However, Windows 10 Mail's reading pane should use browse mode.
+			if isInReadingPane:
+				self.disableAutoPassThrough = False
+				self.passThrough = False
+			else:
+				self.disableAutoPassThrough = True
+				self.passThrough = True
+		super().event_treeInterceptor_gainFocus()
 
 class MailWordDocument(WordDocument):
 


josephsl:
Hi, I’ll also perform tests through my add-on and do something about reviewing this. Thanks for a fix for the mentioned issue.

 

From: Michael Curran ***@***.***> 
Sent: Wednesday, April 21, 2021 4:58 PM
To: nvaccess/nvda ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [nvaccess/nvda] Windows 10 Mail: ensure browse mode is used in the reading pane. (#12320)

 


Link to issue number:


Fixes #12117 <https://github.com/nvaccess/nvda/issues/12117> 


Summary of the issue:


In Windows 10 Mail, a Microsoft Word document control is used to display content of received emails and emails currently being composed. In NVDA 2020.4, NVDA would use browse mode for reading emails, but not for writing emails.
However, after merging of pr #12051 <https://github.com/nvaccess/nvda/pull/12051>  browse mode is no longer used by default when reading emails. This is because the base Microsoft Word document NVDAObject now creates a TreeInterceptor all the time, but set to focus mode, so that elements list is always available in Microsoft Word.
But as the hxMail implementation assumed browse mode would be available for the TreeInterceptor always, and only created the TreeInterceptor in the reading pane, Windows 10 mail ended up getting no treeInterceptor for writing email (ok) but for reading email it got a treeInterceptor but set to focus mode (not okay).


In the hxMail appModule: rather than overriding shouldCreateTreeInterceptor, expose an isInReadingPane property, and in the treeInterceptor's treeInterceptor_gainFocus event, correctly set browse mode or focus mode based on whether we are now in the reading pane, and if we were or not before.
Aso the overriding of the treeInterceptor's isAlive property has been removed, as the treeInterceptor should now stay around as long as that physical control exists.
Note that Mail uses the same physical instance of the same control for both reading and composing mail.


Testing strategy:


Opened Windows 10 mail. Opened a received email, and verified that the arrow keys could again be used to navigate / read the content of the email message and that quick navigation was available. Then pressed control+n to create a new message, tabbed to the message document, and ensured that


Known issues with pull request:


None known.


Change log entries:


None needed.


Code Review Checklist:


*	[ ] Pull Request description is up to date.
*	[ ] Unit tests.
*	[ ] System (end to end) tests.
*	[ ] Manual tests.
*	[ ] User Documentation.
*	[ ] Change log entry.
*	[ ] Context sensitive help for GUI changes.

  _____  


You can view, comment on, or merge this pull request online at:


  https://github.com/nvaccess/nvda/pull/12320


Commit Summary


*	Windows 10 Mail: ensure browse mode is used in the reading pane.
*	Windows 10 mail appModule: use event_treeInterceptor_gainFocus rather than the treeInterceptor's constructor to toggle browse mode.


File Changes


*	M source/appModules/hxmail.py <https://github.com/nvaccess/nvda/pull/12320/files#diff-541ff0e27eecea05308cdd60601073b6e47aac04446b3b578b3128ea49f6c7f2>  (25) 


Patch Links:


*	https://github.com/nvaccess/nvda/pull/12320.patch
*	https://github.com/nvaccess/nvda/pull/12320.diff

leonardder:
I'm pretty sure Outlook with UIA enabled also suffers from this. cc @bramd

michaelDCurran:
Yep, it does. I will address this tomorrow.

leonardder:
I looked at the code and honestly, I find it a little hacky. Basically what's done in #12051 is enforcing the word tree interceptor to start in focus mode and then this pr tries hard to circumvent that. I'd rather see something like moving the __init__ logic that was added in #12051 (i.e. the part copied from ExcelBrowseModeTreeInterceptor) to BrowseModeTreeInterceptor:
```
	def __init__(self,rootNVDAObject):
		super().__init__(rootNVDAObject)
		if self.disableAutoPassThrough:
			self.passThrough = True
			reportPassThrough.last = True
```

Then, this __init__ override can be removed for excel and word. For Excel and Word, disableAutoPassThrough should still be set to True, whereas in the outlook appModule, it should be restored back to False.

leonardder:
AH, I missed the fact that the interceptor is used for both reading and writing in my proposal. Also it doesn't seem to be cleaned up correctly, right?
Still, I hope it will inspire you somehow.

source/appModules/hxmail.py
```diff
@@ -12,13 +12,32 @@
 
 class MailWordDocumentTreeInterceptor(WordDocument.treeInterceptorClass):
 
-	def _get_isAlive(self):
-		return super(MailWordDocumentTreeInterceptor,self).isAlive and self.rootNVDAObject.shouldCreateTreeInterceptor
+	_wasInReadingPane: bool = False
+
+	def event_treeInterceptor_gainFocus(self):
+		isInReadingPane = self.rootNVDAObject.isInReadingPane
```
feerrenrut, line 18: This assumes that `rootNVDAObject` will always be a `MailWordDocument`. If I'm understanding this right, while unlikely, a developer may set this as the `TreeInterceptor` on a different object, resulting in it being constructed with a different type as the `rootNVDAObject`.

This took me quite some time to determine, typing would help, maybe something like this at class level:
``` python
rootNVDAObject: MailWordDocument

def __init__(self, obj:MailWordDocument):
		if not isinstance(obj, MailWordDocument):
				log.error("MailWordDocumentTreeInterceptor depends on rootNVDAObject being a MailWordDocument see event_treeInterceptor_gainFocus")
		super().__init(obj)
```

source/appModules/hxmail.py
```diff
@@ -12,13 +12,32 @@
 
 class MailWordDocumentTreeInterceptor(WordDocument.treeInterceptorClass):
 
-	def _get_isAlive(self):
-		return super(MailWordDocumentTreeInterceptor,self).isAlive and self.rootNVDAObject.shouldCreateTreeInterceptor
+	_wasInReadingPane: bool = False
```
feerrenrut, line 15: Should this be initialized to False? Currently this will only be accurate after the gain focus event is fired the first time. Could this be initialized correctly in the `__init__` method instead.

Could a situation arise where the object is created while focus is already in a readingpane? Thus no initial fire of the gainfocus to set this to True, but rendering it inaccurate?

feerrenrut (changes requested):
There are also some unanswered comments/questions here also from @leonardder who seems much more familiar with this change.

michaelDCurran:
Closing this in favor of pr #12365 which reverts pr #12051.
A cleaner solution should be worked out as Outlook was also affected, not just Windows 10 Mail.

PR #12321, opened 2021-04-22
Fix up from #12210: Excel without UIA enabled: again allow typing and editing in cells 

### LINKED ISSUE #12303: Key echo while typing and speech while editing cells in Excel does not work at all without the experimental UIA setting enabled
### Steps to reproduce:

This happens with default settings (i.e, usage of UIA in Excel is not enabled). The problem goes away when turning the setting on, so this is probably a regression.

1. Open a new blank excel workbook. Type some text into cell A1, e.g "abc"
2. Press f2, and attempt to use the arrow keys to read the text

### Actual behavior:

Nothing is heard when typing into a cell. Pressing the arrow keys to review the text when using f2 to edit the cell contents does not work either.

This makes it virtually impossible to edit complex formulae without resorting to copying and pasting out into notepad, doing the edits there, then pasting it back.

### Expected behavior:

Typing into the cell echos the character that you typed (based on my key echo setting). Using the arrow keys to review text in a cell when pressing f2 works.

Installed

#### NVDA version:
alpha-22295,49a85788

#### Windows version:
Windows 10 20h2

Excel 16.0.13901.20366 64-bit 

Yes

No

Yes

N/A

michaelDCurran:
### Link to issue number:
Fixes #12303

### Summary of the issue:
After merging of pr #12210 editing cells in Excel without UIA enabled became im possible as NVDA did not report / track focus had ented the Cell Edit control. 
This was due to the EXCEL6 window accidentally being marked as having a good UIA implementation. This was testing code left over from the early implementation of #12210.
 
Remove EXCEL6 from the good UIA windows list.
Also ensure that MSAA focus events on this window are ignored when using Excel with UIA enabled, as Excel will fire its own UIA focus event on an edit control within the active cell.

### Testing strategy:
Testing with all combinations of NvDA's use Microsoft Excel with UI Automation setting, and Excel's allow directly editing in a cell setting (4 different variations possible):
In an Excel spreadsheet:
* Started typing into a cell. Ensured that NVDA was announcing the typed characterrs, and that it announced focus moving into the edit control.
* Pressed f2 on a cell, ensured that NVDA announced focus on the edit control, that characters were announced when typed, and that moving the cursor with left and right arrows would speak the characters in the edit control.

This was successful for all 4 variations of the two settings.
 
None known.

### Change log entries:
None needed.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### DIFF
--- a/source/_UIAHandler.py
+++ b/source/_UIAHandler.py
@@ -58,7 +58,6 @@ WDAG_PROCESS_NAME=u'hvsirdpclient'
 goodUIAWindowClassNames=[
 	# A WDAG (Windows Defender Application Guard) Window is always native UIA, even if it doesn't report as such.
 	'RAIL_WINDOW',
-	"EXCEL6",
 ]
 
 badUIAWindowClassNames=[
--- a/source/appModules/excel.py
+++ b/source/appModules/excel.py
@@ -5,6 +5,7 @@
 # For more details see: https://www.gnu.org/licenses/gpl-2.0.html
 
 import time
+import config
 import eventHandler
 import api
 import UIAHandler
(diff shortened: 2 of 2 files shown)


seanbudd (approved):
Tested using the suggested method in the PR, and confirmed that the original issue is fixed. LGTM

PR #12360, opened 2021-05-03
Ensure not to maintain AccessibleTable2 when fetching child nodes when the underlying table has neither rows nor columns

### LINKED ISSUE #12359: Incorrect assumptions in Gecko VBuf break table navigation in some tables in Chromium
Related to #8462, #9760 

### Steps to reproduce:
1. Download [BrokenTableTest.html.txt](https://github.com/nvaccess/nvda/files/6414036/BrokenTableTest.html.txt)
2. Remove the extension and open the file in Chrome/Edge Chromium
3. Try table navigation.

### Actual behavior:
Table navigation doesn't work. NVDA does not announce table row/column count for the table. If reporting of layout tables is enabled, table navigation results in an error.

### Expected behavior:
Row/Column count is announced for the table and navigation works as expected.

Note that this is what happens when using UIA for Edge is enabled.

### Possible cause
There's some logic in gecko_ia2.cpp around Line 794 that makes me think there's an incorrect assumption made about the table information for the actual table can be ignored.

Installed

#### NVDA version:
NVDA alpha-22571

#### Windows version:
Windows 10 21h1

Chrome/Edge 91Edge 91

LeonarddeR:
### Link to issue number:
Fixes #12359 

### Summary of the issue:
When creating a virtual buffer and NVDA encounters a table, it wil render the children of that table. It keeps a reference to the IAccessibleTable2 interface of the table until it has found a cell in that table. However, when a table contains no cells (i.e. when it has neither rows nor columns) and it encounters an inner table in that table, that inner table is ignored and table navigation does not work.

This pr ensures that when row and column count are 0, the IAccessibleTable2 reference isn't used when rendering children. Therefore, for every child, the logic that checks for IAccessibleTable2 will apply and issue #12359 no longer occurs.
As a bonus, I removed support for IAccessibleTable2 support. This support was left in the VBuf code to support older versions of Seamonkey. I verified that Seamonkey now uses a version of Gecko that also uses IAccessibleTable2.

### Testing strategy:
Exampel attached to #12359 as wel as a private environment where this issue was prevalent.

None

### Change log entries:
Bug fixes

* In Chromium based browsers, several cases were solved where table navigation didn't work and NVDA didn't report the number of rows/columns of the table. (#12359)

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.


lukaszgo1:
> As a bonus, I removed support for IAccessibleTable2 support. This support was left in the VBuf code to support older versions of Seamonkey. I verified that Seamonkey now uses a version of Gecko that also uses IAccessibleTable2.


What about [Basilisk](https://www.basilisk-browser.org/)  - it is quite popular in certain circles due to its compatibility with older style add-ons for Firefox and  as far as I can see is still in active development?

leonardder:
I tested it and it supports IAccessibleTable2 as well.

Brian1Gaff:
Waterfox Classic as well?
 Brian

***@***.***
Sent via blueyonder.
in the display name field.
----- Original Message -----

leonardder:
Feel free to try the try build with whatever you like. I think 
IAccessibleTable2 has been supported for around a year of ten in Firefox 
now. IAccessibleTable1 is really old.

leonardder:
A similar case that can be used to reproduce, is by embedding a table in a div with a role of grid. I know that this is bad authoring practice, but currently NVDA gets pretty frustrated by these cases, whereas this pr properly announces the inner table. This is also reproducible in Firefox.

MarcoZehe:
Note that current WaterFox is built on Gecko 78, but is not supported by NVDA's virtual buffers because its main window class is "WaterFoxWindowClass", not "MozillaWindowClass", which NVDA checks for. So if you try WaterFox with this build or any NVDA build really, you won't get virtual buffers, but that has nothing to do with this PR.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -764,56 +723,58 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
...
+			if (curNodePaccTable2->get_nColumns(&colCount) == S_OK) {
+				s << colCount;
+				parentNode->addAttribute(L"table-columncount", s.str());
+				s.str(L"");
+			}
+			if (rowCount > 0 || colCount > 0) {
+				// This table has rows and columns.
+				// Maintain curNodePaccTable2 for child rendering until any table cells are found.
+				paccTable2 = curNodePaccTable2;
+			}
```
seanbudd, line 777: Could this still be done in a helper function? It would be good to continue to split `fillVBuf` up.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -764,56 +723,58 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
...
+	// If paccTable2 is not NULL, we're within a table but not yet within a cell, so don't bother to query for table info.
+	if (!paccTable2) {
 		// Try to get table information.
-		pacc->QueryInterface(IID_IAccessibleTable2,(void**)&paccTable2);
-		if(!paccTable2)
-			pacc->QueryInterface(IID_IAccessibleTable,(void**)&paccTable);
-		if (paccTable2||paccTable) {
-			// We did the QueryInterface for paccTable, so we must release it after all calls that use it are done.
-			releaseTable = true;
+		pacc->QueryInterface(IID_IAccessibleTable2,(void**)&curNodePaccTable2);
```
seanbudd, line 750:
```suggestion
		pacc->QueryInterface(IID_IAccessibleTable2, (void**)&curNodePaccTable2);
```

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
leonardder: It could be done in a helper function theoretically, but I'm afraid it will make the code less readable that way. Note that I removed this from a helper function because when it was a helper, it only fetched the row and column count to write it to the vbuf xml. Now the row and column count are used later on as well.

seanbudd (approved):
LGTM, thanks @leonardder

PR #12376, opened 2021-05-07
Added release blurb and fixed grammar in api breaking sentence

Qchristensen:
### Summary of the issue:
Needed a release blurb for 2021.1 and also there was a grammar error in the "API breaking sentence" in the what's new document.

### Description of how this pull request fixes the issue:

With words

### Testing strategy:

Read text, and ran past others when writing.

### Change log entries:

For Developers

### Code Review Checklist:

- [X] Pull Request description is up to date.
- [ ] Unit tests.
- [ ] System (end to end) tests.
- [ ] Manual tests.
- [ ] User Documentation.
- [X] Change log entry.
- [ ] Context sensitive help for GUI changes.

PR #12412, opened 2021-05-13
Fixing spelling, grammar, and one markup error, in changes.t2t

XLTechie:
Now that #12376  has been merged, I thought it was time for this.

### Summary of the issue:

While reading through changes.t2t looking for something, I began finding lots of spelling, grammar, and other little errors that could be fixed. I decided to run the whole thing through a spelling and capitalization checker, and touch-up various bits I found wanting along the way.

* Fixed a T2T list error.
* Updated URLs from http to https for links that now work with https.
* Numerous spelling, capitalization, and pluralization fixes.
    - Various conversions to American/International English.
    - Combinations using the NVDA key were commonly written uncapitalized; capitalized the "NVDA" portion of such combinations.
    - Corrected the spelling of "Pinnacle TV software" based on probable results in a web search for "Pinacle TV software".
* Made several grammar corrections and clarifications while looking for spelling issues, etc. By no means complete.
    - Corrections from "addon" to "add-on".
    - Removed some unnecessary words from "Attempt to cancel speech for expired focus events" sub-entry.
    - Removed a double negative.

### Testing strategy:

* Strategically read the text around every spelling correction, making certain it wasn't part of a variable, function, etc. name, or other thing that was supposed to look that way.
* Looked up various company names to check spellings and capitalizations, when I didn't already know.

There are two styles for second level lists.
43 entries looking like this:
```
- Outer list item.
  - Inner list item.
```
Which is, as I understand it, the T2T spec way of writing inner lists (two space indent for each level).
Then there is the other way:
```
- Outer list item.
 - Inner list item.
```
(One space indent) of which there are 230 examples.
Presumably this later form works fine, even though it is contra-standard.
I did not change either version of the sublist to the other.

### Change log entries:

Changed the change log? I don't think so. No entry needed.

### Code Review Checklist:
-X[ ] Pull Request description is up to date.
- [ ] Unit tests.
- [ ] Manual tests.
-X[ ] User Documentation.
- [ ] Change log entry.

### DIFF
--- a/user_docs/en/changes.t2t
+++ b/user_docs/en/changes.t2t
@@ -284,8 +284,8 @@ The existence of marked (highlighted) content can be reported in browsers, and t
 - NVDA should no longer sometimes freeze in Microsoft Word when rapidly arrowing up and down or typing characters with Braille enabled. (#11431, #11425, #11414)
 - NVDA no longer appends nonexistent trailing space when copying the current navigator object to the clipboard. (#11438)
 - NVDA no longer activates the Say All profile if there is nothing to read. (#10899, #9947)
-- NVDA is no longer unable to read the features list in Internet Information Services (IIS) Manager. (#11468)
-- NVDA now keeps the audio device open improving performance on some sound cards (#5172, #10721)
+- NVDA is now able to read the features list in Internet Information Services (IIS) Manager. (#11468)
+- NVDA now keeps the audio device open, improving performance on some sound cards (#5172, #10721)
 - NVDA will no longer freeze or exit when holding down control+shift+downArrow in Microsoft Word. (#9463)
 - The expanded / collapsed state of directories in the navigation treeview on drive.google.com is now always reported by NVDA. (#11520)
 - NVDA will auto detect the NLS eReader Humanware braille display via Bluetooth as its Bluetooth name is now "NLS eReader Humanware". (#11561)
@@ -884,7 +884,7 @@ Highlights of this release include Support for tables in Kindle for PC, support 
 
 
 == Changes ==
-- The status column in the addons manager has been changed to indicate if the addon is enabled or disabled rather than running or suspended. (#7929)
+- The status column in the add-ons manager has been changed to indicate whether the add-on is enabled or disabled rather than running or suspended. (#7929)
 - Updated liblouis braille translator to 3.5.0. (#7839)
 - The Lithuanian braille table has been renamed to Lithuanian 6 dot to avoid confusion with the new 8 dot table. (#7839)
 - The French (Canada) grade 1 and grade 2 tables have been removed. Instead, the French (unified) 6 dot computer braille and Grade 2 tables will be used respectively. (#7839)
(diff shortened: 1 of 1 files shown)


josephsl:
Hi, critical checkup: can translators comment on this PR so the resulting what’s new document is easier to translate? Thanks.

XLTechie:
Improving translatability was one of my goals with this, although I could 
probably have spent time to find more items with improvable clarity.
I was also thinking of how long it would take to review, given its length 
as it stands.

I'm not sure it's worth triggering a retranslation of all 290-odd lines of 
changes, as some of them will only be of interest to people who dig far 
back into NVDA's history.

Not being a translator, I can't make an educated guess, only a regular 
one, about what alterations might or might not lead to a clearer document.

I will understand if this is thought to be not worth the effort.

Your opinion?

josephsl:
Hi, I think it would be best to focus on 2021.1 changes for now and think about targeting this for 2021.2 or later due to how massive this change is. Also, some items may require digging into old Git commits (some of the oldest commits are based on GNU Bazaar assumptions, so be aware of that). Having worked on what’s new for a number of years (I used to submit PR’s for fixing what’s new document), I understand the temptation to fix old entries – I once tried submitting a PR like this but decided against it due to its size and vast scope. Also, note that changing old what’s new entries will raise issues for some translators who were told they don’t have to translate old entries if they were written prior to the first version of NVDA in which their translation work was included (Korean is a good example in that I intentionally did not translate entries older than 2012.3). Thanks.

zstanecic:
Hi Joseph and all!

Me, as a translator, I am against these changes even in the future, because I don’t want to translate old version history of NVDA versions. 

Best,

Zvonimir

XLTechie:
I should note the also massive and somewhat related changes proposed by @seanbudd in #12411.

In reading the [Translating Changes](https://github.com/nvaccess/nvda/wiki/TranslatingChanges) wiki page, I misunderstood the "only translate the latest version's changes" section. My assumption was that modifications to the file would only be translated if they occurred in the 2021.1 section. But from the comments above, I gather that all sections since a translation project began would have to be translated retroactively for any changes.

codeofdusk:
>Various conversions to American/International English.

Guessing this is something like [Oxford spelling](https://en.wikipedia.org/wiki/Oxford_spelling)?

XLTechie:
Bill Dengler wrote:

>       Various conversions to American/International English.
> Guessing this is something like Oxford spelling?

Removed the u from colour, behaviour, etc., for those cases where it 
appeared.
Replaced the S version with the Z version, for words like realise and 
synthesiser.

https://www.bbcamerica.com/anglophenia/2014/07/brits-spell-realize-s
https://www.bbcamerica.com/anglophenia/2014/05/america-drop-u-british-spellings/

codeofdusk:
Since NVDA has an international audience, I would do the latter but not the former (i.e. realize, colour, analyse).

XLTechie:
Closing this for now, until some reasonable direction for translations is found.

PR #12416, opened 2021-05-14
A few more suggestions for the 2021 changes file sections

XLTechie:
### Link to issue number:

None
Follow-up to #12414 

### Summary of the issue:

After @seanbudd's corrections to the changes file, I went through the 2021 sections again and came up with these final suggested revisions.

* The changes from #11782 were listed twice, though slightly differently. Picked the best lines from both.
* Added lots of missing backticks around codeish objects.
* Regularized and added settings option quoting.
* Grammar, punctuation, missing words, pluralization.
* Fixed spelling of "preferred".
* Corrected and regularized i.e. and e.g. usage.

### Testing strategy:

None needed.

None known.

### Change log entries:

None
### Code Review Checklist:
- [ ] Unit tests.
- [ ] Manual tests.
- [X] User Documentation.
- [ ] Change log entry.

### DIFF
--- a/user_docs/en/changes.t2t
+++ b/user_docs/en/changes.t2t
@@ -96,13 +96,10 @@ This release also drops support for Adobe Flash.
 - ``REASON_QUICKNAV`` has been removed from ``browseMode`` - use ``controlTypes.OutputReason.QUICKNAV`` instead. (#11969)
 - ``NVDAObject`` (and derivatives) property ``isCurrent`` now strictly returns Enum class ``controlTypes.IsCurrent``. (#11782)
   - ``isCurrent`` is no longer Optional, and thus will not return None.
-  - When an object is not current ``controlTypes.IsCurrent.NO`` is returned.
+    - When an object is not current ``controlTypes.IsCurrent.NO`` is returned.
 - The ``controlTypes.isCurrentLabels`` mapping has been removed. (#11782)
-  - Instead use the ``displayString`` property on a ``controlTypes.IsCurrent`` enum value. EG ``controlTypes.IsCurrent.YES.displayString``
-- ``NVDAObject`` (and derivatives) property ``isCurrent`` now strictly returns ``controlTypes.IsCurrent``. (#11782)
-  - ``isCurrent`` is no longer Optional, and thus will not return None, when an object is not current ``controlTypes.IsCurrent.NO`` is returned.
-- The ``controlTypes.isCurrentLabels`` has been removed, instead use the ``displayString`` property on a ``controlTypes.IsCurrent`` enum value. (#11782)
-  - EG ``controlTypes.IsCurrent.YES.displayString``
+  - Instead use the ``displayString`` property on a ``controlTypes.IsCurrent`` enum value.
+    - For example: ``controlTypes.IsCurrent.YES.displayString``.
 - ``winKernel.GetTimeFormat`` has been removed - use ``winKernel.GetTimeFormatEx`` instead. (#12139)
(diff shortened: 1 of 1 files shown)


XLTechie:
@seanbudd I did have one question here, following from your regular expression for entries.

In the case of a list like this:

```
- Changed these things in that file (#XXXXX)
  - First thing.
  - Second thing.
  - Third thing.
```

My preference would be to write the first line as:
```
- Changed these things in that file (#XXXXX):
```
But I think that doesn't meet with your formatting, which I think would lead to:
```
- Changed these things in that file. (#XXXXX)
```

Which should be used in these cases?

XLTechie:
@feerrenrut Should I re-apply this to beta instead?

feerrenrut:
@XLTechie no, I don't think this change should go into the beta. Changing all these lines will show up as a diff for translators of this file. Before we merge this, there should be a discussion with translators and a method for them to easily apply the same changes to the translated files.

XLTechie:
Reef Turner wrote:

> @XLTechie no, I don't think this change should go into the beta. Changing all these lines will show up as a diff for translators of this file. Before we
> merge this, there should be a discussion with translators and a method for them to easily apply the same changes to the translated files.

@feerrenrut Ah, I thought the 2021.1 section hadn't been translated yet.

feerrenrut:
Oh, my mistake, I thought this PR contained further changes to the rest of the file. I think the majority of these change have already been made and merged to beta via #12411, is that true?

Still, I don't want to hold up the release further for the changes file. Perhaps instead this can be come a discussion of how to make these changes through the full file?

XLTechie:
> Oh, my mistake, I thought this PR contained further changes to the rest of the file. I think the majority of these change have already been made and merged
> to beta via #12411, is that true?

No, but it applies the changes he made to a bunch of places they probably 
should have been in the 2021.1 section.
In other words, it carries on his work to its logical conclusion imo.

With the idea of hopefully suggesting a unified way the file should look 
going forward, at least for some things.

One problem it fixes, is that a few changes were listed multiple times, in 
somewhat different ways, re #11782. I collapsed those into reasonable 
singularities. (See db7c339; I haven't rebased yet)

I think that commit at least, should be considered for beta.

But really, if translation hasn't been done, I do think this is worth 
looking at for the code fencing and quote standardization alone, and the 
fix to the #11782 change listings.

I can try to pull any grammar/spelling fixes if that would help, although 
i don't think there were that many.

> Still, I don't want to hold up the release further for the changes file. Perhaps instead this can be come a discussion of how to make these changes through
> the full file?

We may always run into the translators not wanting to touch any older work 
(not that I blame them). That's why I'm pushing to get this as right as 
possible before translation.
But of course changes aren't as significant as the whole product, so as 
you wish.

I am willing to prune these commits however necessary to make this 
acceptable though.
Even if you'd rather leave the duplicate stuff re #11782 in, and just see 
the code fencing and quoting, I can do that.

feerrenrut:
> One problem it fixes, is that a few changes were listed multiple times, in
somewhat different ways, re #11782. I collapsed those into reasonable
singularities. (See db7c339; I haven't rebased yet)

That looks like a reasonable change. Changes like that would be best on their own. Getting the change made quickly is necessary at this stage.

Translation is currently in progress. Translators haven't finished, but those who have started will have to understand any changes.

I don't think that minor grammar and spelling issues are important at this stage, also code fencing (given how long it has not been done correctly)

Lets close this PR. It would be great if you would like to open a new one targeting beta fixing only the most important items that might affect comprehension.

PR #12435, opened 2021-05-19
Fix-up of #12432: Fix key error when reading appointment with categories

CyrilleB79:
### Link to issue number:

Fix-up of #12432

### Summary of the issue:

In Outlook Calendar, KeyError in the log when tabbing on an appointment having an associated  category.

In #12432, `.format` was used without key word argument, whereas the string to be formatted contained a named interpolated value.

#### Details
To reproduce / test:
* Open Outlook
* Press control+2 to go to calendar view
* Tab or shift+tab to focus an appointment

Note: using up/down arrows will read the appointment, but does not focus it and will not read the category.

Tested in Outlook 2016 and 365
- with Calendar view 'working week' and day.

If you do not have an appointment with a category, you can add one the following way:
* Open Outlook
* Press control+2 to go to calendar view (working week for me)
* Tab to an appointment
* Press context menu key
* Select "categorize" item
* Select a category in the sub-menu

### Description of how this pull request fixes the issue:

Call `.format` with the keyword.

### Testing strategy:
Tested in Outlook that appointments with no category, one category or wo categories are correctly reported.

### Known issues with pull request:
None

### Change log entries:
None; fixing a not released PR.

### Code Review Checklist:

- [x] Pull Request description is up to date.
- [x] Unit tests.
- [x] System (end to end) tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.
- [x] Context sensitive help for GUI changes.

Cc @feerrenrut  @seanbudd

PR #12448, opened 2021-05-21
Remove one and only use of ngettext for now

michaelDCurran:
### Link to issue number:
None

### Summary of the issue:
Recently NVDA began using ngettext to allow translators to provide both singular and plural translations for a particular message.
 However, it seems that our PO validation code in the translation system is not handling plural forms well, and is incorrectly classing translations as missing brace format variables, as it is incorrectly comparing the translation with the message before that one.
It then outputs the following error:
```
Message starting on line 7169
Original: "{startTime} to {endTime}"
Translated: "category {categories}"
Error: missing brace format interpolation, extra brace format interpolation
Expected: these brace format interpolations: {endTime}, {startTime}
Got: these brace format interpolations: {categories} 
```

ngettext was so far only used for one string in NVDA, in source/appModules/outlook.py. Seen by doing `git grep ngettext`.

We have many many strings in NVDA which would benefit from ngettext, but we need to know it works correctly. So until we are sure we can validate this correctly, we should stop using it in the one and only place it exists today.

Stop using ngettext for the one string in the Outlook appModule, instead replacing it with a normal use of `_()` providing just the plural form. 

### Testing strategy:
* Manually tested by Opening Outlook. Going to a calendar. Creating an appointment and adding 0, 1 and 2 categories, verifying each time that focusing the appointment NVDA reports  the categories.

None.

### Change log entries:
None needed.

New features
Changes
Bug fixes
For Developers

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual tests.
- [x] User Documentation.
- [x] Change log entry.

### DIFF
--- a/source/appModules/outlook.py
+++ b/source/appModules/outlook.py
@@ -37,7 +37,7 @@ from NVDAObjects.behaviors import RowWithFakeNavigation, Dialog
 from NVDAObjects.UIA import UIA
 from NVDAObjects.UIA.wordDocument import WordDocument as UIAWordDocument
 import languageHandler
-from gettext import ngettext
+
 
 PR_LAST_VERB_EXECUTED=0x10810003
 VERB_REPLYTOSENDER=102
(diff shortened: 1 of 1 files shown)


seanbudd (approved):

feerrenrut:
Originally added in https://github.com/nvaccess/nvda/pull/11598
Updated in:
- https://github.com/nvaccess/nvda/pull/12432
- https://github.com/nvaccess/nvda/pull/12435

PR #12475, opened 2021-05-28
Fix empty edit fields not speaking in Microsoft Edge

### LINKED ISSUE #12189: Lots of intermittent errors while using Microsoft Edge with NVDA
### Steps to reproduce:
1. Install edge beta/dev/caneri
2. Install NVDA latest snapshot
3. Turn on NVDA
4. Open edge
5. Navigate casually and observe
### Actual behavior:
Whenever we navigate the web or open the edge or go to the address bar, we get tons of errors and error sounds
### Expected behavior:
No errors should be thrown
Installed
#### NVDA version:
Latest alpha snapshot
#### Windows version:
Win 10 21332
Edge
I'm attaching the logs which I got while browsing the web, opening edge and going to address bar.
yes, always
No
Yes
Yes
[edge.txt](https://github.com/nvaccess/nvda/files/6158389/edge.txt)
[Log1.txt](https://github.com/nvaccess/nvda/files/6158450/Log1.txt)
[Log2.txt](https://github.com/nvaccess/nvda/files/6158453/Log2.txt)

### LINKED ISSUE #12474: Microsoft Edge/UIA: Expanding an empty text range to line results in a broken IUIAutomationTextRange
Note, this is likely a bug in Edge, but we might want to consider working around it

### Steps to reproduce:
1. Open Edge
2. Focus the address bar
3. Make it blank
4. Open a python console
5. Observe the following behavior
```
>>> ti=nav.makeTextInfo("all")
>>> ti.text
''
>>> ti.isCollapsed
True
>>> ti.expand("character")
>>> ti.isCollapsed
True
>>> ti.expand("line")
>>> ti.isCollapsed
  File "baseObject.pyc", line 26, in __get__
  File "textInfos\__init__.pyc", line 438, in _get_isCollapsed
  File "NVDAObjects\UIA\__init__.pyc", line 798, in compareEndPoints
  File "comtypesMonkeyPatches.pyc", line 27, in __call__
_ctypes.COMError: (-2147220991, 'An event was unable to invoke any of the subscribers', (None, None, None, 0, None))
```

### Actual behavior:
Expanding to line breaks the range object, makes it unusable. It can't be cloned, collapsed, etc.

### Expected behavior:
Expanding to line behaves the same as expanding to character, word, etc. when the text range is empty, and it should still be a collapsed range as it is empty,

Installed

#### NVDA version:
alpha-22944,31ef7c54

#### Windows version:
Windows 10 21h01

Edge 90.0.818.66 (Official build) (64-bit)

LeonarddeR:
### Link to issue number:
Fixes #12474 
Fixes #12189

### Summary of the issue:
Empty edit fields aren't spoken in Edge. There's a bug in the UIA implementation that causes expand to line to fail when there is no text content in the object

1. Use the ChromiumUIA object overlay for text boxes in the Chrome/Edge interface. For Edge, this applies by default. For Chrome, this only applies to cases where experimental UIA support is enabled, but this is far from stable in Chrome itself.
2. On the ChromeUIATextInfo when expanding to line, check whether there's any text in the document range. If not, don't try to expand as it will break the range.
3. Make sure not to cache the TextInfo property on UIA objects.ChromiumUIATextInfo. This is necessary because the property is fetched before the overlay classes are added that might change the result of evaluating the property. This ensures the proper TextInfo is used on focus.

### Testing strategy:
Tested that the address bar of Edge is read correctly when empty and focused.

None known.

### Change log entries:
Bug fixes
* In the new Chromium based Microsoft Edge, text fields such as the address bar are now announced when empty. (#12474)

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -940,6 +940,9 @@ class UIA(Window):
 				clsList.append(spartanEdge.EdgeList)
 			else:
 				clsList.append(spartanEdge.EdgeNode)
+		elif self.windowClassName == "Chrome_WidgetWin_1" and self.UIATextPattern:
+			from . import chromium
+			clsList.append(chromium.ChromiumUIA)
 		elif self.windowClassName == "Chrome_RenderWidgetHostHWND":
 			from . import chromium
 			from . import web
@@ -1265,8 +1268,9 @@ class UIA(Window):
 
 	_TextInfo=UIATextInfo
 	def _get_TextInfo(self):
-		if self.UIATextPattern: return self._TextInfo
-		textInfo=super(UIA,self).TextInfo
+		if self.UIATextPattern:
+			return self._TextInfo
+		textInfo = super(UIA, self).TextInfo
 		if textInfo is NVDAObjectTextInfo and self.UIAIsWindowElement and self.role==controlTypes.ROLE_WINDOW:
 			import displayModel
 			return displayModel.DisplayModelTextInfo
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -1267,6 +1267,8 @@ class UIA(Window):
 		return self.UIALegacyIAccessiblePattern
 
 	_TextInfo=UIATextInfo
+	_cache_TextInfo = False
+
 	def _get_TextInfo(self):
 		if self.UIATextPattern:
 			return self._TextInfo
--- a/source/NVDAObjects/UIA/chromium.py
+++ b/source/NVDAObjects/UIA/chromium.py
@@ -6,6 +6,8 @@
 import UIAHandler
 from . import web
 import controlTypes
+from comtypes import COMError
+from logHandler import log
 
 """
 This module provides UIA behaviour specific to the chromium family of browsers.
@@ -15,6 +17,16 @@ or UIA.anaheim_edge.
 
 
 class ChromiumUIATextInfo(web.UIAWebTextInfo):
+
+	def expand(self, unit):
+		# #12474: Expanding to line breaks when the underlying text range is empty.
+		copy = self._rangeObj.Clone()
+		super().expand(unit)
+		try:
+			self._rangeObj.Compare(copy)
+		except COMError:
+			log.debugWarning(f"Expand to {unit} failed for {self}, resorting to backup range before expand")
+			self._rangeObj = copy
 
 	def _getFormatFieldAtRange(self, textRange, formatConfig, ignoreMixedValues=False):
 		formatField = super()._getFormatFieldAtRange(textRange, formatConfig, ignoreMixedValues=ignoreMixedValues)
--- a/source/NVDAObjects/UIA/chromium.py
+++ b/source/NVDAObjects/UIA/chromium.py
@@ -1,13 +1,11 @@
 # A part of NonVisual Desktop Access (NVDA)
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
-# Copyright (C) 2020 NV Access limited, Leonard de Ruijter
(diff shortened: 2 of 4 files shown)


leonardder:
I actually think that this pr will mean a significant improvement for Edge users, therefore I'd like to request to have this merged into beta instead.

cary-rowen:
This is a significant improvement and hope to be merged into Beta

alexarnaud:
Thanks a lot @leonardder for fixing this.

I have two questions : 
* Do you still reproduce with Edge version 91 released today?
* Is it possible for you to report the UIA issue against Edge?

Best regards.

leonardder:
@alexarnaud  wrote:

> I have two questions :
> 
> * Do you still reproduce with Edge version 91 released today?

Yes.

> * Is it possible for you to report the UIA issue against Edge?

If I knew where to do that, I'd certainly would. DO I need to do it against Edge or Chromium? Thoughts @michaelDCurran?

josephsl:
Hi, I recommend contacting Microsoft Edge team (hopefully through Feedback hub if possible) and also ask that Chromium/Chrome team look into this issue (because this could also impact other Chromium based browsers). Thanks.

leonardder:
I think it would be better if NV Access pokes their Microsoft contacts about this.

josephsl:
Hi, and Google, too.

leonardder:
Well, as long as Google's implementation is experimental, I guess it's up to Microsoft to push fixes upstream.

Brian1Gaff:
If one could trace it directly to the Chrome team, surely then it would be 
on all and fixed with the same patch otherwise you might end up with a 
different work around for every browser if one is not careful.
 Brian

***@***.***
Sent via blueyonder.
in the display name field.
----- Original Message -----

michaelDCurran (changes requested):
I am concirned that this may introduce performance issues, Especially also as these days it is important to try and keep the number of cross-process calls in UIA to a minimum with scenarios such as WDAG and RAIL.

In ChromiumUIATextInfo.expand: instead of the added calls to clone and compare, why not just call GetText(1) to fetch the first 1 characters and check if it is the empty string or not. If it is empty, then simply don't call expandToEnclosingUnit.
 
Although in general you are probably correct that we should call obj.invalidateCache after applying overlay classes, I would prefer this to be in its own pr, where we can very carefully consider and track any performance issues this may cause. This change obviously affects code across NVDA, not just the Edge address bar.
However, to address the specific issue of TextInfos incorrectly being cached here, perhaps we can at least set `_cache_TextInfo = False` on the UIA NVDAObject which will stop the TextInfo property from being cached. All the major calls it makes internally are cached properties themselves, so this would have no real performance impact, but it would still allow the current value of `_TextInfo` to be used, thus an overlay with a different TextInfo class would correctly be reflected in speakObject after overlay classes were applied but still in the same core cycle.

leonardder:
> In ChromiumUIATextInfo.expand: instead of the added calls to clone and compare, why not just call GetText(1) to fetch the first 1 characters and check if it is the empty string or not. If it is empty, then simply don't call expandToEnclosingUnit.

Is that going to work? I assume GetText returns text in the range and doesn't return anything when the range is collapsed, therefore if the range is collapsed, it will never expand.

michaelDCurran:
Ah, sorry, I mean self.obj.UIATextPattern.documentRange.GetText(1)

leonardder:
Ah, yeah that makes sence. I changed this as requested.
I also tested every UIA unit and it turns out that it only fails when expanding to line as I originally thought. Therefore I changed the code to only apply when expanding to line. I will change the description accordingly.

michaelDCurran (approved):

PR #12544, opened 2021-06-13
winVersion/Windows 10 version fetch: use Windows Registry instead of a static release names to builds

### LINKED ISSUE #12509: Windows 10 versions: use Windows Registry to obtain release Id/display name
Hi,

I'd be happy to provide a pull request regarding this - after all, what I'm proposing was already implemented as part of an add-on.

### Introduction
As part of Windows as a Service (WaaS) model, Microsoft releases new releases of Windows every six months or so. As of June 2021, the latest release is 21H1 (May 2021 Update).

Until Windows 10 RTM (Version 1507/build 10240), one way to determine Windows releases was build number (Windows 10 uses Windows NT version 10.0 internally). With Version 1511 (November Update released in 2015), Microsoft added "ReleaseId" (aka YYMM) to identify releases, with this info displayed as part of About Windows (WinVer) dialog. The release Id can be found under the Windows Registry path HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion.

Then in 2020, Microsoft made two related changes: the codename for a feature update became a public name for a semi-annual update, the best known being Version 20H2 (October 2020 Update). To reflect this change, a new key named "DisplayName" was added to the mentioned Registry path, which is used by WinVer instead of ReleaseId when displaying version information. If you recall, a few months ago we had a discussion about Windows 10 versions to be used (2009 versus 20H2) as part of #11795 (see that issue for details), and the new Registry key is the reason why.

The picture became complicated in 2021. As of June 2021, Microsoft is working on at least four code bases: Vibranium (2004/20H2/21H1), Iron Server 2022 and initial (and now canceled) version of Windows 10 X), Cobalt (a likely 21H2 build), and Nickel (next year and first bits may show up to dev channel Windows Insiders in a few weeks). Consequently, four build ranges were spotted by Windows Insiders: 1904x for Vibranium, 203xx for Iron, 21xxx for Cobalt, 22xxx for Nickel. As a result, it becomes harder to keep track of which release is which in the current Windows 10 version implementation, which is described below.

Currently winversion module includes a record of which Windows 10 release is tied to a specific build. This is used by winVersion.WinVersion class to display Windows 10 release information.

As described above, with multiple code bases being developed, it becomes harder to associate a feature update with a specific build number. This will become an issue when recognizing 21H2 later this year, as one needs to use a tuple or a similar structure to record multiple builds for a given release and look up the correct index based on product type.

Retrieve Windows 10 version information (Release Id/display name) directly from Windows Registry, and to use an updated releases to builds map as a fallback if Registry access is denied. In short, the following solution (source: Resource Monitor add-on which I am currently maintaining) is proposed:

1. For Windows 10 build 10240, return "1507".
2. Look for "DisplayName" key from the Registry path noted above.
3. If it fails, look up "ReleaseId" key.
4. If Registry access is completely denied, use the refined Windows 10 releases to builds map as a fallback. I expect it won't be necessary unless Registry access is denied to user programs such as NVDA.

This solution is applicable to Windows Server releases as well i.e. Server 2019 will say "Windows 10 1809" when winVersion.GetWinVer() is called.

The biggest advantage is that there is no need to add entries to releases to builds map when making NVDA recognize new Windows 10/Server releases - all the developer needs to do is add a WIN10_* constant (for Server 2022, if it turns out the build number is different then client releases, I think it would be better to call it WIN10_VER_SERVER, or perhaps WINSERVER_2022) since the needed info comes from Windows Registry. Plus, for Insiders on dev channel, display name is "Dev", making it easier for developers to determine which Windows 10 release is in use by simply looking for "Windows 10 whatever".

The biggest issue is prerelease builds that are being tested by beta and release preview Insiders. Prior to general availability, Microsoft assigns a specific build as the feature update build, and this will be noticed by beta and release preview Insiders ahead of time by months to weeks, respectively. Therefore, to make sure people can add the correct Windows version information/constants, I propose that people should wait until a feature update is released to release preview Insiders before sending a pull request to update winVersion module (this was the route I took to add 21H1 information a few weeks ago).

Perhaps an unlikely issue is Registry access denial. I doubt this will be the case, but just in case this happens, I think existing releases to builds map should be used as a fallback to some degree.

Leave it as is and refine releases to builds map to account for a given feature release using different build numbers based on product type. One approach (although folks will need to wait until 2022.1 cycle) is change releases to builds map so that it records builds:releases instead of releases:builds.

### Additional context
For a practical implementation, see:
https://github.com/josephsl/resourceMonitor

Don't worry about Windows Server LTS bits as it is not applicable to this issue. Also, if anyone would like to base a PR on bits of resource Monitor, feel free to do so.

Thanks.

josephsl:
Hi,

Mostly in preparation for a possibility that Windows 10 Version 21H2 will be used by multiple builds, including Server 2022, and prepares NVDA to support Windows 11 and future Windows versions:

### Link to issue number:
Closes #12509 

### Summary of the issue:
Currently a static Windows 10 release names to builds map is used to record Windows 10 builds to feature updates (releases). Until 21H1 there was a one-to-one correspondence between releases and builds, but it appears Version 21H2 will be represented by multiple builds:

* Windows 10: 19044
* Server 2022: 20348
* Windows 11: build 22000

Builds subject to confirmation.

Therefore this breaks current static map.

Since Version 1511, Windows Registry holds release name which corresponds to builds (e.g. 1511 = 10586, 20H2 = 19042), housed inside a "Release Id" key. Starting with Version 20H2, a separate "DisplayVersion" key is used to announce public release name. In Windows Insider Preview builds, dev channel has display version of "Dev", making it even easier to determine what kind of build a user is running. This tradition continues with Windows 11 and Server 2022, with Registry recording "21H2", which makes sense given their projected release period. Therefore use Windows Registry instead of the static releases to builds map and deprecate releases to builds map.

### Testing strategy:
Manual testing:

1. On supported Windows 10 releases (1507 LTS, 1607/server 2016 LTS, 1809/Server 2019 LTS, 1909 (Enterprise and Education), 2004, 20H2, 21H1, 21H2/Server 2022 preview, Windows Insider Preview dev channel), open Windows Registry path described in #12509 and confirm that release Id's and display name keys are present (the latter is available starting with 20H2).
2. In Python Console, verify that winVersion.getWinVer() says "Windows 10 YYMM/YYHn" and the last part (release name) matches the value obtained from Windows Registry.
3. Optionally, define a local unit test that checks the version string returned by winVersion.getWinVer() starts with "Windows 10 release" for the local system (e.g. "Windows 10 21H1" on build 19043).

None

### Change log entries:
Likely a change for developers, stating that on Windows 10, release names are obtained from Windows Registry, although note that this is an internal operation. A visible changelog is:

Changes for developers:

On Windows 10 Version 1511 and later (including Insider Preview builds), current Windows feature update release name is obtained from Windows Registry. (#12509)
The releases to builds mapping found in winVersion module is deprecated and will be removed in 2022.1. (#12544)

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### Discussion:
If this PR is adopted, what should we do about the static release names to builds map? There are three possibilities:

1. Swap releases with builds so the map records build : release. This makes Windows Registry method unnecessary - adding future Windows releases will continue to involve editing this map.
2. Deprecate and remove the static map - the only place the map is used is when fetching Windows 10 release names anyway, and since Windows Registry contains needed information, there is no need for this map. If deprecation and removal are chosen, a follow-up PR should be created during 2022.1 development cycle to remove the static map altogether.
3. For releases represented by multiple builds, have Windows 10 release name fetcher check for these builds and return the release name for these. This makes version fetch routine complicated as one needs to hunt for all build numbers represented by a single release name.

Update (2021-07-07): deprecated and replaced by a private builds to releases mapping.

Thanks.


josephsl:
Hi,

By the way, people watching NVDA add-ons may notice a similarity with Resource Monitor - in fact, this PR is a simplified version of Windows 10 version fetch routine found in that add-on, therefore it went through years of real-life usage.

Thanks.

cary-rowen:
@josephsl 
There is a digression, Windows 10 application enhancement is so versatile; after years of extensive use and testing, why not integrate it into NVDA itself?
Thanks.

josephsl:
Hi, not all features from that add-on can be integrated into NVDA because Windows 10 apps (built-in and from Microsoft Store) can change without notice, which also applies to features found in Windows Insider Preview builds (things come and go without notice, making it harder to say something is stable enough to be part of NVDA unless it’s been out there for a while). To be honest, Windows 10 App Essentials is “versatile” because of the state of app and feature accessibility and usability across Windows 10 releases (many commits in this add-on are frustrations about app accessibility and usability issues, which have lessened somewhat in recent years). Another key consideration is support duration for Windows 10 releases – some features are part of Windows 10 releases no longer supported by Microsoft, therefore if support for them ends in the add-on, I promptly remove the code associated with it (the add-on is currently going through such a transition to fully remove support for releases earlier than Version 2004). Thanks.

cary-rowen:
Thanks @josephsl for reply

source/winVersion.py
```diff
@@ -49,12 +50,22 @@ def _windowsVersionToReleaseName(self):
...
-			if self.build in buildsToReleases:
-				return f"Windows 10 {buildsToReleases[self.build]}"
-			else:
-				# Windows Insider build.
-				return "Windows 10 prerelease"
+			# From Version 1511 (build 10586), release Id/display version comes from Windows Registry.
+			# Always return "Windows 10 1507" on build 10240.
+			if self.build == 10240:
+				return "Windows 10 1507"
+			currentVersion = winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, r"Software\Microsoft\Windows NT\CurrentVersion")
```
leonardder, line 57: Please use a with clause

LeonarddeR (changes requested):
I assume it will suffice to remove the release map altogether.

josephsl:
Hi, I once saw Registry access used as a context manager but wasn’t sure if it is applicable; now you mention it, I’ll perform some tests and make necessary changes (this is also applicable to Resource Monitor add-on, come to think of it). Thanks for this tip.

josephsl:
Hi,

In the end, there is no point in keeping something that will soon break (look at various 21H2 build branches), so the releases map will be gone.

Thanks.

source/winVersion.py
```diff
@@ -49,12 +50,23 @@ def _windowsVersionToReleaseName(self):
...
+				# Version 20H2 and later where a separate display version string is used.
+				# For backward compatibility, release Id will store display version string.
+				try:
+					releaseId = winreg.QueryValueEx(currentVersion, "DisplayVersion")[0]
+				except OSError:
+					releaseId = None
+				# Version 1511 and later unless display version string is present.
+				if not releaseId:
+					releaseId = winreg.QueryValueEx(currentVersion, "ReleaseID")[0]
+			return f"Windows 10 {releaseId}"
```
feerrenrut, line 69: This will return the name of the Windows release you are running on regardless of how the `WinVersion` class is constructed right?

Eg, ensure there is a unit test for this. I suspect this will fail on any computer running a later version of Windows.
```python
assertEqual(repr(WIN10_1511), "Windows 10 10586")
```

source/winVersion.py
```diff
@@ -124,21 +136,5 @@ def isUwpOcrAvailable():
 	return os.path.isdir(UWP_OCR_DATA_PATH)
 
 
-WIN10_RELEASE_NAME_TO_BUILDS = {
```
feerrenrut, line 127: This is part of the API, it can be deprecated, but shouldn't be removed.

source/winVersion.py
```diff
@@ -49,12 +50,23 @@ def _windowsVersionToReleaseName(self):
 		elif (self.major, self.minor) == (6, 3):
 			return "Windows 8.1"
 		elif self.major == 10:
-			buildsToReleases = {build: release for release, build in WIN10_RELEASE_NAME_TO_BUILDS.items()}
```
feerrenrut, line 52: If this is the only usage, and it reverses the mapping, you could consider adding a new mapping, deprecating the old one. The new mapping would be internal only, and map builds to release names. Though this isn't necessary if we just construct the `class WinVersion` with the release name populated.

source/winVersion.py
feerrenrut: If you keep this for the `getWinVer` case, please cache it at module level, it doesn't need to happen more than once.

feerrenrut (changes requested):
I'm ok with getting the Release name like this (from the registry) when the `getWinVer` method is called. But for the constants it would be better to stick to an in code mapping, perhaps instead of the `WIN10_RELEASE_NAME_TO_BUILDS` dictionary, add an argument to `class WinVersion` to set the release name and set all the names with the constants. Then there can be a constant for each variation of  "21H2 " EG something like `WIN10_21H2_SERVER_2022 = WinVersion(major=10, minor=0, build=20348, releaseName="Windows 10 21H2 Server 2022")`

josephsl:
> I'm ok with getting the Release name like this (from the registry) when the `getWinVer` method is called. But for the constants it would be better to stick to an in code mapping, perhaps instead of the `WIN10_RELEASE_NAME_TO_BUILDS` dictionary, add an argument to `class WinVersion` to set the release name and set all the names with the constants. Then there can be a constant for each variation of "21H2 " EG something like `WIN10_21H2_SERVER_2022 = WinVersion(major=10, minor=0, build=20348, releaseName="Windows 10 21H2 Server 2022")`

Hi,

Adding "releaseName": this will also benefit Windows 7 and 8.x.

Windows Registry and release name: yes, Windows Registry will return the correct release name depending on the version you've got, so it can be used in winVersion.getWinVer function, I was wondering about unit tests - I'll come up with soething.

Release names to builds map: not needed as Windows Registry and release name argument will suffice.

Caching release name from the module: right, that is a useful optimization strategy.

Would you like me to restore the release names to builds dictionary for 2021.x and mark it as deprecated in the process?

Thanks.

josephsl:
Hi,

For Windows Server 2022, at least "Windows 10 21H2" will be returned, which is fine. The overall PR could change depending on what we learn on June 24th.

Thanks.

josephsl:
Hi,

As for unit test, str.startswith should be checked as repr records release name, build, and product type in one go.

Thanks.

source/winVersion.py
feerrenrut: You'll need to restore this dictionary and add a comment that it is deprecated to be removed in 2022.1. You can also add a comment that it isn't used internally and should not be updated further.

source/winVersion.py
```diff
@@ -26,35 +27,38 @@ def __init__(
...
-			if self.build in buildsToReleases:
-				return f"Windows 10 {buildsToReleases[self.build]}"
-			else:
-				# Windows Insider build.
-				return "Windows 10 prerelease"
+			# From Version 1511 (build 10586), release Id/display version comes from Windows Registry.
+			# Always return "Windows 10 1507" on build 10240.
+			if self.build == 10240:
+				return "Windows 10 1507"
+			return f"Windows 10 {WIN10CURRENTRELEASENAME}"
```
feerrenrut, line 61: I don't think this should fall back to the current release name, that would be quite confusing.

feerrenrut: If we want to support manual construction of these classes rather than only using the pre-constructed ones `WIN10_1607` etc, and automatically supplying the release name where it is known, then instead of passing in the name on construction make a build to release name map and look up the release name from the build.

source/winVersion.py
```diff
@@ -81,22 +85,38 @@ def __ge__(self, other):
...
+	# Version 20H2 and later where a separate display version string is used.
+	# Release Id (older releases) and display version (20H2 and later) will be cached.
+	try:
+		WIN10CURRENTRELEASENAME = winreg.QueryValueEx(currentVersion, "DisplayVersion")[0]
+	except OSError:
+		# Don't set anything if this is Windows 10 1507 or earlier.
+		try:
+			WIN10CURRENTRELEASENAME = winreg.QueryValueEx(currentVersion, "ReleaseID")[0]
+		except OSError:
+			WIN10CURRENTRELEASENAME = None
```
feerrenrut, line 119: Please put this in a function I'd recommend calling it directly from `getWinVer` to supply the `releaseName=` argument when constructing `WinVersion`

source/winVersion.py
feerrenrut: Could you make this happen automatically, like this:
``` python
# Deprecated: Windows 10 releases will be obtained from Windows Registry, no entries will be added.
# WIN10_RELEASE_NAME_TO_BUILDS will be removed in 2022.1.
if version_year < 2022:
	WIN10_RELEASE_NAME_TO_BUILDS = {
```

source/winVersion.py
```diff
@@ -81,32 +70,94 @@ def __ge__(self, other):
...
+			# Version 20H2 and later where a separate display version string is used.
+			try:
+				releaseId = winreg.QueryValueEx(currentVersion, "DisplayVersion")[0]
+			except OSError:
+				# Don't set anything if this is Windows 10 1507 or earlier.
+				try:
+					releaseId = winreg.QueryValueEx(currentVersion, "ReleaseID")[0]
+				except OSError:
+					releaseId = "unknown"
+		return f"Windows 10 {releaseId}"
```
feerrenrut, line 144: Please put this code in a separate function EG `_getRunningVersionNameFromReg`, I don't think it should be called here. It's too easy to conflate the version name for the build given by major, minor, build args with the version name for the currently running version of windows.

In this new function `_getRunningVersionNameFromReg`:
- protect against trying to query the registry before build 1511, raise an exception.
- Cache the result, it shouldn't have to open the registry twice, the running version of windows is not going to change. I'd use [`@lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache).

josephsl:
Hi, ah, similar to what Sean is doing with control types refactor? Sure – give me a sec. Thanks for this tip.

source/winVersion.py
```diff
@@ -81,32 +70,94 @@ def __ge__(self, other):
...
+WIN10_1903 = WinVersion(major=10, minor=0, build=18362, releaseName="Windows 10 1903")
+WIN10_1909 = WinVersion(major=10, minor=0, build=18363, releaseName="Windows 10 1909")
+WIN10_2004 = WinVersion(major=10, minor=0, build=19041, releaseName="Windows 10 2004")
+WIN10_20H2 = WinVersion(major=10, minor=0, build=19042, releaseName="Windows 10 20H2")
+WIN10_21H1 = WinVersion(major=10, minor=0, build=19043, releaseName="Windows 10 21H1")
+
+# Records a mapping between Windows builds and release names.
+# These include build 10240 for Windows 10 1507 and releases with multiple release builds.
+# These are applicable to Windows 10 as they report the same system version (10.0).
+BUILDS_TO_RELEASE_NAMES = {
```
feerrenrut, line 94: Please make this internal 
```suggestion
_BUILDS_TO_RELEASE_NAMES = {
```

source/winVersion.py
```diff
@@ -81,32 +70,94 @@ def __ge__(self, other):
...
-WIN10_1909 = WinVersion(major=10, minor=0, build=18363)
-WIN10_2004 = WinVersion(major=10, minor=0, build=19041)
-WIN10_20H2 = WinVersion(major=10, minor=0, build=19042)
-WIN10_21H1 = WinVersion(major=10, minor=0, build=19043)
+WIN7 = WinVersion(major=6, minor=1, build=7600, releaseName="Windows 7")
+WIN7_SP1 = WinVersion(major=6, minor=1, build=7601, releaseName="Windows 7", servicePack="1")
+WIN8 = WinVersion(major=6, minor=2, build=9200, releaseName="Windows 8")
+WIN81 = WinVersion(major=6, minor=3, build=9600, releaseName="Windows 8.1")
+WIN10 = WIN10_1507 = WinVersion(major=10, minor=0, build=10240, releaseName="Windows 10 1507")
+WIN10_1511 = WinVersion(major=10, minor=0, build=10586, releaseName="Windows 10 1511")
```
feerrenrut, line 79: Now that we have `BUILDS_TO_RELEASE_NAMES` you can remove this. I'd make it an optional argument on `class WinVersion` which should look it up using the function `_getWindowsReleaseName` but don't fall back to `_getRunningVersionNameFromReg`

source/winVersion.py
```diff
@@ -81,32 +70,94 @@ def __ge__(self, other):
...
+	else:
+		raise RuntimeError("Unknown Windows release")
 
 
 def getWinVer():
 	"""Returns a record of current Windows version NVDA is running on.
 	"""
 	winVer = sys.getwindowsversion()
+	try:
+		releaseName = getWindowsReleaseName(major=winVer.major, minor=winVer.minor, build=winVer.build)
```
feerrenrut, line 154: Try `_getRunningVersionNameFromReg` catch exception, then construct without a release name explicitly set, allowing `class WinVersion` to populate it via `_getWindowsReleaseName`.

feerrenrut (changes requested):

source/winVersion.py
```diff
@@ -81,32 +70,94 @@ def __ge__(self, other):
...
+	17763: "Windows 10 1809",
+	18362: "Windows 10 1903",
+	18363: "Windows 10 1909",
+	19041: "Windows 10 2004",
+	19042: "Windows 10 20H2",
+	19043: "Windows 10 21H1"
+}
+
+
+def getWindowsReleaseName(major: int = 0, minor: int = 0, build: int = 0) -> str:
```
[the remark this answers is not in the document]
josephsl: Done - it's a matter of restoring and tweaking the private method from previous commits.

source/winVersion.py
josephsl: Yep - the new WinReg method will return an empty string for release Id, which is then processed inside the private release name method. This is actually a stepping stone toward #12585 once this PR is merged into master. Thanks.

source/winVersion.py
josephsl: Done. Note that maxsize argument is supplied becasue we are using Python 3.7; in 3.9 there is no need to supply maxsize when defining the decorator.

source/winVersion.py
```diff
@@ -34,21 +78,43 @@ def __init__(
...
+		elif self.major == 10:
+			# From Version 1511 (build 10586), release Id/display version comes from Windows Registry.
+			# However there are builds with no release name (Version 1507/10240)
+			# or releases with different builds.
+			# Look these up first before asking Windows Registry.
+			if self.build in _BUILDS_TO_RELEASE_NAMES:
+				return _BUILDS_TO_RELEASE_NAMES[self.build]
+			releaseId = _getRunningVersionNameFromWinReg()
+			if not releaseId:
+				releaseId = "unknown"
```
feerrenrut, line 111: `_getWindowsReleaseName` shouldn't fall back to the registry, the only case where the registry should be used is when getting the version for the version currently running.

```suggestion
			if self.build in _BUILDS_TO_RELEASE_NAMES:
				return _BUILDS_TO_RELEASE_NAMES[self.build]
			if not releaseId:
				releaseId = "unknown"

```

source/winVersion.py
```diff
@@ -12,6 +12,50 @@
...
+	17763: "Windows 10 1809",
+	18362: "Windows 10 1903",
+	18363: "Windows 10 1909",
+	19041: "Windows 10 2004",
+	19042: "Windows 10 20H2",
+	19043: "Windows 10 21H1"
+}
+
+
+@functools.lru_cache(maxsize=128)
```
feerrenrut, line 37: I think this could only every have one value:
```suggestion
@functools.lru_cache(maxsize=1)

```

source/winVersion.py
```diff
@@ -70,94 +136,32 @@ def __ge__(self, other):
...
 	"""
 	winVer = sys.getwindowsversion()
-	try:
-		releaseName = getWindowsReleaseName(major=winVer.major, minor=winVer.minor, build=winVer.build)
-	except RuntimeError:
-		releaseName = "Windows release unknown"
 	return WinVersion(
 		major=winVer.major,
 		minor=winVer.minor,
 		build=winVer.build,
```
feerrenrut: Since this getting the currently running version of windows:
```suggestion
	try:
		releaseName = f"Windows 10 {_getRunningVersionNameFromWinReg()}"
	except RuntimeError:
		releaseName = None
	return WinVersion(
		major=winVer.major,
		minor=winVer.minor,
		build=winVer.build,
		releaseName=releaseName,
```

feerrenrut (changes requested):

josephsl:
Hi,

Ah, these clarify things a lot - in the future I should sleep before working on PR's...

Thanks.

source/winVersion.py
josephsl: When preparing for #12585, we might as well enhance release name or the private WinReg function to say "Windows 11" if running specific builds.

josephsl:
Hi,

Regarding first suggestion (release Id inside get release name method), since release Id is not used, NVDA will simply say, "Windows 10 unknown". This is done to resolve Flake8 errors, remove unused variable, and to prepare for Windows 11 and future releases.

Thanks.

josephsl:
Hi, you’re right – thanks for catching this.

source/winVersion.py
```diff
@@ -26,20 +72,26 @@ def __init__(
 			major: int = 0,
 			minor: int = 0,
 			build: int = 0,
+			releaseName: Optional[str] = "",
```
lukaszgo1, line 75: This  is quite unusual. The standard usage for `Optional` is for arguments which are set to `None` by default yet can accept a given type. Here however you're setting it into an empty string as a default value and passing `None` when constructing the instance. Can this be simplified to either always using empty  string and not accepting `None` or preferably using `None` as default value for this kwarg?

source/winVersion.py
```diff
@@ -124,20 +183,23 @@ def isUwpOcrAvailable():
...
-	"1903": 18362,
-	"1909": 18363,
-	"2004": 19041,
-	"20H2": 19042,
-	"21H1": 19043,
-}
+# Deprecated: Windows 10 releases will be obtained from Windows Registry, no entries will be added.
+# The below map will be removed in 2022.1.
+if version_year < 2022:
+	WIN10_RELEASE_NAME_TO_BUILDS = {
```
lukaszgo1, line 189: Please add a change log entry to the PR description explaining that this map is deprecated.

josephsl:
Hi, not really as this map is only used to look up Windows 10 releases, and no-one uses it. I feel it should have been a private map to begin with, but…

feerrenrut (changes requested):
Code looks good now.

Let's add support for Windows 11 in a follow up PR.
Can you please update the PR description, including deprecations in the section for the changes file.

josephsl:
Hi,

PR intro comment updated with what's new entries (any other changes needed?).

Thanks.

feerrenrut (approved):

feerrenrut:
Yes, looks good @josephsl.

feerrenrut (approved):
I think it is more useful if the deprecation entry in the change log points to the PR rather than the issue.

josephsl:
Hi, agreed. Thanks.

PR #11494, opened 2020-08-15
Provide proper path to the .reg file in COMRegistrationFixes tool

lukaszgo1:
### Link to issue number:
Discussion in #9709 
### Summary of the issue:
In the COMRegistrationFixes tool the path to the .reg  file was incorrect `COMRegistryFixes` rather than `COMRegistrationFixes`. This causes tool to fail in some cases in which performing these steps manually worked. The problem was not detected earlier because `regedit.exe` returns 0 as its exit code even on failure (see [this Stack Overflow question](https://stackoverflow.com/questions/18296926/can-i-get-an-exit-code-from-running-regedit-s)).

### Description of how this pull request fixes the issue:
The proper path to the .reg file is used. Also the function used to apply reg files checks if the file with the given path exists and raises an exception if it does not.
### Testing performed:
None - this is just a path change.
### Known issues with pull request:
We should consider using reg.exe as per the SO question linked above to be able to check if merging file to the registry succeeded or not. I haven't done that in this PR because I believe this PR should be included in 2020.3 and it would be a riskier change. We should also display a message to the user if the tool failed when it is being run manually from the menu.
### Change log entry:

Section: Bug fixes

COM Registration Fixes tool now works in more cases

PR #12560, opened 2021-06-17
Refactor COM Registration Fixing Tool part 1: make the tool effective on 64-bit Windows and fix more problems

### LINKED ISSUE #9039: COM Registration Fixing tool doesn't work
### Steps to reproduce:
1. Install MathPlayer.
2. Uninstall MathPlayer.
3. Read the task bar and notice it says unknown.
4. Run the COM Registration Fixing tool.
5. Read the task bar again and it still says unknown.

### Actual behavior:
The tool doesn't fix NVDA's accessibility support.

### Expected behavior:
The tool fixes NVDA's accessibility support.

Installed

#### NVDA version:
Version: alpha-16405,19bb21c2

#### Windows version:
Version 1809 (OS Build 17763.134)

Yes

No

XLTechie:
### Link to issue number:

Fixes #9039 

### Summary of the issue:

The COM Registration Fixing Tool seems to have never worked fully on 64bit Windows, as a result of path confusion caused by the use of SysWOW64, and the executable names used.

In addition, modern browsers seem to check a typelib registration that could be damaged by certain uninstallations (e.g. older versions of MathPlayer). We didn't previously do anything about this.

### Further background:

In 64 bit Windows, `%WINDIR%\System32` contains 64-bit executables. Meanwhile, `%WINDIR%\SysWOW64` contains 32-bit executables. That seems inverted from what it should be, but there is a good explanation [here](https://www.samlogic.net/articles/32-64-bit-windows-folder-x86-syswow64.htm) for why it's like that.

When a 32-bit application calls something in `System32`, SysWOW64 assumes that it wants the 32-bit version, and redirects the call to the version in the `SysWOW64` directory.
If the 32-bit application really wants the 64-bit version of whatever it's calling, it has to use the virtual directory `%WINDIR%\Sysnative`, which SysWOW64 provides.

We were calling `regedit.exe` and `regsvr32.exe` in contexts where we assumed 32-bit, from either `%WINDIR%` or `%WINDIR%\System32`. Because NVDA is 32-bit, that would result in the 32-bit application being called. That's okay.
However when the test was made for 64-bit, executables in `%WINDIR%\SysWOW64` were being used, which are also 32-bit.

At least one of those should be 64-bit for COM reregistrations, but neither are.

In no case should we ever call anything in %WINDIR%\SysWOW64 ourselves, which we currently do.

Taken from the commit that rewrites `COMRegistrationFixes/__init__.py`:
Repaired the COM Registration Fixing Tool so it is more effective on 64bit Windows (Partial fix of #9039).

* Problems:
    - The `registerServer` function called the 32-bit version of `regsvr32.exe`, even in 64-bit contexts.
    - The `applyRegistryPatch` function called the 32-bit version of `regedit`, even in 64-bit contexts.
    - The Win7 32-bit run did not take into account 32-bit only systems (no `Program Files (x86)` folder).
* Remediation:
    - Replaced `applyRegistryPatch` function with two new functions: `apply32bitRegistryPatch` and `apply64bitRegistryPatch`.
    - Replaced `registerServer` function with `register32bitServer` and `register64bitServer`, to make clear what they do.
    - The new functions don't check 32/64 bitness; they leave that to the caller and log errors on failure.
    - Updated to more descriptive error logging.
    - Adjusted the Windows 7 code to use server registration with proper bitness for each DLL.
* Path remediations:
    - Moved the `OLEACC_REG_FILE_PATH` constant to the top of the file with the rest.
    - Added `Sysnative` path to the list of path constants at the top of the file.
    - Now use `Sysnative` in the path for intentional 64bit calls.
    - Now use `System32` in the path for 32-bit calls on either 32-bit or 64-bit systems.
    - Now use `reg.exe`'s import option to load .reg files instead of `regedit.exe`.
    - Now check whether to use "`Program Files`" or "`Program Files (x86)`" on Win7.
    - Removed now unused `sysWow64` path constant.
* Misc:
    - Added docstring note about 32 and 64 bit functions needing attention if NVDA goes 64-bit in the future.
    - Converted path constants to uppercase-with-underscore style, and corrected case on some Windows paths.
    - Moved comments with discussion links into module docstring, and rearranged.
    - Used `subprocess.STARTUPINFO` to prevent console windows from showing during executions.
    - In `gui/__init__.py` update the message shown on tool completion, to recommend a restart.

Separately, @michaelDCurran found a registry key that we should delete to restore browser functionality under certain circumstances. That is included as a separate joint commit.

### Testing strategy:

Followed the STR in [#9039  (comment)](https://github.com/nvaccess/nvda/issues/9039#issuecomment-857356117).

* With current Alphas, could not restore the registry, as described in that STR.
* With these modifications, the registry key was restored, indicating that the reg patch successfully applied.
* Installed and uninstalled MathPlayer to break browsers, and used the final version of the tool to fix it.

Windows 7 tests would be valuable, but I can not run them.

My original refactor of this tool is still in progress, but that will now be limited to UX enhancements, since this covers the guts of the tool. These are important updates to the functionality, and imo should be accepted apart from the rest of the refactor.

### Change log entries:
Bug fixes
The COM Registration Fixing Tool has been rewritten to solve more problems, and work better on 64bit Windows systems.

Changes for developers:
```
- In the `COMRegistrationFixes` module, the `applyRegistryPatch` function has been replaced by `apply32bitRegistryPatch` and `apply64bitRegistryPatch` functions. The `registerServer` function has been replaced by `register32bitServer` and `register64bitServer` functions.
```

### Code Review Checklist:

- [X] Unit tests.
- [X] Manual testing.
- [X] User Documentation.
- [X] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


michaelDCurran (changes requested):
Code looks all great to me. However, due to careful testing required and NVDA 2021.1 rc just about to come out, I would like this pr rebased on master for NVDA 2021.2 please.

cary-rowen:
Important fixes, I hope it can be incorporated into the upcoming version.

XLTechie:
@michaelDCurran Will do.

There are also two DLLs that are failing to register on some test systems, 
and I am working on figuring out why, so this wouldn't have made it to the 
release anyway I think.
Initially this looked like it would be more straight forward.

Will rebase on master tomorrow and proceed from there.

source/COMRegistrationFixes/__init__.py
```diff
@@ -11,71 +11,110 @@
...
+		apply64bitRegistryPatch(OLEACC_REG_FILE_PATH)
 	# IDispatch and other common OLE interfaces
-	registerServer(os.path.join(system32,'oleaut32.dll'))
-	registerServer(os.path.join(system32,'actxprxy.dll'))
+	register32bitServer(os.path.join(system32, "oleaut32.dll"))
+	register32bitServer(os.path.join(system32, "actxprxy.dll"))
 	if is64bit:
-		registerServer(os.path.join(sysWow64,'oleaut32.dll'),wow64=True)
-		registerServer(os.path.join(sysWow64,'actxprxy.dll'),wow64=True)
+		register64bitServer(os.path.join(sysnative, "oleaut32.dll"))
```
michaelDCurran, line 114: As we are calling the 64 bit version of regsvr32, the dll path we give it must already assume to be 64 bit. Thus, system32 would be the appropriate directory from regsvr32's point of view. In fact, I don't think c:\windows\sysnative is a valid path unless running as a wow64 32 bit process. Same goes for the next line down as well.

michaelDCurran (changes requested):

XLTechie:
@michaelDCurran Great catch! I fell prey to the same problem as the original tool had, just in the opposite direction.
Fixed and rebased.
Also noticed that the Windows 7 checks were trying to pull a DLL from `Program Files (x86)`, but on 32-bit systems there is no such thing as I recall. So in the case of non-64-bit, we need to load from the right place. Fixed in latest.
**Needs re-review.**

michaelDCurran (approved):

michaelDCurran:
@XLTechie Are you still requiring more testing on this? I'm happy to merge this when you are happy.

XLTechie:
@lukaszgo1 I believe you have access to Windows 7: can you do any Windows 7 and/or 32-bit testing of this? Particularly interested in whether the [original STR](https://github.com/nvaccess/nvda/issues/9039#issuecomment-857356117) can be fixed by the tool with this try-build, and whether any errors appear in the log about anything (but especially registration failures).

XLTechie:
@michaelDCurran I have a few tests outstanding that I have requested from people with known COM registration problems, or for whom earlier builds of this fix bugged out. I'd like to hear from at least one or two of them before subjecting this to the wider alpha community.
Also, I didn't find anyone who could test this in either Windows 7 or a 32-bit version of 7 or 10. If you think it doesn't need to wait for that, then we can go ahead and hope they try it in alpha.

source/COMRegistrationFixes/__init__.py
```diff
@@ -11,71 +11,114 @@
...
 	@param fileName: the path to the .reg file
 	@type fileName: str
 	"""
 	if not os.path.isfile(fileName):
-		raise FileNotFoundError(f"Cannot apply registry patch, {fileName} not found.")
-	regedit=os.path.join(sysWow64 if wow64 else systemRoot,'regedit.exe')
+		raise FileNotFoundError(f"Cannot apply 64-bit registry patch: {fileName} not found.")
+	# On 64-bit systems, SysWOW64 provides 32-bit apps with a virtual directory to reach 64-bit executables.
+	# For hard coded historical compatibility, regedt32.exe is still the name of the 64-bit app.
+	regedit = os.path.join(SYSNATIVE, "regedt32.exe")
```
lukaszgo1, line 84: Have you considered using `reg.exe` rather than `regedit` here? As discussed in #11494 the latter always returns 0 as its exit code even on failure and since you're refactoring this code anyway it makes sense to solve more known problems IMO.

lukaszgo1:
> 
> 
> @lukaszgo1 I believe you have access to Windows 7: can you do any Windows 7 and/or 32-bit testing of this? Particularly interested in whether the [original STR](https://github.com/nvaccess/nvda/issues/9039#issuecomment-857356117) can be fixed by the tool with this try-build, and whether any errors appear in the log about anything (but especially registration failures).



Since my only remaining Win 7 machine is my main working one I'm reluctant  to  do any testing there I'll set up a Vm and try the build from this  PR during the weekend. As to 32-bit testing no I don't have any 32-bit Windows install lying around. For what its worth setting up a  32-bit VM should not take long - have you considered testing  yourself?

XLTechie:
@michaelDCurran I am declaring this good in as far as it goes. You are aware of the work still in progress; another PR will be required when that is done, but it should be a small update given the foundation here.
Also the commit message is already the same as the above except for spacing and markdown, and it is singly rebased on current master.

source/COMRegistrationFixes/__init__.py
XLTechie: @lukaszgo1 I didn't use reg.exe for the simple fact that I was unaware of it. @jcsteh recommended regedt32.exe, so that's what I went with. I have updated now to use reg.exe, thanks.
So far I left it loading the values from the .reg file still, instead of directly including the keys in source.

Regarding a 32-bit virtual test machine: I would need to switch to a Win 10 Pro installation on my work machine in order to setup a proper virtual environment. There are various reasons that would be highly inconvenient right now, although it is on my list.

XLTechie:
@michaelDCurran this has been rebased, and now includes your registry key deletion.
Additionally, it adds a strong recommendation to restart the machine, to the tool's completion dialog.

I have no information suggesting that the other typelib issue described in #9039 is not fixed by this, so I think this PR can close that issue now.

lukaszgo1:
@XLTechie I gave the last build from this PR a try on A Windows 7 X64 VM and everything works as expected. In particular:
- No errors in the log
- Registry patch is successfully applied to the registry
- - After running the tool Firefox is once again readable even though I've installed and then uninstalled MathPlayer to which @tspivey  linked to in #9039 
For what its worth in my case no restart was necessary not even of NVDA

source/COMRegistrationFixes/__init__.py
```diff
@@ -1,81 +1,126 @@
...
+	@type fileName: str
 	"""
-	Applies the registry patch with the given file name
-	using regedit.
+	if not os.path.isfile(fileName):
+		raise FileNotFoundError(f"Cannot apply 32-bit registry patch: {fileName} not found.")
+	# On 32-bit systems, reg.exe is in System32. On 64-bit systems, SysWOW64 will redirect to 32-bit version.
+	regExe = os.path.join(SYSTEM_ROOT, "System32", "reg.exe")
+	try:
+		subprocess.check_call([regExe, "import", fileName])
```
lukaszgo1, line 72: For each run of `reg.exe` a console window is shown briefly it seems. Have you considered calling these with a `STARTUPINFO` structure and `wShowWindow` set to `SW_HIDE`?

source/gui/__init__.py
```diff
@@ -348,11 +348,17 @@ def onRunCOMRegistrationFixesCommand(self, evt):
 			log.error("Could not execute fixCOMRegistrations command",exc_info=True) 
 		progressDialog.done()
 		del progressDialog
-		# Translators: The message displayed when the COM Registration Fixing tool completes.
-		gui.messageBox(_("COM Registration Fixing tool complete"),
+		# We get an F405 here because "gui may be undefined".
+		gui.messageBox(  # noqa F405
```
lukaszgo1, line 352: Avoid NOQA. Since `messageBox` is defined in this file you can just do `messageBox(blah)`

lukaszgo1 (changes requested):

source/gui/__init__.py
XLTechie: Done. That's what I get for sleepworking.

source/COMRegistrationFixes/__init__.py
XLTechie: I'll be running some tests with that option later today.
Thanks.

XLTechie: @lukaszgo1 I have done this. I also did it for the regsvr32.exe calls.

michaelDCurran (approved):
Thanks for working on this :)

XLTechie:
@michaelDCurran Thanks for approving it, and for the pick-up of the final 
piece at the end. :)

Issue #10588, opened 2019-12-08
Control key gets stuck after using Clipboard History


MarcoZehe:
### Steps to reproduce:

1. In Windows 10 1809 or later, turn on Clipboard History through Clipboard Settings in Windows Settings.
2. Copy some bits of text.
3. Press WindowsKey+V to open Clipboard History, choose an item, and paste it using Enter.
4. Continue working.

### Actual behavior:

Sometimes, not always, the keyboard now behaves as if the control key got stuck. Arrow keys left and right will move by word instead of character, up and down will move by paragraph, and letters act as if the control key was pressed alongside. You have to tap the control key once to un-stick it.

### Expected behavior:

The keyboard should behave normally.

#### NVDA installed/portable/running from source:

Installed.

#### NVDA version:

Seen it in any version since 2018.4 up to 2019.3 alphas.

#### Windows version:

1809, 1903, 1909, and even the insider builds of 2004.

#### Name and version of other software in use when reproducing the issue:

Firefox, NotePad, NotePad++, Word.

#### Other information about your system:

Happens on two quite different systems, a Dell XPS15-9560 (2017) as well as a Microsoft Surface Laptop 2.

#### Does the issue still occur after restarting your PC?

Yes.

#### Have you tried any other versions of NVDA? If so, please report their behaviors.

See above.

[Microsoft feedback for this issue](https://aka.ms/AA6rjli)

MarcoZehe:
CC @josephsl Maybe you have seen this?

Brian1Gaff:
This is not new, I've seen this sort of thing apparently randomly in Windows 
7 as well, but usually when a word processor is running.

Might be some odd effect due to the way nvda can hook into some routines 
perhaps?
 Brian

bglists@blueyonder.co.uk
Sent via blueyonder.
Please address personal E-mail to:-
briang1@blueyonder.co.uk, putting 'Brian Gaff'
in the display name field.
Newsgroup monitored: alt.comp.blind-users
----- Original Message -----

Qchristensen:
Just to confirm, I can replicate this with NVDA, and pressing control once seems to fix it.

I can't replicate with Narrator, or without any screenreader running.  There is no error or anything unusual in the NVDA log:

IO - inputCore.InputManager.executeGesture (11:01:52.356) - winInputHook (20872):
Input: kb(desktop):enter
IO - inputCore.InputManager.executeGesture (11:01:52.414) - winInputHook (20872):
[...]

PR #9208, opened 2019-01-28
Work around frozen keyboard input while the NVDA core is frozen

LeonarddeR:
### Link to issue number:
Improves #6291 

### Summary of the issue:
When large chunks of text are written to a terminal window, NVDA tends to freeze entirely. The freeze tends to lock up the keyboard in such a way that a user needs to sign out of his session and sign in back again, in which case it is likely that work is lost.

### Description of how this pull request fixes the issue:
This pr checks whether the core is attempting recovery way earlier than the current code does. The current code does the following:

1. Input is received
2. winInputHook executes the keyboardHandler.internal_keyDownEvent function
3. internal_keyDownEvent checks for pass through, sticky keys, etc.
4. It executes inputCore.manager.executeGesture
5. inputCore.manager.executeGesture checks whether the core is attempting recovery. If it is, it raises NoInputGestureAction
6. The finally block of internal_keyDownEvent is executed.

The code in this pull request checks the core's recovery state within the winInputHook module itself and doesn't execute keyboardHandler.internal_keyDown/UpEvent at all when the core is frozen. While it does not fix #6291, it does no longer break keyboard input entirely, allowing a user to restart NVDA by hand.

### Testing performed:
Tested by @lukaszgo1 

### Known issues with pull request:
This approach might be a bit aggressive. We could also check the state of the watchdog in keyboardHandler, for example. However, the current code where the keyboard hook executes a massive amount of code even when the core is frozen, is certainly not something we should keep as it is.

### Change log entry:
* Bug fixes
    + NVDA is now less likely to freeze in such a way that the only way to escape is signing out from your current windows session. (#6291)

PR #12610, opened 2021-07-03
Fix tracking of keyboard modifiers when watchdog is attempting recovery

### LINKED ISSUE #12609: State of keyboard modifiers is not tracked correctly in some cases
### Description
A number of users have previously reported problems with NVDA incorrectly tracking the state of modifiers, e.g. NVDA thinks insert key is down, while it is really up. A couple of examples are #11744 and #11546. In many cases it's hard to define clear steps to reproduce.
I was experiencing this issue as well and found somewhat reliable way of reproducing, although it involves editing a file on a network-mounted share, so it is pretty difficult to configure.
I also think I fixed the problem, so creating this issue to be fixed with my PR.
Please note, that this is not an issue with keyboard.

### Steps to reproduce:
Prerequisites: a network folder with write access, available either via windows network share (SMB) or SSHFS.
1. Mount a network folder. Either via SMB or SSHFS-win.
2.Create a new file on that network share and open it in Notepad++. 
3.Make any change to the file.
4. Quickly press Control+S and then Insert+T.

### Actual behavior:
At this point a number of bugs might happen. Here are a few different scenarios, of what can happen:
1. Pressing NVDA+T doesn't announce window title. While still holding down insert key I might press T again, but still window title is not announced. Then when I alt-tab away and back, I observe multiple letters "T" were actually typed into the document. Which suggests that NVDA didn't catch me pressing insert key.
2. Alternatively, if insert+T correctly announces the title, I may release Insert key. Then I may press some unrelated keystroke, such as Control+Z, and to my surprise this opens NVDA Python console. This suggest that NVDA didn't catch me releasing Insert key and it thinks that insert key is down and Instead of Control+Z NVDA thought I pressed Insert+Control+Z.
3. Alternatively, when I press arrow keys, instead of their usual function, I observe that NVDA does navigate by word when I press left or right, or doesn't move anywhere when I press Up/Down. This is consistent with the hypothesis of NVDA believing that control key is down.
Typically Either of these scenarios might reproduce with at least 10% probability, sometimes as high as 70% (percentage figures are my rough estimation).

### Expected behavior:
NVDA should correctly track the state of control key, insert and other keyboard modifiers. Even when NVDA deals with a window that is not responding, e.g. frozen on network write operation.
### Alternative way to reproduce
When I press Control+S followed by Alt-Tab, I also notice that sometimes Alt key gets stuck.

Tried all three.

#### NVDA version:
2020.4
Also tried 2021.1beta1 - could only reproduce options 1 and 2, but not 3 for some reason.
#### Windows version:
Windows 10 
10.0.17134.2145
Version 1803 (OS Build 17134.2145)

Notepad++ v7.7.1

Yes
Yes
Yes
Yes

mltony:
### Link to issue number:
Fixes: #12609

### Summary of the issue:
NVDA doesn't correctly track state of keyboard modifiers (such as Control, or Insert) in some cases, e.g. when watchdog is recovering.

In `winInputHook.py` , in `def keyboardHook()`, currently there is a condition:
```
if watchdog.isAttemptingRecovery or code!=HC_ACTION:
return windll.user32.CallNextHookEx(0,code,wParam,lParam)
```
So when `watchdog.isAttemptingRecovery` is `True`, then this function returns early and doesn't process any keyboard events, including events related to keyboard modifiers. As a result, when a user happens to press modifier while watchdog is recovering, NVDA won't process this, thus getting out of sync with current keyboard state.
In this PR I propose to check whether current keyboard event is related to keyboard modifiers, and         when it is related, then process it even when watchdog is recovering.
Please note, I reduplicate the list of keyboard modifiers in `winInputHook.py`, since importing the same list from `keyboardHandler.py` makes NVDA to crash on startup, probably due to circular import.

### Testing strategy:
Performed thorough manual testing on locally built version. This PR was applied on top of 2021.1beta1.
I have been running my custom build for about a month and couldn't reproduce the problem even once. For comparison, without this PR< the problem would reproduce several times a day during my normal work process.
I also had `tones.beep()` inserted in multiple versions to verify my hypothesis, and it proved true that when Notepad++ window is stuck on network write, then indeed watchdog enters recovery mode very often, and also while this happens if user happens to press or release any modifiers, then I also verified, that this event is indeed being ignored in current master version.
None observed.
### Change log entries:
Bug fixes:
Fixing tracking state of keyboard modifiers (such as Control, or Insert) when watchdog is recovering.

### Code Review Checklist:

- [x] Unit tests (something unrelated fails regarding to Bulgarian braille table - but master fails too)
- [ ] System (end to end) tests - doesn't apply
- [x] Manual testing.
- [ ] User Documentation - doesn't apply
- [x] Change log entry.
- [ ] Context sensitive help for GUI changes - doesn't apply
- [ ] UX of all users considered - doesn't apply.

### DIFF
--- a/source/keyboardHandler.py
+++ b/source/keyboardHandler.py
@@ -195,7 +195,7 @@ def internal_keyDownEvent(vkCode,scanCode,extended,injected):
 			currentModifiers.discard(stickyNVDAModifier)
 			stickyNVDAModifier = None
 
-		if _watchdogObserver._get_isAttemptingRecovery():
+		if _watchdogObserver.isAttemptingRecovery:
 			# When attempting recovery only process modifiers, but do not execute gesture.
 			return True
 
@@ -212,7 +212,7 @@ def internal_keyDownEvent(vkCode,scanCode,extended,injected):
 	except:
 		log.error("internal_keyDownEvent", exc_info=True)
 	finally:
-		if _watchdogObserver._get_isAttemptingRecovery():
+		if _watchdogObserver.isAttemptingRecovery:
 			return True
 		# #6017: handle typed characters in Win10 RS2 and above where we can't detect typed characters in-process 
 		# This code must be in the 'finally' block as code above returns in several places yet we still want to execute this particular code.
@@ -282,7 +282,7 @@ def internal_keyUpEvent(vkCode,scanCode,extended,injected):
 #Register internal key press event with  operating system
 
 
-def initialize(watchdogObserver=None):
+def initialize(watchdogObserver):
 	"""Initialises keyboard support."""
 	global _watchdogObserver
 	_watchdogObserver = watchdogObserver
--- a/source/keyboardHandler.py
+++ b/source/keyboardHandler.py
@@ -11,6 +11,8 @@ import ctypes
 import sys
 import time
 import re
+import typing
+
 import wx
 import winVersion
 import winUser
@@ -29,7 +31,11 @@ import tones
 import core
 from contextlib import contextmanager
 import threading
-_watchdogObserver = None
+
+if typing.TYPE_CHECKING:
+	from watchdog import WatchdogObserver
+
+_watchdogObserver: typing.Optional["WatchdogObserver"] = None
 ignoreInjected=False
 
 # Fake vk codes.
@@ -282,7 +288,7 @@ def internal_keyUpEvent(vkCode,scanCode,extended,injected):
 #Register internal key press event with  operating system
 
 
-def initialize(watchdogObserver):
+def initialize(watchdogObserver: "WatchdogObserver"):
 	"""Initialises keyboard support."""
 	global _watchdogObserver
 	_watchdogObserver = watchdogObserver
--- a/source/keyboardHandler.py
+++ b/source/keyboardHandler.py
@@ -29,7 +29,7 @@ import tones
 import core
 from contextlib import contextmanager
 import threading
-
+_watchdogObserver = None
 ignoreInjected=False
 
 # Fake vk codes.
@@ -195,6 +195,10 @@ def internal_keyDownEvent(vkCode,scanCode,extended,injected):
 			currentModifiers.discard(stickyNVDAModifier)
 			stickyNVDAModifier = None
 
+		if _watchdogObserver._get_isAttemptingRecovery():
+			# When attempting recovery only process modifiers, but do not execute gesture.
+			return True
+
 		try:
 			inputCore.manager.executeGesture(gesture)
 			gestureExecuted=True
@@ -208,6 +212,8 @@ def internal_keyDownEvent(vkCode,scanCode,extended,injected):
 	except:
 		log.error("internal_keyDownEvent", exc_info=True)
 	finally:
+		if _watchdogObserver._get_isAttemptingRecovery():
+			return True
 		# #6017: handle typed characters in Win10 RS2 and above where we can't detect typed characters in-process 
 		# This code must be in the 'finally' block as code above returns in several places yet we still want to execute this particular code.
 		focus=api.getFocusObject()
@@ -275,8 +281,11 @@ def internal_keyUpEvent(vkCode,scanCode,extended,injected):
 
 #Register internal key press event with  operating system
 
-def initialize():
+
+def initialize(watchdogObserver=None):
 	"""Initialises keyboard support."""
+	global _watchdogObserver
+	_watchdogObserver = watchdogObserver
 	winInputHook.initialize()
 	winInputHook.setCallbacks(keyDown=internal_keyDownEvent,keyUp=internal_keyUpEvent)
 
--- a/source/watchdog.py
+++ b/source/watchdog.py
@@ -39,7 +39,7 @@ safeWindowClassSet=set([
 ])
 
 isRunning=False
-isAttemptingRecovery = False
+isAttemptingRecovery: bool = False
 _coreIsAsleep = False
 
 _coreDeadTimer = windll.kernel32.CreateWaitableTimerW(None, True, None)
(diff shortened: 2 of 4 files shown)


mltony:
CC: @michaelDCurran  @feerrenrut

michaelDCurran (commented):
At first glance, I'm thinking it might be better to not check for watchdog recovery at all in winInputHook. Rather in keyboardHandler.internal_keyDownEvent: Just before the try block with executeGesture,  if watchdog is attempting recovery, return True. Similarly, at the top of the finally block, if watchdog is attempting recovery, return true.
internal_keyUpEvent does not need to check at all I believe.
Essentially, we are trying to avoid queuing anything to the main thread, and or interrigating any NVDAObjects.
I agree the existing watchdog checks are too broad.

mltony:
@michaelDCurran, fixed this according to your suggestions.

source/keyboardHandler.py
```diff
@@ -217,6 +222,7 @@ def internal_keyDownEvent(vkCode,scanCode,extended,injected):
 			and not gestureExecuted 
 			# and not if this gesture is a modifier key
 			and not isNVDAModifierKey(vkCode,extended) and not vkCode in KeyboardInputGesture.NORMAL_MODIFIER_KEYS
+			and not watchdog.isAttemptingRecovery
```
michaelDCurran, line 225: I would prefer to see this as its own check returning true above this if condition, as I fear it gets lost otherwise.

michaelDCurran (dismissed):

mltony:
@michaelDCurran, fixed!

source/keyboardHandler.py
```diff
@@ -29,6 +29,7 @@
 import core
 from contextlib import contextmanager
 import threading
+import watchdog
```
feerrenrut, line 32: I'm not entirely happy to introduce this dependency. Before you make any changes, I'd like to hear what @michaelDCurran has to say about the following alternative:
- Introduce a new class to the watchdog module: `watchdogObserver`
- The `watchdogObserver` can be the interface for observing the state of the watchDog. It could have a getter for the `isAttemptingRecovery` state.
- An instance of the watchdogObserver is passed into the `keyboardHandler.Initialize` function where it is stored as a protected variable on the module.

michaelDCurran:
Um... it really doesn't bother me either way. I think just importing watchdog is much more easier to understand... and directly addresses the issue.  but @feerrenrut  if you want to work with @mltony  to implement your idea then that is okay with me.

feerrenrut:
Whether it is easier to understand is a matter of perspective I think. If you aren't familiar with `watchdog` you'll have to go and understand the origin of that either way to understand this PR. My main reasoning for wanting this separation is to:
- prevent an 'import order dependency'. Currently the import for watchdog is later in core, however it is implicitly imported by `inputCore` before `keyboardHandler`. This makes the import order hard to reason about. This won't matter when/if we get to stage of having no import time behavior. But we aren't there yet.
- encode the intent, `keyboardHandler` is an observer of the `watchdog`, not a controller. It doesn't cause state changes.

I'm happy to work through this with @mltony

mltony:
@feerrenrut, I updated this PR according to your suggestion - I created WatchdogObserver class.

mltony:
I did some more testing, it seems this might also fix #10588 - at least pasting from clipboard history seems to work much better with this PR on my computer. Tagging FYI.

CyrilleB79:
@mltony if you do not have any additional modification to do in this PR, please pass it in "Ready" state rather than letting it in "Draft". This way, it will not be forgotten by NVAccess.

source/watchdog.py
```diff
@@ -370,3 +370,9 @@ def cancellableSendMessage(hwnd, msg, wParam, lParam, flags=0, timeout=60000):
 	result = ctypes.wintypes.DWORD()
 	NVDAHelper.localLib.cancellableSendMessageTimeout(hwnd, msg, wParam, lParam, flags, timeout, ctypes.byref(result))
 	return result.value
+
+
+class WatchdogObserver:
+	def _get_isAttemptingRecovery(self):
```
leonardder, line 376: Could you make this a property?

```suggestion
	@property
	def isAttemptingRecovery(self):
```

source/keyboardHandler.py
```diff
@@ -275,8 +281,11 @@ def internal_keyUpEvent(vkCode,scanCode,extended,injected):
 
 #Register internal key press event with  operating system
 
-def initialize():
+
+def initialize(watchdogObserver=None):
```
leonardder, line 285: This being an optional parameter doesn't follow the intend of the watchdog observer being absolutely required within the keyboard handler. I don't think backwards compatibility has to be taken into account here. However, if it is decided otherwise, all checks for the property on the observer must first check whether there's an observer assigned to the variable.

```suggestion
def initialize(watchdogObserver):
```

LeonarddeR (changes requested):
For reference, the current code that checks the watchdog was introduced by me in #9208. I agree that it is to aggressive, though I hope #6291 won't be made worse this way again. May be @lukaszgo1  could have a look as well, as he performed some testing as part of #9208?

mltony:
@leonardder, addressed your comments.
Also, I marked this PR as ready for review - please let me know if I shouldn't have done that.

seanbudd (approved):
LGTM - @feerrenrut suggested these changes so I'm re-requesting review.

feerrenrut (changes requested):
Yep this looks good. I'd just like to add some type information for WatchdogObserver. I'll push a commit with this.

feerrenrut:
@mltony I have pushed a commit to add some more type information to the changed areas. I tested only by ensuring NVDA runs from source.

feerrenrut (approved):

LeonarddeR (approved):

feerrenrut:
I think the description for this PR is inaccurate. This could be misleading to someone re-visiting in the future. Could you please update it?

PR #12617, opened 2021-07-05
Fix Enum Cyclic Ref

feerrenrut:
### Link to issue number:
Blocking PR #12500

### Summary of the issue:
When creating DescriptionFor enum values, a "ValueError" can be raised.
There is a bug in the Python standard library for enum, causing a cyclic reference.
The exception is referenced by the frame.
The frame is referenced by the traceback.
The traceback is referenced by the exception.
This is detected by NVDA as an unreachable object, logging an error.

This patches the `enum.Enum.__new__` method to no longer reference the exception from the frame.

### Testing strategy:
Merged with #12500, ran with sample, ensured no unreachable object log error.
Unit testing this would be difficult, we'd have to install the NVDA `garbageHandler`, this would cause side affects for all tests.
These errors are reported to the log of system tests already.
In future we plan for system tests to fail if there are ANY errors in the log, however right now there are several
expected errors in the log on appveyor.
These first need to be resolved.
Additionally, we are in the process of resolving a tricky intermittent failure for system tests.
Once the intermittent failure is resolved we can look to make the tests more strict.

None

### Change log entries:
None

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


feerrenrut:
A diff of the python standard library implementation vs the replacement implementation:
[stdLibVsNew.txt](https://github.com/nvaccess/nvda/files/6763530/stdLibVsNew.txt)

source/monkeyPatches/__init__.py
```diff
@@ -0,0 +1,14 @@
+# A part of NonVisual Desktop Access (NVDA)
```
lukaszgo1, line 1: Since you're grouping all monkey patches in one package can our monkey patches for `wx` be moved to it  from `gui.initialize`?

source/monkeyPatches/enumPatches.py
```diff
@@ -0,0 +1,60 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2021 NV Access Limited
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+
+
+def replace__new__():
+	import enum
+	# prevent cyclic references on ValueError during construction
```
lukaszgo1, line 9: Have you tested if this occurs in more recent versions of Python? If it isn't fixed it might be worth reporting this issue against them.

devDocs/unreachableObjects.md
```diff
@@ -0,0 +1,66 @@
+# Garbage collection errors
+NVDA's `garbageHandler.py` monitors Python's cyclic garbage collector and reports
+on objects that are unreachable.
+Cyclic references are typically a symptom of bad design, and can cause major problems for certain objects.
+For instance, cyclic references can cause deadlocks for UIA objects.
```
michaelDCurran, line 5:
```suggestion
For instance, cyclic references involving COM objects may cause a deadlock if the garbage collector happens to break the cycle and release the COM object in the wrong thread.
```

devDocs/unreachableObjects.md
```diff
@@ -0,0 +1,66 @@
...
+Found at least 1 unreachable objects in run
+```
+
+## How to debug a cyclic reference?
+
+Once you can reliably reproduce the log error, you can tell the garbage collector to save all unreachable objects.
+After an unreachable object is detected the references to the unreachable object can be inspected via the python console.
+Inspecting this should give you a fair idea of where the issue is occurring.
+
+1. Open the NVDA console `NVDA+control+z`
```
michaelDCurran, line 25:
```suggestion
1. Open the NVDA Python console `NVDA+control+z`
```

michaelDCurran (changes requested):
Would it not be better for maintainability that replacement__new__ is simply:
```
if value not in cls.__members__.values():
    raise ValueError(f"{value} is not a supported value of {cls}")
return old__new__(value)
```
Though obviously what you wrote will be useful for a pr against cPython. 
But other than the other corrections I and @lukaszgo1  have mentioned, I'd be happy to approve with the code you have here.

feerrenrut:
> Would it not be better for maintainability that replacement__new__ is simply

I'm trying to maintain compatibility with the built-in enum behavior. There are a number of other values that it accepts for various reasons. Using the suggestion would mean that only values for the enum are accepted.

source/monkeyPatches/__init__.py
feerrenrut: I'm not sure of the implications of importing wx so early in the process. Perhaps I'll do that in a separate function and have it called from `gui.initialize`

source/monkeyPatches/enumPatches.py
feerrenrut: The latest implementation in cpython is even more complicated. From inspection it does appear to have the same flaw.  I will be creating an issue / PR to fix it.

michaelDCurran (approved):

PR #11437, opened 2020-07-30
Add experimental option to only process IAccessible events for focus ancestry, etc.

jcsteh:
### Link to issue number:
Fixes #11077. Fixes #10484.

### Summary of the issue:
Currently, NVDA listens for all win events. It drops events for background windows and some other cases, but there can be a lot of events for objects in the foreground window that aren't interesting to the user. As well as being a performance concern, this can cause events for the focus to be dropped from the limiter due to a flood of events from other objects, resulting in problems such as #10484.

### Description of how this pull request fixes the issue:
In this PR, most win events are dropped before they're even added to the limiter unless they're for the focus, a focus ancestor or the desktop object. Certain events have to be allowed regardless of object; e.g. valueChange for progress bars. Show events have to be allowed for certain window classes; e.g. tooltips.

### Testing performed:
Verified that #10484 is fixed with this option enabled.

I wasn't able to find a case where this noticeably improved performance. It previously helped with Slack, but Slack now fires a lot less events (as of a few weeks ago), so this is no longer significant here. I do suspect there might be cases I haven't thought of yet, though.

### Known issues with pull request:
For now, this is an option in advanced settings disabled by default, since there is a risk of unforeseen regressions. It also breaks eventHandler.requestEvent, though this is also the case for selective UIA event registration. Nevertheless, it could be useful to have advanced users test this in various situations to see if it resolves problems.

If we decide to pursue this beyond an experiment, we'll need to re-think eventHandler.requestEvent.

It duplicates the window classes for which we process show events from eventHandler.shouldAcceptEvent. Again, if we pursue this beyond an experiment, we'll need to re-think eventHandler.shouldAcceptEvent.

PR #12622, opened 2021-07-08
Add experimental option to only process IAccessible events for focus ancestry, etc.

### LINKED ISSUE #10484: Div with role checkbox not reading checked/unchecked on performing action
### Steps to reproduce:
Open following codepen : https://codepen.io/sandeepnmenon/pen/NWKOgYo
Tab till the check button in the header of the table.
Check/Uncheck using space button

### Actual behavior:
NVDA not reading checked/unchecked

### Expected behavior:
NVDA should read checked/unchecked

React: 16.8.6
Office fabric: 7.60.0

### LINKED ISSUE #11077: Restrict most event processing to objects of interest to the user
### problem
Some web apps, such as Twitter and Slack, fire a lot of events, most of which aren't things we need to report to the user. For example, when you focus the message list in Slack, it changes the name for every item in the list and adds the Message actions grouping to every item in the list. This results in noticeable performance problems. On my system with Firefox Nightly, focusing the Slack message list with 32 items in the list takes > 700 ms to respond. This is quite a bit faster in Chrome, so clearly there's some work to do in Firefox (and I'm looking into that as well). However, processing less of these events makes things significantly faster (~500 ms or less with my prototype).

We've had similar problems with UIA. For example, 1Password fires many property change events when you switch items in the list (#10508).

### High Level Solution
NVDA already drops events for background windows, etc. However, there can be a lot of events even in the foreground window that aren't important to the user. For example, a nameChange on an object that isn't focused and isn't in the focus ancestry isn't reported in any way. Previously, these events were processed, NVDAObjects were instantiated for them and the events were queued, but they simply did nothing when they were finally handled by the NVDAObject.

To address this, I propose that NVDA only process most IAccessible events for the focus, focus ancestry, navigator object and desktop object. For UIA, we could do the same, but it probably makes sense to only listen for events on specific objects; see below.

### Issues
I've prototyped two versions of this for IAccessible.

1. Do all of this in IAccessibleHandler. Problems:
    - This doesn't integrate well with eventHandler.requestEvent. If a plugin requests an event, it will get dropped before shouldAcceptEvent is called. And we can't do this stuff after shouldAcceptEvent because it will return True for cases we want to block.
    - We'd need a completely separate implementation for UIA; we can't use any common code.
2. Integrate it into shouldAcceptEvent. Pass raw API event params to shouldAcceptEvent and let it ask the relevant NVDAObjects whether the event is for them.
    - The problem is that for UIA, it makes more sense to just only listen for events on relevant objects, rather than listening for all events and dropping the ones we don't care about. In contrast, we can't specify specific objects for IAccessible.

It's really hard to come up with a framework that works well for both. On the other hand, as can be seen from the complexity of shouldAcceptEvent, duplicating all of the logic between IAccessible and UIA wouldn't be sustainable.

I think addressing this for both APIs is probably going to involve a hybrid approach. We'll need some common logic in eventHandler, but this will need to be split up so we can call different pieces in different places. For example, we need to be able to separately ask "did a plugin request this?" and "is this event for the foreground window?", rather than this all being handled in one call. We'll also need to be able to ask "what objects are of interest to the user?", since IAccessible and UIA will handle that differently (filtering vs selective registration).

@michaelDCurran, @feerrenrut, thoughts appreciated. I can push both of my prototype patches somewhere if you're interested (draft pull requests perhaps?), but I didn't want to spam with in-progress work without asking/discussing first.

LeonarddeR:
Original pr by @jcsteh 

### Link to issue number:
Replaces #11437 
Fixes #11077.

### Summary of the issue:
Currently, NVDA listens for all win events. It drops events for background windows and some other cases, but there can be a lot of events for objects in the foreground window that aren't interesting to the user. As well as being a performance concern, this can cause events for the focus to be dropped from the limiter due to a flood of events from other objects, resulting in problems such as #10484.

In this PR, most win events are dropped before they're even added to the limiter unless they're for the focus, a focus ancestor or the desktop object. Certain events have to be allowed regardless of object; e.g. valueChange for progress bars. Show events have to be allowed for certain window classes; e.g. tooltips.

### Testing strategy:
- [ ] Get a proper test case

* For now, this is an option in advanced settings disabled by default, since there is a risk of unforeseen regressions. It also breaks eventHandler.requestEvent, though this is also the case for selective UIA event registration. Nevertheless, it could be useful to have advanced users test this in various situations to see if it resolves problems.
* If we decide to pursue this beyond an experiment, we'll need to re-think eventHandler.requestEvent.
* It duplicates the window classes for which we process show events from eventHandler.shouldAcceptEvent. Again, if we pursue this beyond an experiment, we'll need to re-think eventHandler.shouldAcceptEvent.

### Change log entries:
New features

### Code Review Checklist:
- [x] Unit tests.
- [ ] Manual testing.
- [x] User Documentation.
- [ ] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


leonardder:
I had to close #10484 because the codepen was lost and another example did not reproduce. Therefore I think we need another test case for the impact of this.

leonardder:
@cary-rowen and @lijianwei2019, given you seemed interested in the progress of this pr, could you elaborate on how it benefits you?

codeofdusk:
Given that NVDA is slow to process events (#11002), anything that can be done to filter them seems like it would have at least some performance benefit.

leonardder:
Note that pr doesn't touch UIA at all.
It's not that I don't see that there's likely a benefit for this pr, I only want to have a firm test case to prove a point.

cary-rowen:
HI leonardder,

NVDA's fluency headquarters is satisfactory, which is manifested in some programs with more object levels.
I very much hope that NVDA can improve in these areas.

lijianwei2019:
HI leonardder,

In programs with more objects. Users are not ideal for a smooth experience.

codeofdusk:
@leonardder What is the current status of this PR? Anything blocking this being merged (as a feature disabled by default)?

leonardder:
As long as we don't have a good test case to prove that this pr actually improves things, I'm reluctant to bring this further. It is much more important to invest in UIA performance.

lijianwei2019:
@jcsteh Do you still have good test cases?

jcsteh:
The scenario I described in #11077 involves focusing the message list in Slack; i.e. pressing up arrow from the compose text box to move into the message list. Firefox has become significantly faster since then, but on my system, it still takes around 140 ms from the time i press up arrow to the time NVDA speaks something. I haven't tested this PR yet, but it'd be interesting to know whether that time is still improved.

LeonarddeR:
I'm goin to close this as abandoned. Whoever wants to pick this up, feel free to take it.

PR #12635, opened 2021-07-11
Replace the appveyor deploy ssh key with one encrypted more strongly 

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
Appveyor contains an ssh private key that allows it to log into an NV Access server to deploy snapshots / releases. The current ssh key in the repository, although encrypted with an appveyor secret variable, was not as strong as it could have been, and during decryption, in the appveyor build, openssl would warn about this.
 
This pr replaces the ssh key with a newly generated one, encrypted using a new appveyor secret variable, and the encryption of the key is also significantly stronger
as it uses SHA256 instead of md5, a salt is now used, and a newer secret key derivation algorithm is now used, as recommended by openssl.
This ssh key is now encrypted just as strongly as the authenticode certificate also in this repository. 

### Testing strategy:
Merge to master and ensure the build succeeds.

None known.

### Change log entries:
New features
Changes
Bug fixes
For Developers

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


appveyor/scripts/decryptFilesForSigning.ps1
```diff
@@ -4,7 +4,7 @@ if(!$env:APPVEYOR_PULL_REQUEST_NUMBER) {
 		$errorCode=$LastExitCode
 		Add-AppveyorMessage "Unable to decrypt authenticode certificate"
 	}
-	openssl enc -md md5 -aes-256-cbc -d -pass pass:$env:secure_ssh_pass -in appveyor\ssh_id_rsa.enc -out appveyor\ssh_id_rsa
+	openssl enc -d -md sha256 -aes-256-cbc -pbkdf2 -salt -pass pass:$env:secure_ssh_pass -in appveyor\ssh_id_rsa.enc -out appveyor\ssh_id_rsa
```
michaelDCurran, line 7: We may eventually want to put try catch blocks around this and the decrypting of the authenticode certificate above as we no longer expect any output from standard error. However, for this pr at least, it is best to be merged as is so we can firstly ensure the  whole build chain actually runs without errors.

seanbudd (approved):

PR #9197, opened 2019-01-24
When the background of a display model chunk is transparent, report that the color is unknown when fetching formatting info

LeonarddeR:
### Link to issue number:
Workaround for #6467

### Summary of the issue:
From https://github.com/nvaccess/nvda/issues/6467#issue-183212018

> When text is written with the GDI background mode set to transparent, the display model doesn't necessary report the correct background color for that text. In this case, the GDI hook should look at the actual background color before the text is drawn, rather than using the result of GetBkColor.

### Description of how this pull request fixes the issue:
There have been two attempts to fix this, both causing severe performance regressions.

This implementation implements a better safe than sorry approach. When the background of a device context is in transparent mode, the background color of the device context does most likely not apply. In this case, it is better to say that the background color is unknown when someone wants it reported, rather than reporting the wrong color.

The implementation works as follows:

1. A colorref is a 32 bit value of which only the last 24 bits are used. The first 8 bits are only used for the case an invalid color is returned, in which the value is 0xffffffff.
2. When the background of a dc is transparent, set the 8th bit of the resulting colorref to 1 instead of 0. This means that the fetched background color is still programmatically available. For example, transparent white would be 0x01ffffff, transparent black would be 0x01000000.
3. colors.RGB.fromCOLORREF now returns a RGB object with values of -1, -1, -1 if the colorref is out of the RGB range.
4. colors.RGB(-1,-1,-1).name results into a name of unknown

### Testing performed:
Tested with a corporate application which has a transparent background as part of a grid.

### Known issues with pull request:
We might throw away cases where the dc's background color is equal to the real background color. Still, I prefer not to report wrong information whenever possible.

### Change log entry:
Bugfixes, yet to be written

PR #7440, opened 2017-07-25
displayModel: second try at recording the actual background color for text when a transparant background color is set.

michaelDCurran:
This PR is based on PR #6844 which was reverted due to performance reasons.

It improves upon #6844 by only calling GetPixel if the text being written  is using a bk mode of transparent, greatly reducing the amount of calls.

This extra commit from @mwcampbell was provided as a patch attached to #6467. 
Fixes #6467

PR #6844, opened 2017-02-04
Display model: Try to provide the correct background color for transparent text drawing (issue #6467)

mwcampbell:
As described in issue #6467, the NVDA display model doesn't provide the correct background color for text that is drawn in transparent mode. A complete solution for this problem is difficult, but a good enough fix for some instances of the problem is to get the previous color at the origin (X, Y) coordinates before calling the original `TextOut` or `ExtTextOut` function. Note that this implementation only applies to the `TextOut` and `ExtTextOut` functions, since other cases such as `PolyTextOut` are more difficult.

THe motivating use case, and the only one I've tested, is the Dragon NaturallySpeaking vocabulary editor. In that editor's list view, without this fix, the highlighted line is reported as being white on white. With this fix, it's correctly reported as white on aqua blue.

Issue #6467, opened 2016-10-15
Display model reports the wrong background color for text drawn in transparent mode


mwcampbell:
When text is written with the GDI background mode set to transparent, the display model doesn't necessary report the correct background color for that text. In this case, the GDI hook should look at the actual background color before the text is drawn, rather than using the result of `GetBkColor`.

michaelDCurran:
Our displayModel currently only records chunks of text. If a fill or other non-text drawing action occurs, any existing text in its path is simply removed from the model. Thus the only background color info we have currently is from written text.  I'm assuming you mean we should actually track other non-text drawing actions, including its background and foreground color, so that if a drawing action such as text has a transparent color we can inherit the background color from the existing chunk? 

This would be a large project. What is the impact of this (taking into account GDI becoming less used anyway)?

mwcampbell:
I don't think you need anything that complex. Before you call the original GDI `TextOut` or `ExtTextOut` function, you can just use GetPixel to get the color at the specified X/Y coordinates. That would be good enough for my purpose, at least.

This impacts my ability to make Dragon NaturallySpeaking fully accessible as part of the [DictationBridge](http://dictationbridge.com/) project. Specifically, Dragon has some custom list views where I can't yet track the highlighted item because NVDA thinks the background color of all the text in the list view is white.

michaelDCurran:
If the background contained some kind of pattern or image then it may 
not be accurate. However, I agree that is better than just transparent.


I will look into it at some point, though if you need it quicker I'll 
certainly review  a submitted pull request.

jcsteh:
This was reverted in 47337ef due to severe performance regressions; see #7251.

josephsl:
Hi,

@MichaelDCurran, attached is the path that Matt should have sent as an official PR, but looks like it didn't make it through. I'm attaching in in case you'd like to review it and let Jamie or someone commit this. Thanks.
[displayModel-attempted-fix.txt](https://github.com/nvaccess/nvda/files/1168656/displayModel-attempted-fix.txt)

Adriani90:
There is a work around introduced in #12658. @mwcampbell, @LeonarddeR is there anything else expected here? Or is the workaround good enough to close this issue?

lukaszgo1:
[...]

PR #12658, opened 2021-07-20
Expose transparent background colors for displayModel

feerrenrut:
### Link to issue number:
Resolves concerns about contributed PR: https://github.com/nvaccess/nvda/pull/9197
Workaround for #6467

### Summary of the issue:

From #6467 (comment)

>When text is written with the GDI background mode set to transparent, the display model doesn't necessary report the correct background color for that text. In this case, the GDI hook should look at the actual background color before the text is drawn, rather than using the result of GetBkColor.

There have been two prior attempts to fix this, both causing severe performance regressions.
- https://github.com/nvaccess/nvda/pull/6844
- https://github.com/nvaccess/nvda/pull/7440

This implementation merely exposes information about the potential transparency of the text background.
[Testing with various GDI dependent GUI APIs](https://github.com/nvaccess/testDisplayModel) shows that the `GetBkColor` result can be:
- Opaque and matches visual presentation(rawGDI)
- Transparent and matches visual presentation: (mfcTest, GDItest)
- Transparent and does not match visual presentation: (annecdotal)

The exposure of the transparency (via the `alpha` value on the `colors.RGB` class) will allow appModules 
or add-ons to customize NVDA's presentation as necessary.

For users, there is no change in behavior by default. An option is introcuded to the advanced settings panel to allow reporting of transparency in colors. The intention here is that developers will use this to identify where
color reporting could be improved.

The prior approach used the colorref high-level byte of a `colorref` struct. However, MSDN requires this byte be zero.
This prior approach did not pass the colorref into any Windows API's, so it was safe. But to prevent this from being accidentally introduced in the future, a compatible alternative type is introduced and used instead.

In order to stay compatible with python code interpretting this color value, a single bit is set if the color is considered transparent.

### Testing strategy:
- Built applications from https://github.com/nvaccess/testDisplayModel
- Tested with/without transparency reporting enabled (advanced panel)
   - Ensure color reporting is enabled in document formatting panel of NVDA settings
   - Use `NVDA+Numpad 7` to switch to screen review
   - Use `Numpad 7`, `Numpad 8`, `Numpad 9` to read the dialog

None

### Change log entries:
For Developers
```
- Transparency of text background color, sourced from GDI applications (via the display model) are now exposed for add-ons or appModules.
```

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


seanbudd (commented):
These changes look good to me, this is a great change which might be able to be leveraged in #12372

Would it be too difficult to add system tests here considering the need for an external app (that was used for manual testing)?

Also can you explain why it is okay to remove all the `free` statements here? I'm guessing it is to do with how the construction was changed but looking to confirm.

devDocs/displayModel.md
```diff
@@ -0,0 +1,17 @@
...
+
+For information on working in this area see the `NVDAHelper/readme.md` file.
+
+## Testing GDI applications
+To test:
+- Use `NVDA+control+numPad 7` to enter screen review mode.
+- Use the number pad to read lines (7, 8, 9), read words, (4, 5, 6), or characters (1, 2, 3).
+
+Note:
+- For color reporting enable the option (via Document formatting panel) _"report color"_.
```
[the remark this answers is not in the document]
feerrenrut: This could be clearer, but note that for color reporting an option in "document formatting" needs to be enabled, for transparency reporting a second option in advanced settings needs to be enabled.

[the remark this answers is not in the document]
seanbudd: Ah my mistake

feerrenrut:
> Also can you explain why it is okay to remove all the free statements here? I'm guessing it is to do with how the construction was changed but looking to confirm.

In C `calloc` allocates memory (much like `new` or `new[]` in C++), it must be matched with a call to `free` (in C++ `delete` or `delete[]`) to prevent a memory leak. In complicated code (and even in simple code) free / delete can be missed.

To replace these I tracked the variable being freed back to where the call to `calloc` was made, and checked if it was really necessary to allocate memory. In all cases I could use `std::string` or `std::vector`, these standard library containers manage their own memory. As a benefit, this change can also reduce the number of copies made, since the `std::string` or `std::vector` can be filled directly.

Since C++11 the memory of `std::string` has been guaranteed to be contiguous, this was already the case with `std::vector`. As of C++17, the 'data' method on a non-const `std::string` returns a non-const pointer to its internal buffer. The caveat for this is that there are a variety of ways to invalidate the pointer, anything that might cause the container to reallocate and move its contents to a new memory location. Once the data is filled however, the pointer can be discarded and member functions used again.

> Would it be too difficult to add system tests here considering the need for an external app (that was used for manual testing)?

We'd have to make a build of the demo app that could be fetched by appveyor. At the moment the demo app doesn't let us test very much (only colors and text reading). So the value is limited. I think if we covered a bit more functionality this would be worth it.

There are three demo apps, which I used to test variations in how the GUI APIs use GDI. It isn't immediately obvious how we would structure the tests to interact with these variations.

seanbudd (approved):

devDocs/displayModel.md
```diff
@@ -0,0 +1,18 @@
...
+
+For information on working in this area see the `NVDAHelper/readme.md` file.
+
+## Testing GDI applications
+To test:
+- Use `NVDA+numPad 7` to enter screen review mode.
+- Use the number pad to read lines (7, 8, 9), read words, (4, 5, 6), or characters (1, 2, 3).
+
+Note:
+- For color reporting enable the option (via Document formatting panel) _"report color"_.
```
lukaszgo1, line 14: Would it be worth mentioning here that for quick checking of color NVDA+Shift+F can be used?

Qchristensen (approved):
UserGuide change looks fine, good work!

source/colors.py
```diff
@@ -1,24 +1,62 @@
...
 import re
+from functools import lru_cache
+from typing import Union
+
+#: Flag to indicate color being decoded from displayModelFormatColor_t
+# is transparent.
+# See (displayModel.cpp in nvdaHelper) displayModelFormatColor_t::TRANSPARENT_BIT
+TRANSPARENT_BITFLAG = 0x01 << 24
 
 class RGB(namedtuple('RGB',('red','green','blue'))):
```
leonardder, line 19: I wonder whether we should consider changing this into a dataclass in 2022.1

source/colors.py
```diff
@@ -1,24 +1,62 @@
...
+			Encoding alpha in TT was considered (eg 0xFFbbggrr for opaque),
+			but this would break compatibility with code that continues to pass
+			0x00bbggrr for opaque, all usages would need to be fixed.
+		"""
+		rr = c & 0xFF
+		gg = (c >> 8) & 0xFF
+		bb = (c >> 16) & 0xFF
+		tt = c & TRANSPARENT_BITFLAG
+		rgb = cls(rr, gg, bb)
+		rgb.alphaValue = 0x00 if bool(tt) else 0xFF
```
leonardder, line 45: Could we use constants for this instead?

LeonarddeR (approved):

devDocs/displayModel.md
feerrenrut: Yes, as long as we are confident it reports the color from the display model? I'll have to test to be sure.

source/colors.py
feerrenrut: Yes, or given it's complexity a regular class might be more appropriate.

source/colors.py
feerrenrut: Yep, good point.

devDocs/displayModel.md
feerrenrut: I have confirmed this.

source/colors.py
feerrenrut: I won't include this change here, I think this deserves its own conversation. Happy to see a PR if you would like to experiment with it.

michaelDCurran (approved):

PR #12660, opened 2021-07-20
UI Automation in Windows Console: add "API levels" as a better isImprovedTextRangeAvailable

codeofdusk:
### Link to issue number:
None.

### Summary of the issue:
The Windows 11 inbox console will ship with most, but not all, fixes required to make UIA by default a good experience. However, a new conhost version will be delivered "out-of-box" including additional functionality and bug fixes. Therefore, `isImprovedTextRangeAvailable` (a boolean value) is now insufficient to express the range of possible UIA console versions now in the wild.

* Adds an enumeration, `WinConsoleAPILevel`, with comments describing the meaning of each member.
* Refactors NVDA's current code to use the new enum without changing current functionality. Functionality changes to follow in subsequent PRs.
* Expose apiLevel in dev info to ease debugging.
* Deprecate `isImprovedTextRangeAvailable`, but refactor it for now to use `apiLevel`.

### Testing strategy:
Tested conhost versions meeting each API level and verified that the correct level was being set.

* The `FORMATTED` level is intentionally unused in NVDA code, usage to follow in subsequent PRs.
    * `FORMATTED` consoles are correctly detected when run though.

### Change log entries:
== Changes for Developers ==
- `NVDAObjects.UIA.winConsoleUIA.WinConsoleUIA.isImprovedTextRangeAvailable` has been deprecated. (#12660)
    - Instead use `apiLevel` (see the comments at _UIAConstants.WinConsoleAPILevel for details).

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/UIA/winConsoleUIA.py
+++ b/source/NVDAObjects/UIA/winConsoleUIA.py
@@ -342,7 +342,8 @@ class WinConsoleUIA(KeyboardHandlerBasedTypedCharSupport):
 		needed in a given conhost instance.
 		See the comments on the WinConsoleAPILevel enum for details.
 		"""
-		return _getConhostAPILevel(self.windowHandle)
+		self.apiLevel = _getConhostAPILevel(self.windowHandle)
+		return self.apiLevel
 
 	def _get__caretMovementTimeoutMultiplier(self):
 		"On older consoles, the caret can take a while to move."
--- a/source/NVDAObjects/UIA/winConsoleUIA.py
+++ b/source/NVDAObjects/UIA/winConsoleUIA.py
@@ -11,7 +11,8 @@ import UIAHandler
 
 from comtypes import COMError
 from logHandler import log
-from UIAUtils import _isImprovedConhostTextRangeAvailable
+from UIAUtils import _getConhostAPILevel
+from _UIAConstants import WinConsoleAPILevel
 from . import UIATextInfo
 from ..behaviors import EnhancedTermTypedCharSupport, KeyboardHandlerBasedTypedCharSupport
 from ..window import Window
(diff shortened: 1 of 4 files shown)


codeofdusk:
Cc @carlos-zamora, @DHowett.

codeofdusk:
Also CC @seanbudd.

codeofdusk:
Wow, that was a fast branch for 2021.2!

Any chance of this making it into 2021.2, with functionality changes to follow for 2021.3? This change is very low risk but will make debugging quite a lot easier.

leonardder:
Would it be possible to determine the correct behavior based on the version of conhost available on the system?

codeofdusk:
> Would it be possible to determine the correct behavior based on the version of conhost available on the system?

According to the conhost devs, version number is no longer a reliable indicator of functionality.

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -335,9 +336,20 @@ class WinConsoleUIA(KeyboardHandlerBasedTypedCharSupport):
 	#: a lot of text.
 	STABILIZE_DELAY = 0.03
 
+	def _get_apiLevel(self) -> WinConsoleAPILevel:
+		"""
+		This property shows which of several console UIA workarounds are
+		needed in a given conhost instance.
+		See the comments on the WinConsoleAPILevel enum for details.
+		"""
+		return _getConhostAPILevel(self.windowHandle)
```
leonardder, line 345: I think this can be a fetch once property. In the current situation, this will be cached for only one core cycle.

```suggestion
		self.apiLevel = _getConhostAPILevel(self.windowHandle)
		return self.apiLevel
```

seanbudd (approved):
Thanks for this, looks good to me

DHowett:
I think this is such a good way to do this. I'm always an advocate for feature detection rather than versioning, especially since Windows' servicing story makes it ... difficult to pin down precise time ranges 😄.

Thanks @codeofdusk!

PR #12680, opened 2021-07-24
Optionally detect caret movement with events in terminal applications

codeofdusk:
### Link to issue number:
None

### Summary of the issue:
Using events to detect caret movement increases responsiveness in some scenarios, but we intermitantly get unreliable results when backspacing quickly (see microsoft/terminal#10820).

Add a feature flag to enable `caretMovementDetectionUsesEvents` in terminals. The setting is currently ternary to avoid the need for a config spec upgrade should automatic enablement later be added in some apps.

In a follow-up PR, the checkbox may be changed to a combobox (similar to the approach in #10964).

### Testing strategy:
Tested that caret events are used when the feature flag is enabled, and not used when the flag is disabled in Windows Terminal and both UIA and legacy conhost.

None.

### Change log entries:
== Changes for Developers ==
- Added an advanced setting to enable caret movement detection in terminal applications for testing purposes.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/behaviors.py
+++ b/source/NVDAObjects/behaviors.py
@@ -362,7 +362,7 @@ class Terminal(LiveText, EditableText):
 	def _get_caretMovementDetectionUsesEvents(self):
 		"""Using caret events in consoles sometimes causes the last character of the
 		prompt to be read when quickly deleting text."""
-		return False
+		return config.conf["terminals"]["caretMovementShouldUseEvents"] == "always"
 
 
 class EnhancedTermTypedCharSupport(Terminal):
--- a/source/config/configSpec.py
+++ b/source/config/configSpec.py
@@ -236,6 +236,7 @@ schemaVersion = integer(min=0, default={latestSchemaVersion})
 	speakPasswords = boolean(default=false)
 	keyboardSupportInLegacy = boolean(default=True)
 	diffAlgo = option("auto", "dmp", "difflib", default="auto")
+	caretMovementShouldUseEvents = option("auto", "always", "never", default="auto")
 
 [update]
 	autoCheck = boolean(default=true)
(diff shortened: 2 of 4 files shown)


codeofdusk:
Cc @DHowett, @carlos-zamora, @miniksa.

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -410,4 +417,9 @@ def findExtraOverlayClasses(obj, clsList):
 
 
 class WinTerminalUIA(EnhancedTermTypedCharSupport):
-	pass
+	def _get_caretMovementDetectionUsesEvents(self):
```
leonardder, line 420: Isn't this the default? Also, I don't think this needs to be a property.

codeofdusk:
> Isn't this the default?

[Nope](https://github.com/nvaccess/nvda/blob/master/source/NVDAObjects/behaviors.py#L361).

source/NVDAObjects/UIA/winConsoleUIA.py
codeofdusk: > Isn't this the default?

[Nope](https://github.com/nvaccess/nvda/blob/master/source/NVDAObjects/behaviors.py#L361).

seanbudd (dismissed):

feerrenrut:
At this late stage there needs to be strong justification for why the 2021.3 release can't be made without this change, otherwise it should target 2022.1

codeofdusk:
PR retargetted to master (2022.1).

codeofdusk:
I've rebased/retargetted the PR for 2022.1. Cc @seanbudd and I think @Qchristensen?

feerrenrut (changes requested):

PR #12683, opened 2021-07-25
User guide: mention Windows 11

### LINKED ISSUE #12675: User guide: begin mentioning Windows 11
Hi,

This is mostly for Quentin:

At the moment user guide and other documentation notes Windows 10 as the latest supported Windows release. With Windows 11 being finalized, along iwth NVDA 2021.2 code moving to beta soon (with foundations to support Windows 11 included), it makes sense to start mentioning Windows 11 in user-facing documentation such as the user guide.

At least for the user guide, edit the following sections:

* Supported Windows releases
* Windows 10 OCR
* Windows OneCore voices
* Other parts where Windows 10 mentioned or implied

None

### Additional context
This assumes NVDA 2021.2 stable version will be released around the time Windows 11 will be released to the general public.

Thanks.

josephsl:
### Link to issue number:
Closes #12675 

### Summary of the issue:
According to the user guide, latest supported Windows release is Windows 10. As Windows 11 is on the horizon, it makes sense to mention the new operating system in approrpate places throughout the user guide.

Edited the following contents:
* Supported operating systems (add a note that Windows 11 does not support 32-bit systems but runs 32-bit apps such as NVDA)
* Windows 10 OCR (mention Windows 11 indirectly)
* Windows OneCore voices (indirectly mention Windows 11)
* Other places where Windows 10 is mentioned

### Testing strategy:
User guide generation, verifying information on Windows 11 (build 22000).

Few thoughts:
* Should limitations with screen curtain be mentioned?
* Should Windows 10 OCR be renamed? If so, a new issue/PR pair should be used.

### Change log entries:
None (this is user guide)

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/user_docs/en/userGuide.t2t
+++ b/user_docs/en/userGuide.t2t
@@ -66,9 +66,10 @@ For further details, you can [view the full license. https://www.gnu.org/license
 For details regarding exceptions, access the license document from the NVDA menu under the "help" section.
 
 + System Requirements +[SystemRequirements]
-- Operating Systems: all 32-bit and 64-bit editions of Windows 7, Windows 8, Windows 8.1, Windows 10, and all Server Operating Systems starting from Windows Server 2008 R2.
+- Operating Systems: all 32-bit and 64-bit editions of Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11 (64-bit only), and all Server Operating Systems starting from Windows Server 2008 R2.
  - For Windows 7, NVDA requires Service Pack 1 or higher.
  - For Windows Server 2008 R2, NVDA requires Service Pack 1 or higher.
+ - Windows 11 does not support 32-bit systems but will run 32-bit programs such as NVDA.
 - at least 150 MB of storage space.
 -
 
@@ -797,7 +798,7 @@ When the screen curtain is active some tasks directly based on what appears on t
 
 + Content Recognition +[ContentRecognition]
 When authors don't provide sufficient information for a screen reader user to determine the content of something, various tools can be used to attempt to recognize the content from an image.
-NVDA supports the optical character recognition (OCR) functionality built into Windows 10 to recognize text from images.
+NVDA supports the optical character recognition (OCR) functionality built into Windows 10 and later to recognize text from images.
 Additional content recognizers can be provided in NVDA add-ons.
 
 When you use a content recognition command, NVDA recognizes content from the current [navigator object #ObjectNavigation].
@@ -810,7 +811,7 @@ Pressing enter or space will activate (normally click) the text at the cursor if
 Pressing escape dismisses the recognition result.
 
 ++ Windows 10 OCR ++[Win10Ocr]
-Windows 10 includes OCR for many languages.
+Windows 10 and later includes OCR for many languages.
 NVDA can use this to recognize text from images or inaccessible applications.
 
 You can set the language to use for text recognition in the [Windows 10 OCR category #Win10OcrSettings] of the [NVDA Settings #NVDASettings] dialog.
@@ -1835,7 +1836,7 @@ If you suffer from performance issues in one or more applications, We recommend 
 ==== Use UI automation to access Microsoft Word document controls when available ====[AdvancedSettingsUseUiaForWord]
 When this option is enabled, NVDA will try to use the Microsoft UI Automation accessibility API in order to fetch information from Microsoft Word document controls.
 This includes Microsoft Word itself, and also the Microsoft Outlook message viewer and composer.
- For the most recent versions of Microsoft Office 2016/365 running on windows 10, UI Automation support is complete enough to provide access to Microsoft Word documents almost equal to NVDA's existing Microsoft Word support, with the added advantage that responsiveness is majorly increased.
+ For the most recent versions of Microsoft Office 2016/365 running on windows 10 and later, UI Automation support is complete enough to provide access to Microsoft Word documents almost equal to NVDA's existing Microsoft Word support, with the added advantage that responsiveness is majorly increased.
 However, There may be some information which is either not exposed, or exposed incorrectly in some versions of Microsoft Office, which means this UI automation support cannot always be relied upon.
 We still do not recommend that  the majority of users turn this on by default, though we do welcome users of Office 2016/365 to test this feature and provide feedback.
 
@@ -2263,7 +2264,7 @@ For an even more extensive list of  free and commercial synthesizers that you ca
 
 ++ eSpeak NG ++[eSpeakNG]
 The [eSpeak NG https://github.com/espeak-ng/espeak-ng] synthesizer is built directly into NVDA and does not require any other special drivers or components to be installed.
-On Windows 7, 8 and 8.1, NVDA uses eSpeak NG by default ([Windows OneCore #OneCore] is used in Windows 10 by default).
+On Windows 7, 8 and 8.1, NVDA uses eSpeak NG by default ([Windows OneCore #OneCore] is used in Windows 10 and later by default).
 As this synthesizer is built into NVDA, this is a great choice for when running NVDA off a USB thumb drive on other systems.
 
 Each voice that comes with eSpeak NG speaks a different language.
@@ -2297,9 +2298,9 @@ To use these voices, you will need to install two components:
 -
 
 ++ Windows OneCore Voices ++[OneCore]
-Windows 10 includes new voices known as "OneCore" or "mobile" voices.
+Windows 10 and later includes new voices known as "OneCore" or "mobile" voices.
 Voices are provided for many languages, and they are more responsive than the Microsoft voices available using Microsoft Speech API version 5.
-On Windows 10, NVDA uses Windows OneCore voices by default ([eSpeak NG #eSpeakNG] is used in other releases).
+On Windows 10 and later, NVDA uses Windows OneCore voices by default ([eSpeak NG #eSpeakNG] is used in other releases).
 
 To add new Windows OneCore voices, go to "Speech Settings", within Windows system settings. 
 Use the "Add voices" option and search for the desired language.
@@ -2516,7 +2517,7 @@ If connecting via USB to displays which do not use HID, you must first install t
 The VarioUltra and Pronto! use HID.
 The Refreshabraille and Orbit Reader 20 can use HID if configured appropriately.
 
-The USB serial mode of the Orbit Reader 20 is currently only supported in Windows 10.
+The USB serial mode of the Orbit Reader 20 is currently only supported in Windows 10 and later.
 USB HID should generally be used instead.
 
 Following are the key assignments for these displays with NVDA.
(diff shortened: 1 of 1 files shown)


josephsl:
Hi,

Note: PR based on beta branch as master branch was merged into it last week.

Thanks.

user_docs/en/userGuide.t2t
```diff
@@ -1835,7 +1836,7 @@ If you suffer from performance issues in one or more applications, We recommend
 ==== Use UI automation to access Microsoft Word document controls when available ====[AdvancedSettingsUseUiaForWord]
 When this option is enabled, NVDA will try to use the Microsoft UI Automation accessibility API in order to fetch information from Microsoft Word document controls.
 This includes Microsoft Word itself, and also the Microsoft Outlook message viewer and composer.
- For the most recent versions of Microsoft Office 2016/365 running on windows 10, UI Automation support is complete enough to provide access to Microsoft Word documents almost equal to NVDA's existing Microsoft Word support, with the added advantage that responsiveness is majorly increased.
+ For the most recent versions of Microsoft Office 2016/365 running on windows 10 and later, UI Automation support is complete enough to provide access to Microsoft Word documents almost equal to NVDA's existing Microsoft Word support, with the added advantage that responsiveness is majorly increased.
```
seanbudd, line 1839: Might as well make the capitalisation consistent here
```suggestion
 For the most recent versions of Microsoft Office 2016/365 running on Windows 10 and later, UI Automation support is complete enough to provide access to Microsoft Word documents almost equal to NVDA's existing Microsoft Word support, with the added advantage that responsiveness is majorly increased.
```

user_docs/en/userGuide.t2t
```diff
@@ -66,9 +66,10 @@ For further details, you can [view the full license. https://www.gnu.org/license
 For details regarding exceptions, access the license document from the NVDA menu under the "help" section.
 
 + System Requirements +[SystemRequirements]
-- Operating Systems: all 32-bit and 64-bit editions of Windows 7, Windows 8, Windows 8.1, Windows 10, and all Server Operating Systems starting from Windows Server 2008 R2.
+- Operating Systems: all 32-bit and 64-bit editions of Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11 (64-bit only), and all Server Operating Systems starting from Windows Server 2008 R2.
```
seanbudd, line 69: I have a concern that specifying this might confuse some readers, as the following is still a true statement:

```all 32-bit and 64-bit editions of Windows 7, ..., Windows 11, and all .. ```

File: user_docs/en/userGuide.t2t [hunk 3]
```diff
@@ -66,9 +66,10 @@ For further details, you can [view the full license. https://www.gnu.org/license
 For details regarding exceptions, access the license document from the NVDA menu under the "help" section.
 
 + System Requirements +[SystemRequirements]
-- Operating Systems: all 32-bit and 64-bit editions of Windows 7, Windows 8, Windows 8.1, Windows 10, and all Server Operating Systems starting from Windows Server 2008 R2.
+- Operating Systems: all 32-bit and 64-bit editions of Windows 7, Windows 8, Windows 8.1, Windows 10, Windows 11 (64-bit only), and all Server Operating Systems starting from Windows Server 2008 R2.
  - For Windows 7, NVDA requires Service Pack 1 or higher.
  - For Windows Server 2008 R2, NVDA requires Service Pack 1 or higher.
+ - Windows 11 does not support 32-bit systems but will run 32-bit programs such as NVDA.
```
seanbudd, line 72: Do NVDA users need to know this, considering we have said that Windows 11 is supported? I don't think the NVDA user guide is a good source of truth for Windows 11 functionality.

seanbudd (commented):
Thanks @josephsl, I left some notes on this PR but I would wait to hear what @Qchristensen and others think before changing anything based on my suggestions. I agree that this should be included in beta.

josephsl:
Hi, for the first point, capitalization issue was present before this PR, but it might be best to address it now. As for the second and third point, I think it is better to at least provide truths about Windows 11 requiring 64-bit systems because that’s what Microsoft notes in its specs (one of the first documentation users would read about screen readers is user guide, and if we say NVDA supports 32-bit and 64-bit Windows releases, we might as well provide more accurate info by including a note about Windows 11 here; @BriTechGuy, any thoughts?). Thanks.

britechguy:
My general thoughts are that:

1.  As much as possible, drop references to Windows versions, directly, at all.  There needs to be an understanding, in the age of Windows as a Service, and with Windows 7 now being dead, that "currently supported versions of Windows" is what Windows, not otherwise categorized, means.  With Microsoft ditching it's, "Windows 10 is the final version of Windows," stance, we have no idea how long it will be until Windows 12 (or something else, will follow).  It makes sense to me to avoid version specification to the greatest extent possible.
2. Bit depth of Windows is very rapidly becoming irrelevant.  It has been ages since 32-bit hardware was produced.  32-bit software routinely runs under 64-bit hardware and, as you note, very soon even 32-bit hardware will not be supported by the very latest version of Windows.

There needs to be an acknowledgement that being too specific with regard to Windows versions makes documentation a nightmare.  One could put something either at the beginning, or end, of the documentation that says something like, "Windows is construed to mean Windows 10 (32 or 64-bit), Windows 11 (64-bit only), and any successors."

The main body of the documentation should strive to be as broadly applicable as possible, with specifics defined once, wherever is felt appropriate.

seanbudd:
I would generally support @britechguy's changes here, but perhaps they should be in a follow up PR as this might be contentious. 

We do need to be specific with versions we support somewhere, as it can be why someone may or may not upgrade or buy a device. 

I would support changing this documentation to be more general, but maybe by specifically defining "Legacy Windows (<10)" and "Modern Windows (10+)". That way we can minimise changes for support being dropped for Legacy Windows or new Windows being supported. Most of the changes here used groups of "Modern Windows" or "Legacy Windows". Windows 7 support will be dropped with the next Python upgrade. While these changes occur fairly slowly, it might be helpful for readers to be able to track the changes consistently all in one place. 

I somewhat agree with @britechguy, I'm not sure the value is in specifying the Bit depth of in Windows, other than confirming that all available versions are supported. However, I'm also hesitant to change this, as 32bit support is getting worse as we progress into the future. For example, the screen curtain does not currently work on Windows 7 due to the Windows API not fully supporting WOW64 (it's possible this is the cause of the Windows 11 problem too, I'm currently investigating this). In fact, in general the API we use for screen curtain is (now) no longer support for WOW64: ["The Magnification API is not supported under WOW64; that is, a 32-bit magnifier application will not run correctly on 64-bit Windows."](https://docs.microsoft.com/en-us/windows/win32/winauto/magapi/entry-magapi-sdk). We currently cannot update to python3.8+ due to a bug with 32bit support.

britechguy:
@seanbudd wrote, in part, " Windows 7 support will be dropped with the next Python upgrade."

If I am not mistaken, this will be a part of 2022.1, so in terms of development pipeline that's the very near future.  I know that Freedom Scientific's changes in the same time frame will even drop support for Windows 8.x.  Essentially, everything we consider to be Windows Legacy is going to be completely out of support in the clearly foreseeable and not-too-distant future.  I don't know when NVDA might have Windows 8.x fall out of support.  Might it be even before January 2023?

Perhaps another PR is needed, I'm not arguing about that, or anything, really.  But the days of Windows Legacy are numbered, and very, very close to their end.  Focus should be more on how these references will be purged going forward rather than maintained.

We are so far in to the Windows 10 era (and I include the forthcoming Windows 11 as part of that) and so near the end of all that came before that planning for the demise of all that came before really does need "front and center" consideration.

josephsl:
Hi, regarding Python upgrades, we also need to balance that with implications for things such as SAPI4 support; earlier this year we did experiment with Python 3.8 but ran into problems. If I’m to advocate Python upgrade, I advise 3.0 at the least, or better yet, 3.10, but then lack of wxPython development is going to cause headaches for a while. Thanks.

seanbudd:
@britechguy - the upgrade to Python3.8+ is currently blocked by the issue Joseph described. Python 3.8, which is the last version that supports Windows 7, has an EOL of 14 Oct 2024. So we will drop Windows 7 support sometime before then, but it may not be in 2022.1

britechguy:
@seanbudd,

Thanks for the clarification.

If there's a choice between positioning NVDA to move "boldly into the future" that breaks backward compatibility with Windows 7, and maintaining that compatibility, I fall firmly in support of the former.

Any compatibility, of anything, with Windows 7 at this point in time should not be a priority and should only occur if it is a an accidental compatibility maintenance.  There are lots of things that remain compatible because the next major update that would break it has not occurred yet.  But Windows 7 is dead, and no hoops should be jumped through to maintain backward compatibility.

josephsl:
Hi, it isn’t just Windows 7 – we might as well take this time to end support for SAPI4, knowing that people will complain about breaking speech synthesizers. A similar concern was raised when NV Access dropped support for Adobe Flash. Given that Windows 8.x and Server 2012/2012 R2 will exit support in 2023, I think the best target would be 2023.1, at which point folks would have moved onto Windows 10 and Windows 11 would have been tested in the wild for at least a year. Thanks.

seanbudd (dismissed):

user_docs/en/userGuide.t2t
```diff
@@ -810,7 +810,7 @@ Pressing enter or space will activate (normally click) the text at the cursor if
 Pressing escape dismisses the recognition result.
 
 ++ Windows 10 OCR ++[Win10Ocr]
```
leonardder, line 812: Should we rename this to be more generic?

```suggestion
++ Windows OCR ++[Win10Ocr]
```

user_docs/en/userGuide.t2t
```diff
@@ -2297,9 +2297,9 @@ To use these voices, you will need to install two components:
 -
 
 ++ Windows OneCore Voices ++[OneCore]
-Windows 10 includes new voices known as "OneCore" or "mobile" voices.
+Windows 10 and later includes new voices known as "OneCore" or "mobile" voices.
```
leonardder, line 2300:
```suggestion
Windows 10 and later includes voices known as "OneCore" or "mobile" voices.
```

josephsl:
Hi, I think it would be best to rename Windows 10 OCR to Windows OCR or something sooner than later (this must be done for both docs and UI). Thanks.

seanbudd (approved):

valiant8086:
Hi.


I'm tempted to complain about breaking speech synthesizers. Is there a 
distinct advantage to removing SAPI4 support from NVDA, and are you 
aware that SAPI4 tts engines like True Voice install and work well in 
Windows 10? I have not tried in Windows 11. True Voice is the best 
example I can come up with. It's actually a good sounding tts and is 
well liked at least by people I've ever asked. It reminds me distantly 
of the BDL voice from RH Voice. Only other one we'd care about in 
English anyway to my knowledge would be Mary Mike and Sam which are 
better quality voices than their SAPI5 counterparts at least Sam is. 
There's also Flextalk but I'm not even sure that's supposed to be free 
and it's not very good so I'm ok with not being able to use it.

josephsl:
Hi, SAPI4 issue is a separate matter from this one – we (NVDA contributors and NV Access) learned that SAPI4 support is affected when we use Python 3.8 or later. Besides, the fact that certain Windows API’s are deemed “unstable” for WoW64 apps such as NVDA, coupled with attempts by Microsoft to get folks to use modern technologies are indication that the days of SAPI4 support is slowly coming to a close (this is my comment on it at this point, as the final decision rests with NV Access and market conditions). Thanks.

Issue #12672, opened 2021-07-22
Have a way to configure beep on error


CyrilleB79:
First of all, sorry in case this issue is a duplicate of another one.

### Is your feature request related to a problem? Please describe.
In alpha and beta versions of NVDA, a beep is heard when an error is logged. I am missing this beep when I am doing some tests on a stable NVDA (add-on tests or tests of NVDA itself).

### Describe the solution you'd like

Add an option to configure when the error should make a beep with the 3 following choices:
* "Always": i.e. in NVDA stable as well as alpha/beta
* "Never": neither in NVDA stable nor in alpha/beta
* "Only in dev releases": beeps in alpha/beta but not in stable release. It is the current behaviour and should be the default.

The option may be located in the advanced settings panel.

### Describe alternatives you've considered

For now I have a global plugin in my scratchpad to do a similar job. But it's an issue when I make some tests with add-ons disabled.

### Additional context
N/A

feerrenrut:
Happy for this to be added, but given alpha/beta are really intended for testing, and we want users to be aware of and then report errors, I don't think the "Never" option should be included.

CyrilleB79:
Yes this makes sense. Let's remove the "Never" option.

leonardder:
I really dig this idea, as this generally holds me back from stable releases, as I am that accustomed to error sounds that I really miss them in case there's a real error.

leonardder:
Related to #11634

lukaszgo1:
While I agree with this proposal in general I'd like to make a case for 'never' being an option. Consider the following real life example where having a way to suppress a beep on error can be useful for developers. After PR #11782 got merged accessing Internet explorer controls resulted in many errors from garbage Handler. Since the mailer that I am using displays messages in MSHTML controls I've been getting an error sounnd each time when the message was opened or closed so hardly an comfortable experience. In such case where the given issue has been reported and user awaits a fix it would be pretty nice to be able to disable error sounds in a configuration profile for a given program as the only alternative is to downgrade to the stable build.

feerrenrut:
We really want to discourage relying on alpha versions of NVDA in "production" environments. If there are annoying errors like this, users should fall pack to a prior final release. Either have the prior stable release available as a portable install, or use the alpha as a portable install.

Disabling the error sound makes it too easy to forget to enable it again. We also don't want to train users to be complacent about error sounds, turning off error sounds means that errors in different situations won't be reported.

PR #12689, opened 2021-07-27
Option to configure when the error sound may be played.

CyrilleB79:
### Link to issue number:

#12672

### Summary of the issue:
The error sound that is played in test versions may also be useful to testers on stable and RC releases.
The motivation to have this feature included in NVDA's core rather than in an add-on is that it may be useful to test NVDA stable versions with add-ons disabled.

Added a 'Play a sound for logged errors' option in the advanced settings panel. It is a combo-box with two choices:
- 'Only in test versions': NVDA plays error sounds as today, i.e. only in NVDA test versions (alpha, beta and from source) but not in RC or stable.
- Yes: NVDA plays the error sound whatever its version (test or stable)
According to the discussion in #12672, I do not provide an option to disable error sounds in tests versions, since test versions are aimed to capture potential issues. Thus error notifications should not be disabled for this usage.

If an error occurs before the config is initialized, the default behaviour is applied, i.e. the error sound is played only in test versions.

Note for reviewers:
* I have had to import config in the `handle` method rather than at the top of the module since NVDA could not start in the latter case. I do not know why however, since I do not know anymore how to debug this since the virtual environment has been set up.
* Please native English speakers, check my writings (GUI and user guide)

Additional modification:
While working on this part of the code, I have made the critical error sound played asynchronously upon @leonardder's suggestion. It should allow to spare 2-3 seconds (the time of the error sound) in the code execution time.
NV Access, if you prefer not to include this modification for whatever reason, I or you may revert this commit.

### Testing strategy:

Manual testing:
Triggered an error in the 4 following situations and checked that the error sound is played for the 3 first ones and not the last one:
* NVDA test version (from source) and option set to 'Yes'
* NVDA test version (from source) and option set to 'Only in NVDA test versions'
* NVDA stable version (simulated from source) and option set to 'Yes'
* NVDA stable version (simulated from source) and option set to 'Only in NVDA test versions'

The stable version is simulated running NVDA from source with this branch and entering the 2 following lines in the Python console:
```
import buildVersion
buildVersion.isTestVersion = False
```

The error is triggered by trying to enable the screen curtain while Windows Magnifier is enabled with color inversion activated. But many other ways could have been used.

I will also check context help on the snapshot generated by appVeyor.

If an error occurs before the config is initialized, the default behaviour is applied, i.e. the error sound is played only in test versions whatever the value of this option is in the config.

### Change log entries:
New features (or Changes? Or For developers?)
It is now possible to configure the error sound to be played in any version of NVDA, not only in test versions. (#12672)

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


source/logHandler.py
```diff
@@ -266,8 +266,13 @@ def emit(self, record):
 class FileHandler(logging.FileHandler):
 
 	def handle(self,record):
-		# Only play the error sound if this is a test version.
-		shouldPlayErrorSound =  buildVersion.isTestVersion
+		import config
+		# Only play the error sound if this is a test version or if the config states it explicitely.
```
lukaszgo1, line 270: explicitely -> explicitly
```suggestion
		# Only play the error sound if this is a test version or if the config states it explicitly.
```

source/logHandler.py
```diff
@@ -266,8 +266,13 @@ def emit(self, record):
...
 
 	def handle(self,record):
-		# Only play the error sound if this is a test version.
-		shouldPlayErrorSound =  buildVersion.isTestVersion
+		import config
+		# Only play the error sound if this is a test version or if the config states it explicitely.
+		shouldPlayErrorSound = (
+			buildVersion.isTestVersion
+			# Play error sound: 1 = Yes
+			or config.conf["featureFlag"]["playErrorSound"] == 1
```
lukaszgo1, line 274: You need to protect against `config.conf` being `None` here. This can happen if anything is logged at the error level between `logHandler` being initialized and the call to `config.initialize` in `core.main`.

lukaszgo1 (changes requested):

source/logHandler.py
CyrilleB79: Thank you. Fixed in 9a1cb16.
I'll update the initial description regarding the inability to have the error sound played in the stable release before the log is initialized.

CyrilleB79:
I have checked context help on the appVeyor artifact. And I have addressed @lukaszgo1's comments (thanks Lukasz for the review).
Thus I have passed this PR to ready state.

lukaszgo1:
Looking at this again would it be possible to move the code which is responsible for determining if the error sound should play into a separate method/function? It would allow add-on developers / power users to monkey patch just a single function rather than the entire `emit` and would allow to indirectly add 'never' option for developers.

CyrilleB79:
> Looking at this again would it be possible to move the code which is responsible for determining if the error sound should play into a separate method/function? It would allow add-on developers / power users to monkey patch just a single function rather than the entire `emit` and would allow to indirectly add 'never' option for developers.

Done.

source/logHandler.py
```diff
@@ -266,14 +281,12 @@ def emit(self, record):
 class FileHandler(logging.FileHandler):
 
 	def handle(self,record):
-		# Only play the error sound if this is a test version.
-		shouldPlayErrorSound =  buildVersion.isTestVersion
 		if record.levelno>=logging.CRITICAL:
 			try:
 				winsound.PlaySound("SystemHand",winsound.SND_ALIAS)
```
leonardder, line 286: May be you could fix this to play asynchronously while at it.

```suggestion
				winsound.PlaySound("SystemHand", winsound.SND_ALIAS | winsound.SND_ASYNC)
```

CyrilleB79:
This PR is ready again.

@feerrenrut you may want to review this one since you gave indications in https://github.com/nvaccess/nvda/issues/12672#issuecomment-885357649.

feerrenrut (approved):
The only thing that concerns me is the async sound playing. It is probably fine, and I'm happy for this to get some testing on alpha. But if there is an issue, we'll need to revert that aspect.

I won't merge this immediately, to give others a chance to review.

feerrenrut (approved):

PR #12702, opened 2021-07-30
Make unit tests for oneCore pass  under the versions of Windows where oneCore is not supported

lukaszgo1:
### Link to issue number:
Fixes unit tests introduced in #12647
### Summary of the issue:
Some unit tests introduced as part of #12647 were failing under versions of Windows where oneCore was unsupported. The output was as follows:
```
FAIL: Ensures that if oneCore doesn't support the current language, setSynth("auto") falls back to the
----------------------------------------------------------------------
  File "D:\my_repos\nvda\tests\unit\test_synthDriverHandler.py", line 113, in test_setSynth_auto_fallback_ifOneCoreDoesntSupportDefaultLangauge
    self.assertEqual(synthDriverHandler.getSynth().name, FAKE_DEFAULT_SYNTH_NAME)
AssertionError: 'espeak' != 'defaultSynth'
- espeak
+ defaultSynth


======================================================================
FAIL: Ensures that if oneCore supports the current language, setSynth("auto") uses "oneCore".
----------------------------------------------------------------------
  File "D:\my_repos\nvda\tests\unit\test_synthDriverHandler.py", line 104, in test_setSynth_auto_usesOneCore_ifSupportsDefaultLangauge
    self.assertEqual(synthDriverHandler.getSynth().name, "oneCore")
AssertionError: 'espeak' != 'oneCore'
- espeak
+ oneCore
```

Tests which require oneCore being supported are being skipped on versions of Windows on which oneCore is not supported.
### Testing strategy:
Unit tests passes locally whereas previously they didn't. I'll inspect log on AppVeyor to ensure they still pass on a version of Windows where oneCore is supported.
None known.
### Change log entries:
None needed

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/tests/unit/test_synthDriverHandler.py
+++ b/tests/unit/test_synthDriverHandler.py
@@ -94,31 +94,25 @@ class test_synthDriverHandler(unittest.TestCase):
 			self.assertEqual(synthName, synthDriverHandler.getSynth().name)
 			self.assertEqual(FAKE_DEFAULT_LANG, config.conf["speech"]["synth"])
 
+	@unittest.skipUnless(OneCoreSynthDriver.check(), "Requires oneCore being supported under current OS")
 	def test_setSynth_auto_usesOneCore_ifSupportsDefaultLanguage(self):
 		"""
-		Ensures that if oneCore supports the current language, setSynth("auto") uses "oneCore" on versions
-		of Windows where oneCore is the default - else eSpeak is used regardless of the language  support.
+		Ensures that if oneCore supports the current language, setSynth("auto") uses "oneCore".
 		"""
 		# test setup ensures curLang is supported for oneCore
 		synthDriverHandler.setSynth(None)  # reset the synth so there is no fallback
 		synthDriverHandler.setSynth("auto")
-		if OneCoreSynthDriver.check():
-			self.assertEqual(synthDriverHandler.getSynth().name, "oneCore")
-		else:
-			self.assertEqual(synthDriverHandler.getSynth().name, "espeak")
+		self.assertEqual(synthDriverHandler.getSynth().name, "oneCore")
 
+	@unittest.skipUnless(OneCoreSynthDriver.check(), "Requires oneCore being supported under current OS")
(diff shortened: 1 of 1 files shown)


Brian1Gaff:
Are not the u and a in Language the  wrong way around in the code?
 Brian

***@***.***
Sent via blueyonder.
in the display name field.
----- Original Message ----- 
From: "Łukasz Golonka" ***@***.***>
Cc: "Subscribed" ***@***.***>
Sent: Friday, July 30, 2021 5:36 PM
Subject: [nvaccess/nvda] Make unit tests for oneCore pass under the versions 
of Windows where oneCore is not supported (#12702)

lukaszgo1:
> Are not the u and a in Language the wrong way around in the code?

This has been fixed as part of this PR.

seanbudd:
I think performing logic within the test defeats the purpose of these tests. Generally you should avoid conditional logic within unit testing. I think it might be better to just add these wrappers to the existing oneCore tests

```python
@unittest.case.skipIf(winVersion.getWinVer() < winVersion.WIN10, "oneCore unsupported on this OS")
```

lukaszgo1:
> 
> 
> I think performing logic within the test defeats the purpose of these tests. Generally you should avoid conditional logic within unit testing. 

I agree conditional logic is an anti-pattern for unit tests however currently these test rely on `defaultSynthPriorityList` being set in a particular way which differs between the version of Windows in use which is also an anti-pattern IMO.

> I think it might be better to just add these wrappers to the existing oneCore tests

I've done mostly as you suggested however rather than testing for a version of Windows in use I'm checking for oneCore being supported which guarantees that if we would need to ever drop support  of it and / or it would become broken on a particular version of Windows the check would not need to be  modifiied in more than one place.

seanbudd (approved):
Thanks @lukaszgo1, looks good to me.

PR #10257, opened 2019-09-20
Make NVDA less verbose when  searching in Windows 7 start menu

lukaszgo1:
### Link to issue number:
Fixes #2020
### Summary of the issue:
When performing a search in Windows 7 start menu NVDA is a little bit too verbose.
### Description of how this pull request fixes the issue:
The search results are inside a list which name is 'Items view'. As the name adds no useful information I've removed it as discussed in #2020.
### Testing performed:
- Ensured that all important information's are still spoken when search results are shown.
- Ensured that the name is really gone.
- Ensured that no other list is affected by this change.
### Known issues with pull request:
None known
### Change log entry:
I don't think that such entry is needed here.

Issue #2020, opened 2012-01-02
decrease nvda verbosity in the windows 7 start search area


nvaccessAuto:
**Reported by jmwalton on 2012-01-02 07:21**
when you search in the windows 7 start menu search box, nvda speaks "items view list, programs, program" as an example. i feel that it would be better if it were to just say the search results as they come up. an example there would be you search for "update" and nvda would speak "windows update" not "items view list, programs, windows update."

ehollig:
While I see that this can be annoying at times, it is not life threatening. The information that is provided lets the user know where they are in the search. What do others think?

PR #12706, opened 2021-08-01
Catch `AttributeError` when fetching parent of  an no longer existing object when navigating quickly in Windows Explorer

lukaszgo1:
### Link to issue number:
None - fix-up of #10257

### Summary of the issue:
PR #10257 introduced code which removes  pointless name of list which contains search results in Windows 7 start menu. To determine if we're dealing with a correct control however it is necessary to travel a lot of object ancestry. This causes the following error when navigating quickly in Windows Explorer:
```
ERROR - NVDAObjects.__call__ (22:47:08.606) - Dummy-1 (6416):
Exception in event_NVDAObject_init for <'explorer' (appName 'explorer', process ID 2552) at address 3cc63d0>
  File "NVDAObjects\__init__.pyc", line 157, in __call__
  File "appModules\explorer.pyc", line 393, in event_NVDAObject_init
AttributeError: 'NoneType' object has no attribute 'parent'

```

Since this occurs for objects which are dead I've just started catching `AttributeError` in this block of code.
### Testing strategy:
- Navigated in Windows Explorer for several hours - made sure that no error sound plays
- Made sure that #2020 is not reintroduced.
 
None known
### Change log entries:
None needed- error sounds are not enabled in releases and aside from annoying error sound for time to time this error had no adverse side effects.

### Code Review Checklist:

- [X] Unit tests.
- [X] Manual testing.
- [X] User Documentation.
- [X] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/appModules/explorer.py
+++ b/source/appModules/explorer.py
@@ -389,13 +389,18 @@ class AppModule(appModuleHandler.AppModule):
 			return
 
 		if windowClass == "DirectUIHWND" and role == controlTypes.Role.LIST:
-			if obj.parent and obj.parent.parent:
-				parent = obj.parent.parent.parent
-				if parent is not None and parent.windowClassName == "Desktop Search Open View":
-					# List containing search results in Windows 7 start menu.
-					# Its name is not useful so discard it.
-					obj.name = None
-					return
+			# Is this a list containing search results in Windows 7 start menu?
+			isWin7SearchResultsList = False
+			try:
+				if obj.parent and obj.parent.parent:
+					parent = obj.parent.parent.parent
+					isWin7SearchResultsList = parent is not None and parent.windowClassName == "Desktop Search Open View"
+			except AttributeError:
+				isWin7SearchResultsList = False
+			if isWin7SearchResultsList:
+				# Namae of this list is not useful and should be  discarded.
+				obj.name = None
+				return
 
 	def event_gainFocus(self, obj, nextHandler):
 		wClass = obj.windowClassName


Brian1Gaff:
As a matter of interest, is this the reason for the difference between the 
spoken item and the actual auctioned item in explorer, ie some kind oof lag. 
This can be very irritating when you copy or delete the wrong file, 
believing nvda not what the actual visual screen is saying.
 Brian

***@***.***
Sent via blueyonder.
in the display name field.
----- Original Message ----- 
From: "Łukasz Golonka" ***@***.***>
Cc: "Subscribed" ***@***.***>
Sent: Sunday, August 01, 2021 6:04 PM
Subject: [nvaccess/nvda] Catch `AttributeError` when fetching parent of an 
no longer existing object when navigating quickly in Windows Explorer 
(#12706)

lukaszgo1:
> 
> As a matter of interest, is this the reason for the difference between the spoken item and the actual auctioned item in explorer, ie some kind oof lag. This can be very irritating when you copy or delete the wrong file, believing nvda not what the actual visual screen is saying. 


I have never experienced the issue you're describing but this PR is very unlikely to fix this - it merely prevents error sound to be  played in a very specific situation.

leonardder:
I'm slightly puzzled by how this could raise AttributeError as the original code already checks the boolean value of the three parents.

lukaszgo1:
> 
> 
> I'm slightly puzzled by how this could raise AttributeError as the original code already checks the boolean value of the three parents.


My working theory is that the object exists at the time when it is being checked for parent but no longer does when the parent is accessed since user moved to the different folder and therefore the list dies. To use a  analogy to accessing files calling `os.path.isfile` is not enough since even though this check returns `True` the file may be removed between the point when its existence is being checked and the next access to it.

michaelDCurran:
Generally, if this were being executed from the main thread, the existing code would be completely safe as the parent property would be chaed for the core cycle and therefore not change. However, from the error we can see that it is being executed on another thread "dummy-1" My guess probably a UIA worker thread. And because of this, the execution of this code could overlap the end of a main thread core cycle, thus the property cache may get cleared half way through. 
For now, I think this pr is suitable, but we may wish to consider queuing more if not all of the interactions with NvDAObjects from UIA thread into the main thread. In oother words, a UIA event should really just queue the event and the UIA element to the main thread, before it tries actually creating an NVDA object for it.  But that is something to consider longer-term.

michaelDCurran (approved):

PR #12727, opened 2021-08-11
Prefer the UIA value over the UIA range value when reporting value, and respect minum and maximum values for ranges

### LINKED ISSUE #12724: IN cases where UIA element both has ValuePattern and RangeValuePattern, NVDA uses the latter whereas this might be incorrect
### Steps to reproduce:
1. Download [Surge XT nightly](https://surge-synthesizer.github.io/nightly_XT) and install it.
2. Start the standalone version
3. With object nav, find a slider, e.g. inside global controls, the split point slider.

### Actual behavior:
@pitermach wrote in https://github.com/surge-synthesizer/surge/issues/4616#issuecomment-895431379

> On Windows, NVDA seems to ignore the real value and use a different scale, but as this doesn't happen with the other windows readers I tried (Narrator, JAWS, ZDSR) I suspect this is an NVDA issue especially because it also happens in at least one other Juce plugin

NVDA reads numerical values, e.g. 49.

### Expected behavior:
NVDA reads the actual value, e.g. C4

### Additional details

> It looks like the sliders implement both the ValuePattern and RangeValuePattern. The RangeValuePattern exposes a numerical value whereas the ValuePattern exposes the real value. From what I can see, it looks like code in NVDA needs to be swapped in order for this to work correctly. I have to take a look at whether this is an authoring error in Juce or NVDA.

_Originally posted by me in https://github.com/surge-synthesizer/surge/issues/4616#issuecomment-895749900_

LeonarddeR:
### Link to issue number:
Fixes #12724

### Summary of the issue:
1. For UIA controls that implement a RangeValuePattern and a ValuePattern, the RangeValuePattern gives a numerical presentation of the value in a range whereas the value could also have a textual representation exposed in the ValuePattern. NVDA prefered the numerical representation, which doesn't make much sense and isn't according to how Narrator, JAWS and LeagcyIAccessiblePattern expose the value.
2. NVDA insisted on correcting every range to a 0--100 range, even though UIA could specify different minimum and maximum values in the rangeValue pattern.
3. Range value patterns where rounded up to 4 digits and then converted to an integer. This resulted in 4.9998 resulting into an integer value of 4 instead of a rounded value of 5.

1. Swapped the order of methods the value is constructed. NVDA first tries the ValuePattern, than if that's empty, it uses the RangeValuePattern.
2. No longer take the minimum and maximum value into account when working with a range value and report the value as a percentage, but report the real value exposed by the pattern.
3. The range value pattern value is now rounded properly

### Testing strategy:
Tested the sliders as mentioned in #12724 as well as the sliders in Windows settings.

It could be that the range correction (from 0 to 100) was intentional, but it looks like it was there from the beginning, at least since 2009.

### Change log entries:
Bug fixes
* NVDA now longer reports a numerical value for UIA sliders that have a textual representation of their value defined. (#12724)
* NVDA now longer treats the value of UIA sliders as always percentage based.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


leonardder:
@josephsl raised the point of progress bars. I assume proper progress bars have a min value of 0.0 and a max value of 100.0, so they'll just work.

michaelDCurran:
In a purfect world, yes, progressBars would be already between 0 and 100. But as in the past NVDA has used the min and max of the value to normalize to a percentage, we wouldn't really know if they were already correct or not. Changing this now would break announcement of any progress bar which happened to not conform to what we think is correct. We can either just take the risk and fix broken progress bars as we find them, or my preferrred option would be to make a ProgresssBar overlay class for UIA, which normalized the value property to between 0 and 100 like the old code did.

leonardder:
I created a new overlay for progress bars.

michaelDCurran (approved):

PR #12742, opened 2021-08-17
Fix backtranslation for uncontracted braille when typing a sequence of characters consisting of multiple patterns

### LINKED ISSUE #12667: Braille input with NVDA isn't always correct
### Steps to reproduce:

1. Open NVDA braille settings and choose, for example, "unified English braille grade 1" as an input braille table (generally relevant for all braille tables).
2. (for UEB) input with a braille display keyboard: ⠐⠣⠠⠉⠐⠜ (5-126-6-14-5-345). Another example: ⠐⠣⠼⠁⠐⠜ (5-126-3456-1-5-345). And many other examples.

### Actual behavior:

For the first example: NVDA says: "Input isn't supported" when I try to input the caps indicator (dot 6) after '(' (dots 5-126) and ignores my input.
For the second example: NVDA says "Input isn't supported" when I try to input the number sign (3456) after '(' (5-126) and ignores my input.
Generally: something goes wrong with input multicellular signs after parentheses, quotes, brackets, dashes and other symbols. 

### Expected behavior:

Input should be processed correctly, number signs, caps indicators and other parts of multicellular symbols shouldn't be ignored.

Installed
#### NVDA version:
2021.1
#### Windows version:
10
I use the Focus 40 (4th generation) braille display.
Yes
Yes (including BrailleExtender)
### Other comments

I've found this problem when working on Russian Liblouis tables, but it's actual for all input tables.
It doesn't seem to be a problem of Liblouis because back translation in other programs used Liblouis (including lou_translate.exe) works correctly. I think that the problem is in how NVDA processes Liblouis tables.

LeonarddeR:
### Link to issue number:
Fixes #12667

### Summary of the issue:
Best described with the str in #12667. In short, when typing multiple characters consisting of multiple patterns, NVDA did not correctly clear the buffer after sending the translation of characters to a text field.

It looks like an if statement was incorrect. I will clarifiy in an inline comment.

I also added many debug logging while at it.

### Testing strategy:
Tested the steps to reproduce in #12667

I have time nor ideas to test this for possible regressions.

### Change log entries:
Bug fixes
- NVDA no longer fails to translate braille input when multiple characters are typed that consist of multiple braille patterns (e.g. (1) in UEB). (#12667)

Changes for developers
- Added a new "brailleInput" debugging category to NVDA"s advanced settings to debug input and translation of braille input.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


source/brailleInput.py
```diff
@@ -134,48 +152,64 @@ def _translate(self, endWord: bool) -> bool:
...
+					if _isDebugForBrailleInput():
+						log.debug(f"Clearing newText, was {newText}")
 					newText = u""
 				else:
 					self.emulateKey(newText)
 			else:
 				self.sendChars(newText)
 
-		if endWord or (newText and (not self.currentFocusIsTextObj or self.currentModifiers)):
+		if endWord or (newText and (self.currentModifiers or not self.useContractedForCurrentFocus)):
```
leonardder, line 193: In the old situation, the buffer wasn't cleared when in a text field. In the new situation, the buffer will be cleared when contraction should not apply, either when the table is not contracted or when modifiers are pressed.

seanbudd (approved):

Futyn-Maker:
Hello! Maybe I do something wrong, but in the last Alpha-version of NVDA this bug is still here.

leonardder:
I think it's time to revert this.

Issue #12620, opened 2021-07-07
Error in `tones.beep`


CyrilleB79:
### Steps to reproduce:

Make NVDA generate a beep via with `tones.beep`.

For example with [Speech History](https://addons.nvda-project.org/addons/speech_history.en.html) from @jscholes:
* Use NVDA normally (with speech)
* Press F12 to copy the last spoken text to the clipboard. This script copies the last spoken text and then calls the following beep command to indicate that the copy has occurred:
`tones.beep(1500, 120)`

### Actual behavior:

Rarely, the beep is not heard and the following can be seen in the log:
```
IO - inputCore.executeGesture (09:58:52.937) - winInputHook (11384):
Input: kb(desktop):f12
IO - tones.beep (09:58:52.961) - MainThread (11120):
Beep at pitch 1500, for 120 ms, left volume 50, right volume 50
ERROR - scriptHandler.executeScript (09:58:52.967) - MainThread (11120):
error executing script: <bound method GlobalPlugin.script_copyLast of <globalPlugins.speechHistory.GlobalPlugin object at 0x068FDA70>> with gesture 'f12'
Traceback (most recent call last):
  File "scriptHandler.pyc", line 208, in executeScript
  File "C:\Users\CB232690\AppData\Roaming\nvda\addons\speechHistory\globalPlugins\speechHistory.py", line 57, in script_copyLast
    tones.beep(1500, 120)
  File "tones.pyc", line 59, in beep
  File "nvwave.pyc", line 491, in stop
  File "nvwave.pyc", line 453, in _idleUnbuffered
  File "nvwave.pyc", line 408, in sync
  File "nvwave.pyc", line 111, in _winmm_errcheck
OSError: [Errno 5] Le handle du périphérique spécifié n’est pas valide.
```

The occurence of this issue is very rare (maybe 1/100 or less). But trying to execute the F12 command many times does not help reproducing this issue.

### Expected behavior:
The beep is played and no error in the log.

#### NVDA installed/portable/running from source:
installed
#### NVDA version:
2021.1rc2
#### Windows version:
Windows 10 20H2 (64-bit) build 19042.1052

#### Name and version of other software in use when reproducing the issue:

[Speech History](https://addons.nvda-project.org/addons/speech_history.en.html) add-on from @jscholes
Version modified by myself for 2021.1 compatibility:
* base version: 2020.2
* Manifest updated for 2021.1
* Import of the `speech` module updated according to 2021.1 code reorganization

[...]

PR #12804, opened 2021-09-03
Fix error when using tones.beep

feerrenrut:
### Link to issue number:
#12620

### Summary of the issue:
When exceptions could escape nvWave when an audio device is being changed / removed.
To stress test this, call tones.beep repeatedly and connect / disconnect audio devices.

When nvWave is in error, logHandler would still try to beep. This caused additional errors in the log.
The additional errors make debugging harder.

- The exception escapes from 'sync', wrap this in a try / except
- stop also ignored failures, and could call to _idleUnbuffered, which calls sync. Sync asserts that _waveout is not None.
  The failures are no longer ignored.
- Prevent logHandler calling beep when nvWave is in an error state.

### Testing strategy:
Used the following code in the NVDA python console, and disconnected a bluetooth audio device:
```python
from time import sleep
from tones import beep

sleepMilliSecs = 30
for i in range(5 * (1000 // 30)):
	beep(440, sleepMilliSecs)
	sleep(sleepMilliSecs / 1000.0)
```

This will need careful testing.

### Change log entries:
Bug fixes
`A rare error when changing audio devices has been fixed. (#12620)`

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


feerrenrut:
@CyrilleB79 Could you [test this PR build](https://ci.appveyor.com/api/buildjobs/xxdbj6hy8huk43si/artifacts/output%2Fnvda_snapshot_pr12804-23677%2Ccd4e6028.exe) as a fix for #12620 please.

CyrilleB79:
> @CyrilleB79 Could you [test this PR build](https://ci.appveyor.com/api/buildjobs/xxdbj6hy8huk43si/artifacts/output%2Fnvda_snapshot_pr12804-23677%2Ccd4e6028.exe) as a fix for #12620 please.

I have just done a test with the STR in https://github.com/nvaccess/nvda/issues/12620#issuecomment-906185718. I can confirm that the issue is gone in this case.
Just to inform that once, it took 2 or 3 seconds for the beeps to switch from the computer's speaker to the headset, while the voice was already in the headset. I do not think that this has been introduced with this PR however since I had already experimented some oddity like this in the past (2020.4 / 2021.1)

I am not able however to confirm or not the resolution of the issue described in the initial description of #12620 since it only occurs very rarely.
We may close #12620 with this PR and open a new issue in case the error still appears while testing with alpha.

CyrilleB79:
Just a clarification:
Should this PR fix only the case when the audio device is changed (like STR in [#12620 (comment)](https://github.com/nvaccess/nvda/issues/12620#issuecomment-906185718). Or does also fix other cases with no device change like the initial STR of #12620?
If the initial STR of #12620 applies to an add-on, I have also just had the case with NVDA 2021.2rc1 with a beep reporting indentation:
```
IO - inputCore.executeGesture (10:49:06.965) - winInputHook (15972):
Input: kb(desktop):downArrow
DEBUG - editableText.EditableText._hasCaretMoved (10:49:07.012) - MainThread (1912):
Caret move detected using event. Elapsed: 0 ms
IO - speech.speech.speak (10:49:07.012) - MainThread (1912):
Speaking [BeepCommand(233.08188075904496, 80, left=50, right=50), '- description is up to date\r\n']
IO - tones.beep (10:49:07.032) - MainThread (1912):
Beep at pitch 233.08188075904496, for 80 ms, left volume 50, right volume 50
ERROR - speech.manager.SpeechManager._handleIndex (10:49:07.032) - MainThread (1912):
Error running speech callback
  File "speech\manager.pyc", line 681, in _handleIndex
  File "speech\commands.pyc", line 344, in run
  File "tones.pyc", line 59, in beep
  File "nvwave.pyc", line 491, in stop
  File "nvwave.pyc", line 453, in _idleUnbuffered
  File "nvwave.pyc", line 408, in sync
  File "nvwave.pyc", line 111, in _winmm_errcheck
OSError: [Errno 5] Le handle du périphérique spécifié n’est pas valide.
```

Thus I recommend closing #12620 only if you can confirm that this PR also solves the case when no audio device is changed.

feerrenrut:
I haven't been able to reproduce the exact situation (probably due to the audio devices I have available). Judging from the stack, in the log error, this case should be handled correctly.

I'm a bit confused about why you would get an error "The device handle is invalid" if the audio device is not changed in some way, but I think that is irrelevant to NVDA. If the OS claims the handle is invalid, nvWave needs to reinitialize, and the error should not escape nvWave.

Granted, there may be situations where we may want to ensure that an audio chunk has been heard by the user. I think there would be a lot of work to try to ensure that. Even if the device was reinitialized and chunk sent we can't be sure the new device is appropriate (volume up, actually ready etc).

CyrilleB79:
OK so I guess we can close the issue with this PR and see if we notice some problems.
Cc @lukaszgo1 who also commented in #12620 .

If I understand correctly, we just need to be aware that the error beep may not be heard in the future if such issue occurs.

lukaszgo1:
I agree with #12620 being closed by this PR since we can always reopen. It would still be good to understand why "handle is invalid" error occurs even if the audio device is not changed though. Also from reports I received it seems this is a new behavior though it is hard to say  if caused by NVDA update, new builds of Windows or perhaps affected users simply changed their hardware. cc @jakubl7545 Who reported this problem to me in the first place.

source/nvwave.py
```diff
@@ -640,3 +663,8 @@ def _cleanup():
 	global fileWavePlayer, fileWavePlayerThread
 	fileWavePlayer = None
 	fileWavePlayerThread = None
+
+
+@property
```
lukaszgo1, line 668: I believe this is a mistake. Modules cannot have properties.

source/logHandler.py
```diff
@@ -106,9 +106,13 @@ def getCodePath(f):
 def shouldPlayErrorSound() -> bool:
 	"""Indicates if an error sound should be played when an error is logged.
 	"""
-	
+	import nvwave
+	if not not nvwave.isInError:
```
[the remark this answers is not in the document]
feerrenrut: Weird, not sure how that happened.

source/nvwave.py
feerrenrut: Yes, I think a result of moving code out of the class. I'll fix this.

seanbudd (approved):

PR #12825, opened 2021-09-09
Normalize review by collapsing before expanding

### LINKED ISSUE #12808: Some UIA text ranges have no apparent bottom
### Steps to reproduce:
In various modern UIA text providers (notepad with UIA forceably enabled with `nav.appModule.isGoodUIAWindow = lambda hwnd: True` in console), an upcoming/private build of conhost, and Word/Edgium with UIA enabled):

1. Navigate to the last line of text with the review cursor.
2. Attempt to move to the next line of text.

### Actual behavior:
The review cursor is moved to the end of the current line, and "bottom" is not announced. Technical: `UIATextInfo.move` (`IUIAutomationTextRange::move`) returns 1.

### Expected behavior:
"bottom" is announced.

### Context:
It seems that, per Microsoft Word's behaviour (other UIA providers are copying this): when moving a degenerate (collapsed) range past the end of the document, the correct UIA behaviour is to permit the move and move to the end of the requested unit, whereas expanded ranges should not permit the move when attempting to go past the end. Is this correct UIA behaviour (the docs seem unclear) and therefore an NVDA bug, or are providers incorrectly implementing this (so I should raise it upstream)? Cc @csauls, @michaelDCurran and @leonardder.

codeofdusk:
### Link to issue number:
Closes #12808.

### Summary of the issue:
Due to differences in behaviour between UIA and MSAA text ranges, some UIA text ranges never report "bottom" when moving past the last line of text.

In review, normalize text ranges by collapsing to start before expanding by the unit being reviewed (i.e. move an expanded range, not a collapsed one).

### Testing strategy:
Tested Microsoft Word with UIA enabled, an upcoming UIA Windows Console, and Notepad (both MSAA and UIA proxied MSAA) by moving through multiline text. Verified that "bottom" is properly reported when moving by line in all cases.

I'm not 100% set on this approach, mostly because I don't fully understand this code. This may introduce additional bugs.

I've opened this PR as a possible fix and to start a discussion on resolving this issue as it will continue affecting word (even after UIA) and will soon affect consoles.
### Change log entries:
== Bug Fixes ==
- When reviewing text in Microsoft Word accessed via UI Automation, "bottom" is now reported when moving beyond the last line of text. (#12808)

### Code Review Checklist:

- [ ] Unit tests.
- [x] Manual testing.
- [ ] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -1285,8 +1285,8 @@ class GlobalCommands(ScriptableObject):
 	)
 	def script_review_previousLine(self,gesture):
 		info=api.getReviewPosition().copy()
-		info.expand(textInfos.UNIT_LINE)
 		info.collapse()
+		info.expand(textInfos.UNIT_LINE)
 		res=info.move(textInfos.UNIT_LINE,-1)
 		if res==0:
 			# Translators: a message reported when review cursor is at the top line of the current navigator object.
@@ -1326,8 +1326,8 @@ class GlobalCommands(ScriptableObject):
 	)
 	def script_review_nextLine(self,gesture):
 		info=api.getReviewPosition().copy()
-		info.expand(textInfos.UNIT_LINE)
 		info.collapse()
+		info.expand(textInfos.UNIT_LINE)
 		res=info.move(textInfos.UNIT_LINE,1)
 		if res==0:
 			# Translators: a message reported when review cursor is at the bottom line of the current navigator object.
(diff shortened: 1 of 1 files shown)


leonardder:
Note that this changes behavior in two ways:

1. It collapses before movement
2. It no longer collapses after movement. I think that's still necessary to ensure the cursor is at the start of the new unit.

Shouldn't we still collapse after movement? For UIA and OffsetsTextInfo, I believe after movement, the range is still collapsed, but I don't think that's the case for every single textInfo.

Also note that it's not only IAccessible and UIA we have to consider. There's DisplayModelTextInfo, Java Access Bridge, Word object model, etc. The default move method on OffsetsTextInfo also collapses beforehand when no endpoint is defined.

I think the following questions should be answered for every TextInfo implementation in core before continuing:

1. Does it collapse before a move automatically (OffsetsTextInfo) or does it need an additional collapse (UIA)?
2. Does it collapse after a move automatically (OffsetsTextInfo) or not, or is the collapsed state dependent on the state before movement?

We first need to uniformize how move behaves before changing any review command logic.

codeofdusk:
Closing in favour of a better approach.

PR #12827, opened 2021-09-09
Restart nvda_dmp on exceptions

codeofdusk:
### Link to issue number:
None

### Summary of the issue:
In some situations, nvda_dmp stops functioning. NVDA falls back to Difflib in these cases, so the console continues to function, but frequent error sounds are played and Difflib is forceably used until NVDA is restarted.

When nvda_dmp fails to diff, restart it.

### Testing strategy:
Added code to randomly cause exceptions to the nvda_dmp flow and observed that diffing remained reliable.

None.

### Change log entries:
None.

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/diffHandler.py
+++ b/source/diffHandler.py
@@ -98,6 +98,7 @@ class DiffMatchPatch(DiffAlgo):
 				]
 		except Exception:
 			log.exception("Exception in DMP, falling back to difflib")
+			self._terminate()
 			return Difflib().diff(newText, oldText)
 
 	def _terminate(self):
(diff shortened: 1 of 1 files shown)


seanbudd (approved):

CyrilleB79:
Just a question: 
If we make nvda_dmp restart automatically on error, is there a risk that we do not capture errors in nvda_dmp anymore?
From a user point of view, if nvda_dmp restarts, quite everything is OK and the error may be seen fleetingly.

codeofdusk:
Before this PR, an error would be written to the log and Difflib would be used seemlessly to the user.

CyrilleB79:
Yes, actually you wrote:
> In some situations, nvda_dmp stops functioning. 

The question are:
1. why does it stop functioning?
2. can we do something to avoid that nvda_dmp stops functioning?
IMO the answer to these 2 questions should appear in the PR description to understand how you have investigated the issue.

If the answer to the question is Yes, this PR hides the original rather than resolving it.

PR #7843, opened 2017-12-09
Braille input: emulate keys outside text content and support combining input with modifiers

LeonarddeR:
### Link to issue number:
Follow up of #6213/#7503
Fixes #7306.

### Summary of the issue
Although the braille input system has been vastly improved due to support of contracted input, combining braille input was not yet supported. Also, even though letter quick navigation was supported in the forward direction, navigating backwards did not work as well.

### Description of how this pull request fixes the issue:
This pr addresses the following:

1. No longer always send unicode characters to the system, but in the case of non-editable text, emulate a gesture. See https://github.com/nvaccess/nvda/issues/7306#issuecomment-316555321 . This means that pressing dot1257 in US English 8 dot, which translates to capital H, results in shift+h being sent to the system, which will go to the previous heading.
2. In the case of key emulation, make sure we only emulate one key stroke at once. This is different from sending unicode chars, which allowed sending whole words at once.
3. The point above means that we had to deal with contracted braille in non-editable cases. It turned out that the liblouis partial trans mode works perfectly for this. I tested with both Unified english braille code grade 2 and English US grade 2 that at least all the alphanumeric characters work. They do not translat to their contracted equivalents, so this seems to be the perfect method to use a contracted table in uncontracted mode .
4. Also discussed in #7306 is the possibility to use braille input together with modifiers. As noted by @jcsteh in https://github.com/nvaccess/nvda/issues/7306#issuecomment-319234741 , the implementation for this had to be quite different from #6213, al though I have been able to integrate them as much as possible.

	A. New, unbound scripts have been added to toggle control, shift, alt, windows and NVDA. Here, toggling means that the state of these keys will be enabled or disabled within the scope of braille input. For example, when you assign control to space+dot7 and you press space+dot7, than dot 125, control+h will be emulated. This also works for contracted tables due to partial translation mode.
	B. The toggable modifier scripts integrate in the modifier system that already existed (#6213/#7503), For example, when key1 is bound to toggle control, key 2 to kb:shift and key3 to kb:down arrow, you can press and release key 1 to enable control, than key2+key3 to emulate control+shift+down arrow.

### Testing performed:
* Tested Unified English braille grade 2 and English US grade 2 in non-editable text cases, no contracted translation did occur and thus the implementation worked as expected.
* Tested assigning modifier toggle scripts and combined them with braille input in both editable and non-editable cases.
* Tested combining modifier toggle scripts with emulate scripts, as explained above.
[...]

PR #12850, opened 2021-09-15
Ensure that NVDA no longer checks validity of input for non contracted braille tables when in text controls

### LINKED ISSUE #12667: Braille input with NVDA isn't always correct

lukaszgo1:
### Link to issue number:
Fixes #12667

### Summary of the issue:
When entering text from a braille display NVDA reports that it is unsupported even though entered combination is valid. This check has been introduced in #7843 to prevent a situation in which someone uses contracted table outside text content and enters a combination which cannot be converted to a  single character (a good example is entering dots 356 using English UK Grade 2 which is a contraction for "was"). The check however was too broad and didn't take into account that for literary / computer braille tables single  combination of dots can either always be converted to a single character, or if not it does not result in multiple characters i.e.  number sign.
Now the check applies only when the table in use is contracted.
### Testing strategy:
- With English UK Grade 2 entered dots 356 in the text editor - ensured that after pressing space dots are translated to was.
- Repeated this test in the web browser - made sure that NVDA reports unsupported input.
- With the same table as above entered dots 125 - ensured that in the text editor the combination was back-translated to 'have' whereas in the web browser NVDA navigated to the next heading.
- Repeated tests from #12667 - made sure that no 'unsupported input' was reported
- With Polish Grade 1 table entered dots 3456-12-46-1 - ensured that this  combination was properly back-translated to 2A
- Got confirmation on #12667 from @Futyn-Maker that the issue is fixed for them.

- `brailleInput` is in dire need of unit tests so that future work in this area is safer - I've created #12869 to track this
- According to @aaclause  there are some problems with input for literary tables - they haven't provided more details as of yet other than the fact that problem has not been introduced by this pr.

### Change log entries:
Bug fixes
- Input with literary braille tables should behave more reliably when in edit fields.

### Code Review Checklist:

- [X] Unit tests.
- [X] Manual testing.
- [X] API is compatible with existing addons.
- [X] User Documentation.
- [X] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/brailleInput.py
+++ b/source/brailleInput.py
@@ -1,9 +1,7 @@
-# -*- coding: UTF-8 -*-
-# brailleInput.py
 # A part of NonVisual Desktop Access (NVDA)
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
-# Copyright (C) 2012-2019 NV Access Limited, Rui Batista, Babbage B.V.
+# Copyright (C) 2012-2021 NV Access Limited, Rui Batista, Babbage B.V.
 
 import os.path
 import time
@@ -253,7 +251,7 @@ class BrailleInputHandler(AutoPropertyObject):
 			if self._translate(endWord):
 				if not endWord:
 					self.cellsWithText.add(pos)
-			elif self.bufferText and not self.useContractedForCurrentFocus:
+			elif self.bufferText and not self.useContractedForCurrentFocus and self._table.contracted:
 				# Translators: Reported when translation didn't succeed due to unsupported input.
 				speech.speakMessage(_("Unsupported input"))
 				self.flushBuffer()
--- a/user_docs/en/changes.t2t
+++ b/user_docs/en/changes.t2t
@@ -50,6 +50,7 @@ If you need this functionality please assign a gesture to the appropriate script
 - A rare error when changing audio devices has been fixed. (#12620)
 - Routing keys on Braille displays supported by the HID Braille driver are no longer reversed. (#12860)
 - Ensure the Windows 11 system tray calendar reports the day of the week in full. (#12757)
+- Input with literary braille tables should behave more reliably when in edit fields. (#12667)
 -
 
 


lukaszgo1:
@leonardder  Even though this is still a draft  I'll appreciate a look from you if time permits since you've introduced this code in the first place and you're much more familiar with the brailleInput code.

leonardder:
I think we really need to unit test this somehow. There are simply too much edge cases.

aaclause:
@lukaszgo1 I had made the same patch a short time ago for testing. It seems to me that it broke the good behavior of grade 1 tables.
Also, some input with these tables don't work. For example, with "en-us-g1.ctb" (English (U.S.) grade 1) with the input ⠼⠉⠠⡙.
Sorry, i only read the diff, not tested the build for now.

lukaszgo1:
> 
> 
> @lukaszgo1 I had made the same patch a short time ago for testing. It seems to me that it broke the good behavior of grade 1 tables.

Would you be able to provide an example of input which worked before this  patch and now is broken?

> Also, some inputs with these tables don't work. For example, with "en-us-g1.ctb" (English (U.S.) grade 1) with the input ⠼⠉⠠⠙.

With the code from this PR entering the above combination results in 3D which is I believe an expected input.

aaclause:
@lukaszgo1 Please forget my previous comment. Indeed, your PR fixes the issue for contacted tables. Just that the issue still remains with literary tables, but no regression introduced. Sorry for the noise.

lukaszgo1:
> 
> 
> @lukaszgo1 Please forget my previous comment. Indeed, your PR fixes the issue for contacted tables. Just that the issue still remains with literary tables, but no regression introduced. Sorry for the noise.



Would you be able to give an STR for an issue with literary tables so that it can be fixed and / or unit test created for this?

lukaszgo1:
@aaclause  Have you seen my latest comment? It would be nice to have a STR reproducing your issue as that is the only thing getting this blocked from being converted from a draft.

lukaszgo1:
@leonardder While I agree with your opinion that unit tests for `brailleInput` are very important to have I'm afraid I would not have time to work on them before the release and this PR improves the known bad cases,. I've created #12869  and would be happy to work on this subsequently. Would  you be okay with  this being reviewed  as is and getting unit test at a later point

leonardder:
I guess NV Access will have to decide about that.

seanbudd (approved):
Thanks @lukaszgo1

PR #12857, opened 2021-09-16
Treat tables with 1 col or 1 row as layout tables in Outlook message viewer

### LINKED ISSUE #11430: Suppress layout tables when reading messages in Outlook with UIA for Word enabled

michaelDCurran:
### Link to issue number:
Fixes #11430
Replaces #12820 

### Summary of the issue:
Microsoft Outlook uses a Microsoft word document control to present emails when both reading and composing. NVDA is in the process of switching to using UI Automation to access Microsoft Word document controls due to major performance advantages, and also due to Microsoft no longer maintaining the Office object model.
However, in the UI automation implementation exposed by MS Word document controls, it is impossible to identify the difference between a data table and a layout table. Thus, NVDa reports all layout tables when reading, which is extremely annoying, as the majority of HTML emails use many nested layout tables.

Tables are now classed as layout tables in MS Word document controls if the app is Outlook, and the table is read-only (Outlook is in message reading view) and the table only has either 1 column or 1 row.

### Testing strategy:
With use UIA in MS Word controls enabled in NVDA's advanced settings:
• Create a table in Microsoft Word and ensure that NVDA still reports it (number of rows / columns, cell's row and column number etc).
• Create a table in Microsoft Outlook composition view and ensure that NVDA still reports it (number of rows / columns, cell's row and column number etc).
• Open an HTML email in Outlook that has layout tables, and ensure that the table is not reported when moving into and through the table's content with the arrow keys.

Treating tables with 1 column or 1 row as layout tables catches a good majority of layout tables. However, there will still be a few that will not be (they might have 2 columns say). However, the choice to suppress all tables in #12820 was not popular.

### Change log entries:
Bug fixes
* When reading emails in Outlook and NVDA is accessing the message with UI Automation, certain tables are now marked as layout tables, which means they will no longer be reported by default. (#11430)

### Code Review Checklist:

- [x] Unit tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


michaelDCurran:
@leonardder  Your testing of this pr would be most welcome.

codeofdusk:
> certain tables

Why not say "tables containing only one row or column"?

seanbudd (approved):

user_docs/en/changes.t2t
```diff
@@ -37,13 +37,17 @@ If you need this functionality please assign a gesture to the appropriate script
 - NVDA no longer treats the value of UIA sliders as always percentage based.
 - Reporting the location of a cell in Microsoft Excel when accessed via UI Automation again works correctly on Windows 11. (#12782)
 - NVDA no longer sets invalid Python locales. (#12753)
+- When reading emails in Outlook  via UI Automation, reporting of tables is now suppressed by default. (#11430)
```
leonardder, line 40: I think this one should go away in favour of the latter.

LeonarddeR (approved):
Works great for me. Data tables are properly read whereas most layout tables I've seen are properly silenced. I think I've seen one table that was incorrectly treated as data table compared to at least 50 where all was ok.

PR #12876, opened 2021-09-24
Empty checkable list: trial not to report checkbox.

CyrilleB79:
### Link to issue number:
None

### Summary of the issue:

In an an empty gui.nvdaControls.CustomCheckListBox, the focus is reported as checkbox even if there is none.
Also, when reporting the focus after having removed all the items of the checkable list, the following error occurs:
```
IO - inputCore.executeGesture (21:33:29.011) - winInputHook (20908):
Input: kb(desktop):NVDA+tab
ERROR - unhandled exception (21:33:29.033) - MainThread (27252):
  File "gui\nvdaControls.pyc", line 100, in GetState
wx._core.wxAssertionError: C++ assertion ""IsValid(uiIndex)"" failed at ..\..\src\msw\checklst.cpp(275) in wxCheckListBox::IsChecked(): bad wxCheckListBox index
```

To reproduce the issue, use the following global plugin in the scratchpad.
[CustomCheckListBox.py.txt](https://github.com/nvaccess/nvda/files/7239483/CustomCheckListBox.py.txt)

To reproduce the error:
* press NVDA+A
* Tab to the "Supprimer jour" button
* Press 7 times to remove all the items in the first list.
* Press NVDA+Tab to have the current focus reported.

For now, this PR is an unsuccessful attempt to fix the issue.
I would have expected that `super().GetRole(childId)` and `super().GetState(childId)` would have been successful in the case where the list contains an item with this index. But this does not seem to be the case and in any case, `wx.ACC_OK` is not returned even when the list contains items. I would be curious to know why.
In any case, I would need to find another way to filter the case where GetRole/GetState is called with a childId > 0 whereas the list is empty.

@leonardder would you accept to review the few lines of this PR since you have worked on this topic? Surely, you could have done the PR yourself. But I thought that trying to fix it myself with your help would allow me to understand how this part of code works. Thanks.

### Change log entries:
New features
Changes
Bug fixes
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/nvdaControls.py
+++ b/source/gui/nvdaControls.py
@@ -89,12 +89,14 @@ class ListCtrlAccessible(wx.Accessible):
 	"""WX Accessible implementation for checkable lists which aren't fully accessible."""
 
 	def GetRole(self, childId):
-		if childId == winUser.CHILDID_SELF:
+		(status, role) = super().GetRole(childId)
+		if childId == winUser.CHILDID_SELF or status != wx.ACC_OK:
 			return super().GetRole(childId)
 		return (wx.ACC_OK, wx.ROLE_SYSTEM_CHECKBUTTON)
 
 	def GetState(self, childId):
-		if childId == winUser.CHILDID_SELF:
+		(status, state) = super().GetState(childId)
+		if childId == winUser.CHILDID_SELF or status != wx.ACC_OK:
 			return super().GetState(childId)
 		states = wx.ACC_STATE_SYSTEM_SELECTABLE | wx.ACC_STATE_SYSTEM_FOCUSABLE
 		if self.Window.IsChecked(childId - 1):
@@ -103,7 +105,7 @@ class ListCtrlAccessible(wx.Accessible):
 			# wx doesn't seem to  have a method to check whether a list item is focused.
 			# Therefore, assume that a selected item is focused,which is the case in single select list boxes.
 			states |= wx.ACC_STATE_SYSTEM_SELECTED | wx.ACC_STATE_SYSTEM_FOCUSED
-		return (wx.ACC_OK, states)
+			return (wx.ACC_OK, states)
 
 
 class CustomCheckListBox(wx.CheckListBox):


leonardder:
I'm confused. How is this settingsDialogs.py.txt supposed to behave like a global plugin?

CyrilleB79:
Sorry, I have attached the wrong file. Please put this review on hold until I have modified the attachment; I have not it available for now (it is on another computer).

CyrilleB79:
@leonardder I have updated the initial description replacing the wrong file with the good one. Should be OK now. Sorry again for confusion.

leonardder:
Thanks for that.

It looks to me that there's a bug in the wx control that exposes a single child in the list, even when it is empty. You can see that as follows.

1. Go to the wx checklistbox
2. Remove all the days with the delete day button
3. Notice that NVDA says: `This is a wx.CheckListBox  list unknown`

CyrilleB79:
Yes, the `wx.checklistbox` exposes a degenerated child when it is empty. From devInfo, I can see that IAccessible does not provide properties for it (exception: (-2147024809, 'Paramètre incorrect.', (None, None, None, 0, None)). And thus NVDA reports its role as unknown.
But visually, the focus on this empty first degenerated element can be seen (even if NVDA does not detect it with focus highlight).
On the contrary, in the `gui.nvdaControls.AutoWidthColumnCheckListCtrl`, there is no such visual indication of a degenerated first child when it is empty.

I do not intend to change wx control behaviour. I just wanted that NVDA does not report checkbox for the degenerated first child of an empty `gui.nvdaControls.CustomCheckListBox`. I would rather have unknown type control reported for this degenerated first child as it is done in `wx.CheckListBox`. And that's what I was trying to do with the code of this PR.

But I do not understand why calling `super.GetRole()` and `super().GetState()` is not successful for normal items of the list (i.e. when the list is not empty.

Would you be able to explain it? Obviously, I have missed something.
Thanks.

leonardder:
> But I do not understand why calling `super.GetRole()` and `super().GetState()` is not successful for normal items of the list (i.e. when the list is not empty.
> 
> Would you be able to explain it? Obviously, I have missed something.

I think I know why this happens. When super.GetRole()` and `super().GetState()` are called, they might return their default, which is probably something like not implemented. I think this causes any AT to fall back to the default accessible implementation of the control, which is part of Windows itself.

CyrilleB79:
I am closing this PR because I have not found any solution with this trial and because I do not intend to work on it from now on.
If anyone is interested or has new suggestion, let me know and I can reopen if needed.

PR #12888, opened 2021-09-30
Don't braille or speak duplicate name / content / description

### LINKED ISSUE #12750: Excessive Braille text in web navigation with latest alpha versions
### Steps to reproduce:

1. Go to, for example, to [this wiki page](https://en.wikipedia.org/wiki/English_language);
2. search "Language family" in the page;
3. go down and observe the link: its text content is "Indo-European", its title "Indo-European languages".

### Actual behavior:

Braille shows title and text content, so:
Indo-European languages Indo-European
with cursor (correctly) at starting position of link text content.

### Expected behavior:

Braille should show only link text content, as usual.

Installed.

#### NVDA version:

NVDA version alpha-23598,a4524704, but I have noticed this issue some alphas ago (after 2021.2beta1 was released).

#### Windows version:

Windows 10 20H2 (build SO 19042.1165).

Mozilla Firefox 91.0.1, Google Chrome 92.0.4515.159.

Yes.

NVDA 2021.2beta1, no problems.

Yes.

Yes.

feerrenrut:
### Link to issue number:
Related to #12750

### Summary of the issue:
When a link has a `title` attribute that matches the content, braille (and speech) is duplicated.

An example:
```html
<a href="#" title="apple" style="display:block">apple</a>
```
Two links URL encoded:
```
data:text/html,%3Ca href%3D"%23" title%3D"apple" style%3D"display%3Ablock"%3Eapple%3C%2Fa%3E%3Ca href%3D"%23" title%3D"banana" aria-label%3D"banana" style%3D"display%3Ablock"%3Econtents%3C%2Fa%3E
```

Tests have been added for speech, to show current behavior (duplicated name/content/description)
The duplicate speech/braille is fixed:
- In focus mode, ensuring that description is not reported if it matches name
- In browse mode, when filling the virtual buffer identifying when description matches the node contents and adding an attribute.

### Testing strategy:
- System tests added for speech:
  - browse mode navigation with tab
  - browse mode navigation with down arrow
  - focus mode navigation with tab

Manual testing using the braille viewer with the same sample.
- Further manual testing from more experienced braille users would be appreciated.

None

### Change log entries:

Bug fixes
``` 
- Fixed duplicate braille and speech when 'description' matches 'content' or 'name'. #12888
```

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
+++ b/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
@@ -541,7 +541,7 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 	if(pacc->get_accName(varChild,&name)!=S_OK)
 		name=NULL;
 
-	std::optional<wstring> description{ getAccDescription(pacc, varChild) }; // todo: fix usages to use optional
+	std::optional<wstring> description{ getAccDescription(pacc, varChild) };
 	if (description.has_value()) {
 		parentNode->addAttribute(L"description", description.value());
 	}
--- a/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
+++ b/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
@@ -273,6 +273,15 @@ bool hasAriaHiddenAttribute(const map<wstring,wstring>& IA2AttribsMap){
 	return (IA2AttribsMapIt != IA2AttribsMap.end() && IA2AttribsMapIt->second == L"true");
 }
 
+std::optional<wstring> getAccDescription(IAccessible2* pacc, VARIANT childID) {
+	std::optional<wstring> desc;
+	CComBSTR rawDesc;
+	if (S_OK == pacc->get_accDescription(childID, &rawDesc)) {
+		desc = rawDesc;
+	}
+	return desc;
+}
+
 /**
  * Get the selected item or the first item if no item is selected.
  */
@@ -532,12 +541,9 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 	if(pacc->get_accName(varChild,&name)!=S_OK)
 		name=NULL;
 
-	wstring description;
-	BSTR rawDesc=NULL;
-	if(pacc->get_accDescription(varChild,&rawDesc)==S_OK) {
-		description=rawDesc;
-		parentNode->addAttribute(L"description",description);
-		SysFreeString(rawDesc);
+	std::optional<wstring> description{ getAccDescription(pacc, varChild) }; // todo: fix usages to use optional
+	if (description.has_value()) {
+		parentNode->addAttribute(L"description", description.value());
 	}
 
 	wstring locale;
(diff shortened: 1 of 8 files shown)


lukaszgo1:
@feerrenrut As soon as I start Firefox with the build from this PR and load any browse mode document entire Firefox crashes. [sample crash report](https://crash-stats.mozilla.org/report/index/78b5176b-dfc7-4b93-9076-d22c90210930).

ABuffEr:
Sadly, I confirm Firefox crash, starting or bring it to foreground.

feerrenrut:
Yes, I can reproduce this. Thanks for testing, I'll look into it.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -273,6 +273,16 @@ bool hasAriaHiddenAttribute(const map<wstring,wstring>& IA2AttribsMap){
 	return (IA2AttribsMapIt != IA2AttribsMap.end() && IA2AttribsMapIt->second == L"true");
 }
 
+std::optional<wstring> getAccDescription(IAccessible2* pacc, VARIANT childID) {
+	std::optional<wstring> desc;
+	CComBSTR rawDesc;
+	if (pacc->get_accDescription(childID, &rawDesc) == S_OK) {
+		desc = rawDesc;
+		SysFreeString(rawDesc);
```
feerrenrut, line 281: This is the problem, the string is being freed twice. It is now a `CComBSTR` which will free on destruction.

feerrenrut:
The Firefox crash should now be resolved, please test the [latest PR build to confirm].(https://ci.appveyor.com/api/buildjobs/n8wrf4lcrm3hqn4v/artifacts/output%2Fnvda_snapshot_pr12888-23854%2C2ea82fef.exe)

lukaszgo1:
I can confirm that Firefox crash is fixed on my end. Also I would like to thank @feerrenrut  for this fix - not announcing description when it equals the name has been requested for years and it makes various programs (even outside of the web browser) so more pleasant to use.
Unfortunately this PR does not improve #12750 in any way since the issue described there is not caused by title being equal to the link name. Could you please open the sample from the issue in Firefox and navigate with braille viewer? You would immediately notice that title attribute is different than the link name, it should not be shown in braille at all and what is even worse content of the title attribute is shown before the content of the link.
The markup for this particular case is:
`<a href="/wiki/Indo-European_languages" title="Indo-European languages">Indo-European</a>`

ABuffEr:
Thanks @lukaszgo1  for explanation, I confirm all on my part.
If it can be useful, I additionally noticed that moving focus on a link with the issue, then changing and go back to tab, Braille shows the correct text until I move by line again.

feerrenrut:
Oh thanks @lukaszgo1, I seem to have mixed up my test cases while working on this. 

There are a few aspects to this:
- Description is becoming much more useful due to [Aria Annotations](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Annotations).
- We aim to exclude description with it's source is the `title` attribute, this isn't yet available in Firefox. We suspect this accounts for the majority of noisy cases.
- Historically the option "Object presentation: Report Object Descriptions" (default: true) has been limited to focus mode. In #12500 this was changed to report descriptions always. Historically, the word "object" in the settings category, and the name of this option was supposed to imply "focus mode specific behavior".

I will look at reverting the change in behavior for "Report Object Descriptions". I'll also see if I can add further information to the settings panel / user guide to explain the significance of "object".

Although this PR does not fix the specifically reported but, it does seem to be a general improvement. I suggest that we merge this on it's own (after updating the description / test cases). I'll continue the work in a new PR.

tests/system/robot/chromeTests.py
```diff
@@ -649,3 +649,89 @@ def test_i10840():
 		nextActualSpeech,
 		"column 2  items"
 	)
+
+
+def test_preventDuplicateSpeechFromDescription_browse_downArrow():
```
michaelDCurran, line 654: NVDA has never reported description for title when reading in browse mode with up and down arrow, whether it is duplicate or not. Thus this test might be a bit misleading. I'd personally remove it. But if you don't want to do this, then perhaps add another test before it that tests that a non-duplicate title is also not read.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -532,12 +541,9 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 	if(pacc->get_accName(varChild,&name)!=S_OK)
 		name=NULL;
 
-	wstring description;
-	BSTR rawDesc=NULL;
-	if(pacc->get_accDescription(varChild,&rawDesc)==S_OK) {
-		description=rawDesc;
-		parentNode->addAttribute(L"description",description);
-		SysFreeString(rawDesc);
+	std::optional<wstring> description{ getAccDescription(pacc, varChild) }; // todo: fix usages to use optional
```
michaelDCurran, line 544: Is this comment still valid? What do you mean exactly?

michaelDCurran (changes requested):

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: No, sorry. It was just a reminder to myself. I'll remove.

michaelDCurran (changes requested):
I would much prefer it if the commit to "tidy up the tests" was separated into a separate pr. This changes other tests than aria-description and creates a 300+ line diff. 
But other than that everything before looks good.

feerrenrut:
I have no idea where the 300 is coming from. My local diff shows 26 lines with a + and 18 with a -

michaelDCurran (approved):

Issue #12919, opened 2021-10-08
With latest Master, documents are nomore accessible under LibreOffice Writer


MichelSuch:
### Steps to reproduce:
Open LibreOffice Writer and try to write some text.
### Actual behavior:
NVDA does not see nor the texte neither the cursor position.
### Expected behavior:
We should be able to review the document.
#### NVDA installed/portable/running from source:
NVDA installed
#### NVDA version:
alpha-23903,fb4466ee
#### Windows version:
Windows 10 21H2 64 bits
#### Name and version of other software in use when reproducing the issue:
LibreOffice 7.2.1 64 bits

lukaszgo1:
Confirmed with latest Alpha.
When moving with up and down arrows 'blank' is reported. With left and right error sound plays, and the following is in the log:
```
error executing script: <bound method EditableText.script_caret_moveByCharacter of <compoundDocuments.CompoundDocument object at 0x073B3630>> with gesture 'left arrow'
Traceback (most recent call last):
  File "scriptHandler.pyc", line 208, in executeScript
  File "editableText.pyc", line 228, in script_caret_moveByCharacter
  File "editableText.pyc", line 163, in _caretMovementScriptHelper
  File "editableText.pyc", line 149, in _caretScriptPostMovedHelper
  File "speech\speech.pyc", line 1138, in speakTextInfo
  File "speech\types.pyc", line 42, in __iter__
  File "speech\speech.pyc", line 1335, in getTextInfoSpeech
AttributeError: 'NoneType' object has no attribute 'isspace'

```

This does not occur with 2021.2.

MichelSuch:
Is there a chance this could be repaired for 2021.3?

feerrenrut:
@MichelSuch we are looking into it. The list of issues that are blocking the release can be found by looking at [the 2021.3 milestone](https://github.com/nvaccess/nvda/milestone/57), currently it is just this issue.

PR #12969, opened 2021-10-21
Add Typing to getTextWithFields

feerrenrut:
### Link to issue number:
Noticed during investigation of #12919

### Summary of the issue:
`getTextWithFields` is expected to return `TextWithFieldsT` which is a  `List[Union[str, FieldCommand]]`
Many of the overrides of `getTextWithFields` didn't specify the type, thus did not benefit from type hinting, some of them 
return incorrect types.

- add type info to all definitions of `getTextWithFields`
- Ensure that where the helper function `_getFieldsInRange` is used, it must return `TextWithFieldsT`
- Make `_getFieldsInRange` easier to read, break out the normalization of commands.
- Handle any None values to return TextInfo.TextWithFieldsT

Note `_getFieldsInRange` was in one case returning an empty string, not a list with an empty string, this has been resolved.

### Testing strategy:
Test early on alpha.

The `_getFieldsInRange` empty string may have been treated as an empty iterable, in which case the more correct modification would be to return an empty list.

### Change log entries:
None

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/virtualBuffers/__init__.py
+++ b/source/virtualBuffers/__init__.py
@@ -278,12 +278,12 @@ class VirtualBufferTextInfo(browseMode.BrowseModeDocumentTextInfo,textInfos.offs
 		if not text:
 			return [""]
 		commandList = XMLFormatting.XMLTextParser().parse(text)
-		commandList = list([
+		commandList = [
 			self._normalizeCommand(command)
 			for command in commandList
 			# drop None to convert from XMLFormatting.CommandListT to textInfos.TextInfo.TextWithFieldsT
 			if command is not None
-		])
+		]
 		return commandList
 
 	def getTextWithFields(self, formatConfig: Optional[Dict] = None) -> textInfos.TextInfo.TextWithFieldsT:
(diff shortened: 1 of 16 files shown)


source/virtualBuffers/__init__.py
```diff
@@ -260,23 +263,30 @@ def _getPlaceholderAttribute(self, attrs, placeholderAttrsKey):
...
-				continue  # no need to normalize str or None
-
-			field = command.field
-			if isinstance(field, textInfos.ControlField):
-				command.field = self._normalizeControlField(field)
-			elif isinstance(field, textInfos.FormatField):
-				command.field = self._normalizeFormatField(field)
+			return [""]
+		commandList = XMLFormatting.XMLTextParser().parse(text)
+		commandList = list([
```
michaelDCurran, line 281: This creates an extra temporary list.
I don't think you need to wrap this in `list()` as you already have a list expression `[x for x in y if x is not None]`.
 Of course if it were parentheses (a generator expression) then you would need to wrap it in `list()`.

michaelDCurran (changes requested):
Other than the extra list, I'm happy with all of this.

michaelDCurran (approved):

PR #12988, opened 2021-10-26
Support the Modern Comments side track pane in MS Word when not accessing MS Word documents via UIA 

### LINKED ISSUE #12982: MS Word: comments inaccessible to NVDA via object model from build 13901
### Background
In build 13901, MS Word introduced "Modern comments" which allows comment threads (I.e. an initial comment plus replies) and the ability to resolve or reopen threads. The new UI to do this is exposed as a new comments side pane which shows the comments and replies in a treeview. This UI is only exposed via UI Automation.
 This UI is supposed to completely replace the older Comments story in a document.
As this new UI is imbedded within the document, it shares the same window handle. If NvDA is using the object model to access the document (I.e. the UIA provider of the document's window is ignored) then NVDA never sees the modern comments UI at all. It seems that MS Word does not ever proxy this to MSAA either.
 Also, if MS Word detects that something is trying to access the comments via the object model in the old way (E.g. calls comment.range) then the Modern comments UI is disabled (hidden) and MS Word ends up in a state where querying the range at the selection of the object model is no longer in the comments story, yet the UI is not shown either. This MS Word bug is known to Microsoft however they have no plans or willingness to fix it. Their only recommended solution is to stop using the object model and just use UIA.

### Steps to reproduce:
Using NVDA 2021.2 with the Use UIA in MS Word documents NVDA setting disabled,
Open MS Word build 13901 or above.
Create a new blank document.
Type some text.
Select 1 or more characters, and choose New Comment from the context menu.

### Actual behavior:
After the context menu closes, NVDA reports the focus is back in the document at the position of the selected text.
Pressing alt+f12 to get to the Comments side pane does not work.
There is no way to get into the Comments story of the document either.

### Expected behavior:
After leaving the context menu, one of the following is expected:
* Focus should be in the comments side pane, allowing the user to type the comment and post or cancel it, or
* Focus should be on the document but the selection should be in the older Comments story, allowing you to type the comment.

Installed
#### NVDA version:
2021.2
#### Windows version:
Windows 11 build 22483. 
Microsoft® Word for Microsoft 365 MSO (Version 2111 Build 16.0.14613.20000) 64-bit 
yes.
Recent builds of NVDA alpha (approaching 2021.3) use UIA by defauitl in MS Word from build 13901 upwards to avoid this issue. However, there are still some other issues relating to UIA support which may block UIA by default in NVDA 2021.3.

Yes.
Yes.

### ideas for a fix
1. `NVDAObjects.window.winword.WordDocumentTextInfo` (I.e. text info for MS Word object model):  Dot fetch text or minipulate the selection in any way if selection.storyType is wdCommentsStory.
2. in UIAHandler.handler.isUIAElement: return true if the element's windowHandle is _WwG but an ancestor of the element has a UIA className of NetUIHWNDElement.

If we were able to successfully fix this issue for NVDA 2021.3, then we could push back UIA by default, and also the new support for math in UIA to 2022.1, ensuring that they get more testing time.

michaelDCurran:
### Link to issue number:
Fixes #12982 

### Summary of the issue:
In build 13901, MS Word introduced "Modern comments" which allows comment threads (I.e. an initial comment plus replies) and the ability to resolve or reopen threads. The new UI to do this is exposed as a new comments side pane which shows the comments and replies in a treeview. This UI is only exposed via UI Automation.
This UI is supposed to completely replace the older Comments story in a document.
As this new UI is imbedded within the document, it shares the same window handle. If NvDA is using the object model to access the document (I.e. the UIA provider of the document's window is ignored) then NVDA never sees the modern comments UI at all. It seems that MS Word does not ever proxy this to MSAA either.
Also, if MS Word detects that something is trying to access the comments via the object model in the old way (E.g. calls comment.range) then the Modern comments UI is disabled (hidden) and MS Word ends up in a state where querying the range at the selection of the object model is no longer in the comments story, yet the UI is not shown either. This MS Word bug is known to Microsoft however they have no plans or willingness to fix it. Their only recommended solution is to stop using the object model and just use UIA.

in `UIAHandler.handler.isUIAElement`: return true if the element's windowHandle is _WwG but an ancestor of the element has a UIA className of NetUIHWNDElement.
I.e. this UIA element is a control of a NetUI container embedded in an MS Word document. E.g. a control in the Modern Comments side track pane.
Also, if there is a UIA focus event for the Word document root, even though we are not classing that window as native UIA, if the previously focused object was a control in a NetUI container embedded in the same Word document, then generate an MSAA focus event for the document, as MS Word will not do this itself.

Note that allowing NVDA to track the focus with UIA for these embedded controls is enough to work around the issue, as this then means that focus never hits the Word document while the Comments side track pane should be focused, thus avoids making object model calls MS Word does not like or expect.

### Testing strategy:
This can only be tested in conjunction with PR #12989  as UIA in MS word should not be used by default.
1. Ensure that the NVDA advanced setting: Use UIA in MS Word document controls when available is turned off.
2. Open a new MS Word document.
3. Ensure the build of MS Word is 13901 or later.
4. Type some text. E.g. "this is a test".
5. Select a word, E.g. "This".
6.  In the context menu, choose "New Comment".
7. Ensure that focus lands in the Modern Comments side track pane ready for you to type the comment.
8. Type a comment, and tab to the Post comment button and press enter.
8. press alt+f12 to move back to the document and ensure that the arrow keys cause the caret to move and NVDA speaks the characters.
9. Again, select the same word you created the comment for.
10. Press alt+f12 to move to the Modern comment side track pane for this comment.

None known.

### Change log entries:
New features
Changes
Bug fixes
* The new Modern Comments side track pane is now supported in MS Word when not accessing the document via UIA. Press alt+f12 to move between the side track pane and the document.
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/_UIAHandler.py
+++ b/source/_UIAHandler.py
@@ -25,6 +25,7 @@ from comtypes import (
 
 import threading
 import time
+import IAccessibleHandler.internalWinEventHandler
 import config
 import api
 import appModuleHandler
@@ -533,6 +534,15 @@ class UIAHandler(COMObject):
 			return
 		self.lastFocusedUIAElement = sender
 		if not self.isNativeUIAElement(sender):
+			# #12982: This element may be the root of an MS Word document
+			# for which we may be refusing to use UIA as its implementation may be incomplete.
+			# However, there are some controls embedded in the MS Word document window
+			# such as the Modern comments side track pain
+			# for which we do have to use UIA.
+			# But, if focus jumps from one of these controls back to the document (E.g. the user presses escape),
+			# we receive no MSAA focus event, only a UIA focus event.
+			# As we are not treating the Word doc as UIA, we need to manually fire an MSAA focus event on the document.
+			self._emitMSAAFocusForWordDocIfNecessary(sender)
 			if _isDebug():
 				log.debug("HandleFocusChangedEvent: Ignoring for non native element")
 			return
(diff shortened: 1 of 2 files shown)


seanbudd (dismissed):

LeonarddeR (changes requested):

feerrenrut (dismissed):
Thanks @michaelDCurran

PR #12989, opened 2021-10-26
Revert UI a in ms word by default

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
We have tried to switch to using UI automation to access MS Word documents by default in NVDA 2021.3. However, there are still several issues remaining with our UIA support. Switching to UIA by default should be held back until these are addressed.
Some of these include:
* Support for math: #12946
 * Reporting of line numbers: Requires UIA custom patterns support\
 
Over the last couple of months our UIA support has had a lot of fixes and improvements, mainly spured on by the apparent need to switch due to NVDA and MS Word not playing well together with the introduction of Modern Comments in MS Word build 13901. However, a work around has been found for #12982 now, making the swich less of a high priority.

### Description of how this pull request fixes the issue:
Reverts #12859 
Reverts #12854 
Reverts #12770 

### Testing strategy:
With a build of MS Word 13901 or highera dn NVDA advanced setting: use UIA to access MS Word documents when available turned off, Open a Word document and ensure that UIA is not being used to access the document.
When the document is focused, the name of the focused object is "Microsoft Word document" and the role is "edit".

### Known issues with pull request:
None known.

### Change log entries:
New features
Changes
Bug fixes
For Developers

### Code Review Checklist:

- [x] Pull Request description:
  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] API is compatible with existing add-ons.
- [ ] Documentation:
  - User Documentation
  - Developer / Technical Documentation
  - Context sensitive help for GUI changes
- [x] UX of all users considered:
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers
  - Localization in other languages / culture than English

PR #12992, opened 2021-10-26
What's new: remove mention of UIA by default in MS Word

michaelDCurran:
In pr #12989 NVDA no longer defaults to using UIA in MS Word build 13901 or higher. 
This pr removes the relevant line from changes.t2t.

Note that the release blurb for 2021.3 did not explicitly mention this feature thus did not need to be updated.


feerrenrut (approved):
Looks good to me. Is `There are lots of improvements when using Office with UIA.` still true?

seanbudd (approved):

PR #13009, opened 2021-10-30
Fix-up of PR 12943 - Avoid os.path.commonPath` when comparing paths of the binaries as it does not work for files on different drives

### LINKED ISSUE #13008: word problems 2013
### Steps to reproduce:
I have a friend's computer with problems.
This machine is with office 2013 and windows 10, I didn't install these 2 software.
I updated on this machine only nvda and the add-ons.
In several parts of the word, nvda doesn't say anything:
When we press alt inside word nvda doesn't announce the focus in the menu, and when we walk with tabe it doesn't read anything either.
This also occurred within outlook.
When we close the document without saving in the word question if we want to save the changes when walking with tabe again, nvda does not announce anything.
When opening the word on the first screen where you can choose the type of document to create nvda again is muted when walking with tabe or arrows.
I've already tested it with disabled add-ons and the problem continued.
I already ran the registry fix tool with and the error continues.
Nvda seems to work well in any other area of windows 10 and other programs. 
### Actual behavior:
nvda mute on some word screens 
### Expected behavior:
nvda read screens normally 
instaled
#### NVDA version:
alfa 24160
#### Windows version:
10 10.0.19043.
office 2013
yes
no
yes
yes
log of the various screens where I noticed the error.
Sorry, there are so many errors that I couldn't selessionar only the point where it matters in the log, but I tried to turn on nvda at the time the error occurs to be relatively smaller. 
[exit word.txt](https://github.com/nvaccess/nvda/files/7445713/exit.word.txt)
[select word.txt](https://github.com/nvaccess/nvda/files/7445714/select.word.txt)
[start word.txt](https://github.com/nvaccess/nvda/files/7445715/start.word.txt)

lukaszgo1:
### Link to issue number:
Fixes #13008

### Summary of the issue:
PR #12943 improved code which retrieves version information for a given executable files placed in system32 under 64-bit versions of Windows. To do so it is necessary to check if path of the given binary starts with path of system32 directory for the current system. The recommended way to compare path's in a case insensitive way was `os.path`commonPath`, however it turns out it does not work when path's are from different drives.
In case of reporter of #13008 they had Office installed in a custom location probably on non system disk.
This PR simplifies the code to use standard `str.casefold` for comparisons.
### Testing strategy:
- Made sure that product name and product version can be retrieved for a binary  placed on non system disk - previously it resulted in an exception.
- Got confirmation from reporter of #13008 that they problem is no longer reproducible.
None known
### Change log entries:
None needed - regression not in a release.
### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/fileUtils.py
+++ b/source/fileUtils.py
@@ -58,8 +58,8 @@ def _suspendWow64RedirectionForFileInfoRetrieval(func):
 		nativeSys32 = shlobj.SHGetKnownFolderPath(shlobj.FolderId.SYSTEM)
 		if (
 			systemUtils.hasSyswow64Dir()
-			# `os.path.commonpath` is necessary to perform case-insensitive comparisons
-			and os.path.commonpath([nativeSys32]) == os.path.commonpath([nativeSys32, filePath])
+			# Path's returned from `appModule.appPath` and `shlobj.SHGetKnownFolderPath` often differ in case
+			and filePath.casefold().startswith(nativeSys32.casefold())
 		):
 			with winKernel.suspendWow64Redirection():
 				return func(filePath, *attributes)


lukaszgo1:
cc @seanbudd  Would you be able to review this?

seanbudd (approved):

PR #13013, opened 2021-11-02
MS Word with UIA: heading quicknav no longer keeps repeating the last heading

### LINKED ISSUE #9540: UIA in Word: When document ends with a heading, the heading is duplicated in quick nav
### Steps to reproduce:
1. Open MS Word
2. Create a single line and make it a heading with ctrl+alt+1/3
3. Enable browse mode
4. Press H multiple times.

### Actual behavior:
NVDA keeps reporting the heading. When opening the elements list, NVDA shows the heading twice.

### Expected behavior:
NVDA reports "no next heading". Only one heading is shown in the elements list.

Installed

#### NVDA version:
alpha-17134

#### Windows version:
Windows 10 1903 uild 18362.86

Word 365 16.0.11601.20144

Yes

No

michaelDCurran:
### Link to issue number:
Fixes #9540

### Summary of the issue:
When accessing MS Word documents with UIA, if the last bit of text is formatted as a heading, pressing quicknav by heading (h) in browse mode keeps repeating that heding. Similarly, The elements list when set to show headings, lists the final heading twice.

This is due to the fact that most UIA providers when on the last unit, move to the collapsed exclusive end if asked to move forward 1 by that unit. Only after that does the move fail.
We could add some specific checks in the UIABrowseMode.UIAHeadingQuicknavIterator function to take this into account, or we can rewrite that algorithm to make use of iterUIARangeByUnit which is used in many other places and avoids that bug all together.
 
Rewrite the UIABrowseMode.UIAHeadingQuicknavIterator function to make use of iterUIARangeByUnit. It is now much more similar to the UIABrowseMode.UIATextAttributeQuicknavIterator function, which we know does not have the repeating bug.
  
### Testing strategy:
* Ensure NVDA's use UIA to access MS Word document controls when available is on
* Open a new blank document in MS Word
* Type `Test` and press enter
* Press alt+shift+rightArrow to start formatting as heading 2 style and type `First heading` and press enter
* Type `Test` and press enter
* Press alt+shift+rightArrow to start formatting as heading 3, type `Second heading` and press enter 
* `Second heading` should be the final line of text in the document, I.e. the heading ends at the end of the document.
* Switch to browse mode with NVDA+space
* Move to the top of the document with control+home
* Press h to jump to the `First heading`
* Press h to jump to the `Second heading`
* Press h and ensure that NVDA reports that there is no next heading.
* Press shift+h to move to the `First heading`
* Press shift+h and ensure NVDA reports that there is no previous heading
* Press NVDA+f7 to open the NVDA elements list, and select Headings from the Type radio buttons
* Ensure that just two headings are listed in the treeview: `First heading` and `Second heading`.

None known.

### Change log entries:
New features
Changes
Bug fixes
* MS word with UIA: heading quick nav in browse mode no longer gets stuck on the final heading of a document, nor is this heading shown twice in the NVDA elements list.

For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


leonardder:
Would it help to do some timings with huge documents with many headings to see how performance is affected?

michaelDCurran:
To test timing:
I Created a new blank document.
I generated a paragraph of auto text by typing `=rand()` and pressing enter.
I then copied and pasted this many times to produce a document of 285 pages.
I typed a heading on the very last line of the document.
I then jumped back to the top of the document, switched to browse mode, and pressed h to jump to the heading.
Before this pr, time since input logging shows that it took 7.6 seconds to jump to and start speaking the heading.
With this pr, the time was 4.5 seconds.
This pr reduces the amount of TextInfo instantiations, and I think the iterUIARangeByUnit function requires less calls to move and compare methods.

source/UIABrowseMode.py
```diff
@@ -152,27 +152,34 @@ def isChild(self,parent):
 		return self.level>parent.level
 
 def UIAHeadingQuicknavIterator(itemType,document,position,direction="next"):
```
leonardder, line 154: Would you be able to add type hints? See below

source/UIABrowseMode.py
```diff
@@ -152,27 +152,34 @@ def isChild(self,parent):
 		return self.level>parent.level
 
 def UIAHeadingQuicknavIterator(itemType,document,position,direction="next"):
-	if position:
-		curPosition=position
+	reverse = bool(direction == "previous")
+	entireDocument = document.makeTextInfo(textInfos.POSITION_ALL)
+	if not position:
```
leonardder, line 157: When would this happen? Also, would a check for None be more appropriate here?

```suggestion
	if position is None:
```

LeonarddeR (approved):

source/UIABrowseMode.py
```diff
@@ -152,27 +152,34 @@ def isChild(self,parent):
 		return self.level>parent.level
 
 def UIAHeadingQuicknavIterator(itemType,document,position,direction="next"):
-	if position:
-		curPosition=position
+	reverse = bool(direction == "previous")
+	entireDocument = document.makeTextInfo(textInfos.POSITION_ALL)
+	if position is None:
```
michaelDCurran, line 157: @leonardder  position will be None if called by the Elements list to list all headings. Position is a TextInfo if called by quicknav to jump from a specific location.

LeonarddeR (approved):

source/UIABrowseMode.py
leonardder: Thanks for clarifying!

seanbudd (approved):

Issue #11615, opened 2020-09-16
nvWave: Use Windows Audio Session API


OzancanKaratas:
### Is your feature request related to a problem? Please describe.
NVDA lags due to the old Waveout API. Moreover, not turning the device off to solve this problem is costly for mobile computers.
### Describe the solution you'd like
A modern API named as Windows Audio Session was introduced in Windows Vista. NVDA supports Windows 7 and newer, so it is Windows Audio Session API can be used without any problem.
### Describe alternatives you've considered
WASAPI provides more flexibility for us, so we can use the standard device behaviours without a customization.
### Additional context
Important note: There are signs for Waveout's removal. We must prepare for future Windows releases now.

josephsl:
Hi, we need to consider two things: synthesizers and an actual documentation 9or an equivalent proof in the Windows SDK) that waveout is going away. Thanks.

akash07k:
However, I second this. WASAPI is more versatile and future proof.
Moreover, there's very high probability that it will increase NVDA's
responsiveness across devices.
I agree with Joseph's point though

leonardder:
There is a related issue regarding moving NVWave to c++: #5096

feerrenrut:
This is certainly something we have discussed internally, and even suggested as a project for GSOC, I'm surprised if there isn't already an issue for it. 

I think it is easy to justify this change, but I would like to question the following:
> not turning the device off to solve this problem is costly for mobile computers.

I wasn't able to find any data / reference that indicates that calling `waveOutOpen` results in the computer entering a state where it uses more power. 

Docs reference:
https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/nf-mmeapi-waveoutopen

OzancanKaratas:
Do we have to switch to C to use WASAPI? Can't we do this using Python?

feerrenrut:
@OzancanKaratas I would assume it is not necessary until there is some evidence to the contrary.

beqabeqa473:
i don't think so, as there are objective prediction that this might
improve nvda's performance in terms of sound output.

On 9/21/20, Reef Turner <notifications@github.com> wrote:
> @OzancanKaratas I would assume it is not necessary until there is some
> evidence to the contrary.
>
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/nvaccess/nvda/issues/11615#issuecomment-696123096


-- 
with best regards beqa

[...]

PR #13049, opened 2021-11-10
Again allow closing audio device on idle, opt in

LeonarddeR:
### Link to issue number:
https://github.com/nvaccess/nvda/issues/11615#issuecomment-764619365

### Summary of the issue:
On some Dell laptops, e.g. one owned by @jcsteh and my corporate laptop, NVDA tends to trigger an issue in the Waves Audio component installed on these laptops that causes a severe memory leak. This effectively means that you have to reboot your system at least an hour. There is a workaround to disable the waves audio suite on these machines, but that has several drawbacks.

As an attempt to fix this, I added a flag to advanced settings to restore the old audio behavior (i.e. directly close on idle by default).

### Testing strategy:
t.b.d.

None known

### Change log entries:
New features

### Code Review Checklist:

  - description is up to date
  - change log entries
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/config/configSpec.py
+++ b/source/config/configSpec.py
@@ -48,6 +48,8 @@ schemaVersion = integer(min=0, default={latestSchemaVersion})
 # Audio settings
 [audio]
 	audioDuckingMode = integer(default=0)
+	# 0:default, 1:yes, 2:no
+	leaveDeviceOpenOnIdle = integer(0, 2, default=0)
 
 # Braille settings
 [braille]
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -2796,6 +2796,39 @@ class AdvancedPanelControls(
 		self.cancelExpiredFocusSpeechCombo.defaultValue = self._getDefaultValue(
 			["featureFlag", "cancelExpiredFocusSpeech"]
 		)
+		# Translators: This is the label for a group of advanced options in the
+		#  Advanced settings panel
+		label = _("Audio")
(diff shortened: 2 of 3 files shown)


leonardder:
Turns out this doesn't fix the actually problem I"m having.

CyrilleB79:
On my job's DELL laptop, I work with USB headset always plugged.
When I resume from standby, it happens frequently that NVDA's TTS is played by the internal PC speaker rather than in the USB headset. It continues through the speaker as long as I perform the following action: press NVDA+control+S, do not change anything (the USB headset is already selected) and press Enter.

I have not tested this build (and cannot test it as an installed one). But it should probably allow to have the TTS sound played via the headset as soon as NVDA speaks something for the second time, as it was the case before with older versions of NVDA where the audio device was closed.

Another alternative could be to close/reopen only in case there has been an error while trying to play the audio via the USB without success.

What do you think? Would you mind reopening this PR to consider these use cases?

leonardder:
I'm sure happy to reopen it, at least when it fixes something. However, for your particular issue, I think what you propose sounds like a better approach

jcsteh:
My understanding is that there is code to detect changes to the default device and to switch accordingly. Your issue suggests that isn't working correctly. I'd suggest filing an issue for this, as there's probably a better solution than closing the device for every utterance.

PR #13056, opened 2021-11-14
Fixed a security issue where you could get a browse dialog on a secure screen via addon manager

trypsynth:
### Link to issue number:
None
### Summary of the issue:
If a user bound a gesture to open the addon manager, and copied it to their config at secure screens, someone could press it, get a browse dialog, and run CMD as systemroot, as well as do all sorts of other things
Added a check in globalCommands.py to see if we're on a secure screen before opening the addon manager via a gesture.
### Testing strategy:
1. Ran NVDA from source.
2. Bound a gesture to open the addon manager.
3. Copied my config to secure screens.
4. Pressed windows+l to log out, and attempted to press it
None
### Change log entries:
Not really sure if this deserves a CL entry, but if so...

Bug fixes
* It is no longer possible to open the addons manager from a secure screen, in order to avoid a security exploit.
### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -2558,6 +2558,8 @@ class GlobalCommands(ScriptableObject):
 		category=SCRCAT_TOOLS
 	)
 	def script_activateAddonsManager(self,gesture):
+		if globalVars.appArgs.secure:
+			return
 		wx.CallAfter(gui.mainFrame.onAddonsManagerCommand, None)
 
 	@script(


TheQuinbox:
Oops... I totally just realised I forgot to set this as a draft. Sorry, I can't figure out how to do it.

TheQuinbox:
Never mind, I'm such a newb. This should work

cary-rowen:
To be honest, I didn’t think this PR was very useful before NVDA did not allow the management of addon on the security screen. I can now use this vulnerability to manage addon on the security screen. I admit that there are security risks, but I think NVDA should Provide users with a feasible solution as soon as possible to make it more convenient for users to manage the addon of the security screen.

TheQuinbox:
Yeah, maybe allow disabling/enabling of addons or something, but definitly not the browse dialog. This works for now though, I think.

cary-rowen:
Yes, I hope NVDA can take this into consideration, even if the addon manager is opened on the security screen, the install plugin button is not displayed.

josephsl:
Hi, this assumes that a copy of gestures.ini with a bound gesture for Add-ons Manager is copied to system config folder, and the “browse dialog” refers to add-on install procedure, correct? Also, please provide a hypothetical (or even better, an actual) output from cmd.exe that is run using the procedure listed in this PR for verification purposes (hypothetically, this should be looked at, but I think an actual output would provide a persuasive evidence). Lastly, I don’t know about NV Access, but for security issues, please use issue/PR pair as it provides a venue for folks to discuss security implications in one issue that can be referenced using a follow-up PR. Thanks.

TheQuinbox:
@josephsl, yes. Something like this in gestures.ini
[globalCommands.GlobalCommands]
activateAddonsManager = kb(laptop):a+control+nvda+shift
And I'm slightly confused what you mean about cmd output? Fair enough about the issue/PR thing though. I just figured because it was a 2-line change, I'd just open a PR. But will keep that in mind for the future.

Brian1Gaff:
Would any of this affect the ability to install from a portable copy to the 
main machine complete with add ons?
 Brian

***@***.***
Sent via blueyonder.
in the display name field.
----- Original Message -----

lukaszgo1:
My preference would be either to disable the 'install' button in add-ons manager when in secure mode, or if we really want to disallow opening it on secure screens move this check inside `onAddonsManagerCommand` method. As an aside contributing guide asks not to use master when submitting PR's and create a separate topic branch for each patch. It should be possible to cherry-pick this single commit to a new branch and just change the base for this PR.

TheQuinbox:
OK. I found the lines in the addonGui to spawn the install button. Will close this PR, and open a new one (on a fresh branch) disabling that on secure screens :)

PR #13059, opened 2021-11-14
Disallow opening the browse for addons dialog on secure screens to avoid a security exploit.

trypsynth:
### Link to issue number:
Extension of #13056, but I slightly broke it, so redoing here.
### Summary of the issue:
It was possible for a user to bind a gesture to open the addons manager, copy their config to secure screens, press it, press the install button, and get a browse dialog, allowing them to run CMD as systemroot, and do all sorts of things.
Only show the addon manager if we're not on a secure screen.
### Testing strategy:
1. Run NVDA from source.
2. Copy my config to secure screens.
3. Press Windows+L to lock my machine.
4. Press my hotkey.
None
### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


source/gui/addonGui.py
```diff
@@ -248,8 +248,9 @@ def __init__(self, parent):
 		self.getAddonsButton = generalActions.addButton(self, label=_("&Get add-ons..."))
```
lukaszgo1, line 248: This button should also be disabled when in secure mode since it opens a web browser from which you can easily do a lot of insecure stuff.

TheQuinbox:
OK, getAddons is also now disabled, and translation should work again. Not sure why system tests broke though.

mazen428:
Why not just disable the entire add-ons manager from a secure screen, as far as I know you can't do anything useful in it since NVDA's configuration doesn't save in secure screens.

TheQuinbox:
I did that in the initial PR, but people said they'd prefer this approach.

CyrilleB79:
Since the add-on manager is not reachable via NVDA menu in secure screen, I would also recommend not to be able to open it via a shortcut at all.

TheQuinbox:
That was what my initial PR did, but no one seemed to like it. I'm cool with either way, just want ot patch the security problem.

cary-rowen:
> Why not just disable the entire add-ons manager from a secure screen, as far as I know you can't do anything useful in it since NVDA's configuration doesn't save in secure screens.

Hi, We can at least remove the add-on from it, I actually did this often in the past,

cary-rowen:
In addition to the install button, the button to get addon is also meaningless.
Why can't we allow the add-ons manager to be opened from the menu on the security screen,After all, NVDA could not manage the addon on the security screen now.
I apologize if there is a use case that I did not expect and must do this.

CyrilleB79:
Yes there are two options here:
1. Disable completely add-on management from secured screen
2. Enable it in a limited way, e.g. allow add-on suppression. In this case, the add-on manager should be reachable via the menu and via the shortcut but some buttons should be greyed out (Add, Help, Get add-ons). And clarify the allowed actions: remove, enable/disable?

michaelDCurran:
I think the most appropriate direction here is to disable the add-ons 
manager from being opened at all on secure screens.

TheQuinbox:
@michaelDCurran Done.

leonardder:
How is someone supposed to know what add-ons are active in the secure copy? Not everyone is able to browse to `C:\Program Files (x86)\NVDA\systemConfig\addons`

CyrilleB79:
This PR is aimed to fix a security issue. So I think it should be addressed as soon as possible and not be delayed by "almost new feature" requests.

The best way to address the security issue is to disable the possibility to open the add-on manager dialog in secure screen; it seems it was the original intention of NVDA developers when they implemented the menu item removal.

In a subsequent way we may open a new issue and then PR to discuss various possibilities regarding add-on in secure screen among which:
- view installed add-ons and their state
- view incompatible add-ons and the reason
- uninstall add-ons
- enable/disable add-ons
- open the "About" dialog
In case of issue, this second PR could be reverted without impacting the security concern.

TheQuinbox:
I agree. For now, the manager is just disabled. We can discuss possible implimentations in a different issue/pr.

lukaszgo1:
@michaelDCurran  This PR seems ready to be merged. Are you planning to look into it? Frankly if I were a new contributor and my one line security bug fix would not be looked at in almost three months this would be my first and last contribution. I really believe external contributions from newcomers should be given more attention!

seanbudd (approved):
@michaelDCurran - can I merge this?

TheQuinbox:
Wow, this is still open? I honestly forgot this pR even existed in the first place, it's a single line change and I don't get what's so hard about merging that.
It's sad to me that new features are seemingly prioritized over simple, 1-line bugfixes (especially ones like this). Granted the chances are slim, but if I had my NVDA set up in the right configuration without this patch, someone could get access to my machine when I don't want them to.
If there are any steps I missed that I need to do in order to merge this, please let me know, but after this experience, I'm honestly not sure if I'll be contributing to NVDA again. If a simple PR can't be merged, I'm just trying to imagine a larger one...

feerrenrut:
@seanbudd if you're happy with the change / testing, please go ahead with the merge. @TheQuinbox Sorry this fell through the cracks, we'll review our processes to try to ensure this doesn't happen in the future.

PR #13079, opened 2021-11-22
If AppVeyor install of NVDA fails, ensure installer log can be uploaded successfully

### LINKED ISSUE #13083: CI: When installer fails  to install NVDA but does not crash NVDA log cannot be uploaded successfully and build is not interrupted.
### Steps to reproduce:
1. Open a PR against NVDA repository which causes NVDA installer to hang (for an example see #13079 )

### Actual behavior:
1. There is no comment on the pr saying 'NVDA installer process timed out'
2. Log from the installer is not uploaded.
3. Build tries to continue and system tests are executed against broken NVDA installation and obviously fail.
### Expected behavior:
1. There is a message that NVDA installer failed to install in the specified time.
2. Installer log is uploaded so developers can download it from the artifacts tab and inspect to determine reason for the failure.
3. Build is interrupted since tests cannot be executed if NVDA is not installed.
Affects installer.
#### NVDA version:
Latest master as of November 22-th
#### Windows version:
I believe AppVeyor worker images are running Server 2019
None
N/A
N/A
Irrelevant
Irrelevant.

### Technical:
Points 1 and 3 are caused by the fact that by default exceptions raised by `wait-process` are not terminating - therefore script never gets into the `catch` block and does not execute this code path.
However even when the exception is raised and caught the installer log fails to upload with the following, not very helpful output:
```
Push-AppveyorArtifactInternal : Error uploading artifact to the storage: One or more errors occurred.
At C:\Program Files\AppVeyor\BuildAgent\Modules\build-worker-api\build-worker-api.psm1:209 char:2
+     Push-AppveyorArtifactInternal -FullPath $fullPath -FileName $File ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Push-AppveyorArtifactInternal], Exception
    + FullyQualifiedErrorId : System.Exception,Appveyor.BuildAgent.Api.Utils.PushAppveyorArtifactInternalCmdlet
Command executed with exception: Error uploading artifact to the storage: One or more errors occurred.
```

cc @seanbudd  I believe you've worked on AppVeyor scripts recently.

lukaszgo1:
### Link to issue number:
Fixes #13083
### Summary of the issue:
When NVDA's installer fails to install NVDA and the process is still running the log from the failed installation cannot be uploaded and all tests fail since they're executed against broken installation of NVDA.
1. Exceptions raised by `wait-process` are properly caught which in practice means that the block of code responsible for interrupting the build if installer fails to exit in the specified timeout is executed - previously it didn't.
2. Since AppVeyor function used to upload build artifacts is unable to upload files which are in use if the installer process is still running we copy the log and upload it instead of the, currently locked, original file.
### Testing strategy:
With a commit in which I've simulated a crash of nvda_slave inspected [this build log](https://ci.appveyor.com/project/NVAccess/nvda/builds/42096427) and made sure that:
- The message that install process timed out is shown
- the build is interrupted when NVDA fails to install
- The installer log can be downloaded from the artifacts.
None known
### Change log entries:
None needed - only affects CI
### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/appveyor/scripts/installNVDA.ps1
+++ b/appveyor/scripts/installNVDA.ps1
@@ -10,14 +10,21 @@ $outputDir=$(resolve-path .\testOutput)
 $installerLogFilePath="$outputDir\nvda_install.log"
 $installerProcess=start-process -FilePath "$nvdaLauncherFile" -ArgumentList "--install-silent --debug-logging --log-file $installerLogFilePath" -passthru
 try {
-	$installerProcess | wait-process -Timeout 180
+	$installerProcess | wait-process -Timeout 180 -ErrorAction Stop
 	$errorCode=$installerProcess.ExitCode
+	$installlerLogFilePathToUpload = $installerLogFilePath
 } catch {
 	echo "NVDA installer process timed out"
 	$errorCode=1
 	Add-AppveyorMessage "Unable to install NVDA prior to tests."
+	# Since installer failed to exit in the specified timeout the log file is still in use.
+	# Unforturnately `Push-AppveyorArtifact` is unable to upload a file which is  locked
+	# as a work around create a copy of the log and upload that instead.
+	$installerLogFileCopiedPath = "nvda_install_copy.log"
+	Copy-Item $installerLogFilePath $installerLogFileCopiedPath
+	$installlerLogFilePathToUpload = $installerLogFileCopiedPath
 }
-Push-AppveyorArtifact $installerLogFilePath
+Push-AppveyorArtifact $installlerLogFilePathToUpload -FileName "nvda_install.log"
 $crashDump = "$outputDir\nvda_crash.dmp"
 if (Test-Path -Path $crashDump){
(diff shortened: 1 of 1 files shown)


leonardder:
@lukaszgo1  what's supposed to happen with this pr?

lukaszgo1:
Since it allows to reproduce #13083 I'd  like to keep  it as a draft, and use as a playground when fixing that issue.

CyrilleB79:
You may change this PR's title/description for clarity.
Unless of course it have an impact an appVeyor builds... but I think not.

seanbudd (approved):

PR #13086, opened 2021-11-23
Fix MSVC Warnings in NVDAHelper

feerrenrut:
Note to reviewers, each commit contains a small change, the message describes the issue that triggered the warning.

### Link to issue number:
None

### Summary of the issue:
- When setting up Visual Studio to work with NVDAHelper (as per the NVDAHelper readme), some include paths aren't found.
- Visual studio's warnings show now show up while working with NVDA helper.
- Out of date testing code not part of the build

- Fixed all the includes so that they are found by visual studio.
- Fixed the more severe warnings as I noticed them.
- Unused code removed

### Testing strategy:
Built and run locally while testing out aria-details work.

None

### Change log entries:
None

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


feerrenrut:
The build failure was caused by trying to log from `nvdaControllerClient`. Note that although `log.h` lives in `nvdaHelper/common`, `log.cpp` lives in `NVDAHelper/remote` and is not built for the `nvdaControllerClient`.
I don't think it makes sense to log back to NVDA from the `nvdaControllerClient`, instead we can just return `FALSE` from `DllMain` the RPC binding fails. This will mean that that the call to `loadLibrary` in the client code will return `NULL` and `DllMain` will be called immediately with `DLL_PROCESS_DETACH`.

I had forgotten that it had to be built explicitly (`scons client`), mistakenly thinking it would get built with `scons source`.

feerrenrut:
While looking at the build issue with `nvdaControllerClient` I found: https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-best-practices#general-best-practices

This suggests it is a bad idea to do anything significant in `DllMain`, instead prefer an explicit `initialization` method. I suspect that creating the RPC binding is an example of something that shouldn't happen here.

@michaelDCurran thoughts?

nvdaHelper/remote/ime.cpp
```diff
@@ -108,12 +108,12 @@ DWORD getIMEVersion(HKL kbd_layout, wchar_t* filename) {
 			// Do not know how to extract version number
 			return 0;
 	}
-	DWORD ver_handle;
+	DWORD ver_handle = 0;
 	DWORD buf_size = GetFileVersionInfoSizeW(filename, &ver_handle);
```
michaelDCurran, line 112: apparently the pointer to handle is optional, thus we could just use nullptr?

nvdaHelper/remote/ime.cpp
```diff
@@ -108,12 +108,12 @@ DWORD getIMEVersion(HKL kbd_layout, wchar_t* filename) {
...
 			return 0;
 	}
-	DWORD ver_handle;
+	DWORD ver_handle = 0;
 	DWORD buf_size = GetFileVersionInfoSizeW(filename, &ver_handle);
 	if (!buf_size)  return 0;
 	void* buf = malloc(buf_size);
     if (!buf)  return 0;
-	if (GetFileVersionInfoW(filename, ver_handle, buf_size, buf)) {
+	if (GetFileVersionInfoW(filename, /*ver_handle*/0, buf_size, buf)) {
```
michaelDCurran, line 116: I find the inline comment confusing. MS docs state that the second parameter is not used. So we should just write 0, no matter what it may have been before this pr.

nvdaHelper/remote/ime.cpp
```diff
@@ -108,12 +108,12 @@ DWORD getIMEVersion(HKL kbd_layout, wchar_t* filename) {
 			// Do not know how to extract version number
 			return 0;
 	}
-	DWORD ver_handle;
+	DWORD ver_handle = 0;
```
michaelDCurran, line 111: This line can actually be removed, as getFileVersionInfoSize and getFileVersionInfo don't need it. See comments below.

nvdaHelper/vbufBackends/mshtml/mshtml.cpp
```diff
@@ -90,9 +90,9 @@ template<typename toInterface> inline HRESULT queryService(IUnknown* pUnknown, c
 	}
 	hRes=pServProv->QueryService(siid,__uuidof(toInterface),(void**)pIface);
 	pServProv->Release();
-	if(hRes!=S_OK||!pIface) {
+	if(hRes!=S_OK) {
```
michaelDCurran, line 93: Why remove the check for null pIface? I have experienced many implementations of QueryService and QueryInterface that may return S_OK yet leave pIface as null. Incorrect certainly, but it happens. I would certainly put this check back.

nvdaHelper/vbufBackends/mshtml/node.cpp
```diff
@@ -292,18 +292,39 @@ class CHTMLChangeSink : public IHTMLChangeSink {
...
+	bool isBlock,
+	MshtmlVBufBackend_t* backend,
+	bool isRootNode,
+	IHTMLDOMNode*
+	pHTMLDOMNode,
+	const wstring& lang
+)
+	: VBufStorage_controlFieldNode_t(docHandle, ID, isBlock)
+	, backend(backend)
+	, pHTMLDOMNode(nullptr)
```
michaelDCurran, line 307: Surely if this and other instance variables now have initializers in the h file, it is unnecessary to include them specifically here? Or if I'm wrong, then the other way round... why must we give initial vlaues for these instance variables in the h file if they are never used?

michaelDCurran (changes requested):

michaelDCurran:
I'm fairly sure that rpcBindingFromStringBinding only creates a kind of 
cookie thing out of the string to be then used in further rpc functions. 
I don't think it actually initializes any particular systems or does 
anything heavy. This is handled lazeily when the first rpc call is 
actually made.

To my knowledge, this has not caused us issues in the past. Though I do 
acknowledge and agree with your point about not doing anything major in 
dllmain. You certainly don't want to do something that causes the OS to 
try and load another dll.

nvdaHelper/remote/ime.cpp
feerrenrut: The inline docs are for the sake of developers, to know which argument this is. The value passed in here is 0. I could use a `constexpr` instead, to give it a name?

nvdaHelper/client/client.cpp
```diff
@@ -15,16 +15,18 @@ This license can be found at:
 #include <string>
 #include <sstream>
 #include <rpc.h>
-#include "nvdaController.h"
+#include <remote/nvdaController.h>
```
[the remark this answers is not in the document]
feerrenrut: Good catch. Really this shouldn't compile. Why is `remote` available in the include paths for the client build and local the local build?

nvdaHelper/remote/ime.cpp
feerrenrut: Yes, if we have no need for it.

nvdaHelper/vbufBackends/mshtml/mshtml.cpp
feerrenrut: No, then it will dereference a nullptr. See line 95. Note that we are passing in a `void**` if `pIface` is a `nullptr` the mistake is before the call to `QueryService`, right? Perhaps this code should have been:
```suggestion
	if( hRes != S_OK || *pIface == nullptr) {

```

nvdaHelper/vbufBackends/mshtml/node.cpp
feerrenrut: That's what I thought, though the analyzer was still complaining, [looking this up](https://en.cppreference.com/w/cpp/language/data_members):
>2) Through a default member initializer, which is a brace or equals initializer included in the member declaration and is used if the member is omitted from the member initializer list of a constructor.

I'll try this again, perhaps something else went wrong with the analyzer that I didn't notice.

feerrenrut:
There is a specific list of things not to do during DllMain in that link. It looks pretty risky, and easy for something we rely on to incidentally call one of the methods (that is if the list is even exhaustive). It would be pretty easy to move this code to an explicit `initialize` and `deinit` methods?

michaelDCurran:
An inline comment to name a positional argument is fine with me. But in 
that case you should call it "dwHandle" as that is what Microsoft Docs 
call it. The fact it was ver_handle, which is the same name as the 
variable it used to be was the confusing bit.

michaelDCurran:
>> Why is remote available in the include paths for the client build 
and local the local build?

Well, the include path is just nvdaHelper, which contains client, 
common, remote, local etc.

We could make the include paths much more specific, but then you'd need 
to drop the directory name of each #include.

I'm not bothered which way you go here.

michaelDCurran:
yes you're right. It should have been `*pIface == nullptr`.

However, this assumes that pIface itself was non-null as well. Perhaps 
there needs to be a check for this before the call to QueryService if 
there is not already.

feerrenrut:
>But in that case you should call it "dwHandle" as that is what Microsoft Docs

Ah you are right, not sure how I got that confused.

> We could make the include paths much more specific, but then you'd need
to drop the directory name of each #include.

I looked at this briefly, there are some files with the same (or perhaps just similar) names, the directory really helps to add context. If we were to do this, typically public headers are in an includes directory, grouped by "project" (ie all compilation units used to create a single binary form a project). EG:
- `remoteProj/include/remote/`
- `localProj/include/local/`
- `clientProj/include/`

On the other hand, I'm not sure how many of these projects actually refer to each others headers, and when they do it may just be the generated RPC headers. I think this is a bigger discussion, possibly one that should be had when / if replacing the build system.


>However, this assumes that pIface itself was non-null as well. Perhaps
there needs to be a check for this before the call to QueryService if
there is not already.

I'll look at this.

feerrenrut:
I'm leaving the latest commits separate for ease of review. When the PR is approved I'll auto squash these commits. Given this PR is composed of many disparate commits, I intend to merge (rather than squash merge).

michaelDCurran (approved):

PR #13127, opened 2021-12-03
clean up getTextFromIAccessible

feerrenrut:
### Link to issue number:
Depends on:
- https://github.com/nvaccess/nvda/pull/13126
- #13106

### Summary of the issue:
The moved function `getTextFromIAccessible` doesn't use smart pointers, was missing some explicit initialization, and logic could be simplified.

- Introduce (CCom) smart pointers.
- Initialize all variables.
- Simplify logic where obvious.
- ~[ ] Also split based on "useNewText" https://github.com/nvaccess/nvda/pull/13106#discussion_r765567962~ This has been dropped. It will be kept in mind for future work in the area.

### Testing strategy:
- Builds runs locally
- Testing with annotations changes.

None

### Change log entries:
None

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/nvdaHelper/remote/textFromIAccessible.cpp
+++ b/nvdaHelper/remote/textFromIAccessible.cpp
@@ -32,11 +32,11 @@ bool getTextFromIAccessible(
 ) {
 	if (!pacc2) {
 		return false;
-	bool gotText = false;
-	IAccessibleText* paccText = NULL;
-	if (pacc2->QueryInterface(IID_IAccessibleText, (void**)&paccText) != S_OK) {
-		paccText = NULL;
 	}
+
+	bool gotText = false;
+	CComQIPtr<IAccessibleText, &IID_IAccessibleText> paccText(pacc2);
+
 	if (!paccText && recurse && !useNewText) {
 		//no IAccessibleText interface, so try children instead
 		long childCount = 0;
@@ -120,7 +120,6 @@ bool getTextFromIAccessible(
 			SysFreeString(bstrText);
 			textBuf.append(1, L' ');
 		}
-		paccText->Release();
 	}
 	if (!gotText && !useNewText) {
 		//We got no text from IAccessibleText interface or children, so try name and/or description
--- a/nvdaHelper/remote/textFromIAccessible.cpp
+++ b/nvdaHelper/remote/textFromIAccessible.cpp
@@ -70,8 +70,8 @@ bool getTextFromIAccessible(
 		long startOffset = 0;
 		//If requested, get the text from IAccessibleText::newText rather than just IAccessibleText::text.
 		if (useNewText) {
-			IA2TextSegment newSeg = { 0 };
-			if (paccText->get_newText(&newSeg) == S_OK && newSeg.text) {
+			IA2TextSegment newSeg {};
+			if (S_OK == paccText->get_newText(&newSeg) && newSeg.text) {
 				bstrText = newSeg.text;
 				startOffset = newSeg.start;
 			}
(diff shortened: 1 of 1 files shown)


seanbudd (approved):

nvdaHelper/remote/textFromIAccessible.cpp
```diff
@@ -13,13 +13,53 @@ This license can be found at:
...
+	varChild.vt = VT_I4;
+	varChild.lVal = 0;
+
+	CComBSTR val;
+	pacc->get_accName(varChild, &val);
+	bool valEmpty = isEmpty(val);
+	if (!valEmpty) {
+		gotText = true;
+		textBuf.append(val);
+		textBuf.append(L" ");
```
michaelDCurran, line 51: This appends a space even if there is no description later. Perhaps we could remove this line, and lower down once we have a valid description, just before setting gotText to true, append the space if gotText was already true. 
If you just copied the original logic however, I'm happy for you to leave this.

nvdaHelper/remote/textFromIAccessible.cpp
```diff
@@ -29,45 +69,47 @@ bool getTextFromIAccessible(
...
-	bool gotText = false;
-	IAccessibleText* paccText = NULL;
-	if (pacc2->QueryInterface(IID_IAccessibleText, (void**)&paccText) != S_OK) {
-		paccText = NULL;
+	if (!pacc2) {
+		return false;
 	}
+
+	bool gotText = false;
+	CComQIPtr<IAccessibleText, &IID_IAccessibleText> paccText(pacc2);
```
michaelDCurran, line 77: I think it is possible to not have to specify the IID on CComQIPtr, as the template should calculate it from _uuidof(interface).

michaelDCurran (changes requested):

nvdaHelper/remote/textFromIAccessible.cpp
feerrenrut: Yes, I believe so. I chose to make it explicit because I wasn't sure of any 'gotchas' involved. If your confident about this I'll remove the extra param. In some places we already rely on the calculated UUID.

michaelDCurran:
Yes, I'm confident. As long as the COM interface was generated from an 
IDL file, which it was, then it will certainly have an associated UUID.

nvdaHelper/remote/textFromIAccessible.cpp
feerrenrut: I believe this matches the logic from before (see https://github.com/nvaccess/nvda/pull/13127/files#diff-a8111ab994be8314dfa43a5f725fd2969ac65ad8f68f899f4f7dabf6088b9bb1L137) I'll double check this.

nvdaHelper/remote/textFromIAccessible.cpp
feerrenrut: I can't see any differences in the logic.

nvdaHelper/remote/textFromIAccessible.cpp
feerrenrut: I can't see any differences in the logic.

michaelDCurran: Yep, that is fine. In that case just leave it.

nvdaHelper/remote/textFromIAccessible.cpp
feerrenrut: This has been addressed.

michaelDCurran (approved):

PR #13150, opened 2021-12-09
Try new HID protocol after, not before all other driver matches

### LINKED ISSUE #13153: Unstable auto detect for Humanware Brailliant and APH Mantis Q40 braille displays
### Steps to reproduce:
1. Pair a Mantis Q40 over Bluetooth
2. Start NVDA with braille auto detection enabled.

### Actual behavior:
NVDA sometimes fails detecting the display, because it tries the HID prottocol driver first. This driver reports as available but is not supported by the display. Interesting enough, as soon as you select the HID driver manually, NVDA connects to the display but the displays stays empty. When you open the display dialog again, nothing is selected and the HID driver disappeared.
Note that sometimes, NVDA manages to connect to the display with the Brailliant driver, but behavior is pretty instable on my system. 

### Expected behavior:
Direct connection using the Brailliant driver when using auto detect.

Installed

#### NVDA version:
2021.3

#### Windows version:
Windows 11 22000.348

Yes

Same behavior as current Alpha.

Yes

Yes

LeonarddeR:
### Link to issue number:
Closes #13153 

### Summary of the issue:
Auto detection is broken for Humanware Brailliant and APH Mantis q40 when working over Bluetooth. It looks like they implement an incomplete version of the HID protocol.

to be honest, my belief has always been that the new HID protocol should not overrule a custom protocol offered by a vendor. This could lead to several side effects, including a vendor updating a firmware with support for the new protocol, causing different behavior in NVDA for example.

### Testing strategy:
Tested with the try build that detection of the Mantis Q40 over Bluetooth is instant and stable again.

1. It should be considered to create a point release for this if many people are affected. Therefore I'm filing this against RC for now.
2. It should be investigated whether what I report in #13153 about the HID driver connecting, then disappearing from the list of available displays, is a bug in the display or in the HID driver.

### Change log entries:
```
Changes
- The new HID Braille protocol is no longer preferred when another braille display driver can be used.
Bug fixes
- Fixed unstable braille display auto detection for Humanware Brailliant and APH Mantis Q40 braille displays when using Bluetooth.
```

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


michaelDCurran:
This is probably the right thing to do, but very annoying that Humanware ended up officially releasing firmware with initial or half-finished Braille HID support. We had asked them on many occasions about their plans for this, and after we had supplied them with several bug reports for test firmware which I'm guessing they must not have addressed.
I'm going on leave for 3 weeks, so I'll let @feerrenrut  decide as to whether this should warrant a point release. 
To do the right thing by our users, it probably should.

michaelDCurran (dismissed):

michaelDCurran:
I would however want this tested with an Orbit Reader 40, in both Braille HID and  Orbit HID modes, as Orbit Reader so far is the only real-world official display implementing Braille HID.

source/bdDetect.py
```diff
@@ -123,13 +123,14 @@ def getDriversForConnectedUsbDevices():
...
 	)
 	for match in usbDevs:
-		# check for the Braille HID protocol before any other device matching.
-		if match.type == KEY_HID and match.deviceInfo.get('HIDUsagePage') == HID_USAGE_PAGE_BRAILLE:
-			yield ("hid", match)
 		for driver, devs in _driverDevices.items():
 			for type, ids in devs.items():
 				if match.type==type and match.id in ids:
 					yield driver, match
+		# check for the Braille HID protocol after any other device matching.
```
feerrenrut, line 130: This should say "why" NVDA should do things in this order.

source/bdDetect.py
```diff
@@ -144,15 +145,16 @@ def getDriversForPossibleBluetoothDevices():
...
-		# check for the Braille HID protocol before any other device matching.
-		if match.type == KEY_HID and match.deviceInfo.get('HIDUsagePage') == HID_USAGE_PAGE_BRAILLE:
-			yield ("hid", match)
 		for driver, devs in _driverDevices.items():
 			matchFunc = devs[KEY_BLUETOOTH]
 			if not callable(matchFunc):
 				continue
 			if matchFunc(match):
 				yield driver, match
+		# check for the Braille HID protocol after any other device matching.
```
feerrenrut, line 154: Same comment, "why?"

feerrenrut:
This will mean that braille device manufacturers will need to change device IDs if they wish to support / test HID with NVDA. 
I think this will be fine for 2021.3.1. We may add an advanced toggle for this in 2022.1. Device makers should be changing the ID anyway when they change the expected communications approach.

feerrenrut:
If we are going to have a changes file, then we'll want to give translators a chance, this will need to target beta (where translations happen).

feerrenrut:
It would be good to have some instructions for Braille manufacturers about device IDs. I'd like a strong recommendation that device IDs should be changed when moving from legacy custom driver approach to a HID Braille approach. I'm ok with this going into alpha since it may take some time to get the wording / location of the advice right.

feerrenrut (approved):

PR #13169, opened 2021-12-15
no longer log warnings when setting translations for languages without translations

lukaszgo1:
### Link to issue number:
None
### Summary of the issue:
When starting NVDA with language set to English and logging for debug warnings enabled entries like the following are written to the log:
```
DEBUGWARNING - languageHandler.setLanguage (16:40:00.332) - MainThread (4856):
couldn't set the translation service locale to en
DEBUGWARNING - languageHandler.setLanguage (16:40:00.332) - MainThread (4856):
couldn't set the translation service locale to en

```
In case of English for which there are no translations these exceptions are expected and there is no point in logging in this case.
1. Introduced a list of languages for which we don't have a translations and don't log an error if trying to create a gettext translation for one of these languages.
2. Cleaned up the code which is responsible for setting the locale of the current thread (the failure would be logged only when `SetThreadLocale` would raise a `IOError` which was a mistake as its return value should be inspected).
### Testing strategy:
1. Ensured  that no  warnings are logged with language set to English
2. Ensured that warnings are shown when language is set to 'user default' on a Windows installation with the default language set to `en_US` for the 'en-US' locale and no warning is shown for 'en' i.e. after splitting on underscore.
None known
### Change log entries:
None needed - only affects logging.
### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/languageHandler.py
+++ b/source/languageHandler.py
@@ -18,7 +18,13 @@ import enum
 import globalVars
 from logHandler import log
 import winKernel
-from typing import List, Optional, Tuple
+from typing import (
+	FrozenSet,
+	List,
+	Optional,
+	Tuple,
+	Union,
+)
 
 #a few Windows locale constants
 LOCALE_USER_DEFAULT = 0x400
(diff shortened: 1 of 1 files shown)


feerrenrut (approved):
Looks good thanks @lukaszgo1

PR #13215, opened 2022-01-04
appModules/calculatorapp: add support for Windows 11 Calculator

### LINKED ISSUE #13212: Add support for Windows 11 Calculator
Hi,

A subtle follow-up to #9428:

### Background:
One of the side-effects of Windows as a Service, or for that matter, universal apps, is the changing nature of apps. One of the changes is renamed executables, the best known case being Windows 10/11 Mail (hxmail -> hxoutlook). Another change is UIA tree for apps where the old tree traversal route will not work due to rearranged UIA tree for apps, the best example being emoji panel.

In #9428, NVDA introduced support for Windows 10 Calculator (calculator.exe), mostly to suppress UIA notification event announcement when entering expressions while speak typed characters is turned off. Windows 11 twists this in two ways:

* Calculator app executable was renamed to calculatorapp.exe
* UIA tree is different, so tree traversal to detect UIA notification announcement condition must be tweaked.

Add support for Windows 11 Calculator through a new app module. There are two implementation strategies:

1. Create a brand new Windows 11 Calculator app module that inherits from existing Windows 10 (modern) Calculator app module but tweaks UIA notification event handler's results tree traversal (in Windows 11 Calculator, one must move down one more level when fetching results element).
2. Tweak existing Windows 10 Calculator app module's UIA notification event handler to account for UIA tree differences, then add an alias app module for Windows 11 Calculator.

Both approaches are being tested via Windows App Essentials add-on (currently the first approach is available for users, with the second approach being tested in private and a proof of concept will be deployed to dev channel users within the next 24 hours). The upcoming pull request will "transplant" code from Windows App Essentials add-on to NVDA Core (people reading add-on source code should notice what that PR would look like).

Leave it to add-ons to support Windows 11 Calculator.

### Additional context
App modules for modern Windows apps are subject to change based on latest happenings with apps themselves, especially since apps can be updated anytime (this is the main reason why I wait for a while before proposing an app module for a modern Windows app to be included in NVDA Core). Also, Calculator app is now open-source, so anyone can contribute to it (repo is hosted on GitHub), with newer releases focusing on Windows 11.

Thanks.

josephsl:
### Link to issue number:
Closes #13212 

### Summary of the issue:
Adds support for Windows 11 Calculator (calculatorapp.exe), consisting of an alias app module and responding to UI changes via Windows 10 Calculator app module.

In Windows 11, Calculator not only saw UI changes, but it comes with a renamed executable (calculator.exe -> calculatorapp.exe). To take care of these changes:

1. Windows 11 Calculator app module was created as an alias for Windows 10 Calculator to follow precedence set by other apps such as emoji panel, Windows Search, and Mail.
2. Windows 10 Calculator, the base app module, provides detection of Windows 11 UI (if Windows 11 version is in use, UIA notification event handler's results detection code must descend one more level to locate display element).

An alternative is a dedicated Windows 11 Calculator app module that inherits from Windows 10 Calculator's AppModule class and tweaks UIA notification event handler. Although this makes it easy to identify which app module is which, it introduces code duplication which might be an issue from maintenance point of view. Therefore, an alias app module and a tweaked base app module is more suitable for this situation.

### Testing strategy:
Manual testing: requires Windows 11, make sure NVDA performs as expected while using Calculator app just as in Windows 10 (namely not nanouncing display text unless Enter/Escape/Delete is pressed, unit converter results are announced, and others).

None

### Change log entries:
New features:

Added support for Windows 11 Calculator. (#13212)

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


source/appModules/calculator.py
```diff
@@ -76,6 +76,10 @@ def event_UIA_notification(self, obj, nextHandler, displayString=None, activityI
 			)
 		except AttributeError:
 			resultElement = api.getForegroundObject().children[1].lastChild
+			# Descend one more time in Windows 11 Calculator.
+			calculatorVersion = int(self.productVersion.split(".")[0])
+			if calculatorVersion >= 11:
+				resultElement = resultElement.firstChild
```
feerrenrut, line 82: Why is this all in an exception handler?

josephsl:
Hi, it is a legacy of old releases where UIA notification event can come from multiple elements, and in some cases, elements are next to unit converter interface. After testing the event handler, it turns out attribute error check is not needed in recent releases, therefore I’m going to rework this and apply a point about assigning product version somewhere (likely at the start of the event handler). Thanks for reminding me to stay updated on things.

seanbudd (approved):

josephsl:
@seanbudd, ping, milestone update please i.e. when do you expect this will be merged into master. Thanks.

source/appModules/calculator.py
```diff
@@ -66,28 +66,25 @@ def event_nameChange(self, obj, nextHandler):
...
+		calculatorVersion = int(self.productVersion.split(".")[0])
+		# #12268: for "DisplayUpdated", announce display strings in braille  no matter what they are.
 		if activityId == "DisplayUpdated":
 			braille.handler.message(displayString)
-		try:
-			shouldAnnounceNotification = (
-				obj.previous.UIAAutomationId in
-				("numberPad", "UnitConverterRootGrid")
-			)
-		except AttributeError:
```
feerrenrut, line 77: Removing this effectively changes the scope of the subsequent code, is the indent level as expected?

josephsl:
Hi, yes, the indent level should be kept as we are effectively dealing with a single activity Id. Thanks.

source/appModules/calculator.py
```diff
@@ -66,28 +66,25 @@ def event_nameChange(self, obj, nextHandler):
 		nextHandler()
 
 	def event_UIA_notification(self, obj, nextHandler, displayString=None, activityId=None, **kwargs):
-		# #12268: for "DisplayUpdated", announce display strings in braille and move on.
+		calculatorVersion = int(self.productVersion.split(".")[0])
+		# #12268: for "DisplayUpdated", announce display strings in braille  no matter what they are.
```
feerrenrut, line 70: In that case, @josephsl could you add another comment below this one to say that the only type of activity that should be handled here is "DisplayUpdated". Perhaps an example of other activities that shouldn't be handled, or an explanation about why other activities are not relevant.

josephsl:
Hi, will do so in the next commit. Thanks.

feerrenrut (approved):
Thanks @josephsl

PR #13226, opened 2022-01-10
Scons: provide the number of logical processors as the number of maximum concurrent jobs to speed up builds

LeonarddeR:
### Link to issue number:
None

### Summary of the issue:
Building NVDA can take some time. SCons has the ability to provide the number of concurrent jobs during build.

Specify the maximum number of concurrent jobs when building NVDA.

### Testing strategy:
Tested locally on an Intel Core i7-9850H using PowerShell's `measure-command { .\scons.bat source }`, ensuring the venv was created prerun.

* Without the change, 170 seconds
* With the change, 59 seconds

Output sometimes get a bit scrambled, i.e. the output of the cl command comes before SCons output the name of the file being build.

### Change log entries:
For Developers
- Set up SCons to build with multiple concurrent jobs, equal to the number of logical processors in the system (i.e. 8 on a quat core Intel CPU with hyper threading). This can dramatically decrease build times on multi core systems.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/appveyor/scripts/setSconsArgs.ps1
+++ b/appveyor/scripts/setSconsArgs.ps1
@@ -15,6 +15,8 @@ if(!$env:APPVEYOR_PULL_REQUEST_NUMBER) {
 	$sconsArgs += " certFile=appveyor\authenticode.pfx certTimestampServer=http://timestamp.digicert.com"
 }
 $sconsArgs += " version_build=$env:APPVEYOR_BUILD_NUMBER"
+# Use only 1 concurrent job when building
+$sconsArgs += "-j1"
 # We use cmd to run scons because PowerShell throws exceptions if warnings get dumped to stderr.
 # It's possible to work around this, but the workarounds have annoying side effects.
 Set-AppveyorBuildVariable "sconsOutTargets" $sconsOutTargets
--- a/appveyor/scripts/setSconsArgs.ps1
+++ b/appveyor/scripts/setSconsArgs.ps1
@@ -16,7 +16,7 @@ if(!$env:APPVEYOR_PULL_REQUEST_NUMBER) {
 }
 $sconsArgs += " version_build=$env:APPVEYOR_BUILD_NUMBER"
 # Use only 1 concurrent job when building
-$sconsArgs += "-j1"
+$sconsArgs += " -j1"
 # We use cmd to run scons because PowerShell throws exceptions if warnings get dumped to stderr.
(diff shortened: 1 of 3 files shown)


leonardder:
Note, this doesn't seem to have any noticeable benefit on appveyor.

josephsl:
Hi, probably because it uses virtual machine instances, which can impact processor performance. Thanks.

Brian1Gaff:
So if that failed would the correct errors still occur, Not thatI would be 
doing this, but taking up the whole of the processor in this way is bound to 
slug bits of it?
 Brian

***@***.***
Sent via blueyonder.
in the display name field.
----- Original Message ----- 
From: "Leonard de Ruijter" ***@***.***>
Cc: "Subscribed" ***@***.***>
Sent: Monday, January 10, 2022 6:50 PM
Subject: [nvaccess/nvda] Scons: provide the number of logical processors as 
the number of maximum concurrent jobs to speed up builds (PR #13226)

seanbudd:
Is there a way to disable this for appveyor? Scrambling the output of try/PR builds and official builds without a time improvement seems like a loss.

seanbudd (approved):
Thanks @leonardder

michaelDCurran:
I'm really struggling with the merging of this pr. It is very hard to follow build output now, especially when trying to see errors from cl. This gets worse with the introduction of calling msbuild for the remote ops library.
Note that I have 12 logica processors, so there is a great deal going on at the same time.
I think parallel builds are a great advantage, but I do feel they should be optional. I.e. by specifying -j 12 (or how ever many you want) when running scons.
Pr #13367 has become necessary since this merge. And although we should probably take that one as parallel builds should work if people want them, I don't believe enforcing them by default, especially within a developer context is the right thing here. 

I feel that this pr should be reverted.

seanbudd:
@leonardder @michaelDCurran  can this be overridden with `-j1` eg `scons source -j1`? This could either be a workaround, at least until the PR is reverted. 

Would you like me to revert this? GitHub won't open a revert PR automatically

lukaszgo1:
Have you reported the scrambled output thing to SCons?? While this would probably be pretty difficult to fix not reporting the problem makes it guaranteed not to be fixed.

PR #13257, opened 2022-01-19
Don't log error for RPC_E_DISCONNECTED 

### LINKED ISSUE #13244: NVDA error when navigating subreddits with headings in Firefox
### Steps to reproduce:
1. Navigate to a subreddit, for example 
https://www.reddit.com/r/eluktronics/
Then atempt to navigate posts with headings.
### Actual behavior:
NVDA navigates headings with error tone in dev, and following segment appears in log

ERROR - RPC process 17524 (firefox.exe) (04:50:40.730) - Dummy-446 (12556):
Thread 19316, build\x86_64\vbufBackends\gecko_ia2\gecko_ia2.cpp, GeckoVBufBackend_t::fillVBuf, 955:
AccessibleChildren failed (count: 2), res: -2147417848

### Expected behavior:
NVDA navigates headings normally with no error tone and no error in log
installed
#### NVDA version:
alpha-24471,c6a6df82
#### Windows version:
windows 11  Version 21H2 (OS Build 22000.434)
Firefox 64 bit, version 96.01
yes
no
yes
yes

feerrenrut:
### Link to issue number:
Fixes #13244

### Summary of the issue:
Error sound when navigating headings and log message:
```
ERROR - RPC process 17524 (firefox.exe) (04:50:40.730) - Dummy-446 (12556):
Thread 19316, build\x86_64\vbufBackends\gecko_ia2\gecko_ia2.cpp, GeckoVBufBackend_t::fillVBuf, 955:
AccessibleChildren failed (count: 2), res: -2147417848
```
In another instance:
```
ERROR - RPC process 7696 (firefox.exe) (22:29:53.494) - Dummy-21 (7004):
Thread 6684, build\x86_64\vbufBackends\gecko_ia2\gecko_ia2.cpp, GeckoVBufBackend_t::fillVBuf, 955:
AccessibleChildren failed (count: 1), res: -2147467262
```

- `E_NOINTERFACE` is `DWORD 0x80004002L == -2147467262`
- `RPC_E_DISCONNECTED` is `DWORD 0x80010108L == -2147417848`
- `CO_E_OBJNOTCONNECTED` is `DWORD 0x800401FDL == -2,147,220,995`

Note there are two failures, one requires a fix in Firefox, the other shouldn't be logged at the error level
https://github.com/nvaccess/nvda/issues/13244#issuecomment-1015124347

Log failures due to `RPC_E_DISCONNECTED` or `CO_E_OBJNOTCONNECTED` as a debug messages, this indicates that the parent no longer exists (has been removed from the render). This can be common on dynamic pages.

Continue to log an error for failure due to `E_NOINTERFACE` and any other "unknown" failures.

### Testing strategy:
Locally ran with Firefox and Chrome.
Alpha usage.

None.

### Change log entries:
None

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
+++ b/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
@@ -960,13 +960,17 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 					// This is certainly an error, and indicates a bug in the IA2 provider.
 					LOG_ERROR(msg.str());
 					break;
-				case CO_E_OBJNOTCONNECTED:
-					msg << L" (CO_E_OBJNOTCONNECTED, Object is not connected to server)";
-					// CO_E_OBJNOTCONNECTED indicates that the parent died since the query
+				case RPC_E_DISCONNECTED:
+					msg << L" (RPC_E_DISCONNECTED, object invoked has disconnected from its clients.)";
+					// RPC_E_DISCONNECTED indicates that the parent died since the query
 					// to accChildCount, and returning no children indicates that all children
 					// died since the call to accChildCount.
 					// Even if the children had been rendered, they were removed immediately thereafter.
 					// Therfore log at debug level, this is expected to occur in dynamic content.
+					LOG_DEBUG(msg.str());
+					break;
+				case CO_E_OBJNOTCONNECTED:
+					msg << L" (CO_E_OBJNOTCONNECTED, Object is not connected to server)";
 					LOG_DEBUG(msg.str());
 					break;
 				default:
--- a/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
+++ b/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
@@ -957,38 +957,25 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 				switch (accChildRes) {
 				case E_NOINTERFACE:
 					msg << L" (E_NOINTERFACE, No such interface supported)";
-					// This is certainly an error, and indicates a bug in the IA2 provider.
-					LOG_ERROR(msg.str());
+					LOG_ERROR(msg.str());  // Indicates a bug in the IA2 provider.
 					break;
 				case RPC_E_DISCONNECTED:
 					msg << L" (RPC_E_DISCONNECTED, object invoked has disconnected from its clients.)";
-					// RPC_E_DISCONNECTED indicates that the parent died since the query
-					// to accChildCount, and returning no children indicates that all children
-					// died since the call to accChildCount.
-					// Even if the children had been rendered, they were removed immediately thereafter.
-					// Therefore log at debug level, this is expected to occur in dynamic content.
-					LOG_DEBUG(msg.str());
+					// RPC_E_DISCONNECTED indicates that the parent died since the query to accChildCount.
+					LOG_DEBUG(msg.str());  // This is expected to occur in dynamic content.
 					break;
 				case CO_E_OBJNOTCONNECTED:
 					msg << L" (CO_E_OBJNOTCONNECTED, Object is not connected to server)";
 					LOG_DEBUG(msg.str());
 					break;
-				case S_FALSE:
-					if (varChildren.size() == 0 && childCount > 0) {
-						msg << L" (S_FALSE, expected childcount, got no children, children may have been removed from document.)";
-						// Returning no children indicates that all children
-						// died since the call to accChildCount.
-						// Even if the children had been rendered, they were removed immediately thereafter.
-						// Therefore log at debug level, this is expected to occur in dynamic content.
-						LOG_DEBUG(msg.str());
-						break;
-					}
-					else {
-						// theoretically this cause can not be hit, included only for the benefit of developers
-						msg << L" (S_FALSE, unknown cause)";
-						LOG_ERROR(msg.str());
-						break;
-					}
+				case S_FALSE:  // Success, but unexpeced number of children were returned.
+					msg << L" (S_FALSE, expected childcount, got "
+						<< varChildren.size()
+						<< L". Children may have been removed from document.)";
+					// Returning no children indicates that all children died since the call to accChildCount.
+					// Even if the children had been rendered, they were removed immediately thereafter.
+					LOG_DEBUG(msg.str());  // This is expected to occur in dynamic content.
+					break;
 				default:
 					// Other unknown failures, log at error.
 					LOG_ERROR(msg.str());
--- a/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
+++ b/nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
@@ -952,7 +952,27 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 			// The object has no text, but we do want to render its children.
 			auto [varChildren, accChildRes] = getAccessibleChildren(pacc, 0, childCount);
 			if (S_OK != accChildRes || varChildren.size() == 0) {
-				LOG_ERROR(L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes);
+				std::wstringstream msg;
+				msg << L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes;
+				switch (accChildRes) {
+				case E_NOINTERFACE:
+					msg << L" (E_NOINTERFACE, No such interface supported)";
+					// This is certainly an error, and indicates a bug in the IA2 provider.
+					LOG_ERROR(msg.str());
+					break;
+				case CO_E_OBJNOTCONNECTED:
+					msg << L" (CO_E_OBJNOTCONNECTED, Object is not connected to server)";
+					// CO_E_OBJNOTCONNECTED indicates that the parent died since the query
+					// to accChildCount, and returning no children indicates that all children
+					// died since the call to accChildCount.
+					// Even if the children had been rendered, they were removed immediately thereafter.
+					// Therfore log at debug level, this is expected to occur in dynamic content.
+					LOG_DEBUG(msg.str());
+					break;
+				default:
+					// Other unknown failures, log at error.
+					LOG_ERROR(msg.str());
+				}
 			}
 			LOG_DEBUG(L"got " << varChildren.size() << L" children");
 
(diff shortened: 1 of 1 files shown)


nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -952,7 +952,27 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 			// The object has no text, but we do want to render its children.
 			auto [varChildren, accChildRes] = getAccessibleChildren(pacc, 0, childCount);
 			if (S_OK != accChildRes || varChildren.size() == 0) {
-				LOG_ERROR(L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes);
+				std::wstringstream msg;
+				msg << L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes;
+				switch (accChildRes) {
+				case E_NOINTERFACE:
```
jcsteh, line 958: nit: Should this all be indented one extra level... or are you deliberately putting the cases at the same level as the switch?

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -952,7 +952,27 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
 			// The object has no text, but we do want to render its children.
 			auto [varChildren, accChildRes] = getAccessibleChildren(pacc, 0, childCount);
 			if (S_OK != accChildRes || varChildren.size() == 0) {
-				LOG_ERROR(L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes);
+				std::wstringstream msg;
+				msg << L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes;
+				switch (accChildRes) {
+				case E_NOINTERFACE:
+					msg << L" (E_NOINTERFACE, No such interface supported)";
```
jcsteh, line 959: I don't know how useful it is to differentiate E_NOINTERFACE from other errors (as opposed to warnings), since one would have to look up the error message for any other error code anyway. That said, I don't have a specific objection.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -952,7 +952,27 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
...
+				switch (accChildRes) {
+				case E_NOINTERFACE:
+					msg << L" (E_NOINTERFACE, No such interface supported)";
+					// This is certainly an error, and indicates a bug in the IA2 provider.
+					LOG_ERROR(msg.str());
+					break;
+				case CO_E_OBJNOTCONNECTED:
+					msg << L" (CO_E_OBJNOTCONNECTED, Object is not connected to server)";
+					// CO_E_OBJNOTCONNECTED indicates that the parent died since the query
+					// to accChildCount, and returning no children indicates that all children
```
jcsteh, line 966: Returning no children isn't covered by this case. In that case, varChildren.size() == 0 will be true and the result will be S_FALSE. You'll need to cover that separately.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -952,7 +952,27 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
...
+					// died since the call to accChildCount.
+					// Even if the children had been rendered, they were removed immediately thereafter.
+					// Therfore log at debug level, this is expected to occur in dynamic content.
```
jcsteh, line 969: nit: Therfore -> Therefore

jcsteh (changes requested):
Thanks.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -952,7 +952,27 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
...
-				LOG_ERROR(L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes);
+				std::wstringstream msg;
+				msg << L"AccessibleChildren failed (count: " << childCount << L"), res: " << accChildRes;
+				switch (accChildRes) {
+				case E_NOINTERFACE:
+					msg << L" (E_NOINTERFACE, No such interface supported)";
+					// This is certainly an error, and indicates a bug in the IA2 provider.
+					LOG_ERROR(msg.str());
+					break;
+				case CO_E_OBJNOTCONNECTED:
```
jcsteh, line 963: Ah. The code COM is returning in Firefox is actually RPC_E_DISCONNECTED, not CO_E_OBJNOTCONNECTED. You should handle both the same way, though; i.e. log as warnings.

feerrenrut:
Just for reference (I'll add to the description):
- `E_NOINTERFACE` is `DWORD 0x80004002L == -2147467262`
- `RPC_E_DISCONNECTED` is `DWORD 0x80010108L == -2147417848`
- `CO_E_OBJNOTCONNECTED` is `DWORD 0x800401FDL == -2,147,220,995`

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: I only included it because we are encountering it now. Using the specific name and description should help recognizing it as the same/known issue.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: Thanks

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: This is just the way visual studio formats the code (with `clangFormat`, using VisualStudio style). We don't have some other defined standard, so I figure the default is the best approach. We should really should automatically run `clangFormat` to check for style.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: Thanks, I've updated the description with this now.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: Oh I see, it's the other kind of "and", I misread that the first time, good catch.

feerrenrut: This case `(S_FALSE, no children)` sounds like it could be specific to Firefox. I wouldn't expect any children returned in any other case with `S_FALSE`, how do we differentiate this situation from some "generic failure"?

jcsteh: I don't quite follow the reasoning here. From the [AccessibleChildren documentation](https://docs.microsoft.com/en-us/windows/win32/api/oleacc/nf-oleacc-accessiblechildren):

> S_FALSE
> The function succeeded, but there are fewer elements in the rgvarChildren array than there are children requested in cChildren.

So, if you request 1 child and you get 0 back (in this case, because one of them died), you'll get S_FALSE.

A generic failure should return an E_XXX error, not S_FALSE. S_FALSE is [explicitly not a failure](https://docs.microsoft.com/en-us/windows/win32/learnwin32/error-handling-in-com):

> The success code S_FALSE deserves mention. Some methods use S_FALSE to mean, roughly, a negative condition that is not a failure. It can also indicate a "no-op"—the method succeeded, but had no effect.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -952,7 +952,47 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
...
+				switch (accChildRes) {
+				case E_NOINTERFACE:
+					msg << L" (E_NOINTERFACE, No such interface supported)";
+					// This is certainly an error, and indicates a bug in the IA2 provider.
+					LOG_ERROR(msg.str());
+					break;
+				case RPC_E_DISCONNECTED:
+					msg << L" (RPC_E_DISCONNECTED, object invoked has disconnected from its clients.)";
+					// RPC_E_DISCONNECTED indicates that the parent died since the query
+					// to accChildCount, and returning no children indicates that all children
```
jcsteh, line 966: I'd remove everything after "accChildCount" in this comment. There are no children to return if the parent died. The "no children" is the case you handle below.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -952,7 +952,47 @@ VBufStorage_fieldNode_t* GeckoVBufBackend_t::fillVBuf(
...
+						msg << L" (S_FALSE, expected childcount, got no children, children may have been removed from document.)";
+						// Returning no children indicates that all children
+						// died since the call to accChildCount.
+						// Even if the children had been rendered, they were removed immediately thereafter.
+						// Therefore log at debug level, this is expected to occur in dynamic content.
+						LOG_DEBUG(msg.str());
+						break;
+					}
+					else {
+						// theoretically this cause can not be hit, included only for the benefit of developers
```
jcsteh, line 987: On the contrary, it can be hit if, for example:

1. accChildCount returned 2.
2. One child died.
3. AccessibleChildren will return 1 child and S_FALSE (because 2 were requested but only 1 was returned).

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: Ah looks like I conflated `S_FALSE` and `E_FAIL`, that and most code paths only check for `S_OK` treating anything else as a failure.

I'll take another pass at this.

jcsteh: For reference, Windows has the SUCCEEDED and FAILED macros, the former checking for S_OK and S_FALSE and the latter checking for E_XXX.

Most of the time in this code, it's not so much that we treat S_FALSE as failure, but more that we want no data/no-op and failure to be handled the same way; i.e. we just treat failure the same as no data/no-op. The intention was to be as fault tolerant as possible.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
feerrenrut: Looking at this again, I don't think we should try to distinguish between
- `S_FALSE` and got no children 
- `S_FALSE` and got less than `childCount`

I'm going to merge both `S_FALSE` handlers together.

jcsteh (approved):
Looks good to me. Thanks!

seanbudd (approved):

PR #7497, opened 2017-08-15
Add ConsoleWindowClass to badUIAWindowClasses as the UIA implementation for win32 Console windows is not usable yet.

michaelDCurran:
### Link to issue number:
None

### Summary of the issue:
In Windows 10 build 16257, Microsoft introduced a UIA implementation for Win32 console windows. Not only does this conflict with our existing support, their UIA implementation is not yet complete (I.e. caret support is lacking, and most likely no useful events for new text).
As this implementation will be in the Fall Creaters update released before NVDA 2017.4, we should for now ignore this UIA implementation, and revisit how we should handle it properly if the implementation improves.

### Description of how this pull request fixes the issue:
ConsoleWindowClass is added to _UIAHandler.badUIAWindowClasses.

### Testing performed:
* Started a cmd prompt from the run dialog
* Started Ubuntu from the Start Screen
* Started NVDA from an existing Ubuntu session

All of these tests showed NVDA functioning the way it used to before the new UIA implementation in 16257.

### Known issues with pull request:
None.

### Change log entry:
Probably none needed as getting this in 2017.3 before the general public notice would be good.

PR #13261, opened 2022-01-20
UI Automation in Windows Console and Windows Terminal: block SV2M2 notification events for now and preserve accessibility after UIA class name change

codeofdusk:
### Link to issue number:
Related to microsoft/terminal#12358 among others (to be released).

### Summary of the issue:
In upcoming Windows Terminal and Windows Console, UIA notification events will be sent when new text is inserted (i.e. for text output) to improve Narrator support. This will result in double-reporting of all terminal output: once from `LiveText` and once from UIA notifications (but without appropriate filtering for typed characters/passwords etc).

As part of the implementation of notifications, I asked Microsoft to change the UIA class name to allow new terminal support (including notifications) in a follow-up NVDA PR, since notifications will require a departure from the terminal strategy used by NVDA in the past. Changing the class name will break NVDA's current ability to identify the terminal and implement accessibility.

* Suppress any received UIA notification events in UIA console and terminal for now, but log them for development.
* Support the new UIA class name used by terminals that send notifications.

Ideally, NVDA would use these events in place of `LiveText`. If we could get away without registering for `TextChange` at all, this could be an extreme stability improvement in the terminal as #11002 would be completely circumvented. Also, having new text pushed to us (instead of having to diff) should improve performance considerably in high-volume scenarios.

### Testing strategy:
Tested that support works as expected with the new terminal.

It is imperative that this PR be included in 2022.1, as upcoming builds of Windows Terminal will have severely degraded accessibility without it.

### Change log entries:
None (workaround for bug in unreleased code)

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -1156,7 +1156,12 @@ class UIA(Window):
 		if self.windowClassName == "ConsoleWindowClass":
 			from . import winConsoleUIA
 			winConsoleUIA.findExtraOverlayClasses(self, clsList)
-		elif UIAClassName == "TermControl":
+		elif UIAClassName in ("TermControl", "TermControl2"):
+			# microsoft/terminal#12358: Eventually, TermControl2 should have
+			# a separate overlay class that is not a descendant of LiveText.
+			# TermControl2 sends inserted text using UIA notification events,
+			# so it is no longer necessary to diff the object as with all
+			# previous terminal implementations.
 			from . import winConsoleUIA
 			clsList.append(winConsoleUIA.WinTerminalUIA)
 
--- a/source/NVDAObjects/UIA/winConsoleUIA.py
+++ b/source/NVDAObjects/UIA/winConsoleUIA.py
@@ -400,6 +400,17 @@ class WinConsoleUIA(KeyboardHandlerBasedTypedCharSupport):
 				"probably due to a switch to/from the alt buffer."
 			), exc_info=True)
 
+	def event_UIA_notification(self, **kwargs):
+		"""
+		In Windows Sun Valley 2 (SV2 M2), UIA notification events will be sent
+		to announce new text. Block these for now to avoid double-reporting of
+		text changes.
+		@note: In the longer term, NVDA should leverage these events in place
+		of the current LiveText strategy, as performance will likely be
+		significantly improved and #11002 can be completely mitigated.
+		"""
+		log.debugWarning(f"Notification event blocked to avoid double-report: {kwargs}")
+
 
 def findExtraOverlayClasses(obj, clsList):
 	if obj.UIAElement.cachedAutomationId == "Text Area":
@@ -409,4 +420,13 @@ def findExtraOverlayClasses(obj, clsList):
 
 
 class WinTerminalUIA(EnhancedTermTypedCharSupport):
-	pass
+	def event_UIA_notification(self, **kwargs):
+		"""
+		In an upcoming terminal release, UIA notification events will be sent
+		to announce new text. Block these for now to avoid double-reporting of
+		text changes.
+		@note: In the longer term, NVDA should leverage these events in place
+		of the current LiveText strategy, as performance will likely be
+		significantly improved and #11002 can be completely mitigated.
+		"""
+		log.debugWarning(f"Notification event blocked to avoid double-report: {kwargs}")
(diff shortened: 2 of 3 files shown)


josephsl:
Hi, for reference (for others): SV2 = Sun Valley 2 a.k.a. Windows 11 Version 22H2. I assume this was tested in build 22500 series, specifically 22533 and later? Thanks.

codeofdusk:
Not yet, staying ahead of upcoming changes.

CC @carlos-zamora, @DHowett.

codeofdusk:
CC @seanbudd.

seanbudd:
Is there a way to confirm that the console/terminal changes are guaranteed to be implemented?
Is there a way to test this PR with the proposed console/terminal changes?

codeofdusk:
> Is there a way to confirm that the console/terminal changes are guaranteed to be implemented?

"guarantee" is hard, but `wt` will become default, Narrator support is blocking, and notifications is the current path to enable this, so it is all but certain at this point.

> Is there a way to test this PR with the proposed console/terminal changes?

I don't yet have access to a build with notifications available, but I've tested this event handler in other contexts that emit UIA notifications (Word with UIA) and it works as intended. Merging this PR is strictly better than the current behaviour and results in no change in situations where UIA notifications are not emitted.

seanbudd:
I think it might be worth holding off on merging this until a Windows Insider build comes that it can be tested with.
Considering this is a small, low risk change, it can be merged into alpha late in the release cycle (eg just before we release beta).

codeofdusk:
I'll work on obtaining a Windows Terminal build with which I can test this. Thanks for adding to the 2021.1 release.

josephsl:
Hi, also, ping me when that build is available so I can test it on my VM (at least build 22543)). Thanks.

carlos-zamora:
Hi! I'm over from the Console Engineering team at Microsoft. I just wanted to follow up on some things from this thread.

> `wt` will become default

Yes! We're looking to do this this year! Here's our blog post announcement: https://devblogs.microsoft.com/commandline/windows-terminal-as-your-default-command-line-experience/

> Is there a way to confirm that the console/terminal changes are guaranteed to be implemented? Is there a way to test this PR with the proposed console/terminal changes?

We're making progress. I'll be working on creating a prototype this month and sharing that with Bill, NVDA, and JAWS (I'll be meeting with them next week to provide support). We're hoping that this'll be the fastest way to get Narrator on board since this change will have a pretty severe impact on Narrator users.

Like Bill, I hope to take a look at what other text providers (like Word) are doing. So if this works with Word already, that's very promising.

CC @DHowett

codeofdusk:
I can confirm that, with the build in microsoft/terminal#12358, duplicate (buggy) output is present without this PR and absent (correct behaviour) with this PR.

codeofdusk:
Now that this is known to work with an actual implementation, can this PR please be merged?

seanbudd (approved):
Thanks @codeofdusk. It's great that we can prepare for this change early, so that users won't have to update NVDA from 2022.1 to fix this.

Do we have an estimate at what time or which Windows releases this will be incorporated into?
I think it would be helpful to add a changelog item, as users on older versions of NVDA will want to search for which NVDA version fixes this bug.

codeofdusk:
It will arrive in the Sun Valley 2 console and a Windows Store update. Can't speak to anything beyond that sadly. (there is precident for making fixes to prepare for unreleased software without changelog items, see #7497).

seanbudd:
@codeofdusk Thanks, perhaps for users "Windows 11" is specific enough. It is helpful to add bug fixes to the changelog where possible.
Would it be accurate to add the following:

```t2t
- Prevents a bug causing double-reporting when using Windows 11 Console and Terminal. (#13261)
```

codeofdusk:
I'd remove "11" as I think they'll backport the terminal code to vibranium at least.

josephsl:
Hi, for reference, Vibranium = Windows 10, so a compromise could be to say “recent Windows 10 and 11” (I understand that the what’s new entry was committed, but hope this could be handy in the future). Thanks.

seanbudd:
> Hi, for reference, Vibranium = Windows 10, so a compromise could be to say “recent Windows 10 and 11” (I understand that the what’s new entry was committed, but hope this could be handy in the future). Thanks.

I think the problem here is that "recent" is relative and won't be helpful in the future. With the uncertainty surrounding which versions this will be included in, as @codeofdusk suggested potentially earlier builds to Vibranium, I opted to just go with "Windows".

josephsl:
Hi, makes sense, thanks.

codeofdusk:
Please note that, due to pushback from JAWS, the `TermCTRL2` class name is currently not used. However NVDA should maintain support for it in case this changes in the future.

PR #13268, opened 2022-01-21
Force DictionaryDialog to be subclassed to use it

CyrilleB79:
### Link to issue number:

Follow-up of #12967

### Summary of the issue:

Since the merge of #12800, trying to re-open an already opened dialog allows to focus it rather than issuing an error message box. The dialog is refocused according to its class, i.e. trying to open a dialog of the same class rather refocus the first already opened dialog of the same class.

But with #12800, it was not possible to open 2 dictionaries of different type (e.g. speech and default). This was fixed with #12967 by subclassing DictionaryDialog.
In #12967, however, the following known issue was mentionned:
> Maybe it would be interesting to forbid to instance an object of class DictionaryDialog directly, i.e. force subclassing.

Indeed, if DictionaryDialog's have been subclassed in NVDA's core, add-on may still instance directly a DictionaryDialog. That should be forbidden as explained in #12967.

However, it could not be done before 2022.1 due to add-ons that may break if they instance directly DictionaryDialog. That's the case of applicationDictionary.

Note: emoticons and EnhancedDictionaries add-ons use DictionaryDialog's but they already subclass it.

Option 1 (withdrawn after review):
Raise an error in the `__init__` method of DictionaryDialog if it is instancied from the same class (not a subclass).

Option 2 (implemented after review):
I have abstracted DictionaryDialog as advised in https://github.com/nvaccess/nvda/pull/12967#issuecomment-948089531 and thanks to @lukaszgo1's review.

### Testing strategy:
Manual testing:
1. Created aglobal plugin ([dictTest.py.txt](https://github.com/nvaccess/nvda/files/7919429/dictTest.py.txt)) allowing to open two dictionaries:
   * One directly created from DictionaryDialog
   * One created from a subclass of DictionaryDialog
   And tested that:
   * both dictionaries open with NVDA 2021.3.1
   * Only the subclassed dictionary opens with this PR's code (from source)

2. Tested that NVDA's default, voice and temp dictionaries dialogs still open (from menu).

Should be merged in 2022.1 or any API-breaking release.

### Change log entries:
For Developers

`It is now mandatory to subclass DictionaryDialog to use it. (#13268)`

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -3190,13 +3190,21 @@ class DictionaryEntryDialog(
 		self.typeRadioBox.SetSelection(DictionaryEntryDialog.TYPE_LABELS_ORDERING.index(type))
 
 
-class DictionaryDialog(SettingsDialog):
+class DictionaryDialog(
+	SettingsDialog,
+	metaclass=guiHelper.SIPABCMeta,
+):
+	"""A dictionary dialog.
+	A dictionary dialog is a setting dialog containing a list of dictionary entries and buttons to manage them.
+	
+	To use this dialog, override L{__init__} calling super().__init__.
+	"""
+	
 	TYPE_LABELS = {t: l.replace("&", "") for t, l in DictionaryEntryDialog.TYPE_LABELS.items()}
 	helpId = "SpeechDictionaries"
 
+	@abstractmethod
 	def __init__(self,parent,title,speechDict):
-		if type(self) == DictionaryDialog:
-			raise Exception("DictionaryDialog must be subclassed.")
 		self.title = title
 		self.speechDict = speechDict
 		self.tempSpeechDict=speechDictHandler.SpeechDict()
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -3191,8 +3191,8 @@ class DictionaryEntryDialog(
 
 
 class DictionaryDialog(
-	SettingsDialog,
-	metaclass=guiHelper.SIPABCMeta,
+		SettingsDialog,
+		metaclass=guiHelper.SIPABCMeta,
 ):
 	"""A dictionary dialog.
 	A dictionary dialog is a setting dialog containing a list of dictionary entries and buttons to manage them.
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -3195,6 +3195,8 @@ class DictionaryDialog(SettingsDialog):
 	helpId = "SpeechDictionaries"
 
 	def __init__(self,parent,title,speechDict):
+		if type(self) == DictionaryDialog:
+			raise Exception("DictionaryDialog must be subclassed.")
 		self.title = title
 		self.speechDict = speechDict
 		self.tempSpeechDict=speechDictHandler.SpeechDict()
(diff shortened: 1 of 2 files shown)


source/gui/settingsDialogs.py
```diff
@@ -3195,6 +3195,8 @@ class DictionaryDialog(SettingsDialog):
 	helpId = "SpeechDictionaries"
 
 	def __init__(self,parent,title,speechDict):
```
lukaszgo1, line 3197: Have you considered making the constructor abstract rather than checking type of the dialog? This would be way more Pythonic IMO.

CyrilleB79:
@seanbudd, as the reviewer of #12967 and the author of #12800, you may be interested in reviewing this one.
@CyrilleB79

seanbudd:
Note: I've unchecked "API is compatible with existing add-ons, please leave this unmarked as this is a documented breaking change"

seanbudd (approved):

CyrilleB79:
Thanks for review @seanbudd .

Discussing on process details, you wrote:
> Note: I've unchecked "API is compatible with existing add-ons", please leave this unmarked as this is a documented breaking change

Sorry, that's not the way I was told to use these checkboxes (by Reef in another PR).

Moreover if I read `githubPullRequestTemplateExplanationAndExamples.md`:
> The aim of this checklist is to ensure each item has been considered by both the author and the
reviewer.
and
> Not all items will be applicable for all situations, in this case checking the item lets reviewers
know it's been considered.

Since I have considered the API-breaking change topic, I think I should check the checkbox to indicate that it has been considered.

And in the section _API is compatible with existing add-ons_:
> Ensure proposed API changes are included in the change log (Changes for Developers).
This let me think I should check the checkbox _API is compatible with existing add-ons_, since I have actually provided the required change log item.

Just to clarify how I have used these checkbox and to highlight that there may be a misunderstanding about their use.

seanbudd:
@CyrilleB79  - apologies, my mistake. Perhaps we should improve the wording of the checklist item, it is somewhat misleading for it to be checked in this instance.

CyrilleB79:
In the `PULL_REQUEST_TEMPLATE.md` which is the content of the edit field when you open a pull-request, for the _Code Review Checklist_ paragraph, you have the following comment:

> 

This seem quite clear to me, as a PR author/submitter. But of course, this paragraph is not visible for the reviewer who reads the PR.

Do you have an idea on how this can be made clearer for reviewer? What would you expect as a reviewer? Maybe a sentence out of the comment telling "Author should check each checkbox to indicate that the relevance of the item has been considered; reviewer should ensure that all checkbox are checked." I fear however that this wording makes the template to heavy.
What do you think?

seanbudd:
@CyrilleB79 

I was misled after coming back from break and forgetting the context we added to the template. What I find misleading is "API is compatible with existing add-ons." is not a true statement if there is API breaking changes.

Perhaps it could be "API is compatible with existing add-ons or breaking changes have been documented" or the shorter "API compatibility has been considered"

PR #13277, opened 2022-01-25
Add handling for accSelection returning IDispatch or negative child ID

### LINKED ISSUE #13276: Exception when only a single cell is selected in LibreOffice >= 7.3
### Steps to reproduce:

1) start NVDA from current git master
2) start LibreOffice Calc (version 7.3 or later; issue not reproducible with older versions because the relevant code path that triggers the issue is not taken)
3) press Shift+Right to select cells A1 and B1
4) press Shift+Left to leave only cell A1 selected

### Actual behavior:

An error occurs, NVDA "beeps".

Traceback from the NVDA log:

```
NVDA initialized
ERROR - eventHandler.executeEvent (15:50:59.902) - MainThread (6012):
error executing event: selectionRemove on <NVDAObjects.Dynamic_SymphonyIATableCellEditableTextWithAutoSelectDetectionIAccessible object at 0x0E0AB4F0> with extra args of {}
  File "eventHandler.py", line 284, in executeEvent
    _EventExecuter(eventName,obj,kwargs)
  File "eventHandler.py", line 98, in __init__
    self.next()
  File "eventHandler.py", line 107, in next
    return func(*args, **self.kwargs)
  File "appModules\soffice.py", line 196, in event_selectionRemove
    self.event_selectionAdd()
  File "appModules\soffice.py", line 193, in event_selectionAdd
    curFocus.announceSelectionChange()
  File "appModules\soffice.py", line 204, in announceSelectionChange
    reason=controlTypes.OutputReason.CHANGE
  File "speech\speech.py", line 402, in speakObjectProperties
    **allowedProperties,
  File "speech\speech.py", line 453, in getObjectPropertiesSpeech
    newPropertyValues[name]=getattr(obj,tryName)
  File "baseObject.py", line 42, in __get__
    return instance._getPropertyViaCache(self.fget)
  File "baseObject.py", line 146, in _getPropertyViaCache
    val=getterMethod(self)
  File "appModules\soffice.py", line 174, in _get_states
    if self.hasSelection:
  File "baseObject.py", line 42, in __get__
    return instance._getPropertyViaCache(self.fget)
  File "baseObject.py", line 146, in _getPropertyViaCache
    val=getterMethod(self)
  File "appModules\soffice.py", line 164, in _get_hasSelection
    and 1 < self.selectionContainer.getSelectedItemsCount()
  File "NVDAObjects\IAccessible\__init__.py", line 1342, in getSelectedItemsCount
    return self._getSelectedItemsCount_accSelection(maxCount)
  File "NVDAObjects\IAccessible\__init__.py", line 1318, in _getSelectedItemsCount_accSelection
    enumObj=sel.QueryInterface(IEnumVARIANT)
AttributeError: 'int' object has no attribute 'QueryInterface'
```

### Expected behavior:

Corrdinates of cell A1 should be announced after step 4.

running from source

#### NVDA version:

self-compiled from git master as of commit 7735d96259d02eb8cc8d84f13a870ecfec0635b9

#### Windows version:

Windows 10 Version 21H2 (OS Build 19044.1466)

LibreOffice (version >= 7.3, in my case current development version self-compiled from git master as of commit 86a9c860410351125708543c1050bfc0977d3b97)

n/a

no

n/a

n/a

michaelweghorn:
### Link to issue number:

Fixes #13276

### Summary of the issue:

In case a single child is selected,
`IAccessible::get_accSelection` can either return
the child ID of a selected simple element or
an `IDispatch` for a selected accessible object, s.
https://docs.microsoft.com/en-us/windows/win32/api/oleacc/nf-oleacc-iaccessible-get_accselection
and
https://docs.microsoft.com/en-us/windows/win32/winauto/how-servers-implement-child-ids .

Previously, only positive child IDs were considered
in NVDA, while LibreOffice was returning negative child IDs
for Calc cells, which does not fit the MSAA spec.

NVDA did not yet support the case where an
`IDispatch` is returned.

This PR implements the missing handling for an
`IDispatch` returned as accSelection in NVDA.
LibreOffice was adapted to return such a one from
LibreOffice 7.4 on, corresponding LibreOffice commit:
https://git.libreoffice.org/core/commit/00c0ee8cf0fac0c933c5ae600e99a64b1c7d4397

In addition, negative child IDs are now also accepted.
As pointed out by James Teh in the discussion on the
initial version of the pull request:

> The use of negative child ids doesn't fit well into the IAccessible spec, but
> it has been done by IAccessible2 for a very long time and should be considered
> standard for all intents and purposes. A negative child id should be treated as
> a unique id, while a positive child id should be treated as a child index.

Supporting negative child IDs fixes the issue for LibreOffice 7.3.

### Testing strategy:

Test the steps from issue #13276 and also make sure that the IDispatch case
works in NVDA when only the single cell A1 is selected:

1) start NVDA
2) start LibreOffice Calc 7.3
3) press Shift+Right to select cells A1 and B1
4) press Shift+Left to leave only cell A1 selected
5) verify that cell A1 is properly announced
6) set a breakpoint in the method changed by this PR and repeat 
   the previous steps with LibreOffice 7.4 (current development version) 
   or newer and verify that the IDispatch case is actually handled properly
   in NVDA, which is not the case without this PR

(For LibreOffice 7.4, announcement for the case described in
issue #13276 already works fine from a user perspective without
this PR, but just the LibreOffice change in place: An `IDispatch` is
returned by LibreOffice, the previous exception described in #13276
will no longer be triggered, but a `COMError` is triggered when querying
for the `IEnumVARIANT` interface a few lines below. Since that `COMError`
is handled in `getSelectedItemsCount` (in `source\NVDAObjects\IAccessible\__init__.py`)
and it falls back to calling the `nSelected` methods on the `IAccessibleTable2 interface`
instead, the user-facing end result is still OK.)

none

### Change log entries:
*Bug fixes*
`When retrieving the count of selected children via accSelection, the case where a negative child ID or an IDispatch is returned by IAccessible::get_accSelection is now handled properly. (#13276)`

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


michaelweghorn:
The code (including the initial check that is adapted in this PR) was added in

```
commit b02ed2de682ec2b323282b8e4f67d60a825dd8e2
Author: Michael Curran <michaelDCurran@users.noreply.github.com>
Date:   Wed Oct 31 16:56:51 2018 +1000

    Fix up of: Don't announce 'selected' when the focus moves in Google sheets if the focused cell is the only cell selected (#8898)
    
    * Don't announce 'selected' when the focus moves in Google sheets if the focused cell is the only cell selected.
    
    * Address review comments.
    
    * Address review comments.
    
    * Update what's new
    
    * Limit logic that stops announcing selected for a single selected item, and announces 'not selected' for not selected items, to controls that have the focusable state.
    
    * IAccessible NVDAObject's getSelectedItemsCount_accSelection: handle a childID coming from accSelection, as found in QT tree tables like in Mumble

```
but I (as a non-expert) see no obvious reason to allow for positive child IDs here only.

source/NVDAObjects/IAccessible/__init__.py
```diff
@@ -1313,7 +1313,7 @@ def _getSelectedItemsCount_accSelection(self, maxCount: int) -> int:
 			raise NotImplementedError
 		# accSelection can return a child ID of a simple element, for instance in QT tree tables. 
 		# Therefore treat this as a single selection.
-		if isinstance(sel,int) and sel>0:
+		if isinstance(sel, int):
 			return 1
```
leonardder, line 1317: Shouldn't this still compensate for 0? I don't think we should treat CHILDID_SELF as a selection within the container.

```suggestion
			return 0 if sel == 0 else 1
```

michaelweghorn: You're right, thanks!

LeonarddeR (approved):

michaelweghorn:
Since I'm not familiar with how things work exactly, I've updated the branch/PR to incorporate the suggested change in the original commit, and updated commit message and the PR description.
From other PRs, it seems though, that the commits are finally squashed into a single commit anyway.

Sorry if that was not the right way...

For the future: What's the "right" way to address comments in the review process and what should I update so that the commit message of the final commit is an up to date one? Is that taken from the actual commit(s) or from the pull request description, so it's mostly important to keep that up to date and the single commits of the PR don't necessarily have to be "nice" because they're squashed into a single one anyway?

leonardder:
@feerrenrut May be you can answer this question?
As far as I know, commits are indeed squashed and the pr description is used to create the description of the commit, but I'm not 100% sure.

feerrenrut:
Yes, in most cases we do a squash merge and copy information from the description (with some editorializing).

The only cases we don't do that:
- Merges between beta and master
- For code refactors, where individual commits helps to track moved AND modified code. In this case the commits for modification of the code should be isolated from the commits that move code.

feerrenrut:
What does a negative selection ID mean? That doesn't sound like a valid object.

Unless you are sure and can explain why a negative Id would be returned, this should continue to be treated as an error. This may require new code, specific to OpenOffice / LibreOffice specific to handle this error. 

This sounds like a bug in LibreOffice, what is the negative value? Could it be an integer overflow?

It would be helpful if @michaelDCurran could confirm please?

michaelweghorn:
Thanks for explaining the workflow!

> What does a negative selection ID mean? That doesn't sound like a valid object.
>
> Unless you are sure and can explain why a negative Id would be returned, this should continue to be treated as an error. This may require new code, specific to OpenOffice / LibreOffice specific to handle this error.
> 
> This sounds like a bug in LibreOffice, what is the negative value? Could it be an integer overflow?

The central question seems to be whether or not negative child IDs are valid or not. LibreOffice actually [assigns negative child IDs](https://git.libreoffice.org/core/+/refs/tags/libreoffice-7.3.0.3/winaccessibility/source/service/ResIDGenerator.cxx#43) (that seem to be unique application-wide) and since I did not find any statement when reading the [IAccessible::get_accSelection](https://docs.microsoft.com/en-us/windows/win32/api/oleacc/nf-oleacc-iaccessible-get_accselection) or the ["How Child IDs Are Used in Parameters"](https://docs.microsoft.com/en-us/windows/win32/winauto/how-child-ids-are-used-in-parameters) doc, I was assuming that was probably valid, and only 0 would be a special value (meaning `CHILDID_SELF`).

Now, I've found [a "How Servers Implement Child IDs" page in the MSAA docs](https://docs.microsoft.com/en-us/windows/win32/winauto/how-servers-implement-child-ids), which says:

> # How Servers Implement Child IDs
>
> Server developers can assign child IDs to both simple elements and accessible objects. However, the recommended approach is to support the standard Component Object Model (COM) interface IEnumVARIANT in every accessible object that has children.
> 
> If you implement IEnumVARIANT, you must:
>
> * Enumerate all children, both simple elements and accessible objects. Provide child IDs for all simple elements and provide the IDispatch to each accessible object.
> * For accessible objects, set the vt member of the VARIANT to VT_DISPATCH. The pdispVal member must contain a pointer to the IDispatch interface. Note that the VARIANT is allocated and freed by the client.
> * For simple elements, the child ID is any 32-bit positive integer. Note that zero and negative integers are reserved by Microsoft Active Accessibility. Set the VARIANT structure vt member to VT_I4 and the lVal member to the child ID.
>
> If you do not support IEnumVARIANT, you must assign child IDs and number the children in each object sequentially starting with one.

In the LibreOffice Calc case that #13276 is about, the children (table cells) are accessible objects, not just simple elements.
Given that, it currently seems to me like it would probably be reasonable to have `get_accSelection` return a variant with `vt=VT_DISPATCH` and the `pdispVal` member set to the IDispatch interface of the accessible object representing the cell, rather than just the child ID (`vt=VT_IA` and the `lVal` member set to the child ID).

In addition to changes on LibreOffice side, this might need handling the case where `get_accSelection` returns an IDispatch for a single selected child object here on NVDA side. (But given I'm not very experienced with NVDA development, I don't yet know exactly how that would end up here "on the Python side"). From how I read the [`get_accSelection` doc](https://docs.microsoft.com/en-us/windows/win32/api/oleacc/nf-oleacc-iaccessible-get_accselection), returning an IEnumVARIANT is primarily meant to be used when multiple children are selected.

What do you think?

If you agree this would be a reasonable approach, I would start working on the LibreOffice part first, then potentially come up with an updated PR (if changes in NVDA are needed in addition).

leonardder:
I think an opinion from either @jcsteh or @michaelDCurran would be very welcome here, since they've been working with IA2 for ages. As far as I know, one way to uniquely identify an IA2 object is the IAccessible2 uniqueID, which can certainly be negative. I assume with Child ID, you'd have to be compliant with the MSAA spec so avoid negative IDs if possible.

jcsteh:
The use of negative child ids doesn't fit well into the IAccessible spec, but it has been done by IAccessible2 for a very long time and should be considered standard for all intents and purposes. A negative child id should be treated as a unique id, while a positive child id should be treated as a child index.

That said, as noted in https://github.com/nvaccess/nvda/pull/13277#issuecomment-1024134814, IAccessible2 elements are *always* full IAccessible objects, not "simple elements". Thus, anything that returns an accessible (including accSelection) really should return an object pointer. In the case of accSelection, this means VT_DISPATCH for a single selection or VT_UNKNOWN and iEnumVARIANT (with VT_DISPATCH elements) for multiple selection.

In short, NVDA supporting negative child ids returned from accSelection isn't necessarily "wrong", but ideally, LibreOffice would be fixed to return full objects.

michaelweghorn:
Thanks a lot for the explanation!

I have implemented returning an IDispatch for the selected object in LibreOffice (s. https://gerrit.libreoffice.org/c/core/+/129201 , will merge to master branch later today) and updated this PR to support the IDispatch case on NVDA side.

I'll also update PR #13233 to handle the IDispatch case in addition to the one where child IDs are returned (but that's independent of this PR here).

michaelweghorn:
Strictly speaking, this change is not needed to make the announcement of a single selected cell work in LibreOffice, since that already works with only the LibreOffice change in place: An IDispatch is returned by LO, the previous exception will no longer be triggered, but a COMErrror is triggered when querying for the IEnumVARIANT interface a few lines below.
Since that COMError is handled in `getSelectedItemsCount` (in `source\NVDAObjects\IAccessible\__init__.py`)  and it falls back to calling the `nSelected` methods on the `IAccessibleTable2` interface, the result is still OK, but I think NVDA should properly handle the IDispatch case.

michaelweghorn:
>     * FAIL: System tests. See test results for more information.

In the log, I see this test failure:

```
Robot.chromeTests :: HTML test cases in Chrome                                
==============================================================================
checkbox labelled by inner element :: A checkbox labelled by an in... | FAIL |
Unable to locate 'before sample' marker. Too many attempts looking for 'Before Test Case Marker' See NVDA log for full speech.
```

That test passes for me locally when I run `./runsystemtests.bat`. Could it be that this is some failure unrelated to the PR or is there anything I can do to reproduce that locally?

feerrenrut:
Sorry about that @michaelweghorn, there is an intermittent problem that affects the chrome tests. It affects the first test to be run, which is currently that one. We plan to investigate it. Until that is fixed we have to manually re-run the build or you can push an empty commit to re-run.

I'll manually re-run this PR build. The status should update soon.

michaelweghorn:
@feerrenrut: Thanks!

michaelDCurran (approved):

michaelDCurran:
@michaelweghorn  I see the change to support IDispatch, but there is no change to support negative child IDs... the original line still only checks that the int is positive. Is this change no longer needed? If so then the pr description (including proposed change log line) should be updated before I can merge this.

michaelweghorn:
> @michaelweghorn I see the change to support IDispatch, but there is no change to support negative child IDs... the original line still only checks that the int is positive. Is this change no longer needed? If so then the pr description (including proposed change log line) should be updated before I can merge this.

@michaelDCurran:This is no longer needed, since LibreOffice now (from LibreOffice 7.4 on) returns an IDispatch instead of negative child IDs, which conforms better to the MSAA spec. I have updated the PR description accordingly.

michaelweghorn:
> @michaelDCurran:This is no longer needed, since LibreOffice now (from LibreOffice 7.4 on) returns an IDispatch instead of negative child IDs, which conforms better to the MSAA spec. I have updated the PR description accordingly.

Since I was mostly focusing on the LibreOffice case, I took the feedback from https://github.com/nvaccess/nvda/pull/13277#issuecomment-1024622871 to fix LibreOffice and adapt this PR to handle the IDispatch case as well.
If considered useful, I can also re-add support for negative child IDs given @jcsteh 's comment still mentions
> The use of negative child ids doesn't fit well into the IAccessible spec, but it has been done by IAccessible2 for a very long time and should be considered standard for all intents and purposes. A negative child id should be treated as a unique id, while a positive child id should be treated as a child index.

What do you think?

michaelDCurran:
Yes, I think it would be a good idea to add that extra change to this 
pr. I.e. changing the `> 0` to `!= 0`.

michaelweghorn:
Thanks, I've added a commit doing that on top.

michaelDCurran (approved):

PR #13278, opened 2022-01-25
Disable error sounds even for testing builds

beqabeqa473:
### Summary of the issue:
It is not possible for now to disable error sounds for testing builds of NVDA, however special setting was added some time ago and disabling error sounds completely could also be included.

Added one more config value for disabling sound;
Added corresponding item in GUI combobox;
In logHandler added a check for corresponding "No" option of playing error sounds

### Testing strategy:
Manually tested and confirmed that setting "No" in play error sounds option disables it completely.

### Change log entries:
New features
Changes
It is no possible to disable logging error sounds also for testing copies of NVDA.
Bug fixes
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/config/configSpec.py
+++ b/source/config/configSpec.py
@@ -280,8 +280,8 @@ schemaVersion = integer(min=0, default={latestSchemaVersion})
 [featureFlag]
 	# 0:default, 1:yes, 2:no
 	cancelExpiredFocusSpeech = integer(0, 2, default=0)
-	# 0:Only in test versions, 1:yes
-	playErrorSound = integer(0, 1, default=0)
+	# 0:Only in test versions, 1:yes, 2: no
+	playErrorSound = integer(0, 2, default=0)
 """
 
 #: The configuration specification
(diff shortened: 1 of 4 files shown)


leonardder:
I'm afraid we can't follow up with this. It has explicitly been stated by @feerrenrut in https://github.com/nvaccess/nvda/issues/12672#issuecomment-886400162 that this behavior is undesirable.

> We really want to discourage relying on alpha versions of NVDA in "production" environments. If there are annoying errors like this, users should fall pack to a prior final release. Either have the prior stable release available as a portable install, or use the alpha as a portable install.
> Disabling the error sound makes it too easy to forget to enable it again. We also don't want to train users to be complacent about error sounds, turning off error sounds means that errors in different situations won't be reported.

PR #13283, opened 2022-01-27
MS Word with UIA: use custom extensions to support line number,  layout column number and section number

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
When supporting Microsoft word via UI automation, up to now there was no way to report line numbers, layout column numbers or section numbers. This was because these concepts do not exist in the UI Automation standard.
However, Microsoft recently added custom patterns to UI automation, as a part of its Custom Extensions work, which included custom properties and custom annotations.
Microsoft word has now added its own custom pattern which allows fetching information from a text range, such as line number, layout column number and section number.

Using the microsoft/microsoft-ui-uiautomation Remote Operations library, fetch line number, layout column number and section number in Microsoft word.
Note that the Remote Operations library is only supported in Windows 11 and above, and the custom pattern in Microsoft Word is only available as of build 16.0.14725.

A part from adding support for these specific attributes, this PR also acts as a first commit to add integrations with the UI Automation Remote Operations library. Now that we are building and including this library, there is much more code that can eventually be remoted for a massive speedup.

### Testing strategy:
* Opened a document in Microsoft word
* Turned on reporting of line numbers in NVDA's document formatting settings
* Moved up and down the document with the arrow keys, ensuring that NVDA spoke the current line number before the content of the line.
* Ensured report pages was turned on in NVDA's document formatting settings.
* Set up the document in 2-column layout, and inserted a column break in the content.
* used the  arrow keys to move between the two columns, ensuring that the column number is announced as it changes. 

None known.

### Change log entries:
New features
Changes
Bug fixes
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


.gitmodules
```diff
@@ -30,3 +30,9 @@
 [submodule "include/Detours"]
 	path = include/detours
 	url = https://github.com/microsoft/Detours/
+[submodule "include/microsoft-ui-uiautomation"]
+	path = include/microsoft-ui-uiautomation
+	url = https://github.com/michaeldcurran/microsoft-ui-uiautomation
```
codeofdusk, line 35: Are you intending to refer to your fork here, and not microsoft/microsoft-ui-uiautomation?

michaelDCurran:
At this stage yes. I'm still waiting on some PRs to get merged.

source/UIAHandler/remote.py
```diff
@@ -0,0 +1,27 @@
+# A part of NonVisual Desktop Access (NVDA)
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+# Copyright (C) 2021-2022 NV Access Limited, Zahari Yurukov, Babbage B.V., Joseph Lee
```
lukaszgo1, line 4: This list of contributors is incorrect - probably it was mistakenly copied from some other file
```suggestion
# Copyright (C) 2021-2022 NV Access Limited
```

.gitmodules
michaelDCurran: Yes, there is still one PR I'm waiting to be merged to upstream.

nvdaHelper/UIARemote/UIARemote.cpp
```diff
@@ -0,0 +1,154 @@
...
+    it under the terms of the GNU General Public License version 2.0, as published by
+    the Free Software Foundation.
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+This license can be found at:
+http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+*/
+
+#define _SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING
```
feerrenrut, line 15: Could you add a comment with an example warning and where you see it? Makes it easier to work out when/if this is no longer required.

feerrenrut (changes requested):
There are a few build warnings introduced, these may be outside your control, but it would be good if they were silenced:
```
"C:\work\repo\nvda\PR\include\microsoft-ui-uiautomation\src\uiAutomation\microsoft.ui.uiautomation\Microsoft.UI.UIAutomation.vcxproj" (Build target) (1) ->
(PrepareForBuild target) ->
  C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppBuild.targets(499,5): warning MSB8004: Output Directory does not end with a trailing slash.  This build instance will add the slash as it is required to allow proper evaluation of the Output Directory. [C:\work\repo\nvda\PR\include\microsoft-ui-uiautomation\src\uiAutomation\microsoft.ui.uiautomation\Microsoft.UI.UIAutomation.vcxproj]
```

nvdaHelper/microsoft-ui-uiautomation/sconscript
```diff
@@ -0,0 +1,106 @@
...
+		MSUIA_sourceDir.File('microsoft.ui.uiautomation/Generated Files/winrt/microsoft.ui.uiautomation.h'),
+		MSUIA_sourceDir.File('microsoft.ui.uiautomation/Generated Files/winrt/impl/microsoft.ui.uiautomation.0.h'),
+		MSUIA_sourceDir.File('microsoft.ui.uiautomation/Generated Files/winrt/impl/microsoft.ui.uiautomation.1.h'),
+		MSUIA_sourceDir.File('microsoft.ui.uiautomation/Generated Files/winrt/impl/microsoft.ui.uiautomation.2.h'),
+	],
+	source = [
+		MSUIA_sourceDir.File('microsoft.ui.uiautomation/Microsoft.UI.UIAutomation.vcxproj'),
+		MSUIA_sourceDir.File('microsoft.ui.uiautomation/Microsoft.UI.UIAutomation.idl'),
+		glob.glob(os.path.join(MSUIA_sourceDir.abspath, 'microsoft.ui.uiautomation', '*.cpp')),
+		glob.glob(os.path.join(MSUIA_sourceDir.abspath, 'microsoft.ui.uiautomation', '*.h')),
```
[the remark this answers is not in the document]
michaelDCurran: As these globs are only to find source files for the sake of checking if they are out of date (and a build is required), including too many is worse than missing some. Note that this list is never passed to a compiler for actual building. I would much prefer to keep the glob lines, rather than list out all 30 or more source files, which may very well get added to and we wouldn't know... it just wouldn't rebuild if those new ones changed.

nvdaHelper/UIARemote/UIARemote.cpp
```diff
@@ -104,9 +115,12 @@ extern "C" __declspec(dllexport) bool __stdcall msWord_getCustomAttributeValue(I
...
 	} catch (std::exception& e) {
-		std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> converter;
-		auto what = converter.from_bytes(e.what());
-		LOG_ERROR(L"msWord_getCustomAttributeValue exception: "<<what);
+		const std::string what{e.what()};
+		// convert exception message to unicode
+		int wideLen = MultiByteToWideChar(CP_UTF8, 0, what.c_str(), what.length(), nullptr, 0);
+		auto wideBuf = std::make_unique<wchar_t[]>(wideLen);
+		MultiByteToWideChar(CP_UTF8, 0, what.c_str(), what.length(), wideBuf.get(), wideLen);
+	LOG_ERROR(L"msWord_getCustomAttributeValue exception: "<<(wideBuf.get()));
```
feerrenrut, line 123: This seems like something that belongs in a utility function. It is hard to determine correctness at a glance, and easy to make a mistake. 

I think there is a chance that `wideBuf ` is may not be null terminated here.  Using `MultiByteToWideChar` will include return a size which includes the null termination if `cbMultiByte` is `-1`, however this uses the `size()` from `basic_string` which does not count the null terminating character.

An alternative to explicitly dynamically allocating `wchar` array to store the converted string is to use a `std::vector<wchar_t`. One advantage of the vector approach is the easy of initializing the memory eg:
``` c++
std::vector<wchar_t> wideBuf (wideLen, L'\0');
```
This however sacrifices speed for correctness, given this is exception handling I'd say it is a fair tradeoff.

To keep the current heap array approach, just add 1 to `wideLen` when allocating, then after the second call to `MultiByteToWideChar` set `wideBuf[wideLen]=L'\0'`

```suggestion
		auto wideBuf = std::make_unique<wchar_t[]>(1 + wideLen); // add space for a null char.
		MultiByteToWideChar(CP_UTF8, 0, what.c_str(), what.length(), wideBuf.get(), wideLen);
		wideBuf[wideLen] = L'\0';
		LOG_ERROR(L"msWord_getCustomAttributeValue exception: "<<(wideBuf.get()));
```

nvdaHelper/UIARemote/remoteLog.h
```diff
@@ -29,27 +29,39 @@ class RemoteableLogger {
...
-			auto what = converter.from_bytes(e.what());
-			messageBlock+=L"dumpLog exception: " + what + L"\n";
+			const std::string what{e.what()};
+			// convert exception message to unicode
+			int wideLen = MultiByteToWideChar(CP_UTF8, 0, what.c_str(), what.length(), nullptr, 0);
+			auto wideBuf = std::make_unique<wchar_t[]>(wideLen);
+			MultiByteToWideChar(CP_UTF8, 0, what.c_str(), what.length(), wideBuf.get(), wideLen);
+			messageBlock += L"dumpLog exception: ";
+			messageBlock += wideBuf.get();
+			messageBlock += L"\n";
```
feerrenrut, line 64: See other exception handling comments. I'd extract this out into a shared utility.

nvdaHelper/readme.md
```diff
@@ -6,18 +6,47 @@ While virtual buffers apply to several different application types, it may first
...
  the target API interface.
 For instance:
 - IAccessible2 IDL files are built into `IAccessible2Proxy.dll`
 - ISimpleDOM IDL files are built into `ISimpleDOM.dll`
- 
-The `*local*.dll`'s are built for x86 (ie to match NVDA's arch), others are built for x86, x64, and arm64.
+
+#### nvdaHelperLocalWin10.dll
+Contains code specific to Windows 10 and above, that aides in accessing newer technologies such as Windows OneCore speech synthesis, the Windows in-built OCR service.
+This code is usually either written in C++/CX or C++/WinRT.
```
feerrenrut, line 27: I don't think there is any C++/CX code left now.

nvdaHelper/readme.md
```diff
@@ -6,18 +6,47 @@ While virtual buffers apply to several different application types, it may first
...
 - ISimpleDOM IDL files are built into `ISimpleDOM.dll`
- 
-The `*local*.dll`'s are built for x86 (ie to match NVDA's arch), others are built for x86, x64, and arm64.
+
+#### nvdaHelperLocalWin10.dll
+Contains code specific to Windows 10 and above, that aides in accessing newer technologies such as Windows OneCore speech synthesis, the Windows in-built OCR service.
+This code is usually either written in C++/CX or C++/WinRT. 
+
+#### nvdaHelperLocal.dll
+this dll is loaded directly in to NVDA. It provides the following features:
```
feerrenrut, line 30: missing cap
```suggestion
This dll is loaded directly in to NVDA. It provides the following features:

```

nvdaHelper/readme.md
```diff
@@ -6,18 +6,47 @@ While virtual buffers apply to several different application types, it may first
...
+* Several small utility functions that asist in processing text (which are faster in c++).
+
+#### NVDAHelperRemote.dll
+This dll injects itself into other processes on the system, allowing for in-process code execution by NVDA.
+It provides the following features:
+* Server stub methods for several RPC interfaces, including NVDAInprocUtils, VBufBackends and displayModel.
+* Client stub methods for several RPC interfaces, allowing in-process code to execute code back in NVDA. These interfaces include NVDAController and NVDAControllerInternal
+
+#### UIARemote.dll
+this dll is loaded by NVDA, providing utility functions that perform certain tasks or batch procedures on Microsoft UI Automation elements.
```
feerrenrut, line 43: Missing cap
```suggestion
This dll is loaded by NVDA, providing utility functions that perform certain tasks or batch procedures on Microsoft UI Automation elements.

```

nvdaHelper/readme.md
```diff
@@ -6,18 +6,47 @@ While virtual buffers apply to several different application types, it may first
...
+It makes use of the UI Automation Remote Operations capabilities in Windows 11, allowing to declaritively define code  to  access and manipulate UI Automation elements, that will be Just-In-Time compiled by Windows and executed in the process providing the UI Automation elements.
+
+##### microsoft-ui-uiAutomation remote ops library
+As a dependdency of UIARemote.dll, the open source [Microsoft-UI-UIAutomation Remote Operations library](https://github.com/microsoft/microsoft-ui-uiautomation)  is also built.
```
feerrenrut, line 47: typo
```suggestion
As a dependency of UIARemote.dll, the open source [Microsoft-UI-UIAutomation Remote Operations library](https://github.com/microsoft/microsoft-ui-uiautomation)  is also built.

```

feerrenrut (changes requested):

nvdaHelper/microsoft-ui-uiautomation/sconscript
```diff
@@ -0,0 +1,106 @@
...
+#This program is distributed in the hope that it will be useful,
+#but WITHOUT ANY WARRANTY; without even the implied warranty of
+#MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#This license can be found at:
+#http://www.gnu.org/licenses/old-licenses/gpl-2.0.html
+###
+
+import os
+import glob
+
```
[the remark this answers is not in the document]
feerrenrut: Could you action this? What are the outputs of this `sconscript`. Looking at the additions to the readme, an explanation doesn't really belong there, but a doc string at the top of this file would help to know what it is responsible for.

nvdaHelper/microsoft-ui-uiautomation/sconscript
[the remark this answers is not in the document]
feerrenrut: Ok that is fine. The `msbuild` in the action looks like these source files are used for building though?

nvdaHelper/microsoft-ui-uiautomation/sconscript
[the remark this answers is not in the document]
michaelDCurran: Yep, added a top-level docstring now.

nvdaHelper/microsoft-ui-uiautomation/sconscript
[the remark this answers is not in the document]
michaelDCurran: the msbuild command references `$SOURCE`. This variable only ever contains the first source file. You'd need to use `$SOURCES` to have all of them appear in the command.

nvdaHelper/UIARemote/remoteLog.h
```diff
@@ -14,6 +14,14 @@ This license can be found at:
 
 #pragma once
 
+// Converts a utf8 encoded string into a utf16 encoded wstring
+std::wstring stringToWstring(const std::string& from) {
+	int wideLen = MultiByteToWideChar(CP_UTF8, 0, from.c_str(), from.length(), nullptr, 0);
+	std::vector<wchar_t> wideBuf (wideLen + 1, L'\0');
+	MultiByteToWideChar(CP_UTF8, 0, from.c_str(), from.length(), wideBuf.data(), wideLen);
+	return std::wstring(wideBuf.data());
```
feerrenrut, line 22: Looks like C++ standard library has you covered here. No need to use the Windows C API: https://stackoverflow.com/a/38383389

``` C++
#include <iostream>
#include <string>
#include <locale>
#include <codecvt>
std::string str = "your string in utf8";
std::wstring_convert<std::codecvt_utf8_utf16<char16_t>> converter;
std::wstring wstr = converter.from_bytes(str);
```


For reference, since C++17 `basic_string` returns a non-const pointer via `data()` you can use it as a buffer for string types to save a copy. In this case you don't need to explicitly reserve space for the trailing null char.
```suggestion
	std::wstring wideBuf (wideLen, L'\0');
	MultiByteToWideChar(CP_UTF8, 0, from.c_str(), from.length(), wideBuf.data(), wideLen);
	return wideBuf;

```

feerrenrut (commented):
I'll leave it up to you about this, I think the C++ standard library approach is less error prone and sets a good example.

nvdaHelper/UIARemote/UIARemote.cpp
```diff
@@ -57,26 +56,38 @@ extern "C" __declspec(dllexport) bool __stdcall msWord_getCustomAttributeValue(I
...
+							/* else */ [&]() {
+								logger<<L"No guid_msWord_getCustomAttributeValue extension supported"<<endl;
+							}
+						);
+					},
+					/* else */ [&]() {
+						logger<<L"Could not fetch guid_msWord_extendedTextRangePattern pattern"<<endl;
+					}
+				);
+			},
```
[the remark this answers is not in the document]
feerrenrut: I don't see how that can be true (yet!). It makes sense that this would happen in the `Resolve` call, but I think that in this case it can only result in two cross platform calls. It won't know about the second `scope.if` until the lambda is executed, which must happen locally. It can't execute the lambda until it has the result of the first condition.

I suspect that within `Resolve` when an additional `scope.if` is called from a handler, it is recursively resolved immediately.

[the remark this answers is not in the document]
feerrenrut: After chatting about this. The missing piece is that both the body and the else lambda's are executed immediately with the goal of constructing remote ops instructions. This means the instruction tree produced can be run in its entirety on in the remote process, ensuring that all logic is executed with a single process call. The log message are recorded on both sides of the condition, but only emitted based on the result of the condition.

Note: this means that modifying any local (non-remote ops type) variable within these lambdas is likely a bug. 
We'll have to watch out for local variables sneaking into these lambdas.

feerrenrut (approved):

feerrenrut:
The test failure looks unrelated. I have kicked off the build again.
Old build link: https://ci.appveyor.com/project/NVAccess/nvda/builds/42592615/tests

PR #13353, opened 2022-02-17
Disable context help in secure screen to avoid a security exploit.

CyrilleB79:
### Link to issue number:
None
Can be considered a follow-up of #13059.

### Summary of the issue:
If you open NVDA parameter from secure screen and press F1, context help opens in Internet explorer.
I am not an expert in security but I think that it's not secure to have access to the browser from secure screen.
More specifically, via the browser open dialog, I am able to open any file in notepad and modify it, e.g. allowing me to activate NVDA console.

Since context help is displayed in a browser, disable it on secure screen.

### Testing strategy:
Manual testing
#### Test 1
Simulate secure screen and checked that context help is not available anymore (nothing happens):
* Launch NVDA
* Open NVDA Python console
* Enter the following command:
  `import globalVars;globalVars.appArgs.secure = True`
* Press NVDA+control+G to open NVDA settings window
* Press F1 and checked that nothing ahppens
#### Test 2
Checked that context help is still working in normal mode:
* Launch NVDA
* Press NVDA+control+G to open NVDA settings window
* Press F1 and checked that nothing ahppens

Context help is not available anymore on secure screen.
If this is a need, we may open an issue for it and try to address it later.
But I think that the security concern should be addressed first and should not wait for a solution to have context help on secure screen.

### Change log entries:

Changes
In the existing change log, replace:
`- Security fix: The addons manager dialog is now disabled on secure screens. (#13059)`
by:
`- Security fixes: The addons manager dialog and context help are now disabled on secure screens. (#13059, #13353)`

### Additional note

I think `ui.browseableMessage` is also based on a browser. Is there any security concern allowing it on secure screen?
I have not been able to do anything from it since there is no menu or shortcut to do anything it this windows.
Just mentionning it in case someone knows more than I on this topic.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/contextHelp.py
+++ b/source/gui/contextHelp.py
@@ -10,6 +10,7 @@ import typing
 import wx
 from logHandler import log
 import documentationUtils
+import globalVars
 
 
 def writeRedirect(helpId: str, helpFilePath: str, contextHelpPath: str):
@@ -73,6 +74,9 @@ def bindHelpEvent(helpId: str, window: wx.Window):
 
 
 def _onEvtHelp(helpId: str, evt: wx.HelpEvent):
+	if globalVars.appArgs.secure:
+		# Disable context help in secure screens to avoid opening a browser with system-wide privileges.
+		return	
 	# Don't call evt.skip. Events bubble upwards through parent controls.
 	# Context help for more specific controls should override the less specific parent controls.
 	showHelp(helpId)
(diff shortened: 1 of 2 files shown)


CyrilleB79:
Also cc @seanbudd since you have finally reviewed #13059 instead of @michaelDCurran.

TheQuinbox:
Nice to see other people also helping with security problems.

From my testing, and a look at the code, I don't think ui.browsableMessage has any serious security concerns. You can't control O or alt D, and the only menu is a system menu with 4 items:
1. Move
2. size
3. A sepperator
4. close.
So I think we're safe as far as that goes.

michaelDCurran (approved):

Issue #12266, opened 2021-04-03
NVDA does not focus listbox aria-role in browse mode


brunoprietog:
### Steps to reproduce:

1. In Microsoft Edge, go to this [link](https://www.w3.org/TR/wai-aria-practices-1.2/examples/listbox/listbox-scrollable.html) for a listbox in W3C.
2. Look for the listbox located below the level 2 header that says example and press enter.
3. Try to move with up and down arrows through the items.

### Actual behavior:

NVDA does not focus the listbox, so pressing enter calls the action of the last focused item, not the listbox, in order to read the items inside.

### Expected behavior:

NVDA should focus the listbox when passing through it in browse mode, so that when you press enter on it, NVDA automatically switches to focus mode and you can read the items in that listbox using the up and down arrows.

#### NVDA installed/portable/running from source:

Installed

#### NVDA version:

NVDA 2020.4

#### Windows version:

Windows 10 20H2 (64 bits) build 19042.867

#### Name and version of other software in use when reproducing the issue:

Microsoft Edge 89.0.774.63

#### Does the issue still occur after restarting your computer?

Yes

#### Have you tried any other versions of NVDA? If so, please report their behaviors.

Same behavior with NVDA 2020.3

#### If add-ons are disabled, is your problem still occurring?

Yes

#### Did you try to run the COM registry fixing tool in NVDA menu / tools?

Yes

Brian1Gaff:
I believe there is a list like this on most Amazon pages as well, there are 
ways around it not working but that is not really the point.
 Brian

***@***.***
Sent via blueyonder.
Please address personal E-mail to:-
***@***.***, putting 'Brian Gaff'
in the display name field.
Newsgroup monitored: alt.comp.blind-users
----- Original Message ----- 
From: "Bruno Prieto" ***@***.***>
To: "nvaccess/nvda" ***@***.***>
Cc: "Subscribed" ***@***.***>
Sent: Saturday, April 03, 2021 9:57 AM
Subject: [nvaccess/nvda] NVDA does not focus listbox aria-role in browse 
mode (#12266)

MarcoZehe:
This is a browser bug: The listbox, AKA the ul with role "listbox", is marked by Edge/Chromium as read-only, whereas in Firefox it is not. A listbox is not read-only, so this is clearly a bug in the way Chromium exposes the combination of these HTML elements and ARIA roles overrides.

MarcoZehe:
[...]

Issue #12896, opened 2021-10-01
NVDA doesn't toggle virtual cursor on when focus is moved into an ARIA document inside an ARIA gridcell


kloots:
### Steps to reproduce:

1.  Open the [CodePen example](https://codepen.io/toddlr/pen/WNOKJgG)
2. Press tab to move focus to the first ARIA `gridcell` in the ARIA `grid` in the example
3. With focus on the `gridcell` press tab to move focus into the ARIA `document` nested inside the ARIA `gridcell`. (Focus should be on one of the anchor elements.)

### Actual behavior:
When focus moves inside the ARIA `document` NVDA doesn't automatically toggle the virtual cursor on to allow for object navigation or text to be reviewed within the document to be reviewed.

### Expected behavior:
When focus moves inside the ARIA `document` NVDA should automatically toggle the virtual cursor on to allow for object navigation or text to be reviewed  within the document to be reviewed.

#### NVDA version:
2021.2

#### Windows version:
Windows 10 Home, OS Build 19043.1237

#### Does the issue still occur after restarting your computer?
Yes

#### Have you tried any other versions of NVDA? If so, please report their behaviors.
No

#### If NVDA add-ons are disabled, is your problem still occurring?
I don't have any NVDA add-ons installed

kloots:
Some additional information the could help. I reviewed this issue with @jcsteh who had this to say:

> As to why NVDA is misbehaving here, you're hitting a bug/quirk with regard to browse mode focus handling. NVDA has logic to ignore focus changes for the purposes of browse mode if the focus overlaps with the previous cursor position. This is done to avoid spurious mode switches or cursor bouncing where focus is set to one thing and ends up being bounced to a descendant; e.g. if you focus a list box, focus might get bounced to a child list item. If the focus was set due to cursor movement, NVDA doesn't want to snap into focus mode in that case. I think they've tried to replace that logic a few times and run into problematic edge cases. Fixing it will be a bit tricky because of potential regressions and edge cases, but I don't think it should be a multi month project or anything like that.

@jcsteh also suggested a partial workaround:

> FWIW, you can partially work around it by ensuring there's something before the first focusable element in the gridcell. For example:

`<div role="grid"><div role="row"><div role="gridcell" tabindex="0">&nbsp;<div role="document"><a href="https://example.com">bar`

[...]

PR #13357, opened 2022-02-17
Switch to focus mode if read-only list items and their list parents are focusable and gain focus

### LINKED ISSUE #13221: NVDA doesn't toggle browse mode off when focus is moved to an interactive list
I believe I have discovered a regression in NVDA. @jcsteh can correct me if I am misremembering, but I recall NVDA _used_ to toggle browse mode off if focus was moved to an HTML list made interactive with `tabindex` and keyboard shortcuts implemented in JavaScript. Reason I remember this: while I was working at Twitter I made the timeline an interactive list so that customers could use keyboard shortcuts provided by Twitter to navigate their home timeline. And I recall working with @jcsteh on ensuring this worked in NVDA. Interestingly, the same code example I provided in this bug report works fine in JAWS in that it toggles browse mode off when focus lands on an ARIA `listitem` instance.

### Steps to reproduce:

1. Start NVDA
2. Open https://codepen.io/toddlr/pen/mdBjoJV
3. Use tab-key navigation to move focus to the first ARIA `listitem` in the ARIA `list`

### Actual behavior:
NVDA doesn't toggle browse mode off

### Expected behavior:
NVDA should toggle browse mode off so the up/down arrow keys can be used to navigate (move focus) between listitems in the list

Yes

No

I'm not using any add-ons

Yes

MarcoZehe:
### Link to issue number:

Fixes #13221

### Summary of the issue:

In the Microsoft Edge Downloads manager (CTRL+J), as well as in the channel and DM outputs of the Slack app, HTML lists and list items are used to display the items. These are made focusable via either aria-activedescendant, or the rowing tabIndex pattern. The reason is that for compatibility, these can be used to host rich content, such as whole sub documents or controls. Until now, NVDA would not automatically switch to focus mode when focus lands on such a focusable list item.

This PR adds a rule for enabling passthrough mode if a list item that is read-only, has state focused, and at the same time its paarent is of role list and is focusable.

### Testing strategy:

1. Run NVDA.
2. Go into Microsoft Edge.
3. Press Control+J to invoke Downloads Manager.
4. Focus should be on the first downloaded item, and passthrough mode should automatically be enabled.

None known.

### Change log entries:
New features
Changes
Bug fixes

* In the Microsoft Edge downloads manager, NVDA will now automatically switch to focus mode once the list item with the most recent download gains focus. #13221 

For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation: Not needed since expected behavior is introduced where user would previously have to manually switch pass-through mode on.
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


TheQuinbox:
Nice, looks good to me. In NVDA changelog entries though, it's typical to link to the issue that it solves (if applicable).

MarcoZehe:
Thanks @TheQuinbox, I just updated the initial description with that, as well as the checkbox for the testing part, since the manual check checks out nicely.

CyrilleB79:
Typo to fix in the change log:
#13321 -> #13221

lukaszgo1:
Have you considered adding Chrome system test demonstrating the behavior based on the code snippet in #13221?

MarcoZehe:
Thanks @CyrilleB79, corrected the typo.

@lukaszgo1 I'll try that system test thing, but this is all still very confusing, so may take a while. And it's made difficult by the fact that I can't currently even run local builds, and that includes tests. See my message in the nvda-devel mailing list. But I'll try nevertheless.

leonardder:
Thanks Marco, this is a nice improvement! One question: Isn't it just enough that a list item has focus? Why check the focusable state on the parent list as well??

MarcoZehe:
@leonardder I was going with @michaelDCurran's statement from https://github.com/nvaccess/nvda/issues/13221#issuecomment-1039904489:

> We will switch to focus mode if a listItem gets focus, but only if the list itself is also focusable. I.e. the list has at least tabindex of -1.

This is how MS Edge's Download Manager list works, and it means that @kloots will have to make a slight adjustment in Slack to put tabindex="-1" on the actual list, which Slack currently doesn't do yet.

michaelDCurran:
For reference, the reasoning behind requiring focusable on the list is to limit where this code will apply.
1. Making ul (or role="list) and (li or role="listitem") interactive like this does not quite fit standards as really these are presentational read-only lists. Interactive lists should be role="listbox" and role="option".
2. There may already exist some web pages that deliberately set focus to a presentational list item when the page is scrolled to an arbitrary location. In this case switching to focus mode would be annoying and inappropriate.

But, it has been made clear that the standard does not really fit some particular use cases, and therefore this change is needed. Therefore it is being added, but with the limitation that the list must also be focusable. This idea was proposed by Slack themselves.

tests/system/robot/chromeTests.robot
```diff
@@ -120,3 +120,6 @@ Quick Nav reports target first
 Focus reports target first
 	[Documentation]	Focus target should always be reported before ancestors. Ancestors should be reported from inner to outer.
 	test_focusTargetReporting
+focus mode is turned on on focused read-only list item
+	[Documentation]Focused list items with a focusable list container should cause focus mode to be turned on automatically.
```
michaelDCurran, line 124: Missing tab character after `[Documentation]`
```suggestion
	[Documentation]	Focused list items with a focusable list container should cause focus mode to be turned on automatically.
```

michaelDCurran (changes requested):

michaelDCurran:
I think the title of this pr should refer to Focus mode, not Browse mode.

MarcoZehe:
> I think the title of this pr should refer to Focus mode, not Browse mode.

Done.

MarcoZehe:
Sorry this is going so slowly. I cannot run the tests on my local machine because it is not an English language Windows, and therefore all NVDA tests fall on their face. So I have to peel this like an onion.

MarcoZehe:
@michaelDCurran I don't understand this error message. The previous test failures were clear, but this one confuses me. It says:

> Error message:
focus mode - focus link in document and turn off focus mode
speech Actual != Expected: message
Todd Kloots  link !=   message
Todd Kloots  link
Browse mode

Any help is appreciated.

MarcoZehe:
@michaelDCurran I went with your suggestion of simplifying the test case after all, and it now passes. I can't clear the "changes requested" flag from the issue even though I applied your suggested code change, and also updated the PR title. So, tossing this back for review in hopes that things look good so far. I'd also be happy to help out with #12896 if that's desired.

michaelDCurran:
Thanks for your work on this @MarcoZehe.
I shall review this  sometime in the next couple of days.

brunoprietog:
This is great! Can you think of a way to fix #12266? It's a similar list focus problem. Thanks!

MarcoZehe:
I just commented on the issue: It is a different bug, which will not be fixed by this pull request. The bug is in the Chromium engine, not NVDA in that case. The markup is also radically different from what is being used here.

michaelDCurran (approved):

kloots:
@michaelDCurran I see this has been added to the 2022.1 release milestone. Does the team have a release date spec'd for 2022.1?

michaelDCurran:
@kloots  perhaps end of March / early April.

kloots:
@michaelDCurran curious about the release date for this fix: any update as to when this will go out to production?

michaelDCurran:
It is currently in NVDA 2022.1beta3. There are plans for a beta4 in the 
coming week or so, and then the release will most likely be 3 or so 
weeks after that. So rough guess is mid May.

kloots:
Great. Thanks @michaelDCurran. Dumb question: how do I get the beta?

michaelDCurran:
Link to beta3 post: https://www.nvaccess.org/post/nvda-2022-1beta3/

kloots:
Thanks @michaelDCurran. One last question, do you have a sense for how quickly NVDA users upgrade?

michaelDCurran:
Users upgrade pretty slowly I'm sorry to say.
NVDA 2021.3.5 has been out for about 3 weeks now, and only 30% of users 
have upgraded. The rest are all on random older versions. The next most 
popular version is 2021.3.1 at just under 10%.

We alert users of updates within NVDA when ever they are available.

Issue #13354, opened 2022-02-17
NVDA appears not to apply browse mode preferences when reading messsages in Outlook


geoffshang:
My work just moved from using a web-based system to using Outlook, and so I have just moved to using the Outlook application for mail.  I noticed a few differences which I wanted to report.

I admit that there may be ways for me to tame Outlook to improve my experience that I'm not aware of.

### Steps to reproduce:

Open an email in Outlook that has relevant HTML features.

### Actual behavior:

NVDA uses screen layout and appears to show layout tables, even if these features are disabled in browse mode preferences.

I don't have any configuration profiles.

### Expected behavior:

NVDA should use the same browse mode preferences as in web browsers.

#### NVDA installed/portable/running from source:

Installed

#### NVDA version:

2021.3.1

#### Windows version:

Edition	Windows 10 Enterprise
Version	21H1
Installed on	‎10/‎10/‎2020
OS build	19043.1526
Experience	Windows Feature Experience Pack 120.2212.4170.0

#### Name and version of other software in use when reproducing the issue:

Microsoft Outlook 16.0.14326.20784

Note that this is as reported in the NVDA log viewer.  I have no idea where to find this in outlook itself.

#### Does the issue still occur after restarting your computer?

Yes

#### Have you tried any other versions of NVDA? If so, please report their behaviors.

No

#### If NVDA add-ons are disabled, is your problem still occurring?

Haven't tried but I doubt it - I don't have many installed.

#### Does the issue still occur after you run the COM Registration Fixing Tool in NVDA's tools menu?

I just ran it but haven't restarted.  I doubt it will make a difference but I'll report if it does.

CyrilleB79:
Hello @geoffshang

This issue describes two unexpected behaviors.
1. NVDA uses screen layout in Outlook whatever the option value.
2. NVDA reports layout tables in Outlook whatever the option value.

Regarding issue 1:
For now, screen layout only applies to webpages in browsers, not in Word documents or Outlook e-mails. However, I admit that it does not seem to be indicated anywhere. At least a comment in the user guide would be helpful.

But more generally, I think that some options in Browse mode settings apply only to virtual buffers (i.e. on webpages in browsers), whereas other options apply to all documents supporting a browse mode, i.e. webpages but also Word documents and Outlook e-mails. This would really need to be clarified, in the user guide and/or maybe better, in the browse mode settings panel layout.

Regarding issue 2:
[...]

PR #13373, opened 2022-02-22
Rewrote and expanded the Browse mode: Use Screen Layout section in the user guide to correct missing information

XLTechie:
### Link to issue number:

Addresses a concern raised in #13354 

### Summary of the issue:

In [this comment](https://github.com/nvaccess/nvda/issues/13354#issuecomment-1043570903), @CyrilleB79 pointed out that the user guide does not indicate that screen layout is always used in Outlook, irrespective of the "Use screen layout" setting.
It seems reasonable for the user to assume, since it isn't otherwise indicated, that turning off screen layout will be effective anywhere browse mode is used, when in fact this is not the case.

I added a note to the section in the user guide, explaining that this option is not applicable to Microsoft Office applications.

While looking at the section, I thought that it didn't really explain well the difference between screen layout and not screen layout, and used terms like "fields" in a way that most users might not associate with the elements changed by the option.

### Testing strategy:

N/A

I'm not convinced that the result is the best way to present screen layout or the option, but I think it is preferable to what was there, at least until something better comes along.

### Change log entries:

User Guide: Improved the explanation of the browse mode "Use screen layout" option, now including that it doesn't apply to Outlook, Word, etc.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/user_docs/en/userGuide.t2t
+++ b/user_docs/en/userGuide.t2t
@@ -1709,7 +1709,7 @@ This field sets the amount of lines you will move by when pressing page up or pa
 ==== Use screen layout ====[BrowseModeSettingsScreenLayout]
 Key: NVDA+v
 
-This option allows you to specify whether browse mode should place clickable content (links, buttons and fields) on its own line, or if it should keep it in the flow of text as it is visually shown. Note that this only applies to browse mode as it is used in web browsers. Apps such as Microsoft Word and Outlook are unaffected by this option, and always use screen layout.
+This option allows you to specify whether browse mode should place clickable content (links, buttons and fields) on its own line, or if it should keep it in the flow of text as it is visually shown. Note that this option doesn't apply to Microsoft Office apps such as Outlook and Word, which always use screen layout.
 When screen layout is enabled, page elements will stay as they are visually shown. For example, a visual line of multiple links will be presented in speech and braille as multiple links on the same line. If it is disabled, then page elements will be placed on their own lines, which may be easier to understand during line by line page navigation, and may make items easier to interact with for some users.
 
 ==== Enable browse mode on page load ====[BrowseModeSettingsEnableOnPageLoad]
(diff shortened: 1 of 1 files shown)


XLTechie:
CC @Qchristensen

XLTechie:
@feerrenrut  The system test failure has no connection to what was done in this PR.
> Robot . chromeTests . checkbox labelled by inner element
> A checkbox labelled by an inner element should not read the label element twice.
> Unable to locate 'before sample' marker. Too many attempts looking for 'Before Test Case Marker'

user_docs/en/userGuide.t2t
```diff
@@ -1709,8 +1709,8 @@ This field sets the amount of lines you will move by when pressing page up or pa
 ==== Use screen layout ====[BrowseModeSettingsScreenLayout]
 Key: NVDA+v
 
-This option allows you to specify whether content in browse mode should place content such as links and other fields on their own line, or if it should keep them in the flow of text as it is visually shown.
-If the option is enabled then things will stay as they are visually shown, but if it is disabled then fields will be placed on their own line.
+This option allows you to specify whether browse mode should place clickable content (links, buttons and fields) on its own line, or if it should keep it in the flow of text as it is visually shown. Note that this only applies to browse mode as it is used in web browsers. Apps such as Microsoft Word and Outlook are unaffected by this option, and always use screen layout.
```
lukaszgo1, line 1712: Given that screen layout also applies to the PDF content in Acrobat, all Electron based apps and mails in Thunderbird wouldn't it be better to say that it does not apply for Word and Outlook rather than saying that it applies only in browsers?

XLTechie: > Given that screen layout also applies to the PDF content in Acrobat, all Electron based apps and mails in Thunderbird wouldn't it be better to say that it does not apply for Word and Outlook rather than saying that it applies only in browsers?

Yes it would. I have amended it to say that it does not apply to MS office apps such as Word and Outlook.

Qchristensen (approved):
Thanks @XLTechie for the PR and @CyrilleB79 for your insight as well!  The idea of rewriting the Browse mode section of the User Guide came up recently and I have intentions of going through it in more detail.

In the meantime, great work on this.

Issue #8057, opened 2018-03-05
Ignore punctuation level when moving by word or greater, and the new caret location starts on punctuation


Qchristensen:
This is particularly noticeable in Word, as control+left or right arrow moves to the start of each word, but also to punctuation.  For instance, in the sentence:

The quick, brown fox!

(with a comma after quick and an exclamation mark after fox), from the start, pressing control+right arrow moves to the start of each word, but also to the comma and exclamation mark.  Whether these are read depends on the punctuation level.  Some other programs like Notepad ignore the punctuation and just jump to the start of the next or previous word so it isn't as noticeable.

With punctuation set to "some", for instance, nothing is read when control+right arrow lands on the comma or the exclamation mark.

I propose that when navigating by word (or line or any other method), if the only text to read is punctuation, NVDA should ignore the punctuation level as it does when moving by character.  Currently, it is possible to move the caret in Word particularly, but get no feedback from NVDA in this situation.  Reading as if punctuation level was set to all when landing on non-alphanumeric characters would ensure the user is provided with feedback when navigating.

zahra21:
hi quentin.
i set punctuation to none and nvda announces punctuations in notepad plus plus and libreoffice when i navigate word by word.
its not specific for word and i wanted to report it before, but i did not thought that it maybe is a bug in nvda speech.

Qchristensen:
Interesting, the behaviour does seem to vary slightly between programs then.

Another inconsistency I found was in my sentence, if you misplaced the comma and had quick space comma brown, it behaves differently in Word as well:

the hungry, quick ,brown fox.

In word as you control+right arrow through reads:
hungry
(nothing)
quick
comma
brown
fox
dot

NotePad++ treats each of those commas and the full stop as individual words and reads them out.
Windows Notepad ignores all the punctuation, just reading the words themselves.  At the end it does move to the full stop, but reads the word "fox" again.

Adriani90:
This is better now after #11167 has been merged. But Notepad is still not working here. It seems the editor treats only a space bar as word separator. For notepad a coma or a fullstop is not a word separator. For example, if you write "hello,how" in notepad, the whole expression is read although there are actually three word in there. "hello", "fullstop" and "how".
@Qchristensen I think this should be reported to Microsoft, or NVDA could build its own asumption on what a word should be in this application.

XLTechie:
But should that be a word break?

There is no valid grammar use that I know of, for using a comma without a 
[...]

PR #13483, opened 2022-03-15
Return "blank" if text empty due to processing

### LINKED ISSUE #13431: Tab characters should be considered blank
### Steps to reproduce:

Read the two lines between "start" and "end":

```
start
    
		
end
```

(Note: The first line after start has 4 spaces and the next line has 2 tabs)

### Actual behavior:

They are read as:

```
start
blank
*nothing is read*
end
```

Further, if indentation reporting is on, then they are read as:

```
start
4 space blank
2 tab blank
end
```

### Expected behavior:

I would expect the line with the 2 tabs should be read as "blank" when indentation reporting is turned off, since it is a blank line when indentation reporting is turned on. Nothing being read at all causes confusion when reading code, since it can seem as if NVDA has stopped responding.

### Cause of problem

In `speech.py`, whether text is considered blank is determine by this code:

```
#: If a chunk of text contains only these characters, it will be considered blank.
BLANK_CHUNK_CHARS = frozenset((" ", "\n", "\r", "\0", u"\xa0"))
def isBlank(text):
	"""Determine whether text should be reported as blank.
	@param text: The text in question.
	@type text: str
	@return: C{True} if the text is blank, C{False} if not.
	@rtype: bool
	"""
	return not text or set(text) <= BLANK_CHUNK_CHARS
```

The tab character is missing from `BLANK_CHUNK_CHARS `, which is why  it isn't considered blank. The solution would be to also include it in the set.

Not sure if there is something I am missing and this is actually the intended behavior, but the inconsistency between the line being blank when indentation reporting is on/off makes me think it is a bug.

If it is agreed this is a bug, I would be happy to provide the PR as my first contribution to the NVDA project.

SamKacer:
closes #13431 (yay! a palindrome)

### Summary of the issue:

After processing a text for dictionary substitutions and symbol level, the result might be an empty string, which leads to nothing being spoken at all (not even "blank"). Something should always be spoken when reading text, since nothing being spoken gives the appearance of NVDA freezing.

When a speech sequence is submitted to `speech.speak`, if all strings of the sequence become empty due to symbol level and dictionary substitution processing, then the sequence is considered blank and "blank" is appended to it.

The function now also takes an additional argument, `suppressBlanks`, which is false by default. This parameter allows callers to suppress the "blank" being appended to the sequence, even if the conditions are met. The main intention is to suppress this behavior during say all. Usages of `speech.speak` have been updated to specify this parameter when appropriate.

### Testing strategy:

Manually tested on following examples using NVDA built by appveyor. Also updated relevant system tests.

Some lines to test changes on:

```
# 2 tabs:
		
# 4 spaces:
    
***
hello *** world
```

Before changes:
- at symbol level < All and indentation reporting off, 2 tab line, nothing is spoken
  * when indentation reporting is on, read as "2 tab blank"
  * when indentation off and symbol level Allread as "tab tab"
- in contrast, 4 space line is spoken as "blank" with indentation reporting off, and "4 space blank" when turned on
(Note: as you can read in #13431, this is because by `isBlank()` it is considered blank, however tabs cant be considered blank because they are actually spoken when symbol level is All, whereas spaces and other `BLANK_CHUNK_CHARS are not spoken at any symbol level)

- for 3 stars line, nothing is spoken at symbol level None
- also when reading by word, the 3 stars between hello and world have nothing spoken

after changes:
- now when indentation reporting off and symbol level less All, 2 tab line spoken as "blank", other cases same as before
- 4 space line spoken same as before in all scenarios
- 3 star line spoken as "blank" when symbol level is None
- when moving by word, 3 stars between hello and world spoken as "blank"

1. I am thinking that "blank" has been presented to translators as meaning an empty line, so maybe a lot of languages present it as literally "empty line". In which case, the last example of "hello *** world", when moving by word and the *** is read as "empty line" it could be a little confusing.
1. Another possible issue is that addons that use the `speech.speak` function might get different output than they used to. Addons are able to set `suppressBlanks=True` to get the previous behavior. In many cases this updated behavior could be welcomed by addon authors. A way to ensure API compatibility is maintained would be to have the default value be `True` instead of `False`, but then the new behavior, which should be correct in majority of cases, isn't automatically used.
1. A final consideration is whether appending "blank" to the speech sequence is the right solution, because if the sequence has an empty string between two commands such as `[BeepCommand, "", BeepCommand]`, then "blank" will be spoken after the two commands are executed whereas the text it is replacing would have been spoken before the last command was executed.
   - A better solution could be to instead of appending "blank", to replace the last empty string with "blank" instead, which would speak "blank" at the expected position, but this wouldn't cover cases when multiple empty strings and command are interleaved such as `["", Command, "", Command]`.

### Change log entries:
New features
Changes
Bug fixes

- text sequences that are empty due to dictionary or symbol level processing are spoken as "blank" instead of not being spoken at all (#13431)

For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


Adriani90:
I would rather prefer a sound introduced via speech refactor feature in NVDA which indicates that this symbol or text is not spoken due to symbol level or missing dictionary entry. Maybe two sounds, one for symbols and one for empty text linked to dictionary rule. Speaking "blank" could introduce new confusing problems, especially during navigation or say all.

SamKacer:
@Adriani90 I don't think just a sound would be enough. It could be an alternative to the text spoken like for indentation reporting there is the option of speech or beeps, but a speech option would have to be there, since that is what some users might prefer.

Thanks for bringing up say all, since I forgot to consider that. Testing it manually just now, when doing say all and there is a line that is blank due to processing, such as the line with just two tabs, then it is read as "blank" during say all. This is wrong since lines being blank shouldn't be read during say all.

I will have to rethink how to implement this, so that say all is taken into account, since the last thing I want to do is fix one bug only to introduce another.

I'm not closing this right away in case some other people provide feedback on things I might not have considered

feerrenrut:
@SamKacer Could you take a look at the system tests: `tests/system/robot/symbolPronunciationTests.py`
These cover the behavior here. You'll need to update these to match your expectations, and please feel free to extend with new test cases to cover the outcomes you wish for.

You can run just these tests by doing the following:
- in `tests/system/robotArgs.robot` remove `--include NVDA`
- run the system tests with `runsystemtests.bat --include symbols`. The tag `symbols` and others can be found by browsing the `*.robot` files.

The system tests will launch NVDA and another application (eg chrome or notepad), interact collecting speech, then exit. Any mouse / keyboard interaction may interfere with the tests. A work around for this is running the tests in a virtual machine.

I'm going to convert this PR to a draft until it is ready again, please feel free to press the "ready for review" button when this is addressed.

feerrenrut:
@SamKacer You might be interested in #12710. When looking at a this area (however with a focus on symbols), my conclusion was that there is a loss of information, the initial intent of the `str` is lost. It gets processed multiple times, once the initial value is lost, subsequent processing can no longer differentiate and provide the correct outcome. My solution was to encapsulate the `str` with type that carries the original (or modified) intent separate from the contents. This didn't go ahead, there were still many edge cases and the community wasn't convinced by the approach.

SamKacer:
@feerrenrut thanks for the feedback! Sorry for the late response. My life situation suddenly became quite hectic, so I've had to put this on hold, but will return to it, although might be a couple of weeks.

In regards to running the system tests with the setup you outlined, does this still need me to build the NVDA project? I suspect one thing I will have to sort out before I return to this is setup the whole suite of tools for building the project including the 7 gigs or so of MS Visual Studio build tools, which is a blocker for me at the moment. But I am in the process of setting up a new laptop where I will hav enough space to install everything.

feerrenrut:
@SamKacer Yes, you'll need to be able to build NVDA locally.

SamKacer:
One thing that could be changed in the implementation is that the only place `suppressBlanks` needs to be set to `True` is when calling `speak` in sayall. Since when initializing sayall, the speak function that is passed to it has `suppressBlanks=True` set, all of the `suppressBlanks=reason == OutputReason.SAYALL` are redundant and could be safely removed, thus keeping the code simpler.

What do others think? Keep those in there just to be extra safe or remove them to make it simpler?

The one other place I would keep it is `speakText`, since that is passed a `suppressBlanks` parameter itself, so the inner `speak` needs to respect it as well.

SamKacer:
@feerrenrut Hi, I just wanted to ask since this is my first PR on this repo and so am unfamiliar with the general timelines here. I marked this ready for review a little over 2 weeks ago and haven't gotten any response yet. Is this normal an to be expected? Just asking since not sure if this PR was accidently overlooked

feerrenrut:
@SamKacer It is common but not intended, unfortunately we are quite time constrained. An exception to this is when we identify that the change should target a later release. But in those cases we try to communicate it on the issue / PR.

source/speech/speech.py
```diff
@@ -1149,7 +1164,11 @@ def speakTextInfo(
 
 	speechGen = GeneratorWithReturn(speechGen)
 	for seq in speechGen:
-		speak(seq, priority=priority)
+		speak(
+			seq,
+			priority=priority,
+			suppressBlanks=suppressBlanks or reason == OutputReason.SAYALL
```
feerrenrut, line 1170: Why is there an exception for say all? Can you add a comment to explain.

tests/system/robot/symbolPronunciationTests.py
```diff
@@ -195,14 +195,14 @@ def test_moveByLine():
 		symbolLevel=SymLevel.NONE,
 		expectedSpeech=[
 			'Say', '(quietly)', 'Hello,', 'Jim .', "don't",  # Expect:
-			'',  # todo: Expect 'right-pointing arrow'
```
feerrenrut, line 198: Don't remove these expected symbol todo comments. These tests were added when assessing the current behavior for symbols. The todo comments document what we intend the output to become. 

So while "blank" is better than `""`, it still isn't the final goal.

feerrenrut (commented):
I'm a bit worried about known issue 3. It would be good to know if this is a real world case that will be encountered.

SamKacer:
> I'm a bit worried about known issue 3. It would be good to know if this is a real world case that will be encountered.

I haven't spotted this in the wild somewhere or come across a place in the code where such a case is created. Know issue 3 is pure hypothesis on my part.

The logic behind considering a speech sequence blank and appending "blank" if it is is, adapted from similar logic in `speech.speech.getTextInfoSpeech`. So, if this issue exists, then it exists already and if the proposed amelioration should probaby be implemented there as well to keep the "blank" logic consistent. Although the logic there around blanks is quite a bit more complex, so it is also possible it does handle this possible issue somehow and I missed it.

source/speech/speech.py
SamKacer: > Why is there an exception for say all? Can you add a comment to explain.

This is consistent with logic in `speech.speech.getTextInfoSpeech`, where "blank" is added to the speech sequence only if it isn't a sayall. As I understand, blanks are not meant to be spoken during a sayall and that this was something generally known, so not requiring a comment.

snippet from `speech.speech.getTextInfoSpeech`:

```
if not suppressBlanks and reason != OutputReason.SAYALL and shouldConsiderTextInfoBlank:
	# Translators: This is spoken when the line is considered blank.
	speechSequence.append(_("blank"))
```

Also, just before asking about timeline on review, I added a comment pointing out that comparing `reason == OutputReason.SAYALL` is redundant since in `speech\__init__.py`, when SayAll is initialized, it is passed a modified speech function where `suppressBlanks` is overridden to be true. So, I was thinking of removing all these comparisns like this, since sayall is already handled directly at the source. But, I wanted to get feedback on that first. What do you think?

tests/system/robot/symbolPronunciationTests.py
SamKacer: > Don't remove these expected symbol todo comments. These tests were added when assessing the current behavior for symbols. The todo comments document what we intend the output to become.
> 
> So while "blank" is better than `""`, it still isn't the final goal.

This was actually on purpose.  I wanted to spark a discussion about this. 

I would propose that this perhaps should be the final goal. If symbol level is none, then as a user I would expect no symbols would be spoken. If a speech sequence doesn't contain anything to be spoken, "blank" should be spoken instead. 

Otherwise, I don't understand why "'right-pointing arrow" should be spoken when symbol level is none. Shouldn't it be considered a symbol and so not spoken? If this was discussed and this is indeed the intended final goal then at least for me it is confusing and I think the test case would benefit from having a comment with a small explanation about the justification

feerrenrut: Blank is misleading. It suggests there are no visible characters in the range. We've had quite some discussions about this already. The should be given the most accurate picture as a priority. This might mean announcing the name of a single symbol, or "symbols" when there are multiple, perhaps like spaces / tabs the number should be spoken "5 symbols". A mix of whitespace should perhaps be announced as "mixed whitespace".

Our last work in this area was really focusing on symbols and the inconsistency of reporting.

source/speech/speech.py
feerrenrut: >As I understand, blanks are not meant to be spoken during a sayall and that this was something generally known, so not requiring a comment.

I'd prefer a comment. It's very easy for priors to change, reasoning helps promoting questioning the behavior. I.E. does this reason still exist / make sense in the current context. There is also the aspect of making it easy for developers with different histories to get up to speed on the reasoning.

>So, I was thinking of removing all these comparisns like this, since sayall is already handled directly at the source. But, I wanted to get feedback on that first. What do you think?

Yes, I think that would be preferable.

feerrenrut:
>I haven't spotted this in the wild somewhere or come across a place in the code where such a case is created. Know issue 3 is pure hypothesis on my part.

One option if we are concerned about this, is to put the feature behind a feature flag. This makes it easy for us to change the default  / and allow alpha users to experiment more easily. It lowers the risk of having to revert the change entirely.

lukaszgo1:
> ```
> ***
> hello *** world
> ```

These two examples really worry me. In general NVDA should never lie to user about the text it presents and saying "blank" for a non blank line/word is going to be pretty confusing. In general this introduces a uncertainty about what "blank" means and overloading a very well known concept of it being reported for empty lines may upset users. Has it been considered not to report "blank" in cases of hunks where  no text is spoken due to punctuation level being too low, rather try to increase it until the entire hunk produces speech?

feerrenrut:
Yes, @lukaszgo1 this is a known issue (reporting "blank" or no speech at all are equally unhelpful). I was trying to address this with #12710. That PR didn't go ahead, instead tests were added to demonstrate this behavior and currently desired outcome. We have been discussing this already, see comment thread: https://github.com/nvaccess/nvda/pull/13483#discussion_r874693611

Qchristensen:
This might also close #8057 (Moving by word, or greater, where the whole new block of text to be read is characters at a higher punctuation level than set, so nothing is read).

I can also see [@lukaszgo1's point](https://github.com/nvaccess/nvda/pull/13483#issuecomment-1131550214). I do think it is better to say something rather than remain silent, however, I can't think of a good, short, phrase to read in lieu of "This utterance would all be text at a higher puncutation level than would be read" (I don't think NVDA should say that).

tests/system/robot/symbolPronunciationTests.py
SamKacer: hmm, okay. This reasoning was missing in those tests making it unclear why that was the expected output. I will add back those todos along with short comment on the why, as you just explained it, so others are not confused as I was.

However, what about the case of a sequence that only contains tabs? In that case isn't "blank" an appropriate final solution, since tabs are whitespace and so would generally be considered as blank?

SamKacer:
> > I haven't spotted this in the wild somewhere or come across a place in the code where such a case is created. Know issue 3 is pure hypothesis on my part.
> 
> One option if we are concerned about this, is to put the feature behind a feature flag. This makes it easy for us to change the default / and allow alpha users to experiment more easily. It lowers the risk of having to revert the change entirely.

I think this might be overly careful, but it can be done if you think that would be best. Could you please advise how to implement the feature flag?

SamKacer:
> Yes, @lukaszgo1 this is a known issue (reporting "blank" or no speech at all are equally unhelpful). I was trying to address this with #12710. That PR didn't go ahead, instead tests were added to demonstrate this behavior and currently desired outcome. We have been discussing this already, see comment thread: [#13483 (comment)](https://github.com/nvaccess/nvda/pull/13483#discussion_r874693611)

I'd just like to contest that "blank" or saying nothing at all are equally unhelpful. Since, if this is the case, let us close this PR, since all it does is in situations when nothing would be spoken at all due to symbol level or dictionary substitution, "blank" is spoken instead. So, if both states ar equally bad, then this adds no value at the cost of more lines of code.

For me as a user, I feel although perhaps not the ideal final solution, "blank" is much better than nothing being said, because in the scenario that nothing is said after navigating, it gives the impression that NVDA has stopped responding and the user is left momentarily unsure about what just happened. This has happened to me several times and if "blank" was spoken (or anything at all), then it would have been preferable to avoid confusion.

I think, aiming for a better final solution is a good idea, but I find this to be an incremental improvement.

SamKacer:
Based on feedback so far, I have these todos:

- [ ] update comments in tests
- [ ] remove all the redundant `reason == OutputReason.SAYALL` checks
- [ ] comment on why `suppressBlanks` is set to `True` for sayall

Qchristensen:
A small point, but I would contest that saying nothing at all is more unhelpful than "Blank".  If NVDA reads "Blank" then I know it has responded to my instruction (pressing down arrow for instance), and I then need to interpret what "Blank" means.  If, however, nothing at all is said, then I have no way of knowing whether the line is blank, my keypress wasn't registered, the computer has frozen, my speakers have just died, or any number of other possible issues.

Perhaps an earcon could be an option, but then that still brings us back to, do we give the user different feedback for a completely blank line (two new lines in a row) vs a line with three spaces and nothing else.  Or the same feedback for both a blank line and one with only three spaces on it.

XLTechie:
> Perhaps an earcon could be an option, but then that still brings us back to, do we give the user different feedback for a completely blank line (two new
> lines in a row) vs a line with three spaces and nothing else. Or the same feedback for both a blank line and one with only three spaces on it.

Or what for a line with punctuation only that is above the current punct level?

It's going to have to be user preference.

Speech for lines that would otherwise be silent:
Say nothing
Say blank
Increase level until line can be spoken
Earcon

mwhapples:
I get the problem with nothing being said, but to refer to a line containing characters as "blank" feels very wrong as its suggesting something other than the case. The characters may be whitespace or may even be punctuation marks or other symbols. depending upon cursor placement, when navigating to such a line and then just typing may lead to different results (IE. if the inserted text comes before or after those existing ignored characters), but for a truly blank line then the result will always be the same due to only one possible placement on a blank line.
A phrase like "ignored characters" or if keeping it as short as possible "ignored" would be better than "blank" as it makes it clear there is content on the line but that the settings lead to them not being spoken.

Qchristensen:
"Ignored" or "no text" could work 👍

XLTechie:
"ignored" doesn't make it clear at all. It might just be the word "ignored" on 
the line, how does the user know he has to verify?

As a default, either silence (which is unusual enough to warrant checking), or 
situationally elevating the punctuation level until it reads the formerly 
suppressed content, would seem like the logical options.

Anything else is misleading as a default.

mwhapples:
> "ignored" doesn't make it clear at all. It might just be the word "ignored" on the line, how does the user know he has to verify? As a default, either silence (which is unusual enough to warrant checking), or situationally elevating the punctuation level until it reads the formerly suppressed content, would seem like the logical options. Anything else is misleading as a default.

In that case we should get rid of blank for entirely empty lines as well, might that be the word blank on a line? The suggestion of "ignored" is following the "blank" convention. One could take this to the logical, if somewhat extreme, conclusion and say any announcement which is not actually present in the text (eg. saying link to alert the presence of a link, etc) are misleading/ambiguous. Common sense says that some additional information announcements will be needed, but it seems different people feel differently as to what is "required" additional information.
A setting probably is going to be needed to satisfy all. My suggestion of "ignored" was for those who want something spoken where it would clarify it from a truly blank line.

XLTechie:
> In that case we should get rid of blank for entirely empty lines as well, might that be the word blank on a line?

It might be, and in fact I have had the problem where it was, and I thought the 
line was actually blank, until I investigated further.

But it was a rather easy, and rapid, problem to solve, once I thought of its 
potential. Because "blank" for empty lines, has been a screen reader convention, 
for better or worse, for decades. Across many more platforms than just NVDA on 
Windows.

The only one I can think of that violates it is Voiceover on iOS, which says 
"Newline."

> The suggestion of "ignored" is following the "blank" convention.

But that's my point: "ignored", is not a convention. "Blank" is.

I understand why you suggested it. I Just personally think introducing new and 
unfamiliar terms is a bad direction to go in if it can be avoided.

This is why I like the speak nothing approach, or temporarily raising the 
punctuation level to speak the line, as I said.

mwhapples:
> In that case we should get rid of blank for entirely empty lines as well, might that be the word blank on a line?
> It might be, and in fact I have had the problem where it was, and I thought the line was actually blank, until I investigated further. But it was a rather easy, and rapid, problem to solve, once I thought of its potential. Because "blank" for empty lines, has been a screen reader convention, for better or worse, for decades. Across many more platforms than just NVDA on Windows. The only one I can think of that violates it is Voiceover on iOS, which says "Newline."
> The suggestion of "ignored" is following the "blank" convention.
> But that's my point: "ignored", is not a convention. "Blank" is. I understand why you suggested it. I Just personally think introducing new and unfamiliar terms is a bad direction to go in if it can be avoided. This is why I like the speak nothing approach, or temporarily raising the punctuation level to speak the line, as I said.
At one time "blank" would not have been a convention, they all have been new at some point. So to me lack of existing convention does not hold for me or we would end up with no evolution.
As I said it probably should be put behind a setting as some will want this and others won't. The fact this discussion is even happening shows that different users have different views and so demonstrates the importance of putting it behind a setting.

lukaszgo1:
It looks like there are two issues:
- what to report when the  given string is empty  due to the punctuation level set by the user - my personal opinion is that  raising punctuation level in that case is the only logical solution and trying to invent a better replacement string would only introduce confusion
- What to report when the given string is empty after dictionary substitution - I don't have a good answer for that problem

Also @feerrenrut  mentioned #12710 to be related but it is not, at least from the user perspective. This PR deals with a case in which after substitution the given string is empty, whereas #12710 unconditionally speaks words with a higher punctuation levels as to avoid situations where symbols such as closing quotes or parenthesis are missed.

SamKacer:
Hello, so I finally got around to make the cleanups I promised and update new tests etc. This is ready for review again.

I think an apt summary of the discussion is that there is no universally prefferable solution to the problem of what if all the text is removed due to either symbol level or dictionary substitution. Therefore, I think the final solution will need to make this configurable, so that a user can choose the one they prefer.

However, I think most of us can at least agree that the current behavior of NVDA speaking nothing, giving the impression that it froze, isn't a behavior that anyone finds the most preferable.

So, although this wouldn't be the final solution, I propose speaking "blank" as an straightforward, incremental improvement over the current behavior of speaking nothing.

seanbudd (approved):
Thanks @SamKacer

cary-rowen:
Hello @SamKacer 

Can you take a look at #15024?

Thanks!

SamKacer:
I was just thinking, if the issue with this implementation is that it isn't possible to tell if "blank" means an actually empty line with no chars or if it is a line with chars but they were filtered out, then how about in this scenario "empty" is spoken instead?

That way if you hear "blank" then you know it is a line with no chars and if you hear "empty" you know it is a line with chars that were all filtered out

Someone might argue that then when you hear "empty" then you will be wondering whether it is a line that has the text empty or if it is a line with filtered out text. I will just preemptively remind that is already the case with  "blank", "heading level 1", "link", "emphasis", and all the other features that speak something that isn't actually there.

ruifontes:
Header level 1, bold, emphasis, link and so on are attributs and they are visually presents...
So, if you accept the visually  present attributes are voiced, why not blank characters?
If you want something different from just "blank", why not "blank characters"?

ruifontes:
Or better, "blank chars", for spaces and tabs, and "filtered chars", for filtered characters...

SamKacer:
@ruifontes Ok, so if I understand correctly, you are suggesting there should be 3 different things spoken:

1. "blank" when the line just consists of a \n or \r
2. "blank chars" when a line only contains spaces or tabs
3. "filtered chars" when the line has all its text filtered out

I think I'd prefer to keep discussion in this thread to just be about the 3rd case since that is what this PR was about.

My concern with "filtered chars" is that it is quite verbose. Even "empty" isn't that ideal because it is two syllables compared to the one syllable of "blank". I am not sure what the right phrase is, but I think shorter is better

XLTechie:
>  2. "blank chars" when a line only contains spaces or tabs

"chars" is not a concept that users hear much. You and I know what it means, but 
if you go with something along these lines, that doesn't seem like language that 
will be approved.

"whitespace" is one possible option.

Or the actual characters, like I suggested--"4 tab", "4 space", "6 tab 3 space".

>  3. "filtered chars" when the line has all its text filtered out

And now we have two words that users will need defined or explained for them.
"filtered" in this context, is another term that is not used.

Is silence suddenly not looking so bad? :)

This is really an ideal case for an earcon, or some little blip or beep, if 
people don't want to hear the punct level raised to speak lines that would 
otherwise render as silent.

ruifontes:
For me, silence is good, since I can check in Braille display.
But people without a Braille display can think: Did I really press the arrow?
So, to be quick and informative, maybe the idea of two hearcons is the best:
One for only white spaces and another for the presence of filtered chars...

PR #13531, opened 2022-03-24
Only remove XML tags if JAB widget's text starts with <html> tag

### LINKED ISSUE #13102: NVDA seems to apply some kind of extra processing on label text before reading it
### Steps to reproduce:

I have a problem is with a combo box which holds a history of regular expressions in a Java application, NVDA does not properly read the text for each entry.
As an example I have a Java application which I'm running with the Java access bridge enabled:

    public static void main(String[] args) {
    JFrame fr = new JFrame("abc");
    fr.setSize(500, 500);
    JComboBox combo = new JComboBox(new String[] {"abc", "(<image[^>\\n]*)\\n([^>]*>)"});
    combo.setEditable(false);
    BasicComboBoxRenderer renderer = (BasicComboBoxRenderer) combo.getRenderer();
    renderer.putClientProperty( javax.swing.plaf.basic.BasicHTML.propertyKey, null);
    combo.putClientProperty( javax.swing.plaf.basic.BasicHTML.propertyKey, null);
    
    fr.getContentPane().add(combo, BorderLayout.CENTER);
     fr.setVisible(true);
    }

When I move the focus in the combo drop down list on the second entry NVDA should read "less than image square bracket....". But somehow it skips the content all the way to the first ">". So NVDA might to apply some kind of regular expression which removes what it considers to be "&lt;tags>" from the content before reading it.

### Actual behavior:
NVDA does not read the entire content of the "(<image[^>\\n]*)\\n([^>]*>)" value set on the label.

### Expected behavior:
NVDA should read the label content as it is, as the label is not HTML aware.

Yes
#### NVDA version:
nvda_2019.3.1
#### Windows version:
10

Yes
No

mwhapples:
### Link to issue number:
Fixes #13102 
### Summary of the issue:
NVDA was incorrectly stripping text from Java widgets if the text looked like an XML tag. According to the Oracle Java docs, when using HTML in a Java widget you should start the text with the &lt;html&gt; tag.
This change gets NVDA to check if the text starts with &lt;html&gt; and only then will remove any text which appears to be XML tags.
### Testing strategy:
Manually tested, some unit tests.
I haven't checked exactly what the regular expression does, but I suspect it does not handle entity substitution (eg. &amp;lt; with &lt; and such like). Ideally this should also be fixed, but we may want to track that as a separate issue as its something slightly different to the original issue this is designed to fix.
### Change log entries:
New features
Changes
Bug fixes
Fix issue where NVDA may incorrectly remove text from Java widgets when presenting to the user.
For Developers

### Code Review Checklist:

  - description is up to date Yes
  - change log entries Done
- [x] Testing:
  - Unit tests Unit tests for the function which detects and removes HTML.
  - System (end to end) tests Not done, unsure how to do for this issue.
  - Manual testing Yes
- [x] Documentation:
  - User Documentation N/A
  - Developer / Technical Documentation N/A
  - Context sensitive help for GUI changes N/A
  - Speech Yes
  - Braille Yes
  - Low Vision Yes
  - Different web browsers N/A
  - Localization in other languages / culture than English N/A


michaelDCurran (changes requested):
The actual change looks fine to me.
But, Some unit tests would be great for testing _processHTML. Especially a unit test at very least for the label content in the issue. And perhaps a plain text label, and of course a label with valid html.
Add a new file to tests/unit
Name it something like test_javaAccessBridge.py
You could use a file like test_winVersion.py as an example, so you can see how to write the test class and methods.
Then you can run `rununittests` and ensure all of them pass.

mwhapples:
I am on creating a test for this now. In fact I realised that by refactoring out to _processHTML would help me test it as it separates the logic for detecting and removing HTML tags from the accessibility API stuff. Also testing this in the future would be more important should we do things like replacing character entities and such like as the code could become more involved.

mwhapples:
Tests now written.

tests/unit/test_javaAccessBridge.py
```diff
@@ -0,0 +1,28 @@
+# A part of NonVisual Desktop Access (NVDA)
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+# Copyright (C) 2021 NV Access Limited, American Printing House for the Blind
```
michaelDCurran, line 4:
```suggestion
# Copyright (C) 2022 NV Access Limited, American Printing House for the Blind
```

michaelDCurran (approved):
Great work, thanks :)

PR #13563, opened 2022-03-30
Advanced settings panel: restore accelerator key

CyrilleB79:
### Link to issue number:
Fix-up of #13437.
### Summary of the issue:
With #13437, the option "Use UI Automation to access Microsoft Word document controls" in Advanced settings panel has become a combo-box; it was a checkbox before.
In the same time, the accelerator key (mapped to W) has been removed for this option: "&" removed from the label.
When testing Word with and without UIA, it was handy to focus the Advanced panel, check the first checkbox and press alt+W to jump directly to this option.
Now tabbing between the options is required.

Just add the "&" in the label to restore the accelerator key for this option.

### Testing strategy:
Opened the adv settings panel and checked that alt+W jumps to the target option.

* I would have wished to open this PR against beta to fix the issue immediately before the stable release. However, we are already in the translation freeze, thus this is not possible to change a translatable string, unfortunately. If a new translation freeze should occur, please reconsider merging this PR in beta.
* There does not seem to be a rule to define which options has an accelerator key and which has not.
Advanced settings panel have a lot of options missing accelerator key.
If you have a good reason to add them, I will be happy to do it in this PR.

### Change log entries:
Not needed: minor fix.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -2655,7 +2655,7 @@ class AdvancedPanelControls(
 		label = pgettext(
 			"advanced.uiaWithMSWord",
 			# Translators: Label for the Use UIA with MS Word combobox, in the Advanced settings panel.
-			"Use UI Automation to access Microsoft Word document controls"
+			"Use UI Automation to access Microsoft &Word document controls"
 		)
 		wordChoices = (
 			# Translators: Label for the default value of the Use UIA with MS Word combobox,


XLTechie:
I very much support this one. I have to change that option very often. I would 
actually prefer an unmapped gesture for it, but I suppose not much chance of 
that.

CyrilleB79:
> I very much support this one. I have to change that option very often. I would actually prefer an unmapped gesture for it, but I suppose not much chance of that.


Unfortunately we will have to wait until 2022.2 for this to be fixed due to translation freeze.

Regarding the unassigned gesture proposal, it is not a good idea to have it in core because, by assigning and using it, you would bypass the warning checkbox of the advanced settings panel.
As a developer and if you know what you are doing, you can of course develop a plugin for your own use to accomplish this task.

XLTechie:
The advanced dialog warning checkbox is more necessary for some things in that 
dialog than it is for others.

IMO, toggling UIA support in word is not a potentially damaging change that 
needs to be behind a scary checkbox. It is just an average advanced feature.
But there is no "normal advanced" and "really advanced" separation, so it too 
has to be behind that checkbox.

CyrilleB79:
> The advanced dialog warning checkbox is more necessary for some things in that dialog than it is for others. IMO, toggling UIA support in word is not a potentially damaging change that needs to be behind a scary checkbox. It is just an average advanced feature. But there is no "normal advanced" and "really advanced" separation, so it too has to be behind that checkbox.

I have not the same opinion: using UIA in older Word 2016 versions (maybe also some 2013 versions) may give really unsatisfying results.

Anyway, we have gone off-topic a bit. If you want to discuss further the addition of unbound gestures to control one or more options in the advanced settings panel, I recommend that you open a new issue for it, and I will be happy to continue this discussion there.

For the current PR, let's focus again on the accelerator key in the advanced settings panel (alt+letter).

seanbudd (approved):

PR #13589, opened 2022-04-07
Still report / read loaded virtualBuffer after too early focus event in Chromium 100

### LINKED ISSUE #13570: treeInterceptor.documentConstantIdentifier doesn't work in Chrome v100
### Description
This is mltony, author of BrowserNav add-on. Certain functions stopped working in Chrom v100. In particular, I was using treeInterceptor.documentConstantIdentifier property to retrieve URL of current page and it no longer seems to work on some websites in Chrome v100.
I can't figure out what's the pattern, e.g. amazon.com fails, while google .com seems to work fine.
### Steps to reproduce:
1. Open Google Chrome v100.
2. Navigate to Amazon.com
3. Open NVDA Python console by pressing NVDA+Control+Z
4. Type: `focus.treeInterceptor.documentConstantIdentifier`
### Actual behavior:
Returns empty string.
### Expected behavior:
Should return URL, something like `https://www.amazon.com/`.
Installed
#### NVDA version:
2021.3
#### Windows version:
Windows 10
Google Chrome Version 100.0.4896.60 (Official Build) (64-bit)
Used to work fine in Chrome before v100. Also works fine in Firefox and Edge.
yes
Yes, reproduces in 2022.1beta3
yes
yes

### LINKED ISSUE #13571: Google Chrome 100: NVDA no longer reads the title of the newly loaded page and automatic SayAll is broken
Google Chrome 100, and presumably MS Edge 100 which hasn't been released at the time of this issue creation, has broken some aspects of NVDA's browsing capabilities and features, resulting in the suppression of the announcement of the page title upon load, and malfunctioning of the Automatic SayAll on Page Load feature. This also tampers with the capabilities of some add-ons -- namely BrowserNav.
### Steps to reproduce:
While the issues mentioned here can be duplicated with all websites, this is just one sample selected for ease of duplication:
1. With Chrome 100, go to https://osara.reaperaccessibility.com
### Actual behavior:
NVDA just says: "Document." Moreover, if NVDA's Automatic SayAll on Page Load check box is checked, it doesn't work and NVDA remains silent. And if, I mean only if the BrowserNav add-on is enabled, moving around with Up/Down results in the second readout of many HTML elements such as links, buttons, etc., and extra error messages (not included in my attached log) will be added to the log. Pressing Control+F5 resolves the add-on issue.
### Expected behavior:
NVDA should read the title of the page upon page load. Moreover, if the relevant check box is checked, Automatic SayAll should be triggered.
Installed, also duplicated with portable
#### NVDA version:
2021.3.5, 2022.1 Beta 2, and Alpha 25159
#### Windows version:
Windows 11 Pro 64-Bit, 21 H2, 
Google Chrome 100 official
Yes
mentioned above: the latest alpha/official/beta release
Yes. However, with the BrowserNav add-on, more issues will be encountered -- the add-on won't detect its bookmarks and Up/Down will add a second readout of many HTML elements unless Control+F5 is pressed to refresh the page.
Not tried
[Log.txt](https://github.com/nvaccess/nvda/files/8399727/Log.txt)

michaelDCurran:
### Link to issue number:
Fixes #13571 
Fixes #13570 

### Summary of the issue:
Google Chrome 100, and presumably MS Edge 100 which hasn't been released at the time of this issue creation, has broken some aspects of NVDA's browsing capabilities and features, resulting in the suppression of the announcement of the page title upon load, and malfunctioning of the Automatic SayAll on Page Load feature. This also tampers with the capabilities of some add-ons -- namely BrowserNav.
It seems that in Chromium 100 there is a definite change to the events we receive when a page loads.
It looks like the following happens:
1. the user causes a page to load by pressing enter in the location bar or activating a link
2. A new document accessible is created with no name, no value, and states of focused, focusable and readonly.
3. a focus event is fired on this document.
3.1. NVDA receives the focus event, and tries to load a virtualBuffer for the document, but it is blank (no content) and NVDA cannot get its URL (documentConstantIdentifier) as the document's accValue is empty still.
4. Content starts to load, and the name, value, and descendants are added.
5. A documentLoadComplete event is fired on the document.
6.1. Currently NVDA does not pay attention to this event, as it expects that the first or another focus event would happen when most if not all of the document was loaded.
Previously, focus was not fired on the document until the name, value, and at least some of the descendants were loaded. And therefore, NVDA was able to load a virtualBuffer with at least some content.

Added an implementation of event_documentLoadComplete to the gecko_ia2 virtualBuffer class, which performs the initial reporting / sayAll on virtualBuffer load, if this has not yet occurred.
Chromium fires a documentLoadComplete event once the page is fully loaded. Thus by then all the content will be rendered in our virtualBuffer.
Also, BrowseModeDocument's documentConstantIdentifier is no longer forcefully cached on construction, rather each of the implementations of this property, cache themselves the first time they fetch a non-empty value. This works around the fact that Chromium 100 has an empty accValue when the document is first created.

### Testing strategy:
* Opened google Chrome 100
* In a new tab, visited: https://www.nvaccess.org/
* Ensured that once the page loaded, NVDA automatically announced the page title and started reading the page.
* Tested several other pages in Mozilla Firefox to ensure that browsers with existing page load logic still function correctly.

None known.

### Change log entries:

Bug fixes:
```
- NVDA will no longer fail to report the page title and start automatically reading, when a page loads in Google chrome 100. (#13571)
```

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


feerrenrut (dismissed):
Looks good. You might consider adding comments to the `_get_documentConstantIdentifier` to point out that ID is being cached intentionally only if it is valid and why.

I suspect this change will resolve the occasional system test failures for chrome also.

michaelDCurran:
@feerrenrut  I've added a comment to gecko_ia2 virtualBuffer's documentConstantIdentifier property now. However, I have chosen to remove the if condition from all other implementations of documentConstantIdentifier, caching ID as is without checking, as this mirrors the original behavior.

feerrenrut:
I note that the tests are now passing. The older appveyor comments are now hidden.

michaelDCurran:
@feerrenrut  I've added a few more changes.
Firstly, the gecko_ia2 vbufBackend now forces any pending updates to the buffer to run straight away if a documentLoadComplete event is received. This is similar to focus and alert events. This is necessary as NVDA expects that the buffer is fully up to date on these events, as it may try to start a sayAll or restore the caret position etc. Previously in Chrome when going back to a previously visited page NVDA would just say blank as the buffer was not yet fully updated.
I have also re-thought the changes to documentConstantIdentifier, both removing the need to needlessly change all the implementations, but also solving the issue of NVDA not being able to restore the virtual caret  position when going back to a page in a SPA (single page app) such as Github pull request queries. 
In short: Originally we would forcefully override the documentConstantIdentifier property with its own value on class initialization, essentially caching it for ever, so that on terminate we could save off the last caret position in the appModule keyed by the documentConstantIdentifier for later restoring the virtual caret if the user went back to that pgage. that value had to be cached as in terminate the document accessible would already be dead.
However, this never took into account two things:
1. The document's URL could change (E.g. due to an SPA) and
2. 2. the document may not have a valid URL until it is fully loaded (such as with Chromium 100).

As the change is in the base BrowseModeDocumentTreeInterceptor, all browsers will be able to take advantage of these fixes, even though we've only seen the second issue in Chromium 100 so far.

nvdaHelper/vbufBackends/gecko_ia2/gecko_ia2.cpp
```diff
@@ -1208,8 +1209,12 @@ void CALLBACK GeckoVBufBackend_t::renderThread_winEventProcHook(HWINEVENTHOOK ho
 			continue;
 		LOG_DEBUG(L"found active backend for this window at "<<backend);
 
-		//For focus and alert events, force any invalid nodes to be updated right now
-		if(eventID==EVENT_OBJECT_FOCUS||eventID==EVENT_SYSTEM_ALERT) {
+		//For focus, documentLoadComplete and alert events, force any nodes already marked as invalid  to be updated right now,
+		if(
+			eventID == EVENT_OBJECT_FOCUS
+			|| eventID == IA2_EVENT_DOCUMENT_LOAD_COMPLETE
+			||eventID==EVENT_SYSTEM_ALERT
```
feerrenrut, line 1216: nit: space after boolean OR 
```suggestion
			|| eventID==EVENT_SYSTEM_ALERT

```

source/browseMode.py
```diff
@@ -1778,7 +1795,10 @@ def doResult(result):
 			obj = container
 		return doResult(False)
 
-	def _get_documentConstantIdentifier(self):
+	documentConstantIdentifier: Optional[str]
```
feerrenrut, line 1800: Auto property docstring
```suggestion
	documentConstantIdentifier: Optional[str]
	""" Typing information for auto-property: _get_documentConstantIdentifier"""
```

source/browseMode.py
```diff
@@ -1778,7 +1795,10 @@ def doResult(result):
 			obj = container
 		return doResult(False)
 
-	def _get_documentConstantIdentifier(self):
+	documentConstantIdentifier: Optional[str]
+	# Mark documentConstantIdentifier property   for caching during the current core cycle
+	_cache_documentConstantIdentifer = True
```
feerrenrut, line 1803: I think doc strings like this are supposed to be a string after the property. I'd also refer to "auto-property" for consistency with the typing doc strings.
```suggestion
	_cache_documentConstantIdentifer = True
	""" Cache auto-property: documentConstantIdentifier during a core cycle. """
```

source/browseMode.py
```diff
@@ -1778,7 +1795,10 @@ def doResult(result):
 			obj = container
 		return doResult(False)
 
-	def _get_documentConstantIdentifier(self):
+	documentConstantIdentifier: Optional[str]
+	# Mark documentConstantIdentifier property   for caching during the current core cycle
+	_cache_documentConstantIdentifer = True
+	def _get_documentConstantIdentifier(self) -> Optional[str]:
 		"""Get the constant identifier for this document.
```
feerrenrut, line 1806: Given this ID can change, the 'constant' in the name is a bit deceptive. Can we at least point this out in the doc-string?

feerrenrut (commented):
So this will re-cache the doc ID on gain focus, set selection. Now, internally `_lastCachedDocumentConstantIdentifier` will be used.

- What about code that still uses the `documentConstantIdentifier` property? Just checked, it seems usage is limited to overrides and `browseMode.py` for core code. What about addons?
- Is there a situation where the cache should be invalidated, I.E. it has previously been set, but `documentConstantIdentifier` now returns None?

feerrenrut (approved):

XLTechie:
@michaelDCurran In browseMode.py, line 1804:
```
	_cache_documentConstantIdentifer = True
```

Shouldn't that be spelled "Identifier"?

I tried to comment this on the PR itself, but it was already closed so couldn't.

michaelDCurran:
@XLTechie   Thanks for noticing that. Fixed in commit 66b5f24fa936

PR #13618, opened 2022-04-20
Ignore hidden text in richEdit controls accessed with ITextDocument

michaelDCurran:
### Link to issue number:
None.

### Summary of the issue:
Currently if NVDA accesses a RichEdit control with the `ITextDocument objectinterface`, it does not ignore hidden text. I.e. hidden text is both read, and can be navigated through with the review cursor.
An example of this is the hidden text included at the start of hyperlinks in RichEdit controls. Usually looking something like:
```
HYPERLINK "<URL>"
```

So therefore for a link for NV access the user would here:
```
link HYPERLINK "https://www.nvaccess.org" NV access out of link
```
Where as they should really just hear:
```
link NV Access out of link
```

1. Remove the need for the user to turn on the Fetch Formatting after the cursor option in NVDA's document formatting settings, in order to fetch all formatting in a range of text. In Windows XP this was extremely slow. It seems these days it is much faster now. We specifically need this to correctly split on the boundaries of hidden text, and also to report the start and end of links within a line.
2. If a particular format run is marked as hidden, then do not produce a `formatField` or text for that run. In other words, skip over the hidden text within the text range.
3. when moving by a particular unit (character, word etc), skip over any hidden text. I.e. after the move, if the range is marked as hidden, keep moving by the given unit until outside the hidden text. this ensures the review cursor does not land on / report any hidden text.

### Testing strategy:
1. Open a new document in Microsoft word
2. Type the words: `NV Access`
3. Select these words and press `control+k` to insert a hyperlink. Give a URL of `https://www.nvaccess.org/`.
4. Copy the link to the clipboard.
5. Open Wordpad
6. Paste the copied link into a new document.
7. Read the line containing the link with NVDA. Verify that it announces "link NV Access out of link"
8. Move through the line by character with the review cursor. Ensure that NVDA only reports the letters `NV Access` and not any of the initial hidden text (I.e. HYPERLINK).

There may be a performance hit in ignoring the fetch formatting after the cursor option on some systems.
If users report such a performance hit, then some alternatives are:
* Move all this code in-process, similar to Microsoft Word's old object model support. Though this would be a lot of work for very little gain.
* Only ignore the fetch formatting after the cursor option if the richEdit control is in NVDA's process (I.e. part of NVDA's own UI). In this case there would be no cross-process performance cost, and this also answers the original motivation of fixing this bug which was to make links more presentable in richEdit controls in NVDA's UI.

### Change log entries:
Bug fixes
```
- Hidden text is no longer announced in Wordpad and other ``richEdit`` controls. (#13618)
```

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


seanbudd (approved):

source/NVDAObjects/window/edit.py
```diff
@@ -684,16 +684,17 @@ def getTextWithFields(self, formatConfig: Optional[Dict] = None) -> textInfos.Te
 			formatConfig=config.conf["documentFormatting"]
 		textRange=self._rangeObj.duplicate
 		textRange.collapse(True)
-		if not formatConfig["detectFormatAfterCursor"]:
```
feerrenrut, line 687: Is this exposed in the GUI? Can it be removed if no longer observed?

feerrenrut (approved):

source/NVDAObjects/window/edit.py
feerrenrut: From another conversation with @michaelDCurran, if this is the only usage, we can remove it from the settings dialog.

feerrenrut: There is another usage at: `source/textInfos/offsets.py:570`
Snippet:
```python
	def getTextWithFields(self, formatConfig: Optional[Dict] = None) -> textInfos.TextInfo.TextWithFieldsT:
		if not formatConfig:
			formatConfig=config.conf["documentFormatting"]
		if self.detectFormattingAfterCursorMaybeSlow and not formatConfig['detectFormatAfterCursor']:
			field,(boundStart,boundEnd)=self._getFormatFieldAndOffsets(self._startOffset,formatConfig,calculateOffsets=False)
			text=self.text
			return [textInfos.FieldCommand('formatChange',field),text]
		commandList=[]
		offset=self._startOffset
		while offset<self._endOffset:
```

michaelDCurran:
Based on this finding, I would keep this setting available. It looks like at very least Java Access Bridge, old winConsoles, displayModel, and RichEdit without ITextDocument could all be affected. In other words, pretty much anything that inherits from `textInfos.offsets.OffsetsTextInfo`  but not IAccessible2, as it sets `self.detectFormattingAfterCursorMaybeSlow` to False. I have no idea whether performance is still an issue with any of them, but at very least, definitely would require further investigation.

codeofdusk:
After this PR:

1. Open Outlook.
2. press CTRL+n.
3. Enter an email address (in the to: field), then press enter.
4. Press right arrow.

Observed: NVDA says "out of link, blank".
Expected: NVDA says "blank" (there is no indication that the email address is a link).

feerrenrut:
@codeofdusk Is the email address exposed as a link? It probably is, please report to Microsoft. By the way, adding a comment to a closed PR is very likely to get ignored. Please open a new issue so it can be triaged.

Brian1Gaff:
I quite like the ability to hear the link in that way. Very handy for hacked 
web sites. grin.
 Brian

***@***.***
Sent via blueyonder.
in the display name field.
----- Original Message ----- 
From: "Michael Curran" ***@***.***>
Cc: "Subscribed" ***@***.***>
Sent: Wednesday, April 20, 2022 7:00 AM
Subject: [nvaccess/nvda] Ignore hidden text in richEdit controls accessed 
with ITextDocument (PR #13618)

PR #13859, opened 2022-06-30
Add feature flag helper

feerrenrut:
### Link to issue number:
As per PR https://github.com/nvaccess/nvda/pull/13851
Also merges in support for multi value feature flags: https://github.com/nvaccess/nvda/pull/13871

### Summary of the issue:
NVDA increasingly makes use of feature flags to expose experimental changes to users.
There are already several approaches already in use.
It is easy to miss use-cases for changing default behavior, potentially overriding an advanced users preferred config.

Some feature flags require more than just a Boolean enabled/disabled outcome (in addition to 'default'), E.G.:
- `AllowUiaInMSWord` [`WHEN_NECESSARY`, `WHERE_SUITABLE, ALWAYS`]
- `AllowUiaInChromium` [`WHEN_NECESSARY`, `YES`, `NO`]

Several approaches have been tried:
- Store an `int` in the config that match with `enum` values, as per (`AllowUiaInMSWord` and `AllowUiaInChromium`)
- Use "options" in config (several `string` values). Literal strings in code to compare refer to these.

These approaches can hide errors:
- Looking at the config, what are the valid values when an integer is used.
- Using string "options":
  - what are the valid options?
  - how do you find all usages?
  - Change the strategy for default?
- For both, when looking at the `config` or `configSpec`, how can the behavior of default be determined?

The target user of this change are developers.
This introduces constructs and utilities to streamline the approach to adding feature flags.

Adding a new feature flag:
- Most feature flags can use the `BoolFlag`.
- If more values are required, define a new enum class in `featureFlagEnums.py` (include a `DEFAULT` member) E.G.: `MyNewFeature`
- Add the entry to the `configSpec.py` E.G. `shouldUseMyNewFeature = featureFlag(optionsEnumClass="MyNewFeature", behaviorOfDefault="ON_WEEKENDS")`
- Add a GUI option for it, using `FeatureFlagCombo`
- Wrap the code for the feature with `config.conf["section"]["shouldUseMyNewFeature "] == MyNewFeature.ON_WEEKENDS`

Introduction of a generic `BoolFlag` type (enabled/disabled) which needs to account for 3 states:
- The current NVDA default (IE user has no preference)
- User has a preference of enabled
- User has a preference of disabled

To facilitate this, one of the three values (default, enabled, disabled) is stored in the config.
Spec/config transformations are provided to streamline the support.
A `featureFlag` type in the spec is automatically converted to a `FeatureFlag` class instance.
The `FeatureFlag` class abstracts the consideration of "default", allowing the `FeatureFlag` to treated as a boolean within internal NVDA logic.

A utility GUI control is added which presents the `FeatureFlag` option to the user, and allows developers to give translated names to the logical values (enabled, disabled), the name for default incorporates the `behaviorOfDefault` (and is also translated).

Note:
- The `configSpec` explicitly defines what the behavior of default is.
- The enum class that backs the config is defined, allows for type / value checking.
- Existing flags have not been upgraded, these would require upgrade code for the config, this introduces unnecessary complexity.

### Testing strategy:
- Unit tests added to illustrate and test spec / config transform and validation.
- Manually tested various options, including updating from config in #13851

This approach bypasses the handling for defaults provided by configObj.
It was considered whether the default not written on read behavior (of the configObj library) to facilitate the third (default) state.
However:
- This would require explicit code to test for the existence of the key to know if the value returned by `conf["section"]["myKey"]` was just the default or an explicit setting.
- It also requires careful diligence to ensure that the value is not ever written accidentally.
- Code to delete the key is required whenever attempting to return to the default.
- The way validation checks are created in `configObj` makes transforming the type (to provide an abstraction) difficult. It is not possible to get access to the `default` without a similar approach used in this PR. If adopting the complexity of a spec and config transform step, then it is preferable to have explicit values for default. Explicit values make the config easier to inspect and make the approach more robust.

### Change log entries:
For Developers
```
- A config spec type ``featureFlag`` has been created for use with experimental features in NVDA. See devDocs/featureFlag.md for more information.
```

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,6 +17,10 @@ py2exe==0.11.1.0
 sphinx==3.4.1
 sphinx_rtd_theme
 
+# Requirements for type checking.
+# typing_extensions is incorporated in py3.8+, also available via mypy
+typing_extensions==4.3.0
+
 # Requirements for automated linting
 flake8 ~= 3.7.7
 flake8-tabs == 2.1.0
--- a/source/NVDAObjects/IAccessible/chromium.py
+++ b/source/NVDAObjects/IAccessible/chromium.py
@@ -90,7 +90,7 @@ class Document(ia2Web.Document):
 
 	def _get_treeInterceptorClass(self) -> typing.Type["TreeInterceptor"]:
 		shouldLoadVBufOnBusyFeatureFlag = bool(
-			config.conf["virtualBuffers"]["loadChromiumVbufOnBusyState"]
+			config.conf["virtualBuffers"]["loadChromiumVBufOnBusyState"]
 		)
 		vBufUnavailableStates = {  # if any of these are in states, don't return ChromeVBuf
 			controlTypes.State.EDITABLE,
--- a/source/config/featureFlag.py
+++ b/source/config/featureFlag.py
@@ -1,0 +1,226 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2022 NV Access Limited
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+
+"""Manages NVDA configuration.
+Provides utility classes to make handling featureFlags easier.
+"""
+import enum
+import typing
+
+from . import featureFlagEnums
+from .featureFlagEnums import (
+	BoolFlag,
+	FlagValueEnum,
+)
+from typing import (
+	Optional,
+)
+from configobj.validate import (
+	ValidateError,
+	VdtParamError,
+)
+from logHandler import log
+
+
+class FeatureFlag:
+	"""A FeatureFlag allows the selection of a preference for behavior or its default state.
+	It's typically used to introduce a feature that isn't expected to handle all use-cases well
+	when initially introduced.
+	The feature can be disabled initially, some users can manually enable and try the feature
+	giving feedback.
+	Once developers have confidence in the feature, the default behavior is changed.
+	This change in default behavior should not have any impact on users who have already tried the feature,
+	and perhaps disagreed in principle with it (I.E. wished to disable the feature, not because it is buggy,
+	but because even if it works perfectly it is not their preference).
+	The default option allows users to explicitly defer to the NVDA default behaviour.
+	The default behaviour can change, without affecting a user's preference.
+	"""
+	def __init__(
+			self,
+			value: FlagValueEnum,
+			behaviorOfDefault: FlagValueEnum
+	):
+		self.value = value
+		self.enumClassType: typing.Type[FlagValueEnum] = type(value)
+		assert self.enumClassType == type(behaviorOfDefault)
+		assert behaviorOfDefault != value.DEFAULT
+		self.behaviorOfDefault = behaviorOfDefault
+
+	def __bool__(self) -> bool:
+		if not isinstance(self.value, BoolFlag):
+			raise NotImplementedError(
+				"Only BoolFlag supported. For other types use explicit checks"
+			)
+		if self.isDefault():
+			return bool(self.behaviorOfDefault)
+		return bool(self.value)
+
+	def isDefault(self) -> bool:
+		return self.value == self.value.DEFAULT
+
+	def calculated(self) -> FlagValueEnum:
+		if self.isDefault():
+			return self.behaviorOfDefault
+		return self.value
+
+	def __eq__(self, other: typing.Union["FeatureFlag", FlagValueEnum]):
+		if isinstance(other, type(self.value)):
+			other = FeatureFlag(other, behaviorOfDefault=self.behaviorOfDefault)
+		if isinstance(other, FeatureFlag):
+			return self.calculated() == other.calculated()
+		return super().__eq__(other)
+
+	def __str__(self) -> str:
+		"""So that the value can be saved to the ini file.
+		"""
+		return self.value.name
+
+
+def _validateConfig_featureFlag(
+		value: Optional[str],
+		optionsEnum: str,
+		behaviorOfDefault: str
+) -> FeatureFlag:
+	""" Used in conjunction with configObj.Validator
+	param value: The value to be validated / converted to a FeatureFlag object.
+	Expected: "enabled", "disabled", "default"
+	param behaviorOfDefault: Required, the default behavior of the flag, should be "enabled" or "disabled".
+	"""
+	log.debug(
+		f"Validating feature flag: {value}"
+		f", optionsEnum: {optionsEnum}"
+		f", behaviorOfDefault: {behaviorOfDefault}"
+	)
+	if not isinstance(optionsEnum, str):
+		raise ValidateError(
+			'Spec Error: optionsEnum must be specified as a string'
+			f" (got type {type(optionsEnum)} with value: {optionsEnum})"
+		)
+	try:
+		OptionsEnumClass = dict(featureFlagEnums.getAvailableEnums())[optionsEnum]
+	except KeyError:
+		raise ValidateError(
+			"Spec Error: optionsEnum must be an enum defined in the config.featureFlagEnums module."
+			f" (got {optionsEnum})"
+		)
+
+	if not isinstance(behaviorOfDefault, str):
+		raise ValidateError(
+			'Spec Error: behaviorOfDefault must be specified as a valid'
+			f' {OptionsEnumClass.__qualname__} member string'
+			f" (got type {type(behaviorOfDefault)} with value: {behaviorOfDefault})"
+		)
+	try:
+		behaviorOfDefault = OptionsEnumClass[behaviorOfDefault.upper()]
+	except KeyError:
+		raise ValidateError(
+			"Spec Error: behaviorOfDefault must be specified as a valid enum member string for enum class "
+			f"{OptionsEnumClass.__qualname__} (got {behaviorOfDefault})"
+		)
+	if behaviorOfDefault == OptionsEnumClass.DEFAULT:
+		raise ValidateError("Spec Error: behaviorOfDefault must not be 'default'/'DEFAULT'")
+
+	if not isinstance(value, str):
+		raise ValidateError(
+			'Expected a featureFlag value in the form of a string. EG "disabled", "enabled", or "default".'
+			f" Got {type(value)} with value: {value} instead."
+		)
+
+	try:
+		value = OptionsEnumClass[value.upper()]
+	except KeyError:
+		raise ValidateError(
+			"FeatureFlag value must be specified as a valid enum member string for enum class "
+			f"{OptionsEnumClass.__qualname__} (got {value})"
+		)
+
+	return FeatureFlag(value, behaviorOfDefault)
+
+
+def _transformSpec_AddFeatureFlagDefault(specString: str, **kwargs) -> str:
+	""" Ensure that default is specified for featureFlag used in configSpec.
+	Param examples based on the following spec string in configSpec:
+		loadChromiumVBufOnBusyState = featureFlag(behaviorOfDefault="enabled", optionsEnum="BoolFlag")
+	@param specString: EG 'featureFlag(behaviorOfDefault="enabled", optionsEnum="BoolFlag")'
+	@param kwargs: EG {'behaviorOfDefault': 'enabled', 'optionsEnum':'BoolFlag'}
+	@return 'featureFlag(behaviorOfDefault="ENABLED", optionsEnum="BoolFlag", default="DEFAULT")'
+	@remarks Manually specifying 'default' in the configSpec string (for featureFlag) will result in a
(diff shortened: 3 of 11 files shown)


feerrenrut:
Review actions from #13851 are in commit 7ac7eaf76db45aa2a5300e44b459b30da75b1fc9

source/gui/settingsDialogs.py
```diff
@@ -2890,6 +2894,31 @@ def __init__(self, parent):
...
+		# Translators: This is the label for a group of advanced options in the
+		#  Advanced settings panel
+		label = _("Virtual Buffers")
+		vbufSizer = wx.StaticBoxSizer(wx.VERTICAL, self, label=label)
+		vbufGroup = guiHelper.BoxSizerHelper(vbufSizer, sizer=vbufSizer)
+		sHelper.addItem(vbufGroup)
+
+		self.loadChromeVbufWhenBusyCombo: nvdaControls.FeatureFlagCombo = vbufGroup.addLabeledControl(
+			labelText=_(
+				# Translators: This is the label for a checkbox in the Advanced settings panel.
```
CyrilleB79, line 2906:
```suggestion
				# Translators: This is the label for a combo-box in the Advanced settings panel.
```

source/gui/nvdaControls.py
```diff
@@ -391,3 +401,125 @@ def ScrollChildIntoView(self, child: wx.Window) -> None:
...
+				conf = conf[nextKey]
+
+	def _createOptionsDict(
+			self,
+			translatedOptions: OrderedDict[FeatureFlagValue, str]
+	) -> OrderedDict[FeatureFlagValue, str]:
+		behaviorOfDefault = self._getConfigValue().behaviorOfDefault
+		translatedStringForBehaviorOfDefault = translatedOptions[behaviorOfDefault]
+		# Translators: Label for option in the 'Load Chromium virtual buffer when document busy.'
+		# combobox in the Advanced settings panel. {} is replaced with the label of the option
```
CyrilleB79, line 517:
```suggestion
		# Translators: Label for the default option in some comboboxes in the Advanced settings panel (e.g.
		# 'Load Chromium virtual buffer when document busy'). {} is replaced with the label of the option.
```

source/config/featureFlag.py
```diff
@@ -0,0 +1,141 @@
...
+class FeatureFlagValue(enum.Enum):
+	""" The explicit DEFAULT option allows developers to differentiate between a value set that happens to be
+	the current default, and a value that has been returned to the "default" explicitly.
+	"""
+	DEFAULT = enum.auto()
+	DISABLED = enum.auto()
+	ENABLED = enum.auto()
+
+
+class FeatureFlag:
```
CyrilleB79, line 30: Even if not handled in this PR, it would be interesting to indicate how you imagine to handle non-boolean feature choices in the future, e.g. advanced option for UIA in Word documents which has 3 choices plus the default.
Would it be worth to reserve the name `FeatureFlag` for generic flag that may have any number of values (choices) and rename the `FeatureFlag` class to something like `BooleanFeatureFlag`? Or even better, implement `FeatureFlag` with all the generic stuff and implement a subclass `BooleanFeatureFlag` containing all what is specifically related to a 2-value choice (excepted Default)?
A `MultiChoiceFeatureFlag` class could be implemented subsequently.

seanbudd (dismissed):

source/config/featureFlag.py
feerrenrut: Please take a look at https://github.com/nvaccess/nvda/pull/13871

source/NVDAObjects/IAccessible/chromium.py
```diff
@@ -90,21 +90,21 @@ class Document(ia2Web.Document):
 
 	def _get_treeInterceptorClass(self) -> typing.Type["TreeInterceptor"]:
 		shouldLoadVBufOnBusyFeatureFlag = bool(
-			config.conf["virtualBuffers"]["loadChromiumVbufOnBusyState"]
+			config.conf["virtualBuffers"]["loadChromiumVBufOnBusyState"]
```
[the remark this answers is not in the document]
feerrenrut: Please see this discussion: https://github.com/nvaccess/nvda/pull/13851#issuecomment-1170982580

The short version.
The case of "Vbuf" changed to a capital B. Different keys is actually an advantage: If this PR (adding the feature flag helper) is reverted, the user's config file would become invalid if the same key name were used. This would result in their config being reset to defaults.
In the case of a revert with different keys, the config will just contain a key that is no longer used.

The similar names is a bit confusing when looking at the diff, however I'm not sure this is worth trying to clarify.

feerrenrut:
I've merged in #13871 and updated the description.

requirements.txt
```diff
@@ -17,6 +17,10 @@ py2exe==0.11.1.0
 sphinx==3.4.1
 sphinx_rtd_theme
 
+# Requirements for type checking.
+# typing_extensions is incorporated in py3.8+, also available via mypy
+typing_extensions==4.2.0
```
lukaszgo1, line 22: Any particular reason for not using the latest version:
```suggestion
typing_extensions==4.3.0
```

source/NVDAObjects/IAccessible/chromium.py
```diff
@@ -90,21 +90,21 @@ class Document(ia2Web.Document):
...
 		)
 		vBufUnavailableStates = {  # if any of these are in states, don't return ChromeVBuf
 			controlTypes.State.EDITABLE,
 		}
 		if not shouldLoadVBufOnBusyFeatureFlag:
 			log.debug(
 				f"loadChromiumVBufOnBusyState feature flag is {shouldLoadVBufOnBusyFeatureFlag},"
-				" vbuf WILL NOT be loaded when state of the document is busy."
+				" vBuf WILL NOT be loaded when state of the document is busy."
```
lukaszgo1, line 101: Changes to spelling of VBuf in the log messages should probably be in #13851 rather than here.

source/config/featureFlag.py
```diff
@@ -0,0 +1,226 @@
...
+			name_or_msg=(
+				f"Param '{behaviorOfDefaultKey}' should be one of: {[o.name for o in OptionsEnumClass]}"
+				f" but was {behaviorOfDefaultVal}. {usage}"
+			),
+			value=specString
+		)
+	if len(kwargs) != 2:
+		raise VdtParamError(
+			name_or_msg=(
+				f"Unexpected number of params."
```
lukaszgo1, line 214: No need for f-string here
```suggestion
				"Unexpected number of params."
```

source/config/featureFlag.py
```diff
@@ -0,0 +1,226 @@
...
+				f" Got {kwargs}. {usage}"
+			),
+			value=specString
+		)
+	# ensure there is the expected default
+	retString = (
+		f'_featureFlag('
```
lukaszgo1, line 221:
```suggestion
		'_featureFlag('
```

source/config/featureFlagEnums.py
```diff
@@ -0,0 +1,70 @@
...
+	def __bool__(self):
+		if self == BoolFlag.DEFAULT:
+			raise ValueError(
+				"Only ENABLED or DISABLED are valid bool values"
+				", DEFAULT must be combined with a 'behavior for default' to be Truthy or Falsy"
+			)
+		return self == BoolFlag.ENABLED
+
+
+def getAvailableEnums() -> typing.Generator[typing.Tuple[str, DisplayStringEnum], None, None]:
```
lukaszgo1, line 63: Given you had to add an exception for `DisplayStringEnum` to this function already wouldn't it be simpler not to try to detect available enums, rather create a class decorator which would be used to register known implementations? I'm worried that someone may import here something for which these conditions matches and fails to notice.

source/gui/nvdaControls.py
```diff
@@ -391,3 +404,129 @@ def ScrollChildIntoView(self, child: wx.Window) -> None:
...
+
+
+class FeatureFlagCombo(wx.Choice):
+	"""Creates a combobox (wx.Choice) with a list of feature flags.
+	"""
+	def __init__(
+			self,
+			parent: wx.Window,
+			keyPath: List[str],
+			conf,
```
seanbudd, line 416:
```suggestion
			conf: config.ConfigManager,
```

source/gui/nvdaControls.py
```diff
@@ -391,3 +404,129 @@ def ScrollChildIntoView(self, child: wx.Window) -> None:
...
+			if value != self._optionsEnumClass.DEFAULT
+		})
+		if self._optionsEnumClass.DEFAULT in translatedOptions:
+			raise ValueError(
+				f"The translatedOptions dictionary should not contain the key {self._optionsEnumClass.DEFAULT!r}"
+				" It will be added automatically. See _setDefaultOptionLabel"
+			)
+		self._translatedOptions = self._createOptionsDict(translatedOptions)
+		choices = list(self._translatedOptions.values())
+		super(FeatureFlagCombo, self).__init__(
```
seanbudd, line 449:
```suggestion
		super().__init__(
```

source/config/featureFlag.py
```diff
@@ -0,0 +1,226 @@
...
+	def __bool__(self) -> bool:
+		if not isinstance(self.value, BoolFlag):
+			raise NotImplementedError(
+				"Only BoolFlag supported. For other types use explicit checks"
+			)
+		if self.isDefault():
+			return bool(self.behaviorOfDefault)
+		return bool(self.value)
+
+	def isDefault(self):
```
seanbudd, line 60:
```suggestion
	def isDefault(self) -> bool:
```

source/config/featureFlagEnums.py
```diff
@@ -0,0 +1,70 @@
...
+_displayString = typing.TypeVar("_displayString", bound=DisplayStringEnum)
+
+
+class FeatureFlagEnumProtocol(Protocol):
+	""" All feature flags are expected to have a "DEFAULT" value.
+	"""
+	DEFAULT: enum.Enum  # Required enum member
+
+
+FlagValueEnum = typing.Union[FeatureFlagEnumProtocol, _displayString]
```
seanbudd, line 31: I believe this should be the following, based off of [this stackoverflow answer](https://stackoverflow.com/a/70907644).

```suggestion
class FeatureFlagEnumProtocol(Protocol):
	""" All feature flags are expected to have a "DEFAULT" value.
	"""
	DEFAULT: enum.Enum  # Required enum member


class FlagValueEnum(_DisplayStringEnumMixin, FeatureFlagEnumProtocol):
	pass
```

In absence of a thorough check with mypy `reveal_type`, this is what my IDE shows when using this typing structure.
image

source/config/featureFlagEnums.py
feerrenrut: I'm hesitant to complicate it further. There are unit tests that ensure that all collected enums have a `DEFAULT` which should catch most similar cases. If there was an import that brought along an enum that inherited from `DisplayStringEnum` and had a `DEFAULT` member, then it still doesn't do any harm really. I want to keep it as simple as possible to add new feature flag config.

requirements.txt
feerrenrut: No. I'll update that.

source/config/featureFlagEnums.py
feerrenrut: Yes, it wasn't ideal for pycharm either. I haven't found an example for how to enforce this.

With the suggested approach it still isn't ideal, the type system doesn't know it is an enum (thus has `name` and `value` properties. I don't want to get held up on what might just be a limitation of the type system. For now I'll just add name and value to the `Prototype` for now.

feerrenrut: One issue was that the IDE didn't understand interactions with the collection, E.G. Errors with the following:
```python
	v: FlagValueEnum = get()
	theName = v.name
	theValue = v.value
	for i in v:
		print(i)
	a = v["bla"]
	isIn = a in v
```

This seems to be fixed with:
```python
class FeatureFlagEnumProtocol(Protocol, typing.Collection):
	""" All feature flags are expected to have a "DEFAULT" value.
	"""
	DEFAULT: enum.Enum  # Required enum member
	name: str  # comes from enum.Enum
	value: str  # comes from enum.Enum


class FlagValueEnum(enum.EnumMeta, _DisplayStringEnumMixin, FeatureFlagEnumProtocol):
	pass
```

seanbudd (approved):

feerrenrut:
To ensure this arrive in alpha all at once I'm going to squash the `loadVbufStateSystemBusy` branch and squash merge this PR into that one, then merge that into master.

PR #13881, opened 2022-07-08
Support non-interactive use of ensureVenv script

LeonarddeR:
### Link to issue number:
https://github.com/nvaccess/nvda/pull/13048#issuecomment-1178807953

### Summary of the issue:
#13048 introduced another prompt to adres when dependencies in requirements.txt have changed. This blocks non-interactive use of the ensureVenv script, e.g. when running "scons source" non-interactively.

As proposed by @feerrenrut in https://github.com/nvaccess/nvda/pull/13048#issuecomment-1178807953, just use defaults when running in non-interactive mode.

See https://github.com/nvaccess/nvda/pull/13048#issuecomment-1178807953. Note that isatty is not always avaliable on sys.stdout, e.g. when it is set to the NVDA python console.

### Testing strategy:
Tested running `scons source > d:\output` and observed that the venv was dropped and recreated.

Do we want to present the question and also print that the question has been answered as default due to running in non-interactive, I figured it would be a bit overkill for that purpose.

### Change log entries:
None needed

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/venvUtils/ensureVenv.py
+++ b/venvUtils/ensureVenv.py
@@ -25,24 +25,26 @@ venv_python_version_path: str = os.path.join(venv_path, "python_version")
 isInteractive = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
 
 
-def askYesNoQuestion(message: str, default: bool = True) -> bool:
+def askYesNoQuestion(message: str) -> bool:
 	"""
 	Displays the given message to the user and accepts y or n as input.
 	Any other input causes the question to be asked again.
-	If isInteractive is False, the default is always returned.
+	If isInteractive is False, y is assumed and True is returned.
 	@returns: True for y and False for n.
 	"""
-	while isInteractive:
-		answer = input(
-			message + " [y/n]: "
-		)
+	question: str = f"{message} [y/n]: "
+	while True:
+		if isInteractive:
+			answer = input(question)
+		else:
+			answer = "y"
+			print(f"{question}{answer} (answered non-interactively)")
 		if answer == 'n':
 			return False
 		elif answer == 'y':
 			return True
 		else:
 			continue  # ask again
-	return default
 
 
 def fetchRequirementsSet(path: str) -> Set[str]:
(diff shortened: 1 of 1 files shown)


venvUtils/ensureVenv.py
```diff
@@ -19,15 +19,20 @@
...
 venv_orig_requirements_path: str = os.path.join(venv_path, "_requirements.txt")
 venv_python_version_path: str = os.path.join(venv_path, "python_version")
+#: Whether this script is run interactively,
+#: i.e. whether user input is possible to answer questions.
+#: Value is True if interactive (i.e. stdout is attached to a terminal), False otherwise.
+isInteractive = hasattr(sys.stdout, "isatty") and sys.stdout.isatty()
 
 
-def askYesNoQuestion(message: str) -> bool:
+def askYesNoQuestion(message: str, default: bool = True) -> bool:
```
feerrenrut, line 28: I think the default should be explicit. There are only two usages.

feerrenrut:
> Do we want to present the question and also print that the question has been answered as default due to running in non-interactive, I figured it would be a bit overkill for that purpose.

I think it would be a good idea to have a record of this in the build output. If this does the wrong thing in some situation it will be much easier to track it down.

venvUtils/ensureVenv.py
```diff
@@ -25,24 +25,26 @@
...
 
-def askYesNoQuestion(message: str, default: bool = True) -> bool:
+def askYesNoQuestion(message: str) -> bool:
 	"""
 	Displays the given message to the user and accepts y or n as input.
 	Any other input causes the question to be asked again.
-	If isInteractive is False, the default is always returned.
+	If isInteractive is False, y is assumed and True is returned.
 	@returns: True for y and False for n.
 	"""
```
feerrenrut, line 34: This is still implicit.

Instead:
```python
def askYesNoQuestion(message: str, default: bool) -> bool:
	"""
	Displays the given message to the user and accepts y or n as input.
	Any other input causes the question to be asked again.
	If isInteractive is False, the default is always returned, the question and outcome will still be sent to stdout for inspection of the build.
	@param default: the return value when the user can not be prompted.
	@returns: True for y and False for n.
	"""
```

leonardder: Ah I interpreted explicit the other way around 🤣

feerrenrut (approved):
Thanks for your work here @leonardder

PR #13985, opened 2022-08-04
Add-on store

feerrenrut:
This is intended to act as a base for "stacked pull requests" to build up this feature until it is considered ready to merge into NVDA.

### Link to issue number:
None

### Summary of the issue:
Historically, NVDA users have had to browse websites or use add-ons, not directly run/developed by NV Access, in order to learn about, download, and install add-ons. For a long time, easier manual updates for add-ons, and automatic updates for add-ons has been desired by NVDA users. Additionally, with the more common and structured approach to add-on API compatibility breaks, lowering the friction for delivery of updated (and compatible) add-ons is more important than ever.

User stories enabled by this work:
- As an NVDA user, I want to easily browse add-ons which are compatible with my version of NVDA, to learn if they may fulfill an unmeet user need.
- As an NVDA user, I want to be able to install compatible NVDA add-ons that I have determined will fulfill an unmet user need.

Additionally, this work unblocks the following user stories which may yet be incorporated into this PR:
- As an NVDA user, I want to be notified when there are updates to add-ons I have installed, so that I can manually update the add-ons when I am ready to do so.
- As an NVDA user, I want NVDA to automatically keep add-ons up to date with the latest version available.
- When NVDA notifies of an update (to NVDA), I want to be able to check the compatibility status of the add-ons I have installed, so that I can determine whether the NVDA update will break my core requirements.

### Description of changes for the user:
A new dialog is introduced (NVDA menu, tools, addon store) which shows the latest versions of add-ons ([submitted to the add-on store](https://github.com/nvaccess/addon-datastore)).

The list of functionality for the add-on store is expansive.
This should be covered in the user guide and manual test plan
- User guide: https://ci.appveyor.com/api/buildjobs/0ljhmekotff09mw7/artifacts/output%2FuserGuide.html
- Manual test plan: https://github.com/nvaccess/nvda/blob/addonStore-base/tests/manual/addonStore.md

Images of the add-on store dialog:
image
image

### Testing strategy:
The manual test plan exists to cover common user paths and a general smoke test of the add-on store interface:
https://github.com/nvaccess/nvda/blob/addonStore-base/tests/manual/addonStore.md

Unit tests include testing a new data-structure: CaseInsensitiveSet, intended to manage tracking add-on IDs, which should be done in a case insensitive manner.

There are many known issues with this PR.
They should be documented as issues against NVDA, or unresolved comments in https://github.com/nvaccess/nvda/discussions/14912

### Change log entries:
New features
```
- Add-on Store has been added to NVDA.
  - Browse, search, install and update community add-ons.
  - Manually override incompatibility issues with outdated add-ons.
  - The Add-ons Manager has been removed and replaced by the Add-on Store.
  - For more information please read the updated user guide.
  -
```

Deprecations:
```
- gui.MainFrame.onAddonsManagerCommand is deprecated, use gui.MainFrame.onAddonStoreCommand instead.
```

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


CyrilleB79:
It's a very good news to see this topic arrive in NVDA's repo!

The PR is not in draft. Is it already testable?

I get the following error:

```
ERROR - external:addonStore.dataManager.DataManager._getLatestAvailableAddonsData (11:25:34.047) - getAddonData (13724):
Unable to get data from API (https://nvaccess.org/addonStore/en/all/2022.4.0.json), response (400): b'Addon API version not available'
ERROR - stderr (11:25:34.104) - getAddonData (13724):
Exception in thread getAddonData:
  File "C:\Users\Cyrille\AppData\Local\Programs\Python\Python37-32\lib\threading.py", line 926, in _bootstrap_inner
    self.run()
  File "C:\Users\Cyrille\AppData\Local\Programs\Python\Python37-32\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\Cyrille\Documents\DevP\GIT\nvda\source\gui\addonStoreGui\viewModels.py", line 569, in _getAddonsInBG
    addons = self._dataManager.getLatestAvailableAddons()
  File "C:\Users\Cyrille\Documents\DevP\GIT\nvda\source\addonStore\dataManager.py", line 222, in getLatestAvailableAddons
    return self._availableAddonCache.availableAddons
AttributeError: 'NoneType' object has no attribute 'availableAddons'
```

feerrenrut:
@CyrilleB79 theoretically it should be testable.

However, I had previously been testing it before the NVDA version change, and with a prior datastore cache this error doesn't cause a major issue. I'll look into improving the robustness in this case.

There is a known issue to allow support for "in development" NVDA versions on the API side.

For testing purposes, you can safely override `currentVersion` in `_getCurrentApiVersionForURL` (`source/addonStore/dataManager.py`):
``` python
def _getCurrentApiVersionForURL() -> str:
	#currentVersion = addonAPIVersion.CURRENT
	currentVersion = (2022, 2, 0)
```

CyrilleB79:
Thanks. Done and the first  error is actually gone.

Now, I have no previous dataStore cache and this seems to cause another issue:
```
ERROR - stderr (12:13:30.476) - getAddonData (8340):
Exception in thread getAddonData:
  File "C:\Users\Cyrille\AppData\Local\Programs\Python\Python37-32\lib\threading.py", line 926, in _bootstrap_inner
    self.run()
  File "C:\Users\Cyrille\AppData\Local\Programs\Python\Python37-32\lib\threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\Cyrille\Documents\DevP\GIT\nvda\source\gui\addonStoreGui\viewModels.py", line 569, in _getAddonsInBG
    addons = self._dataManager.getLatestAvailableAddons()
  File "C:\Users\Cyrille\Documents\DevP\GIT\nvda\source\addonStore\dataManager.py", line 217, in getLatestAvailableAddons
    self._cacheAddons(addonData=decodedApiData, fetchTime=fetchTime)
  File "C:\Users\Cyrille\Documents\DevP\GIT\nvda\source\addonStore\dataManager.py", line 191, in _cacheAddons
    with open(self._cacheFile, 'w') as cacheFile:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\Cyrille\\Documents\\DevP\\GIT\\nvda\\source\\userConfig\\addonStore\\_cachedLatestAvailableAddons.json'
```

feerrenrut:
@CyrilleB79 The issue creating the cache file should now be resolved.
You'll still need to override the `currentVersion` until we resolve this on the server. We haven't decided on the best way to handle pre-release NVDA.

josephsl:
Hi,

The PR has a lot of potential.

Updating add-ons (in the future): at least the PR recognizes when updates are available. Would it be possible to present current installed version somewhere so users can compare current versus update version, similar to Add-on Updater implementation?

Also, I don't know if it is me, but when downloading add-ons, some kind of a dialog is briefly shown.

Thanks.

nvdaes:
Here's feedback from me:

1. Installation is fast and very comfortable, even when various add-ons have been selected.
2. I'd like that Install button has a shortcut, and even that multiple add-ons can be selected, for example with a list of checkboxes.
3. When the install button is pressed, it would be nice if the add-ons list was focused instead of further details.
4. Some add-ons perform onInstall tasks, and the add-on name maybe presented in this cases, or this may be documented so that authors mention it explicitly. For example, clipContentsDesigner and reportPasswords.

Congrats for this pull request. It's really comfortable how ad-ons are downloaded and installed.

feerrenrut:
Thanks for the feedback.

@josephsl
> Updating add-ons (in the future): at least the PR recognizes when updates are available. Would it be possible to present current installed version somewhere so users can compare current versus update version, similar to Add-on Updater implementation?

Our next phase will be looking at the updating use-cases more carefully. Hopefully this will address these concerns.

>Also, I don't know if it is me, but when downloading add-ons, some kind of a dialog is briefly shown.

Can you get me a log or some information about the dialog? I haven't noticed this, and don't know what it could be.

@nvdaes 
>I'd like that Install button has a shortcut, and even that multiple add-ons can be selected, for example with a list of checkboxes.

Note you can access "actions" for the selected add-on via the application/context menu. I intend to open this menu with the enter key is press while an add-on is selected. This is still a todo item. When we are confident about the full list of actions we can also add in accelerator keys.

>When the install button is pressed, it would be nice if the add-ons list was focused instead of further details.

Is the concern about how long it takes to get back to the list, perhaps an accelerator key (for the list) would be sufficient?

>Some add-ons perform onInstall tasks, and the add-on name maybe presented in this cases, or this may be documented so that authors mention it explicitly. For example, clipContentsDesigner and reportPasswords.

Yes, this is a concern. It isn't always obvious which add-on is presenting a dialog. We intent to move the install tasks running to after NVDA is restarted, which will make this worse. This might resolve itself through add-on updates. Let's give it some time, if the problem still exists when we get to that stage we can look into providing more information about which add-on install is completing.

source/addonStore/dataManager.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def _getCurrentApiVersionForURL() -> str:
+	currentVersion = addonAPIVersion.CURRENT
+	year, major, minor = currentVersion
+	return f"{year}.{major}.{minor}"
+
+
+def _getAddonStoreURL(channel: Channel, lang: str, nvdaApiVersion: str) -> str:
+	_baseURL = "https://nvaccess.org/addonStore/"
```
leonardder, line 47: Can this be a constant?

source/addonStore/dataManager.py
```diff
@@ -0,0 +1,227 @@
...
+			self._getCacheFilenameForAddon(addonData)
+		)
+		inProgressFilePath = cacheFilePath + ".download"
+		if addonData not in self.progress:
+			log.debug("the download was cancelled before it started.")
+			return None  # The download was cancelled
+		try:
+			with requests.get(addonData.addonURL, stream=True) as r:
+				with open(inProgressFilePath, 'wb') as fd:
+					for chunk in r.iter_content(chunk_size=128000):
```
leonardder, line 123: Is this chunk size chosen arbitrarily?

source/gui/addonStoreGui/viewModels.py
```diff
@@ -0,0 +1,663 @@
...
+	# noinspection PyArgumentList
+	DOWNLOAD_SUCCESS = enum.auto()
+	# noinspection PyArgumentList
+	INSTALLING = enum.auto()
+	# noinspection PyArgumentList
+	INSTALL_FAILED = enum.auto()
+	# noinspection PyArgumentList
+	INSTALLED = enum.auto()  # installed, requires restart
+	# noinspection PyArgumentList
+	ENABLED = enum.auto()  # enabled after restart
```
leonardder, line 62: Is there a particular reason why DISABLED is missing? I assume that an update can be available for disabled add-ons. I think it should update but its state should remain disabled.

source/gui/addonStoreGui/views.py
```diff
@@ -0,0 +1,609 @@
...
+	import comtypes.automation
+	import comtypes.client
+	import comInterfaces.tom
+	hwnd = ctrl.GetHandle()
+	obj = oleacc.AccessibleObjectFromWindow(
+		hwnd,
+		winUser.OBJID_NATIVEOM,
+		interface=comtypes.automation.IDispatch
+	)
+	# No definitions available (in NVDA) for ITextRange2 interface, I.E. they are not in the msftedit type
```
leonardder, line 215: Have you looked into this? I recall I did at some point, but no longer recall how and why. I guess it is preferred to use the proper interface directly.

LeonarddeR (commented):
Great to see this flying!

source/gui/addonStoreGui/viewModels.py
feerrenrut: We're doing this work in phases, currently the use-case being handled is browsing and installing new add-ons. To see disabled add-ons, you need to look at the add on manager dialog.

We plan to integrate all features into a single dialog. The next phase will be looking at updates.

source/addonStore/dataManager.py
feerrenrut: Yes, essentially.

source/addonStore/dataManager.py
feerrenrut: I assume you mean module level constant? There would need to be a good rational to reduce the encapsulation.

source/gui/addonStoreGui/views.py
feerrenrut: No, I think I had a discussion with Mick about it at the time, but it was really beyond the scope of this work.

source/addonStore/dataManager.py
leonardder: See https://stackoverflow.com/questions/23369625/ideal-chunk-size-for-python-requests

it looks like bigger chunk sizes are better for performance. Given we're dealing with add-ons that are usually smaller than 1MB, setting the chunk size to 1mb will probably mean that it's written to disk in just one write, wherea's 128 kb means that we might have multiple unnecessary writes.

source/addonStore/dataManager.py
leonardder: As long as there's only one function that needs the base url, I guess it is ok as is.

source/addonStore/dataManager.py
feerrenrut: Smaller writes allows it to be interrupted, and provides a more responsive user experience. This is a tradeoff on throughput.
On very slow connections this would be important.

josephsl:
Hi, upon further testing, looks like there is no separate dialog shown when downloading add-ons. Thanks.

nvdaes:
@feerrenrut 
> @nvdaes
I'd like that Install button has a shortcut, and even that multiple add-ons can be selected, for example with a list of checkboxes.
Note you can access "actions" for the selected add-on via the application/context menu. I intend to open this menu with the enter key is press while an add-on is selected. This is still a todo item. When we are confident about the full list of actions we can also add in accelerator keys.

OK, maybe better just to go on with this idea without the possibility of selecting multiple add-ons at the same time, since for each one we can also see further details, so multiple selections seems tricky

> Is the concern about how long it takes to get back to the list, perhaps an accelerator key (for the list) would be sufficient?

No, the list is not focused after pressing the Install button. I think it maybe related to this code:

```
def _makeButtonClickedEventHandler(_action: AddonActionVM) -> Callable[[wx.CommandEvent, ], None]:
			"""Get around python binding to the latest value in a for loop, create a new lambda
			for each value with an explicit binding to the addon details.
			"""
			# evt: wx.CommandEvent
			return lambda evt: _action.actionHandler(self._detailsVM.listItem)
```

nvdaes:
Regarding the further details view, hyperlinks are reported but seems there's nothing to do with them, for example copy URLs or open them pressing Enter. Is this intended or a todo?

source/gui/addonStoreGui/viewModels.py
```diff
@@ -0,0 +1,663 @@
...
+		):
+			# already set, exit early
+			return
+		self._listItem = newListItem
+		# ensure calling on the main thread.
+		core.callLater(delay=0, callable=self.updated.notify, addonDetailsVM=self)
+
+
+class AddonListVM:
+	presentedAttributes = [
```
nvdaes, line 157: Shouldn't this be a tuple, not list?
```suggestion
	presentedAttributes = (

```

nvdaes (changes requested):
presentedAttributes is a list. Is this needed or a tuple would be enough?

feerrenrut:
> No, the list is not focused after pressing the Install button. I think it maybe related to this code:

I see, STR:
- Tab to an add-on description
- Tab one more time to the install button
- Press enter

Actual:
- install button becomes disabled
- focus shifts to the next control: "Other details"

Expected:
Shift focus somewhere more useful.

I think It's debatable where focus should be shifted. Given this is the standard Windows behavior, I think there should be a pretty strong rational for changing it. Note that when there are other buttons in this group, focus will shift to one of those buttons.

At this stage, this is a low priority usability concern but may be worth considering further later.

source/addonStore/dataManager.py
feerrenrut: This has been addressed with an in code comment ([d1677a4](https://github.com/nvaccess/nvda/pull/13985/commits/d1677a46410b432822a687099062d82b1f69eca9)). There is certainly room for more elaborate schemes.

source/gui/addonStoreGui/views.py
```diff
@@ -0,0 +1,609 @@
...
+		# Instead, add a hidden label for the textBox, Windows exposes this as the accessible name.
+		self.otherDetailsLabel = wx.StaticText(
+			self,
+			# Translators: Label for the text control containing extra details about the selected add-on.
+			# In the add-on store dialog.
+			label=pgettext("addonStore", "Other Details:")
+		)
+		self.contents.Add(self.otherDetailsLabel, flag=wx.EXPAND)
+		self.otherDetailsLabel.Show(False)
+		self.otherDetailsTextCtrl = wx.TextCtrl(
```
lukaszgo1, line 322: It looks like using the `wx.TextCtrl` is causing several issues:
1. Failure to create links under Windows 7 as described above.
2. Links cannot be easily activated as mentioned in the known issues of the PR
3. The presentations of links in speech is pretty strange (there are some silent symbols before it when navigating character by character, and the url of the link seems to be reported twice).


Given these issues has it been considered to use either [wx.html.HtmlWindow](https://docs.wxpython.org/wx.html.HtmlWindow.html#wx-html-htmlwindow) or [wx.html2.WebView](https://docs.wxpython.org/wx.html2.WebView.html)?
These would be standard virtual buffers, with known working behavior for links activation and presentation. They behavior would also don't depend on the version of Windows in use, and styling should be significantly easier when done with CSS.
The only slight issue seems to be accessibility of these web views when focusing, but the [work around in this SO question](https://stackoverflow.com/questions/55345724/wx-html2-webview-and-screen-readers) should  help.

lukaszgo1:
@feerrenrut  From the discussion on the PR it seems you're planning to make the install tasks execute after  NVDA is restarted, rather than, as it is now, during the add-on install. This is going to make most usages of install tasks pretty difficult and make them much less useful. I don't want to go into specific examples here to avoid getting off topic, and also because you gave no reasoning for the change, so it is hard to discuss pros and cons. Could you create a separate issue / discussion outlining what, exactly, you intend to modify and why, so that add-on and core developers can be informed in advance and check if this is going to meet their needs?

feerrenrut:
@lukaszgo1 I've created https://github.com/nvaccess/nvda/issues/14430 it would be helpful if you could explain how changing the timing of the add-on installation and install-tasks would affect add-on authors.

source/gui/addonStoreGui/views.py
feerrenrut: Due to the accessibility issues with the HTML window controls and the minimal requirements for displaying the information, I chose the `wx.TextCtrl`. Converting this into a HTML window in the future might be reasonable, however security of the content will need to be carefully considered.

For now, it is much simpler to resolve the two minor issues found:
1. On old versions of Windows, the URLs won't be converted into links.
2. Activating links is simple but was considered a minor polish issue.
3. There will be a fix for this separately, there was an attempt already: [Ignore hidden text in richEdit controls accessed with ITextDocument #13618](https://github.com/nvaccess/nvda/pull/13618), however it was reverted.

source/gui/addonStoreGui/views.py
```diff
@@ -0,0 +1,609 @@
...
+	iTextRange2_disp = iTextDocument2_disp.range(0, 0)
+	iTextRange2_disp.FindText(
+		labelText,
+		comInterfaces.tom.tomForward,
+		comInterfaces.tom.tomMatchCase,
+	)
+	# move passed last character of the label string.
+	iTextRange2_disp.moveEnd(comInterfaces.tom.tomCharacter, 1)
+	iTextRange2_disp.collapse(comInterfaces.tom.tomEnd)
+	iTextRange2_disp.text = linkText
```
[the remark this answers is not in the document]
feerrenrut: Thanks, for OS before windows 8 the URLS will be inserted instead of converted to links.

source/gui/addonStoreGui/views.py
```diff
@@ -0,0 +1,609 @@
...
+	def OnItemSelected(self, evt: wx.ListEvent):
+		newIndex = evt.GetIndex()
+		log.debug(f"item selected: {newIndex}")
+		self._addonsListVM.setSelection(index=newIndex)
+
+	def _updateContextMenuItem(self, addonActionVM: AddonActionVM):
+		menuItem = self._actionMenuItemMap[addonActionVM]
+		menuItem.Enable(enable=addonActionVM.isValid)
+
+	# noinspection PyMethodMayBeStatic
```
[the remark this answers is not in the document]
feerrenrut: This work is only a prototype at this stage. The goal was to gather some feedback on usage patterns and UX. The warnings were distracting while developing.

source/gui/__init__.py
```diff
@@ -328,6 +334,24 @@ def onAddonsManagerCommand(self,evt):
 		d.Show()
 		self.postPopup()
 
+	@blockAction.when(
```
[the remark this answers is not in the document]
feerrenrut: There might be scope for that, however given the very low numbers for Windows Store users, I don't think it is worth the maintenance cost. I'll disable it.

source/gui/addonStoreGui/viewModels.py
```diff
@@ -0,0 +1,663 @@
...
+		if localAddon.isRunning:
+			return AvailableAddonStatus.RUNNING
+
+		for storeState, handlerStateCategory in addonStoreStateToAddonHandlerState.items():
+			if model.addonId in addonHandler.state[handlerStateCategory]:
+				return storeState
+	else:
+		# todo:
+		#  Handle update available, perhaps 'sideloaded' add-ons can't get updates (because numeric version
+		#  isn't available in the manifest). This will require tracking information from the add-on store.
```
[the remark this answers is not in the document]
feerrenrut: It's reasonable to consider this, and I'm sure it is common. However, this work needs to focus on the core goal first, then adapt to the common edge cases.

source/gui/addonStoreGui/views.py
```diff
@@ -0,0 +1,609 @@
...
+		)
+
+		self.SetMinSize(self.scaleSize((400, 600)))
+
+		# Translators: The name of the column that contains names of addons. In the add-on store dialog.
+		self.InsertColumn(0, pgettext("addonStore", "Name"))
+		# Translators: The name of the column that contains the addons version string. In the add-on store dialog.
+		self.InsertColumn(1, pgettext("addonStore", "Version"))
+		# Translators: The name of the column that contains the addons publisher. In the add-on store dialog.
+		self.InsertColumn(2, pgettext("addonStore", "Publisher"))
```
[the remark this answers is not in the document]
feerrenrut: This is documented as part of the submission validation, see the schema: https://github.com/nvaccess/addon-datastore-validation/blob/main/_validate/addonVersion_schema.json#L94

>Currently the add-on publisher is always the GitHub user name of the author

This was an idea in an earlier proposal but is no longer true. If there are still references to this somewhere, please point them out.

The system is described via the following repositories:
- https://github.com/nvaccess/addon-datastore
- https://github.com/nvaccess/addon-datastore-validation
- https://github.com/nvaccess/addon-datastore-transform

Contributions are welcome.

amirmahdifard:
@feerrenrut this addon stor is really grate, but if you remove manage addons dialogue from the tools menu, then how we can disable / enable or remove our addons? because currently, we are doing it from manage addons.

Qchristensen (dismissed):
Userguide change looks good.

josephsl:
Hi,

If a user disables one or more add-ons, installs add-on store build, then moves back to a previous build, the following traceback may result:

CRITICAL - __main__ (08:33:07.417) - MainThread (19836):
core failure
  File "core.pyc", line 526, in main
  File "addonHandler\__init__.pyc", line 167, in initialize
  File "addonHandler\__init__.pyc", line 78, in load
AttributeError: Can't get attribute 'AddonStateCategory' on <module 'addonHandler' from 'C:\\Program Files (x86)\\NVDA\\library.zip\\addonHandler\\__init__.pyc'>

Thanks.

nvdaes:
@seanbudd When checksum of an add-on doesn't match, a critical error is presented,but NVDA doesn't read it.
It can be tested trying to install Unigram 2.0.0 from the store,checking out this PR.
Here's NVDA's log.
[nvdaLog.zip](https://github.com/nvaccess/nvda/files/11364463/nvdaLog.zip)

seanbudd:
@nvdaes - this should be fixed as of [c70809c](https://github.com/nvaccess/nvda/pull/13985/commits/c70809cadaaaa417010cdddebfa25543ee73c9b5)

seanbudd:
Hi all - this PR is now ready for code review.
The userGuide documentation is still under review: https://github.com/nvaccess/nvda/pull/14883
Here's a [try build of the addonStore with userGuide documentation](https://ci.appveyor.com/api/buildjobs/ar11cf2d9hoy8spq/artifacts/output%2Fnvda_snapshot_try-addonStoreWithDocs-28177%2Ce7077b04.exe).
For a guide on smoke testing, review the [manual test plan](https://github.com/nvaccess/nvda/blob/addonStore-base/tests/manual/addonStore.md).

seanbudd:
Assuming no blocking issues are raised, this should get merged to alpha next week.

nvdaes:
Hi @seanbudd , now, when I try to download Unigram 2.0.0, I can read the message. Then, if I press tab, NVDA says the word Panel, and pressing Enter, it appears that the dialog to install from external source can be opened from this extra unlabelled control.

nvdaes:
Hello, I think that we should send this to the add-ons mailing list.
I'll do it unless NV Access wants to do it before.

2023-05-03 4:56 GMT+02:00, Sean Budd ***@***.***>:

seanbudd:
>Hi @seanbudd , now, when I try to download Unigram 2.0.0, I can read the message. Then, if I press tab, NVDA says the word Panel, and pressing Enter, it appears that the dialog to install from external source can be opened from this extra unlabelled control.

I cannot reproduce this issue. Could you write up more specific steps to reproduce?

>Hello, I think that we should send this to the add-ons mailing list.
As usual, we will post an announcement when we believe the feature is ready to be announced.
We plan to announce this in `nvda-devel` and `nvda-addons` when this has been merged to master and the feature is ready for alpha testing.
I'm not sure what value there is in announcing this now to the add-ons mailing list, or providing early warning that this will be merged soon.
The feature is not ready for testing for general users, and the process has not changed for add-on authors/publishers.
In case blocking issues are found, it may be some time before this is merged to master.
Could you explain what you thinks needs to be announced and why and we can consider making the announcement early?

nvdaes:
Hi Sean, sorry, I have announced this thinking that you wouldn't have concerns with this, but the next time I'll wait. My intention was to have more testers since the try build has been published, and this particular feature has been expected long time, and for add-on authors and developers it maybe of interest.
I'll try to reproduce the mentioned bug with the panel and will come back.

nvdaes:
[nvdaLogWithUnlabelledControlInStore.zip](https://github.com/nvaccess/nvda/files/11379395/nvdaLogWithUnlabelledControlInStore.zip)

nvdaes:
Hi @seanbudd , here's log reproducing the commented bug about an unlabelled control, running a portable copy of NVDA on Windows 11. This can be reproduced just sometimes. Cheers.

CyrilleB79:
Hello

I get the following error when trying to download the try build:
`{"message":"Artifacts download limit (1024 MB/day) exceeded."}`

Can NVAccess or someone else who already has downloaded this build provide an alternate link? Thanks.

seanbudd:
@CyrilleB79 - here is a google drive link of the try build: https://drive.google.com/file/d/1qDhoKsJg7SBOBt2fL00ZKybaofJs9316/view?usp=sharing

ruifontes:
Can we comment here?


Best regards,

Rui Fontes
NVDA portuguese team



Às 08:47 de 03/05/2023, Sean Budd escreveu:

ruifontes:
Thanks!
It is not code review, but only an opinion...


One feature is designed to the "normal user", if such exist, and so the 
default chanel, for me, should be Stable and not all.


Best regards,

Rui Fontes
NVDA portuguese team


Às 00:17 de 04/05/2023, Sean Budd escreveu:

seanbudd:
@nvdaes  - could you still write up specific steps to reproduce? are these correct?

1. Open add-on store
1. Disable internet connection
  - this is a more sure way to get the download fail dialog. the SHA for unigram is fixed, you must have a cached file
1. Try to install an add-on: an error dialog opens announcing the add-on download has failed.
1. Dismiss the error dialog.
1. Press tab, NVDA says the word Panel
1. Press Enter, the external add-on install file open dialog opens

nvdaes:
Yes, these steps to reproduce are correct, but the install from source dialog is not always opened. Sometimes the focus is placed in the status readonly edit box. I suppose that the problematic panel maybe disabled to avoid thisEnviado desde mi iPhoneEl 4 may 2023, a las 1:55, Sean Budd ***@***.***> escribió:﻿
@nvdaes  - could you still write up specific steps to reproduce? are these correct?

Open add-on store
Disable internet connection


this is a more sure way to get the download fail dialog. the SHA for unigram is fixed, you must have a cached file


Try to install an add-on: an error dialog opens announcing the add-on download has failed.
Dismiss the error dialog.
Press tab, NVDA says the word Panel
Press Enter, the external add-on install file open dialog opens

CyrilleB79:
Here is a list of issues reported for an UX review:

1. The option "Available add-ons" in the field "Filter by status" makes the interface confusing. Indeed if you ask "What is filtered?", the answer is twofolds depending on what is displayed in the list.
 there are two answer depending on what is displayed in the list.
   * For the options "Installed addtons", "Updatable add-ons" and "Disabled add-ons" what is displayed in the list and thus filtered is the list of add-ons installed on the system of the user
   * For the option "Available add-ons", what is displayed in the list are the add-ons available in the store and already installed add-ons are filtered out.

   IMO this is a confusing UX. There should be a choice (radio buttons or tabs) to indicate what is displayed in the list. And for the add-ons available in the store, the filter choices could be "All" or "Add-ons not yet installed on this system".
You could also have two items to open the add-on store in the Tools menu:
   - View installed add-ons
   - Search for new add-ons to install

2. The add-on list should have an accelerator, e.g. alt+L which would match all types of list.

3. For add-ons to update, it would be nice to see both the installed version and the version of the update, as it is done in add-on updater:
   - the best solution would be a new column in the list as done in Add-on Updater
   - another solution is to have it in the "Other details", e.g. "Current version" and "Update version"

4. Sorting order of the list:
"Zoom Enhancements" appears before "eMule". Case should be ignored. Or better, `strxfrm` should be used so that letters with diacritic (e.g. "é") may be ordered correctly according to the language, e.g. "é" with "e" in French.

5. I have the following add-on installed:
Read Feeds; Version: 19.0.0; Channel: Stable; Publisher: nvdaes; Status: Enabled
In updatable add-ons, I find:
Read Feeds; Version: 19.1.0; Channel: Beta; Publisher: nvdaes; Status: Update Available
I have read feeds 19.1.0 selected in updatable add-ons list and I open the add-on's help.
Actual: Help from 19.0.0 on my local disk is opened
Expected: Help from 19.1.0 is opened (better) or help should not be available.

6. It would be nice to have a shortcut to jump to the field "Other Details", e.g. "alt+O"

7. When there is no homepage, do not make "None" a link. E.g.:
Easy Table Navigator
Version: 2.4
Homepage: None
None should not be an hyperlink

8. On the last line of "Other details" field if you read the line, you have e.g.:
`Homepage: HYPERLINK "https://github.com/nvda-es/updateChannel"https://github.com/nvda-es/updateChannel`
Expected instead:
`Homepage: link https://github.com/nvda-es/updateChannel`
Same for other links.

9. The hyperlink in the "other details" field cannot be clicked in some way: neither with space or Enter, nor with the mouse. Why?
Please make them clickable (better). Or if they cannot be active, better make them plain text instead of links.

10. In the "other details" field, the value in some fields are too long, e.g. links. Thus the text wraps to the next line. For blind people, this is a bit confusing and it would be better not to wrap line at all. Unfortunately, on the opposite, sighted people probably prefer wrapped text than having to use sliders to see it all.

11. In the "other details" field, go to the following line:
"License: GPL v2"
And go throught this line character by character. You can hear:
L i c e n s e : [nothing] H P L space v 2"
so "HPL" instead of "GPL"
That's probably the "H" of "HYPERLINK".

12. When the window of the add-on store is already opened, try to open a second one via NVDA menu.
Expected: The already opened window should be put to foreground.
Actual: Nothing happens.

13. When add-on store dialog is open in the background with last focus not in the list, e.g. on the "Filter by status" combobox, if you alt+tab back to it:
Actual: you can hear that the foreground window, the selected add-on and the focused control are reported, e.g.:
"Add-on Store  dialog  Add-ons documentation  Filter by status:  combo box Available add-ons  collapsed  Alt+f"
Expected: you should hear that the foreground window and the focused control are reported, not the currently selected add-on, e.g.:
"Add-on Store  dialog  Filter by status:  combo box Available add-ons  collapsed  Alt+f"

14. When I have installed 2 add-ons and close the add-on store dialog, I get the following message dialog:
"Now installing 2 add-ons."
With only "OK" button.
Just after, I get the restart dialog.
Issue: the first dialog is not useful at all since the user can only press OK and is immediately followed by the second one. If there is a possibility that the second dialog does not show up immediately after closing the add-on store dialog, rather show a progress dialog that will disappear without any interaction of the user when the installation is complete. E.g. "Please wait, add-ons are being installed".

15. If I uninstall an add-on but do not close yet the dialog, I can see that the "Disable" item in the contextual menu of the add-on in pending install state is still enabled. This does not make sense and I do not know what it does actually.

16. In the context menu of an add-on in the list, the shortcut key for "Install (override incompatibility)" is "I" as for "Install"; please find another one. Same for "Enable" with "E".

17. Issue with restart window and NVDA not speaking:
STR:
    * In add-on store window, uninstall an add-on which will be flagged as pending install; do not close the add-on store window.
    * Alt+tab to another window, e.g. windows explorer
    * NVDA+Q and restart NVDA
    
    Actual: The dialog indicating that changes require a restart shows up but NVDA does not speak anymore.
Expected: Do not show this dialog and restart since a restart has been requested anyway.
Note: This issue is the same as #14525 reported for the old add-on manager.

19. Context help (F1) does not seem to be working in the add-on store. Will probably be fixed when the PR #14883 will be merged?

nvdaes:
Hello:

STR:

1. Open the add-on store dialog.
2. Select an installed add-on.
3. Press the disable button.
4. Press the remove button.
5. Confirm the Remove dialog.

Expected:
The status should reflect that the add-on will be removed after restarting NVDA.

Actual:
The status will be Disabled.

nvdaes:
Replying to @CyrilleB79 
> for the add-ons available in the store, the filter choices could be "All" or "Add-ons not yet installed on this system".

I would prefer the option Addons not installed on this system.

> You could also have two items to open the add-on store in the Tools menu:
◦ View installed add-ons
◦ Search for new add-ons to install

For me the combo box is good, and radio buttons aren't needed, neither various dialogs in a submenu.
I guess that many aspects of the gui will be controversial and each user would have their own opinion.

CyrilleB79:
In addition to https://github.com/nvaccess/nvda/pull/13985#issuecomment-1534456103, here are other points for an UX review:

19. It would be nice to have the type of list displayed in the window title, i.e.
"Add-on Store - Installed add-ons", "Add-on Store - Available add-ons", etc.

20. The checkbox to include outdated incompatible add-on is alone in its settings panel. Is there a project to have other controls in this panel?
IMO, it would make sense to have it directly in the add-on store dialog where we have all the other controls used to determine which add-ons are shown in the list or not.
On the contrary, if you think that it's a setting that shoud not be played with easily by novice users, rather put it in the advanced settings dialog. But IMO, that not a good idea given the number of persons who have been playing with the manifests.

nvdaes:
Replying to @CyrilleB79 👍 
> It's not a matter of preferring radio buttons or combo boxes. If at the end we use radio buttons or not is not important.
What is important is that the GUI should provide a natural answers to the following questions:...

I agree.

> It would be nice to have the type of list displayed in the window title, i.e.
"Add-on Store - Installed add-ons", "Add-on Store - Available add-ons", etc.

Agreed.

> The checkbox to include outdated incompatible add-on is alone in its settings panel. Is there a project to have other controls in this panel?
IMO, it would make sense to have it directly in the add-on store dialog where we have all the other controls used to determine which add-ons are shown in the list or not.

I think that it makes sense to have this as a configurable option, to be used in add-ons, for example. I ad thought about the possibility to add a button in the add-on store dialog to open the add-on store settings dialog, which may be also available from NVDA's Preferences submenu,  like I did in the readFeeds add-on.

cary-rowen:
For normal user,, it might be better to put the checkbox "include outdated incompatible" in the advanced settings.
I'm concerned that some add-on incompatibilities might cause problems that normal users can't fix, like NVDA restarting wirelessly, or synths being completely silent.
Learning to check a checkbox is far easier than solving these tricky problems.

nvdaes:
I made a comment in PR #14883 by mistake, due to autocompletion in the browser address bar:

An issue to be considered, maybe later, is that the addonId is not presented in other details. In the About button of the old add-on manager, this is reflected along the display name.
Also, what would happen if someone releases an add-on with the same id than a known add-on from other author? For example, let's suppose that I release characterInformation, making it updatable, but presenting a fake version, and even making that the authentic add-on from Cyrille is not presented as the last version.
I know that there's an opened issue on add-on datastore repo related to add-ons origin. But I don't know if this should be addressed before mergint this PR.

Now I want to add that, if providing a mechanism to confirm add-on origin is hard to implement, some good practices maybe recommended, like providing public key and signature in the description, for example in the link to the release.
I'm signing my add-ons with GPG, using GitHub Actions, and including sha256, the public key and the add-on signature as assets.
Providing a link to releases in the description may give some feeling of security, since at least people know what's the GitHub account of the release, and someone with knowledge to verify signatures can do it.
IMO, if a different procedure is provided in the store, this should be optional for publishers.

XLTechie:
With regards to the questions from @nvdaes: isn't the add-on ID now mostly the 
same as the name without spaces? Listing it seems redundant. But maybe you have 
thought of a use case I haven't?

As for your other question (same name submissions): this is indeed concerning.

There are two possibilities: someone releases an add-on with the same name, to 
"cover over" an existing one. Probably a hostile action.

Equally, though, someone could release an add-on with the same name, that does 
something else entirely, or isn't trying to cover up the other add-on, but just 
wants to use the same name.

I had thought, in the original store proposal, that they were going to do 
namespacing by some criteria to avoid this sort of thing. But it's been a long 
long time since I read through that.

Still though, if the store could be re-organized to namespace on publisher, this 
problem could be avoided. Although, publisher would have to be made a 
filesystem-safe field, so that it could be used in directory names.

Alternatively, namespace on the first GitHub handle that submitted an add-on. 
Anyone else submitting an add-on not from that handle would either have it show 
up as a new add-on, or would have to go through some merge authorization 
process to get the new one put in as an alias (human review, perhaps).
The alias procedure could be added later, as it would probably require another 
JSON file and such.

nvdaes:
### to @XLTechie 👍 
> isn't the add-on ID now mostly the 
same as the name without spaces? Listing it seems redundant. 

The display name maybe whatever. There arent restrictions for this and it's not mandatory that needs to be the name without spaces, though generally it's done like that.

For the other questions, even maybe that different people releases legitim versions of an add-on. I think that the best solution maybe to send public keys to NV Access, or something similar, to confirm that the person who released an add-on is verified.
I think that, if each add-on needs to have an unique id, to allow even that different maintainers take care of it, if someone wants to publish an add-on with the sane name, this person needs to use another ID, like we need different IDs for accounts and so on.

CyrilleB79:
Also, the add-on name is translated whereas the add-on ID is always the same, no matter NVDA's interface language. So the add-on ID may be a useful disambiguation when people using two different language speak together.

josephsl:
Hi, regarding publishers: a few weeks ago, I and the community demonstrated why organizing around publishers might not be an effective strategy: change of maintainers and their GitHub handles. As for the confusion surrounding names and ID’s, it is important to inform add-on authors and publishers that add-on ID (name) is what NVDA uses to distinguish add-ons, and summary (title) is the actual name shown to users and can be translated. Thanks.

Qchristensen (dismissed):
Great work everyone on such a large project.  User guide changes for the settings section here look good.  I've commented on the main user guide changes in the issue for that and am happy with those overall.

I look forward to seeing the feedback once this becomes more widely tested and eventually in stable and I think while we might still make changes, everyone involved can be proud of their work on such a large project.

nvdaes:
Hello: should be the download URL presented as plain text? We can provide a false source URL, but not a download url. I don"t think that someone does it since it would be known quickly by people, but just in case, since it"s a way to confirm the origin of add-ons before updating them.

seanbudd:
@CyrilleB79 - I have attempted to respond to your first round of feedback. 

> 1. The option "Available add-ons" in the field "Filter by status" makes the interface confusing. Indeed if you ask "What is filtered?", the answer is twofolds depending on what is displayed in the list.
>What is displayed in the add-on list? (add-ons on the system, add-ons that can be downloaded from the store)
>Are all the add-ons displayed in the list or only a subset? And which subset?

To me, the options you describe as solutions seems a lot more verbose and confusing.

I think the answers to your questions raised here should be obvious based on the combobox option label.
When selecting "installed", "disabled" it is clear to me that this would fetch add-ons from the system.
When selecting "updatable" it is clear to me that this shows installed add-ons with an online store version for update.
"available" is somewhat vague - this could perhaps be renamed to "downloadable"?

> 2. The add-on list should have an accelerator, e.g. alt+L which would match all types of list.

Fixed via ac7e12f

> 3. For add-ons to update, it would be nice to see both the installed version and the version of the update, as it is done in add-on updater:

This still needs to be done

> 4. Sorting order of the list:

Fixed via caf9b23

> 5. I have the following add-on installed:
>    Read Feeds; Version: 19.0.0; Channel: Stable; Publisher: nvdaes; Status: Enabled
>    In updatable add-ons, I find:
>    Read Feeds; Version: 19.1.0; Channel: Beta; Publisher: nvdaes; Status: Update Available
>    I have read feeds 19.1.0 selected in updatable add-ons list and I open the add-on's help.
>    Actual: Help from 19.0.0 on my local disk is opened
>    Expected: Help from 19.1.0 is opened (better) or help should not be available.

This has been fixed by making the help unavailable. 
Without including this in the submission JSON, showing the 19.1.0 version is not possible.
We could address this when we merge the manifest and the submission JSON schemas.
Fixed via c3c3e3b

> 6. It would be nice to have a shortcut to jump to the field "Other Details", e.g. "alt+O"

Fixed via 04c3363

> 7 - 11

The issues with links were known issues with the PR, listed in the description.
I have now removed links from the "other details" section, in favour of using action buttons to open the links.
Fixed via 66b8cda

> 12. When the window of the add-on store is already opened, try to open a second one via NVDA menu.

Fixed via 90a8933

> 13. When add-on store dialog is open in the background with last focus not in the list, e.g. on the "Filter by status" combobox, if you alt+tab back to it:

Not sure how to fix this - seems like a wx bug. Is this a serious and misleading problem? Or a nice to have?

> 14. When I have installed 2 add-ons and close the add-on store dialog, I get the following message dialog:

This has been turned into a progress dialog via ade51ed

> 15. If I uninstall an add-on but do not close yet the dialog, I can see that the "Disable" item in the contextual menu of the add-on in pending install state is still enabled. This does not make sense and I do not know what it does actually.

Fixed via 5eabc23

> 16. In the context menu of an add-on in the list, the shortcut key for "Install (override incompatibility)" is "I" as for "Install"; please find another one. Same for "Enable" with "E".

Could you explain why? What do you suggest instead?
Are you concerned about the collision of keys? This should not happen as only one of the buttons should be visible for "install" or "Install (override incompatibility)".

> 17. Issue with restart window and NVDA not speaking:

Given this is an existing and documented bug I am leaving this as out of scope.

> 18. Context help (F1) does not seem to be working in the add-on store. Will probably be fixed when the PR [Add-on store: Add User Guide documentation #14883](https://github.com/nvaccess/nvda/pull/14883) will be merged?

The try build I posted a link to should include the documentation and have working context help for the add-on store settings. Did you test the try build or something different?

seanbudd:
Hi all - discussion here is getting impossible to follow and track. I am locking conversation.
Please provide feedback and discussion in this thread: https://github.com/nvaccess/nvda/discussions/14912

seanbudd:
Hi all,

This PR is ready for another review.

Rather than UX feedback, we are asking primarily for code review and review of documentation.
Please provide code review through the GitHub code review process.

At this stage, only significant UX issues will block the merge to alpha.
Please provide any significant UX feedback in this discussion: https://github.com/nvaccess/nvda/discussions/14912.
For less urgent/blocking feedback such as feature improvements, please open new NVDA issues.
A try build for this PR is available here: https://drive.google.com/file/d/1lQc72Mqy1gW9fuA4-97aePhm46mhk8eg/view

source/gui/_addonStoreGui/viewModels/store.py
```diff
@@ -0,0 +1,437 @@
...
+					and aVM.model.canOverrideCompatibility
+				),
+				listItemVM=selectedListItem
+			),
+			AddonActionVM(
+				# Translators: Label for an action that removes the selected addon
+				displayName=pgettext("addonStore", "&Remove"),
+				actionHandler=self.removeAddon,
+				validCheck=lambda aVM: (
+					aVM.model.isInstalled
```
nvdaes, line 187:
```suggestion
					aVM.model.isInstalled
					and self._filteredStatusKey in (
						# Showing remove in the updatable add-ons view is useless
						_StatusFilterKey.INSTALLED
					)
					and aVM.status != AvailableAddonStatus.PENDING_REMOVE
				),
```

source/gui/addonGui.py
nvdaes, line 1: @seanbudd Updated add-ons cannot be removed, but a different version is installed. So I think we should exclude updatable add-ons.

source/gui/_addonStoreGui/viewModels/store.py
seanbudd: this should also include `_StatusFilterKey.INSTALLED` as well

user_docs/en/userGuide.t2t
```diff
@@ -2580,63 +2581,100 @@ To toggle the braille viewer from anywhere, please assign a custom gesture using
...
+The Add-on Store displays a list of add-ons.
+You can jump to the list with ``alt+l``.
+If you have not installed an add-on before, the add-on store will open on a list of add-ons available to install.
+If you have installed add-ons, the list will display currently installed add-ons.
 
-To browse and download available add-ons online, press the Get add-ons button.
-This button opens the [NVDA Add-ons page https://addons.nvda-project.org/].
-If NVDA is installed and running on your system, you can open the add-on directly from the browser to begin the installation process as described below.
-Otherwise, save the add-on package and follow the instructions below.
+You can sort add-ons by their various fields by pressing ``ctrl+1,2,3...``, where the number is the column index you wish to sort by.
```
leonardder, line 2596: These actions don't give user feedback when sorting in the GUI. While this is UX related, I'd suggest implementing sort feedback first before publicly announcing this feature in the documentation.

nvdaes (changes requested):
Add-on ID should be shown and it's not a new feature, since it can be seen in the current add-ons manager pressing the about button

LeonarddeR (commented):
One concern I have, is that starting a build of NVDA with the add-on store in place, then starting a build without the add-on store seems to generate a new addons state, as disabled add-ons are enabled again. While downgrading is discouraged, I think this should at least be documented as known behavior.

Overall this is very, very awesome work. Thanks a lot for this!

source/_addonStore/network.py
```diff
@@ -0,0 +1,218 @@
...
+"""
+
+
+def _getCurrentApiVersionForURL() -> str:
+	year, major, minor = addonAPIVersion.CURRENT
+	return f"{year}.{major}.{minor}"
+
+
+def _getAddonStoreURL(channel: Channel, lang: str, nvdaApiVersion: str) -> str:
+	_baseURL = "https://nvaccess.org/addonStore/"
```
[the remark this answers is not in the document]
seanbudd: I'm not sure why that's needed? I think it makes sense to keep it in the scope where it is used.

user_docs/en/userGuide.t2t
seanbudd: How do you suggest giving feedback for these sorts?

source/gui/_addonStoreGui/controls/details.py
```diff
@@ -0,0 +1,309 @@
...
+				# of AppendText) because AppendText has been overridden to use SetValue(GetValue()+newStr)
+				# which drops formatting. Instead, set the text, then the style.
+				self.descriptionTextCtrl.SetValue(details.description)
+				self.descriptionTextCtrl.SetStyle(
+					0,
+					self.descriptionTextCtrl.GetLastPosition(),
+					self.defaultStyle
+				)
+
+				self._appendDetailsLabelValue(
```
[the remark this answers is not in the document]
seanbudd: Please suggest this in the add-on store discussion: https://github.com/nvaccess/nvda/discussions/14912

user_docs/en/userGuide.t2t
seanbudd: I would also encourage this sort of feedback to be provided via https://github.com/nvaccess/nvda/discussions/14912

seanbudd: >One concern I have, is that starting a build of NVDA with the add-on store in place, then starting a build without the add-on store seems to generate a new addons state, as disabled add-ons are enabled again. While downgrading is discouraged, I think this should at least be documented as known behavior.

Also responding to this here, please raise this in https://github.com/nvaccess/nvda/discussions/14912.
Handling downgrading is definitely out of scope of this PR

source/config/configSpec.py
```diff
@@ -232,6 +232,9 @@
 [documentNavigation]
 	paragraphStyle = featureFlag(optionsEnum="ParagraphNavigationFlag", behaviorOfDefault="application")
 
+[addonStore]
```
nvdaes, line 235:
```suggestion
```

source/config/configSpec.py
```diff
@@ -232,6 +232,9 @@
 [documentNavigation]
 	paragraphStyle = featureFlag(optionsEnum="ParagraphNavigationFlag", behaviorOfDefault="application")
 
+[addonStore]
+	incompatibleAddons = boolean(default=false)
```
nvdaes, line 236:
```suggestion
```

nvdaes (changes requested):
I think that the addonStore section in configSpec should be removed if it's not used, to prevent unexpected usage.

source/gui/_addonStoreGui/controls/storeDialog.py
```diff
@@ -0,0 +1,337 @@
...
+			self.includeIncompatibleCtrl.Enable()
+			self.includeIncompatibleCtrl.Show()
+		else:
+			self._storeVM._filterChannelKey = Channel.ALL
+			self.enabledFilterCtrl.Show()
+			self.enabledFilterCtrl.Enable()
+			self.includeIncompatibleCtrl.Hide()
+			self.includeIncompatibleCtrl.Disable()
+
+	def onListTabPageChange(self, evt: wx.EVT_CHOICE):
```
nvdaes, line 279: This method allows to update the same add-on multiple times, selecting the update action in the updatable add-ons page, moving to a different page and again there.

seanbudd: As this is UX feedback, could you please write up a comment in https://github.com/nvaccess/nvda/discussions/14912.

seanbudd: additionally - this code snippet is not the cause. We need to ensure the correct status is fetched when loading add-ons with an update, pending installation.

user_docs/en/userGuide.t2t
```diff
@@ -311,6 +311,7 @@ If you have add-ons already installed there may also be a warning that incompati
 Before you're able to press the Continue button you will have to use the checkbox to confirm that you understand that these add-ons will be disabled.
 There will also be a button present to review the add-ons that will be disabled.
 Refer to the [incompatible add-ons dialog section #incompatibleAddonsManager] for more help on this button.
+You are able to re-enable incompatible add-ons at your own risk.
```
michaelDCurran, line 314:
```suggestion
You are able to re-enable incompatible add-ons from within the Add-on Store dialog after installation at your own risk.
```

user_docs/en/userGuide.t2t
michaelDCurran: I was originally very confused with this feature. As I would have expected that ctrl+1 sorted by the first column (I.e. the add-on's name) however I think ctrl+1 matches version? And ctrl+0 sorts by name?  I also agree that this feature could be better if  the sorting was exposed somehow. My feeling is that these shortcuts should be removed from the user guide before this is merged. And consider perhaps before 2023.2 adding another combo box for sorting.

michaelDCurran (approved):
Just two small changes to the user guide. But otherwise I'm good.

user_docs/en/userGuide.t2t
nvdaes: I agree with @michaelDCurran regarding sorting columnas. I'd prefer that a combo box would be included and I think that the current shortcuts can be removed from the user guide.

Qchristensen (approved):
Looks good.  In the user guide section about incompatible add-ons, it mentions that "If the incompatible add-on causes issues, you can disable or remove it."  Is it worth mentioning here or linking to the ways you can start NVDA with add-ons disabled?  (NVDA+q, down arrow to "restart with add-ons disabled", or the --disable-addons command line option)?

nvdaes:
Quentin wrote:

> In the user guide section about incompatible add-ons, it mentions that "If the incompatible add-on causes issues, you can disable or remove it." Is it worth mentioning here or linking to the ways you can start NVDA with add-ons disabled?

I think this is a good idea for completeness.

user_docs/en/userGuide.t2t
leonardder: +1

user_docs/en/userGuide.t2t
```diff
@@ -2590,63 +2591,101 @@ To toggle the braille viewer from anywhere, please assign a custom gesture using
...
-+
 
-The Incompatible add-ons manager also has an "About add-on..." button.
-This dialog will provide you with the full details of the add-on, which is helpful when contacting the add-on author.
+For older add-ons, you can override the incompatibility at your own risk.
+Incompatible add-ons may not work with your version of NVDA, and can cause unstable or unexpected behaviour including crashing.
+You can override compatibility when enabling or installing an add-on.
+If the incompatible add-on causes issues, you can disable or remove it.
+
+To restart NVDA with all add-ons disabled, chose the appropriate option when quitting NVDA.
```
nvdaes, line 2684:
```suggestion
To restart NVDA with all add-ons disabled, choose the appropriate option when quitting NVDA.

```

nvdaes, line 2684:
```suggestion
To restart NVDA with all add-ons disabled, choose the appropriate option when quitting NVDA.

```

nvdaes (changes requested):

nvdaes (approved):

seanbudd:
Now that https://github.com/nvaccess/nvda/pull/13985 has been merged, please open new issues for add-on store issues, as you would for any other issue on alpha.

Issue #12882, opened 2021-09-28
Setting Automatic row and column header reading does not work in Excel with UIA enabled


msviolafangirl:
### Steps to reproduce:
1. Go to advanced settings and tick "Use UI Automation in Excel spreadsheets) option
2. Open an Excel spreadsheet.
3. Enter text in a cell.
4. On the cell where you just entered text, press NVDA+shift+C to make it a column header.

### Actual behavior:
When I press NVDA+shift+C or NVDA+shift+R to set automatic row or column header reading in Excel with UI automation enabled, I wind up in an edit box to edit the contents of the cell instead of creating row and column headers for automatic reading.

### Expected behavior:
Pressing NVDA+shift+C or NVDA+shift+R should set automatic row and column header reading.

#### NVDA installed/portable/running from source:
Installed.

#### NVDA version:
2021.2

#### Name and version of other software in use when reproducing the issue:
Excel 16/365

#### Does the issue still occur after restarting your computer?
yes

#### Have you tried any other versions of NVDA? If so, please report their behaviors.
no

#### If NVDA add-ons are disabled, is your problem still occurring?
yes

#### Does the issue still occur after you run the COM Registration Fixing Tool in NVDA's tools menu?
yes

jage9:
I also notice that there is no script in Input Gestures that appears to be associated with either of these commands. Keyboard help does not give information as well.

Adriani90:
This affects also MS Word.

CyrilleB79:
@Adriani90, reading row/columns headers in Excel UIA is currently not possible at all with NVDA; and Excel UIA is still an experimental feature that is not enabled by default.
On the opposite, Word UIA is not experimental anymore and is enabled for recent versions of Word and Windows. And there is a way to have row/column headers read with Word UIA; the UX is just different than the one used for legacy Word documents (see https://github.com/nvaccess/nvda/issues/16335#issuecomment-2019830373 for more details).

So we should avoid to factorize the issues with table headers reading between Excel UIA and Word UIA (if any).

PR #14132, opened 2022-09-12
Reporting table headers can now be configured separately for rows and columns

### LINKED ISSUE #14075: Independent setting for report Row/column headers (Document Formatting panel)
Just a small inconvenience. Sometimes users want to read just row, or colum header, but not both at the same time.

NVDA should allow the user to configure if the user wants to read the header of colums, rows, both, or disabled.

I checked the documentation to know if this can be done with some option.

### Additional context
In my opinion, the posibility to choose between read just colum or row headers, but not both aht the same time, can be useful when the user is reading a big table, for example in excel.

CyrilleB79:
### Link to issue number:
Closes #14075
### Summary of the issue:
It was asked in #14075 to be able to have table row/column headers reporting configured separately.

* A combo-box in Documentation formatting settings allows to select among 4 options to have table headers reported: Off, Columns and rows, Rows and Columns; it replaces the previous checkbox for table headers reporting.
* Setting row/column headers in Word or Excel (no UIA) is now possible even if the option to report table headers is disabled. Indeed, it is clearer from a UX point of view to allow it than to disallow it only when reporting table headers in the same direction (row/column) is disabled.
* The toggle script for table headers reporting has also been adapted to cycle through the 4 options.

* Implemented as per https://github.com/nvaccess/nvda/issues/14075#issuecomment-1240263166.
* Created an enumeration to store the possible options for this parameter, containing the numeric values for the config file and the display strings for the UI.
* Not directly linked but a a `noqa` exception for code complexity of `getObjectSpeech` has been removed because it was not relevant anymore. Excepted the removal of this #`noqa` directive, `getObjectSpeech` is not modified in this PR.

### Testing strategy
Manual testing of table navigation commands combining all the following possibilities:
* Tested in the following applications/controls:
  * Browsers (Chrome, Firefox, Edge, `ui.browseableMessage`)
  * Word with and without UIA
  * Excel without UIA (not possible to set table headers when using UIA, as described in #12882 )
  * Outlook message list
  * list view In task manager's tab "Details" (of processes)
* Tested with the 4 options: "Off", "Rows and columns", "Rows" and "Columns".
* Tested that headers were reported with speech with the appropriate options
* Tested that columns headers were reported with braille when "Rows and columns" or "Columns" option is enabled.

It has also been checked in `nvda.ini` that saving the config works as expected.

None

### Change log entries:
New features (or Changes ?)
`Reporting table headers can now be configured separately for rows and columns. (#14075)`

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/IAccessible/winword.py
+++ b/source/NVDAObjects/IAccessible/winword.py
@@ -13,7 +13,6 @@ import speech
 import braille
 import controlTypes
 import config
-from config.configFlags import ReportTableHeaders
 import tableUtils
 import textInfos
 import eventHandler
--- a/source/NVDAObjects/IAccessible/winword.py
+++ b/source/NVDAObjects/IAccessible/winword.py
@@ -13,6 +13,7 @@ import speech
 import braille
 import controlTypes
 import config
+from config.configFlags import ReportTableHeaders
 import tableUtils
 import textInfos
 import eventHandler
@@ -205,7 +206,7 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 
 	def script_setColumnHeader(self,gesture):
 		scriptCount=scriptHandler.getLastScriptRepeatCount()
-		if not config.conf['documentFormatting']['reportTableHeaders']:
+		if config.conf['documentFormatting']['reportTableHeaders'] == ReportTableHeaders.OFF:
 			# Translators: a message reported in the SetColumnHeader script for Microsoft Word.
 			ui.message(_("Cannot set headers. Please enable reporting of table headers in Document Formatting Settings"))
 			return
@@ -233,7 +234,7 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 
 	def script_setRowHeader(self,gesture):
 		scriptCount=scriptHandler.getLastScriptRepeatCount()
-		if not config.conf['documentFormatting']['reportTableHeaders']:
+		if config.conf['documentFormatting']['reportTableHeaders'] == ReportTableHeaders.OFF:
 			# Translators: a message reported in the SetRowHeader script for Microsoft Word.
 			ui.message(_("Cannot set headers. Please enable reporting of table headers in Document Formatting Settings"))
 			return
--- a/source/NVDAObjects/IAccessible/winword.py
+++ b/source/NVDAObjects/IAccessible/winword.py
@@ -206,10 +206,6 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 
 	def script_setColumnHeader(self,gesture):
 		scriptCount=scriptHandler.getLastScriptRepeatCount()
-		if config.conf['documentFormatting']['reportTableHeaders'] == ReportTableHeaders.OFF:
-			# Translators: a message reported in the SetColumnHeader script for Microsoft Word.
-			ui.message(_("Cannot set headers. Please enable reporting of table headers in Document Formatting Settings"))
-			return
 		try:
 			cell=self.WinwordSelectionObject.cells[1]
 		except COMError:
@@ -234,10 +230,6 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 
 	def script_setRowHeader(self,gesture):
 		scriptCount=scriptHandler.getLastScriptRepeatCount()
-		if config.conf['documentFormatting']['reportTableHeaders'] == ReportTableHeaders.OFF:
-			# Translators: a message reported in the SetRowHeader script for Microsoft Word.
-			ui.message(_("Cannot set headers. Please enable reporting of table headers in Document Formatting Settings"))
-			return
 		try:
 			cell=self.WinwordSelectionObject.cells[1]
 		except COMError:
--- a/source/config/configFlags.py
+++ b/source/config/configFlags.py
@@ -1,0 +1,37 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2022 NV Access Limited, Cyrille Bougot
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+
+"""Flags used for the configuration.
+"""
+from enum import unique
+from utils.displayString import DisplayStringIntEnum
+
+
+@unique
+class ReportTableHeaders(DisplayStringIntEnum):
+	"""The possible config values to report table headers.
+	"""
+	
+	OFF = 0
+	ROWS_AND_COLUMNS = 1
+	ROWS = 2
+	COLUMNS = 3
+	
+	@property
+	def _displayStringLabels(self):
+		return {
+			# Translators: This is the label for a combobox in the
+			# document formatting settings panel.
+			ReportTableHeaders.OFF: _("Off"),
+			# Translators: This is the label for a combobox in the
+			# document formatting settings panel.
+			ReportTableHeaders.ROWS_AND_COLUMNS: _("Rows and columns"),
+			# Translators: This is the label for a combobox in the
+			# document formatting settings panel.
+			ReportTableHeaders.ROWS: _("Rows"),
+			# Translators: This is the label for a combobox in the
+			# document formatting settings panel.
+			ReportTableHeaders.COLUMNS: _("Columns"),
+		}
--- a/source/config/configFlags.py
+++ b/source/config/configFlags.py
@@ -3,15 +3,18 @@
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
 
-"""Flags used for the configuration.
-"""
+"""Flags used for the configuration."""
+
 from enum import unique
 from utils.displayString import DisplayStringIntEnum
 
 
 @unique
 class ReportTableHeaders(DisplayStringIntEnum):
-	"""The possible config values to report table headers.
+	"""Enumeration containing the possible config values to report table headers.
+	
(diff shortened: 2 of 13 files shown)


source/config/configFlags.py
```diff
@@ -0,0 +1,37 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2022 NV Access Limited, Cyrille Bougot
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+
+"""Flags used for the configuration.
+"""
```
lukaszgo1, line 7: [Python's conventions for docstrings](https://peps.python.org/pep-0257/) recommends to place closing triple quotes on the same line as the docstring if it is only a single line long.

source/config/configFlags.py
```diff
@@ -0,0 +1,37 @@
...
+	
+	OFF = 0
+	ROWS_AND_COLUMNS = 1
+	ROWS = 2
+	COLUMNS = 3
+	
+	@property
+	def _displayStringLabels(self):
+		return {
+			# Translators: This is the label for a combobox in the
```
lukaszgo1, line 25: The translator comment for these options seems wrong - these are not labels for the combo box.

source/config/profileUpgradeSteps.py
```diff
@@ -118,3 +118,19 @@ def upgradeConfigFrom_6_to_7(profile: Dict[str, str]) -> None:
...
+	In Document formatting settings, "Row/column headers" check box has been replaced with "Headers" combobox.
+	"""
+	try:
+		reportTableHeaders = profile['documentFormatting']['reportTableHeaders']
+	except KeyError:
+		# Setting does not exist, no need for upgrade of this setting
+		log.debug("reportTableHeaders not present, no action taken.")
+	else:
+		if reportTableHeaders:
+			profile['documentFormatting']['reportTableHeaders'] = 1  # Rows and columns
```
lukaszgo1, line 134: Can't you import the enum with possible values here and use `member.value` rather than an integer literal and a comment explaining what the value means?

source/globalCommands.py
```diff
@@ -667,19 +667,17 @@ def script_toggleReportTables(self,gesture):
...
-		if config.conf["documentFormatting"]["reportTableHeaders"]:
-			# Translators: The message announced when toggling the report table row/column headers document formatting setting.
-			state = _("report table row and column headers off")
-			config.conf["documentFormatting"]["reportTableHeaders"]=False
-		else:
-			# Translators: The message announced when toggling the report table row/column headers document formatting setting.
-			state = _("report table row and column headers on")
-			config.conf["documentFormatting"]["reportTableHeaders"]=True
-		ui.message(state)
+		from config.configFlags import ReportTableHeaders
```
lukaszgo1, line 674: Why is this imported lazily?

CyrilleB79: It's imported lazily as done in more recent scripts of `globalCommands.py` such as `script_toggleScreenCurtain`. It seems relevant to avoid to have too much import statements at the top of this file when the imported variable will not be used elsewhere.

Let's wait for the opinion of NVAccess reviewer. I may put back the import at the top of the file if people prefer this way to do.

source/config/configFlags.py
CyrilleB79: Done in 450cfec36e9fca0c58713c7ce35ac32565296beb.

source/config/profileUpgradeSteps.py
CyrilleB79: Done in 450cfec36e9fca0c58713c7ce35ac32565296beb.

seanbudd:
I think a more intuitive UX for the option would be checkboxes:

- [ ] Report table headers (checking this checks both, if one option is checked it is indeterminate)
  - [ ] Rows
  - [ ] Columns
  
What do you think?

seanbudd:
Sorry disregard my previous comment, missed the suggestion in the issue

source/speech/speech.py
```diff
@@ -611,12 +612,7 @@ def speakObject(
 		speak(sequence, priority=priority)
 
 
-
-
-# C901 'getObjectSpeech' is too complex
-# Note: when working on getObjectSpeech, look for opportunities to simplify
-# and move logic out into smaller helper functions.
-def getObjectSpeech(  # noqa: C901
```
seanbudd, line 619: why was this removed?

seanbudd: in fact, where possible, please factor out the new changes into helper function(s)

seanbudd (commented):
Thanks @CyrilleB79, generally looks good.

source/speech/speech.py
CyrilleB79: @seanbudd wrote:
> why was this removed?

As written in initial description
> Not directly linked but a a noqa exception for code complexity has been removed because it was not useful anymore.

To be more specific, I have not modified `getObjectSpeech` in this PR but had to reread it while developing. I have figured that `getObjectSpeech` is not complex anymore and thus have removed this `#noqa` directive that was already useless before this PR.

@seanbudd wrote:
> in fact, where possible, please factor out the new changes into helper function(s)

Where do you suggest me to factor code?

seanbudd: I was hoping to confirm that the function is under the limits.

> >in fact, where possible, please factor out the new changes into helper function(s)
> 
> Where do you suggest me to factor code?

apologies, I thought the below changes were in the function

seanbudd (dismissed):

source/speech/speech.py
CyrilleB79: I have updated initial description to clarify this point.

source/globalCommands.py
CyrilleB79: @seanbudd what is your opinion on this lazy import?

seanbudd: Since config is already imported at the top level, this should probably be moved out.
I believe the screen curtain case (and others) is more about avoid circular dependencies.

Qchristensen (approved):
User Guide change looks fine, good work.

leonardder:
Please don't merge yet as I have some suggestions I will post later today.

source/NVDAObjects/IAccessible/winword.py
```diff
@@ -205,7 +206,7 @@ def fetchAssociatedHeaderCellText(self,cell,columnHeader=False):
 
 	def script_setColumnHeader(self,gesture):
 		scriptCount=scriptHandler.getLastScriptRepeatCount()
-		if not config.conf['documentFormatting']['reportTableHeaders']:
+		if config.conf['documentFormatting']['reportTableHeaders'] == ReportTableHeaders.OFF:
```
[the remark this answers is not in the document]
CyrilleB79: Actually, setting or removing column headers while only row headers are reported is working. If you remove this "if" statement, i.e. allow this script to work regardless of this reportTableHeaders configuration, it works too.

For simplicity (regarding the error message), I am thinking to remove this if statement, i.e. setting column headers even when reporting table headers is disabled. @leonardder, @seanbudd, Can you see any drawback to this solution?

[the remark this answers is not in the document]
CyrilleB79: I have finally allowed setting/removing row/column headers even if reporting table headers is disabled. I have updated the initial description accordingly.
@seanbudd, @leonardder, let me know if this is OK for you.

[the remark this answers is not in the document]
leonardder: Great!

LeonarddeR (approved):

seanbudd (approved):
LGTM, thanks @CyrilleB79

CyrilleB79:
@seanbudd, I just realize that this PR was an API breaking change, due to the configuration modification and upgrade, but it has been merged in 2022.4 which was not intended to be an API breaking release.

I don't know of an add-on that uses this config keys.

This has no sense to revert it now. However, should this API breaking change be advertised somewhere? I would say at least on the API mailing list. And maybe in the change log, even if usually the log of older releases is not amended.

josephsl:
Hi, StationPlaylist add-on uses this config key, but as of now the add-on code will work on both 2022.3 and 2022.4 as integer values above 0 are seen as “true” in Python. I do plan to edit the add-on code son to use 2022.4 changes anyway. Thanks.

paulber19:
Hi,
I would like to know how an API break is defined.
It seems to me that any modification of NVDA's code is an API break for 
add-ons that rely on it.
For example, documentBase.py has undergone a major overhaul in version 
2022.2 compared to 2022.1 to add in particular the commands for moving 
tables in tables.
And that involved a serious upgrade in one of my add-ons.
  So it seems to me that for developers of add-ons embedded in NVDA, any 
new version of NVDA is an API break and it is necessary to 
systematically check that the NVDA code on which it was based did not 
not undergone any modification.
That's why I don't quite understand the principle of API breaking only 
when releasing the first version of the year.
Best regards.

josephsl:
Hi, it didn’t affect StationPlaylist ad-on, so all is good with me. Thanks.

paulber19:
Really sorry for intervening in this thread which was about API breaking.
All is well, forget my intervention.
Best regards.
Paul.

seanbudd:
Hi Cyrille,
I plan to announce this on the API mailing list.

Is this an accurate summary of the API breakage?

```
The configuration specification has been altered in the [documentFormatting] section.
The value of reportTableHeaders is now a member of configFlags.ReportTableHeaders, instead of a boolean.  (#14132)
 ```

PR #14222, opened 2022-10-08
Make support for search suggestions more generic, supporting Excel comment mentions and Google Chrome location bar suggestions

### LINKED ISSUE #13522: Chrome: Omnibox: User is not notified of useful information in the first item of keyword search
### Steps to reproduce:
1. Turn on NVDA and Chrome Browser
2. Activate keyword search by typing a letter that corresponds to a saved keyword search in the Omnibox

### Actual behavior:
As user types, they don’t hear any useful/friendly information that the sighted user sees in item 0, such as the title of the page they will get, or the fact that there are additional buttons

### Expected behavior:
The user should be notified of the title of the page and of the buttons one can tab to and activate as part of suggestions

Yes

2021.3.3. Same behavior

N/A

N/A

### LINKED ISSUE #13764: [Excel- Add @mention]: Screen Reader Users are not Notified when Search Suggestions Appear after Typing "@" in the comment
Untitled

### Steps to reproduce:
Step 1. Turn on NVDA.
Step 2. Open the Excel document.
Step 3. Navigate to 'Comments' by using 'Ctrl+F6', 'Tab' & 'Right/Left' arrow keys.
Step 4: Press Enter key to activate the Comment button.
Step 5: Navigate to any Comment & activate it.
Step 6: Navigate to '@mention or reply,' edit field, start typing "@" and observe the issue.

### Actual behavior:
NVDA is not announcing search suggestions are available after typing some characters in '@mention or reply,' edit field and while navigating on search results NVDA just announcing as "N" and it's not announcing the username and user mail ID.

### Expected behavior:
For better screen reader user experience NVDA should announce the Serach results information, so that user can select the desired person and send the comment easily.
For example, NVDA should announces "10 Search suggestions available" and while navigating those results NVDA should announce the user information like name of the person and his mail ID as well.

Test Environment:
OS Version: Dev (OS Build 22458.1000)
Application Version: 2111 (Build 14606.20000) (Dogfood) (InnerRing)
Application: Excel
Screen Reader: NVDA

#### NVDA version:
2021.3.5

#### Windows version:
OS Version: Dev (OS Build 22458.1000)

### LINKED ISSUE #13772: Office backstage Search: NVDA does not Read Search Suggestions when Arrow Up and Down the List of Document Suggestions.
NVDA does not announce focused search suggestion list item

### Steps to reproduce:
1. Turn on NVDA.
2. Open Word Application and open blank file.
3. Navigate to Ribbon using f6 key.
4. Navigate to File tab using arrow key and activate it.
5. Backstage will be appeared.
6. Navigate to Search Edit in Home using tab key and enter some text in search edit field.
7. Suggestion will be appeared and navigate using up/down arrow key.
8. Observe the issue with NVDA.

### Actual behavior:
NVDA does not read search suggestions when arrow up and down the list of document suggestions. Screen reader user will be impacted, they will be confused and think there is no suggestion list and screen reader focus are stuck in search edit field if NVDA does not read search suggestions while navigating suggestions.

### Expected behavior:
Screen reader should announce the focused search suggestion list item when user navigate the list item using up/down arrow key.

### System configuration
Windows 11 Enterprise

Installed

#### NVDA version:
Version 2022.1

#### Windows version:
Windows 11 
OS build: Version 21H2 Build 22000.675

Word Version: 2206 Build 15330.20020

Yes

Version 2020.4. issue is also reproducible with this NVDA version.

michaelDCurran:
### Link to issue number:
Fixes #13772
Fixes #13764
Fixes #13522

### Summary of the issue:
* When starting to type an @mention in a Microsoft Excel comment, NVDA does not report available suggestions when arrowing through them. (#13764)
* When typing in the Google Chrome location bar, NVDA does not report rich information for each suggestion, nor does it report when a suggestions control (switch to tb, remove suggestion) is selected. (#13522)
* As a developer, adding support for new suggestion scenarios is tedious as specific mappings must be always made.

* NVDA will now report the selected suggestion when arrowing through suggestions when typing an @mention in a Microsoft Excel comment.
* NVDA will report the selected suggestion or suggestion control (switch to tab, remove suggestion) in the Google Chrome location bar.

Make support for suggestions on editable text fields more generic, allowing it to be supported in more places without having to specifically choose particular overlay classes.

This adds support for reporting @mention suggestions in Microsoft Excel comments, and reporting of suggestions and related controls in the Google Chrome location bar.

Specific changes:
* IAccessible NVDAObject: implement the controllerFor property making use of IAccessible2 relations
* NVDAObject: add an isDescendantOf method which should return True if the object this method is called on is a descendant of the given obj argument. The base method raises NotImplementedError. Efficient implementations for UIA and Ia2Web have been provided.
* Implement event_selection on the base NVDAObject, which reports the object being selected, if the current focus is controlling an ancestor of the selected object. This replaces SuggestionListItem UIA NVDAObject's event_UIA_elementSelected logic. Now it is more generic, and works purely on tree topology, rather than checking the name and appModule etc.
* Remove event_selection implementation from IAccessible NvDAObject as the base handles this now.
* Add an event_controllerForchange method to the base NVDAObject which does nothing.
* UIA NVDAObject: implement event_UIA_controllerforchange to directly call event_controllerforchange, allowing for more generic code, but still backwards compatibility for subclasses to still override event_UIA_controllerforchange if they need to.
* behaviours:
	* Rename EditableTextWithSuggestions to InputFieldWithSuggestions and inherit directly from NVDAObject making this class more generic.
	* InputFieldWithSuggestions class: implement event_controllerForChange which will fire the suggestionsOpened and suggestionsClosed events based on whether there are objects in the controllerfor list. this code was taken from SearchField UIA NVDAObject's event_UIA_controllerforchange.
	* Rename EditableText to EditabletextBase
	* Add a replacement EditableTextWithSuggestions class that inherits from both InputFieldWithSuggestions and EditableTextBase
	* Add a replacement EditableText class which inherits from EditableTextwithSuggestions and EditableTextBase.
	Essentially all this builds suggestion support into EditableText, but keeps the EditableTextWithSuggestionsClass specifically available for compatibility.

### Testing strategy:
* Followed steps in #13764
* Followed steps in #13522
* Confirmed that  suggestions still report in the Windows Start screen when typing content and arrowing down through the suggestions.
* Confirmed that suggestions still report in the Windows Settings screen, when typing content into the Settings search and arrowing down through the suggestions.
* Further testing would be welcome by @josephsl please.

* Suggestion open and suggestion close sounds are not played for the Google Chrome location bar suggestions as IAccessible2 does not have a controllerForchange event. This feature would be nice, but probably not expected by users anyway.
* UIA_invalidatedLayout event has not been abstracted, thus auto reporting of suggestion count is still only available in specific UIA situations such as Settings search. this could be addressed separately at a later date if possible.

### Change log entries:
New features
Changes
Bug fixes
* Suggestions are reported when typing an @mention in in Microsoft Excel comments. (#13764)
* In the Google Chrome location bar, suggestion controls (switch to tab, remove suggestion etc) are now reported when selected. (#13522)
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


josephsl:
Hi,

Preliminary test results: works well in People app where contacts are announced automatically (test condition: Windows App Essentials add-on is disabled). I'll take a look at code this weekend.

Thanks.

josephsl (approved):
Should we mark empty classes as deprecated and targeted for removal in 2024.1? Also, I think it woudl be best to talk about NVDAObject.isDescendantOf and the reorganized class hierarchy in behaviors module somewhere in "changes for developers" section. Thanks.

josephsl:
By the way, for @JCSTeh, do you think if the approach taken in this PR can be adopted for use in Firefox and other Mozilla products? I imagine it will help when reviewing suggestions in Firefox address bar.

CC @DerekRiemer in case you would like to employ this change in Notepad++ enhancements.

Thanks.

jcsteh:
Firefox already exposes the controllerFor relation on the address bar. That should mean this already works, but I"m guessing it doesn't based on your comment and the fact that Firefox wasn't mentioned in the PR description.

Perhaps Firefox doesn't fire the selection event on the first item; I'd need to check. If that's true, the reason would be that the list gets shown with the item already selected, so we don't fire the selection event because it would be redundant given that it was always selected since the list came into existence. I can see why that's a problem for NVDA, but I'm also not really sure how to solve it without special casing or causing a bunch of redundant events elsewhere.

That said, per the ARIA spec for autocompletes, Firefox fires focus on the selected item when you use the up and down arrow keys, which would override this new code for items other than the first. I'd love to see this changed, but it can't be changed without breaking backwards compatibility all over the place. I've had an idea for a while to hack around this in NVDA, ignoring those focus events and instead keeping focus on the text box, but that's a much bigger and riskier change.

> • Suggestion open and suggestion close sounds are not played for the Google Chrome location bar suggestions as IAccessible2 does not have a controllerForchange event.

It's worth noting that Firefox exposes the collapsed/expanded state on the address bar depending on whether the list is showing and fires a stateChange event when that changes. That seems like a reasonable proxy for the lack of a native controllerForChange event. Chrome doesn't expose the expanded/collapsed state, but it seems like something they might be willing to do.

> This feature would be nice, but probably not expected by users anyway.

I'm curious as to why users wouldn't expect this. Is this simply because of the previous state of things or is there some broader reason here? It seems to me that the lack of consistency would be unexpected, particularly for new users.

source/NVDAObjects/IAccessible/__init__.py
```diff
@@ -1655,9 +1661,6 @@ def _get_groupName(self):
 		else:
 			return super(IAccessible,self)._get_groupName()
 
-	def event_selection(self):
```
seanbudd, line 1658: does removing this break add-on compatibility?

source/NVDAObjects/IAccessible/ia2Web.py
```diff
@@ -26,6 +26,20 @@ class Ia2Web(IAccessible):
 	IAccessibleTableUsesTableCellIndexAttrib=True
 	caretMovementDetectionUsesEvents = False
 
+	def isDescendantOf(self, obj):
```
seanbudd, line 29: add type annotations

```suggestion
	def isDescendantOf(self, obj: "NVDAObjects.NVDAObject") -> bool:
```

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2006,6 +2007,27 @@ def _get_positionInfo(self):
 	def scrollIntoView(self):
 		pass
 
+	def isDescendantOf(self, obj):
```
seanbudd, line 2010: add type annotations

```suggestion
	def isDescendantOf(self, obj: "NVDAObjects.NVDAObject") -> bool:
```

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2006,6 +2007,27 @@ def _get_positionInfo(self):
 	def scrollIntoView(self):
 		pass
 
+	def isDescendantOf(self, obj):
+		if isinstance(obj, UIA):
+			# As both objects are UIA,
+			# We can search this object's  ancestors for obj with a UIA treeWalker
```
seanbudd, line 2013: remove extra space

```suggestion
			# We can search this object's ancestors for obj with a UIA treeWalker
```

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2265,15 +2290,9 @@ def event_stateChange(self):
 
 class SearchField(EditableTextWithSuggestions, UIA):
 	"""An edit field that presents suggestions based on a search term.
+	This is now an empty class as Functionality has been moved to the base EditableText behaviour.
 	"""
 
-	def event_UIA_controllerFor(self):
```
seanbudd, line 2270: does removing this function break add-on compatibility?

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2265,15 +2290,9 @@ def event_stateChange(self):
 
 class SearchField(EditableTextWithSuggestions, UIA):
 	"""An edit field that presents suggestions based on a search term.
+	This is now an empty class as Functionality has been moved to the base EditableText behaviour.
```
seanbudd, line 2293: Fix mis-capitalization in functionality

```suggestion
	This is now an empty class as functionality has been moved to the base EditableText behaviour.
```

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2302,20 +2321,11 @@ def event_UIA_layoutInvalidated(self):
 class SuggestionListItem(UIA):
 	"""Recent Windows releases use suggestions lists for various things, including Start menu suggestions, Store, Settings app and so on.
 	Unlike suggestions list class, top suggestion is automatically selected.
+	Note that support for reporting the selection is now handled generically on the base NVDAObject.
 	"""
 
 	role = controlTypes.Role.LISTITEM
 
-	def event_UIA_elementSelected(self):
```
seanbudd, line 2309: same question here re backwards compatibility

source/NVDAObjects/__init__.py
```diff
@@ -1177,6 +1181,29 @@ def event_mouseMove(self,x,y):
 					speech.cancelSpeech()
 				speech.speakText(text)
 
+	def event_selection(self):
+		# This object has been selected.
+		# If this object's container / parent is being controlled by the focus,
+		# Then report this selection.
```
seanbudd, line 1187: fix capitalization

```suggestion
		# then report this selection.
```

source/NVDAObjects/__init__.py
```diff
@@ -1177,6 +1181,29 @@ def event_mouseMove(self,x,y):
...
 
+	def event_selection(self):
+		# This object has been selected.
+		# If this object's container / parent is being controlled by the focus,
+		# Then report this selection.
+		focus = api.getFocusObject()
+		controls = focus.controllerFor
+		for control in controls:
+			# The focus is controling one or more objects.
+			# If possible, Check if this object is a decendant of the object being controlled.
```
seanbudd, line 1192: Fix typos

```suggestion
			# If possible, check if this object is a descendant of the object being controlled.
```

source/NVDAObjects/behaviors.py
```diff
@@ -163,7 +163,42 @@ def _get_isPresentableFocusAncestor(self):
 		self.isPresentableFocusAncestor = res = super(Dialog, self).isPresentableFocusAncestor
 		return res
 
-class EditableText(editableText.EditableText, NVDAObject):
+
+class InputFieldWithSuggestions(NVDAObject):
+	"""Allows NvDA to announce appearance/disappearance of suggestions as content is entered.
```
seanbudd, line 168: Fix typo

```suggestion
	"""Allows NVDA to announce appearance/disappearance of suggestions as content is entered.
```

source/NVDAObjects/behaviors.py
```diff
@@ -873,30 +922,6 @@ def event_alert(self):
...
 	event_show = event_alert
 
-class EditableTextWithSuggestions(NVDAObject):
-	"""Allows NvDA to announce appearance/disappearance of suggestions as text is entered.
-	This is used in various places, including Windows 10 search edit fields and others.
-	Subclasses should provide L{event_suggestionsOpened} and can optionally override L{event_suggestionsClosed}.
-	These events are fired when suggestions appear and disappear, respectively.
-	"""
-
-	def event_suggestionsOpened(self):
```
seanbudd, line 883: same questions here re backwards compatibility

seanbudd:
@josephsl 
>Should we mark empty classes as deprecated and targeted for removal in 2024.1? 

The current policy is that if aliases for deprecated objects can remain without a major maintenance cost, then they shouldn't be removed.

seanbudd (commented):
general approach looks good, only minor fixes/questions raised

michaelDCurran:
@jcsteh just the previous state of things. I think I'm just trying to 
say that I wouldn't hold back this pr simply because we can't make the 
sounds play, that's all.

source/NVDAObjects/IAccessible/__init__.py
michaelDCurran: No, as event_selection is still available and still does the same thing, I.e. calls event_stateChange, it is just that it has been moved to a base class.

josephsl:
Hi,

Update: this PR also resolves #13772. Thanks.

source/NVDAObjects/UIA/__init__.py
michaelDCurran: No, as it has been moved to the base.

source/NVDAObjects/UIA/__init__.py
michaelDCurran: Again moved to the base.

source/NVDAObjects/behaviors.py
michaelDCurran: These were moved to InputFieldWithSuggestions, which now EditableTextWithSuggestions inherits from.

michaelDCurran:
@seanbudd  have I missed any review actions from you here?

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2268,15 +2294,10 @@ def event_stateChange(self):
 
 class SearchField(EditableTextWithSuggestions, UIA):
 	"""An edit field that presents suggestions based on a search term.
+	This is now an empty class as functionality has been moved to the base EditableText behaviour.
+	@warning Deprecated, may be removed in future versions.
```
leonardder, line 2298: Couldn't we just remove this as we're at an API breaking release?

source/NVDAObjects/behaviors.py
```diff
@@ -236,6 +271,20 @@ def event_typedCharacter(self, ch: str):
...
+class EditableTextWithSuggestions(InputFieldWithSuggestions, EditableTextBase):
+	""" Represents an editable text field that shows suggestions as you type.
+	This is an empty class as functionality has been moved to the base InputFieldWithSuggestions class.
+	"""
+
+
+class EditableText(EditableTextWithSuggestions, EditableTextBase):
+	""" Represents an editable text field.
+	This is an empty class as functionality has been moved to the base EditableTextBase class.
+	This class also supports reporting of the appearance and disappearence of suggestions
```
leonardder, line 283:
```suggestion
	This class also supports reporting of the appearance and disappearance of suggestions
```

source/NVDAObjects/UIA/__init__.py
michaelDCurran: The policy now days is to only remove something if its continued existence actually conflicts with or stops a newer improved implementation. Keeping this SearchField class is not harmful.

seanbudd (commented):
frustratingly I noticed my pending review never published

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2023,8 +2046,11 @@ def _get_controllerFor(self):
 				objList.append(obj)
 		return objList
 
+	def event_UIA_controllerFor(self):
```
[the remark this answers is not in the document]
michaelDCurran: It is None.
```suggestion
	def event_UIA_controllerFor(self) -> None:
```

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2293,7 +2294,8 @@ def event_stateChange(self):
 
 class SearchField(EditableTextWithSuggestions, UIA):
 	"""An edit field that presents suggestions based on a search term.
-	This is now an empty class as Functionality has been moved to the base EditableText behaviour.
+	This is now an empty class as functionality has been moved to the base EditableText behaviour.
+	@warning Deprecated, may be removed in future versions.
```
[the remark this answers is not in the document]
michaelDCurran: So since it is not going to be removed, then I guess we simply don't mention deprecation at all?
```suggestion
```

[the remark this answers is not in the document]
seanbudd: Yeah, if we want to deprecate the interface, you can use a code snippet to warn users that the function is deprecated. The deprecation warning does not require a plan for removal.

seanbudd (approved):

PR #14260, opened 2022-10-17
Open chrome prior to system tests

feerrenrut:
### Link to issue number:
Splitting up PR https://github.com/nvaccess/nvda/pull/14054

### Summary of the issue:
The first system test to run that requires chrome occasionally fails.
This has been replicated on developer machines after chrome updates, however, it is difficult to reproduce, and difficult to detect/recover-from in an automated way.

None 

Theory for cause:
Chrome is busy with post install tasks, so start chrome in the background ahead of the tests.

It may be possible to handle this better within NVDA, however it seems to actually affect users infrequently, and they seem to be able to recover by waiting a few seconds and de/refocus chrome.
In the meantime, the system tests failing delays development.

### Testing strategy:
This will require many builds on appveyor to confirm an improvement.

This is intentionally working around an issue that may affect users.

### Change log entries:
None

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


appveyor/scripts/tests/beforeTests.ps1
```diff
@@ -2,4 +2,14 @@ New-Item -ItemType directory -Path testOutput
...
+# The first system test to run that requires chrome occasionally fails.
+# This has been replicated on developer machines after chrome updates,
+# however it is difficult to reproduce, and difficult to detect/recover-from
+# in an automated way.
+# It may be possible to handle this better within NVDA, however it seems to actually affect users
+# infrequently, and they seem to be able to recover by waiting a few seconds and de/refocus chrome.
+# In the mean time, the system tests failing delays development.
+#
+# Theory: Chrome is busy with post install tasks, so start chrome in the background ahead of the tests.
+cmd /c start /min chrome.exe
```
CyrilleB79, line 14: In `ChromeLib.py`, Chrome is opened with the following parameters:
* `--force-renderer-accessibility`
*  `--suppress-message-center-popups`
* `--disable-notifications`
* `--no-experiments`
* `--no-default-browser-check`
* `--lang=en-US`

Regarding `--lang=en-US`, I remember that this parameter only has effect when Chrome is first opened, not if there are other Chrome windows already opened. This parameter is not important on appVeyor since Chrome is already in English on appVeyor.

However, if the other parameters behave the same way, the risk is that they will not be taken into account during future tests since Chromme will already be pre-opened with no parameters before.

Please double-check and pass the same parameters as in `ChromeLib.py` if needed.

feerrenrut: Yep, this seems like a good idea.

appveyor/scripts/tests/beforeTests.ps1
```diff
@@ -2,4 +2,19 @@ New-Item -ItemType directory -Path testOutput
 New-Item -ItemType directory -Path testOutput\unit
 New-Item -ItemType directory -Path testOutput\system
 New-Item -ItemType directory -Path testOutput\lint
+# The first system test to run that requires chrome occasionally fails.
+# This has been replicated on developer machines after chrome updates,
+# however it is difficult to reproduce, and difficult to detect/recover-from
+# in an automated way.
+# It may be possible to handle this better within NVDA, however it seems to actually affect users
```
CyrilleB79, line 9: I do not understand this specific sentence. Any clarification?

tests/system/libraries/_chromeArgs.py
```diff
@@ -0,0 +1,28 @@
...
+	File path can be appended to this string.
+	"""
+	return (
+		"chrome"  # Start Chrome
+		" --force-renderer-accessibility"  # don't rely on chrome detecting a screen reader.
+		" --suppress-message-center-popups"  # prevent popups that may interfere with automated tests.
+		" --disable-notifications"  # prevent notifications that may interfere with automated tests.
+		" --no-experiments"  # Stable behavior is preferred.
+		" --no-default-browser-check"  # Don't bother to check if Chrome is the default browser.
+		" --lang=en-US"
```
CyrilleB79, line 22: While at it, add a description that I hadn't added when I have added this parameter:

```suggestion
		" --lang=en-US"  # GUI in English to ensure tests pass on non-English systems
```

tests/system/libraries/_chromeArgs.py
```diff
@@ -0,0 +1,28 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2020-2022 NV Access Limited
```
CyrilleB79, line 2: A detail: Remove year 2020 for a file that has just been created.

```suggestion
# Copyright (C) 2022 NV Access Limited
```

feerrenrut: I'm not sure about this. It's the code that is covered by copyright, and the code existed through that whole time.

appveyor/scripts/tests/beforeTests.ps1
feerrenrut: It's saying that the bug is legitimate, it has been replicated on developer environments, it is not only theoretical or a quirk of running on Appveyor. Despite it being a real issue, the problem occurs very rarely and working around the issue is very easy, the impact on users is minimal. We won't attempt to fix this given the small impact on users, and high time cost to investigate and fix. However, the issue does affect the system tests, occasionally causing the first chrome test to fail. This is a problem for developers, they have to wait for a new build. This change (starting chrome early) is an attempt to work around the problem, purely to prevent delays for developers.

feerrenrut: Clarification:
> Despite it being a real issue, the problem occurs very rarely and working around the issue is very easy

Very easy for end users. When I have seen this occur, pressing `alt+tab` to bring a different app to the foreground, then return to chrome resolves the issue.

feerrenrut:
Dismissing the test failure:
```
Robot.chromeTests.ARIA treegridRobot4 sec 965 ms
Error message:
Unable to locate 'before sample' marker. See NVDA log for full speech.
```
This is a symptom of another intermittent problem, an attempt to resolve is coming another PR.

appveyor/scripts/tests/beforeTests.ps1
CyrilleB79: This description is mixing two issues, hence my confusion:
1. Chrome fails to start normally
2. First Chrome test fails (due to 1.)

The normal user experience with Chrome and the behaviour of Chrome during system tests are also mixed in this paragraph and sometimes, it's unclear what apply to users and to system tests.

Maybe you could reword the paragraph like this (to be improved and described more precisely):
The first system test to run that requires chrome occasionally fails due to Chrome abnormal startup [please describe more precisely].
Chrome sometimes does not starts normally, preventing users to use it normally. This has been replicated on user machines after chrome updates,
however it is difficult to reproduce, and difficult to detect/recover-from
in an automated way in system tests.
It may be possible to work around this abnormal Chrome startup better within NVDA, however it seems to actually affect users
infrequently, and they seem to be able to recover by waiting a few seconds and de/refocus chrome.
In the mean time, the system tests failing delays development.


Explaining somewhere what happens

feerrenrut: >occasionally fails due to Chrome abnormal startup

No, we don't know that for sure. All we know:
- Occasionally the first chrome system test on Appveyor fails. NVDA does not load a virtual buffer.
- Very rarely this symptom has been observed by developers on their own machines.

seanbudd (approved):

appveyor/scripts/tests/beforeTests.ps1
feerrenrut: @CyrilleB79 I've updated this explanation.

appveyor/scripts/tests/beforeTests.ps1
```diff
@@ -2,4 +2,22 @@ New-Item -ItemType directory -Path testOutput
 New-Item -ItemType directory -Path testOutput\unit
 New-Item -ItemType directory -Path testOutput\system
 New-Item -ItemType directory -Path testOutput\lint
+
+# The first Chrome system test to run that occasionally fails.
```
CyrilleB79, line 6: Typo:
```suggestion
# The first Chrome system test to run occasionally fails.
```

CyrilleB79 (approved):
The explanation looks clearer now. Thanks.
Just a typo to fix.

PR #8596, opened 2018-08-06
Logging: add an option to turn logging off altogether

josephsl:
### Link to issue number:
Fixes #8516

### Summary of the issue:
Add an option to turn off logging altogether.

### Description of how this pull request fixes the issue:
Add a new option in General settings panel to turn off logging. If this option is chosen, only a startup message (along with config path) will be logged, and log level will be set to 100.

### Testing performed:
Tested on source code with logging level set to off and other values.

### Known issues with pull request:
Even with logging level set to off, startup info messages will be shown up to config path. Ideally this too should not be logged, but for now, it is acceptable.

### Change log entry:
New features: It is possible to turn off NVDA's logging features off by setting logging level to "disabled" from General settings panel. (#8516)

Thanks.

PR #14406, opened 2022-11-26
Use only supported log levels for command line option

CyrilleB79:
Context: issue found while reading the code for #14398.

### Link to issue number:
None
### Summary of the issue:

When starting NVDA with an unsupported command line parameter, you get the following suggestion for log level in the error message:
> [-l {10,12,15,20,30,40,50,100}]

Launching NVDA with log level set to 30, 40 or 50 actually indicates NVDA using info level.

* Only supported log levels are indicated in the error message when a wrong parameter is passed to NVDA.
* Only supported log level can now be used from command line, i.e. 10=debug, 12=I/O, 15=debugWarning, 20=info and 100=off

* Changed the information passed to the command line parser.
* While at it, changed default from "en" to "Windows" in the parser's `--lang` parameter; this seem to have no impact however since "Windows" was already used when this parameter is missing.
### Testing strategy:
Manual tests:
* NVDA starts normally when `-l` parameter is not passed to NVDA
* NVDA starts normally with `-l` command line param set to 10, 12, 15, 20 and 100 and the appropriate log level is used
* NVDA does not start and displays the usage message when launched with `-l` set to 30, 40, 50 and 99.

Note: also tested `--lang=windows`; this fails but it was already failing before this PR, e.g. in NVDA 2022.4beta3. I will open an issue for that.

None
### Change log entries:
Probably not needed.
Or maybe something in the bugfixes category? What do you think?

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/nvda.pyw
+++ b/source/nvda.pyw
@@ -1,6 +1,7 @@
 # -*- coding: UTF-8 -*-
 # A part of NonVisual Desktop Access (NVDA)
-# Copyright (C) 2006-2022 NV Access Limited, Aleksey Sadovoy, Babbage B.V., Joseph Lee, Łukasz Golonka
+# Copyright (C) 2006-2022 NV Access Limited, Aleksey Sadovoy, Babbage B.V., Joseph Lee, Łukasz Golonka,
+# Cyrille Bougot
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
 
@@ -148,12 +149,23 @@ quitGroup = parser.add_mutually_exclusive_group()
 quitGroup.add_argument('-q','--quit',action="store_true",dest='quit',default=False,help="Quit already running copy of NVDA")
 parser.add_argument('-k','--check-running',action="store_true",dest='check_running',default=False,help="Report whether NVDA is running via the exit code; 0 if running, 1 if not running")
 parser.add_argument('-f','--log-file',dest='logFileName',type=str,help="The file where log messages should be written to")
-parser.add_argument('-l','--log-level',dest='logLevel',type=int,default=0,choices=[10, 12, 15, 20, 30, 40, 50, 100],help="The lowest level of message logged (debug 10, input/output 12, debugwarning 15, info 20, warning 30, error 40, critical 50, off 100), default is info")
+parser.add_argument(
+	'-l',
+	'--log-level',
+	dest='logLevel',
+	type=int,
+	default=20,
+	choices=[10, 12, 15, 20, 100],
+	help=(
+		"The lowest level of message logged (debug 10, input/output 12, debugwarning 15, info 20, off 100),"
+		" default is 20 (info)"
+	),
+)
 parser.add_argument('-c','--config-path',dest='configPath',default=None,type=str,help="The path where all settings for NVDA are stored")
 parser.add_argument(
 	'--lang',
 	dest='language',
-	default="en",
+	default="Windows",
 	type=stringToLang,
 	help=(
 		"Override the configured NVDA language."


seanbudd (approved):

leonardder:
I'm somehow missing the logging category from NVDA"s general settings. Could this pr be related?

CyrilleB79:
> I'm somehow missing the logging category from NVDA"s general settings. Could this pr be related?

Could you elaborate?

When you run nvda with the `--lang` parameter, the logging level item is disabled in General settings; but I guess you know it.

leonardder:
Sure. When launching NVDA from my desktop shortcut, logHandler.isLogLevelForced() returns True. shortcut is: `"C:\Program Files (x86)\NVDA\nvda_slave.exe" launchNVDA -r`. Even without -r, it returns True.
globalVars.appArgs.logLevel is set to 20 and therefore the following clause applies:
`or globalVars.appArgs.logLevel != 0`

source/nvda.pyw
```diff
@@ -148,12 +149,23 @@ quitGroup = parser.add_mutually_exclusive_group()
 quitGroup.add_argument('-q','--quit',action="store_true",dest='quit',default=False,help="Quit already running copy of NVDA")
 parser.add_argument('-k','--check-running',action="store_true",dest='check_running',default=False,help="Report whether NVDA is running via the exit code; 0 if running, 1 if not running")
 parser.add_argument('-f','--log-file',dest='logFileName',type=str,help="The file where log messages should be written to")
-parser.add_argument('-l','--log-level',dest='logLevel',type=int,default=0,choices=[10, 12, 15, 20, 30, 40, 50, 100],help="The lowest level of message logged (debug 10, input/output 12, debugwarning 15, info 20, warning 30, error 40, critical 50, off 100), default is info")
+parser.add_argument(
+	'-l',
+	'--log-level',
+	dest='logLevel',
+	type=int,
+	default=20,
```
leonardder, line 157: This line breaks it for me.

CyrilleB79: OK. I can see the issue, thanks. I will fix it.

CyrilleB79:
In addition, I just realize that in this PR, I have not updated the user guide command line option.

Moreover, I have now come across [this comment in ](https://github.com/nvaccess/nvda/pull/8596#issuecomment-411061826) and more generally the discussion in #8596. It seems that there were (for NVDA 2022.4beta3 and before) many log levels and only a subset of them were accessible from the GUI, as explained by @josephsl .

Thus we have two solutions:

### Solution 1: Fix-up this PR
Fix the issue with log level combo-box not usable anymore in the GUI: the default value in the parser is to be restored to 0.

We would end-up with the following command-line log levels being removed: 30 (= WARNING), 40 (= ERROR), 50 (= CRITICAL).

Since INFO is not so much verbose, are there really persons using the `-l` switch with values of 30, 40 or 50 today?

Note: DEBUGWARNING (12) remains available and is not to be confused with WARNING (30).

### Solution 2: Revert this PR
On the opposite, we may revert this PR.

On the contrary to what I have written in the description of this PR, levels 30, 40, 50 seem to work correctly and NVDA does not log INFO messages at these levels.

There would be some fixes to do in the future however:
* In the GUI, when `-l` is 30, 40 or 50, the combobox for log level is disabled and blank. Even if it is disabled, it should indicate the current log level passed as parameter.
* At level 30, 40 and 50, NVDA+F1 does not allow to open the log viewer; it should instead.
* Maybe indicate more explicitly in the User Guide that the levels 30 to 50 are command-line only
* Refactor the code to list all the log levels in logHandler and associate a boolean value or a function indicating if the level is accessible from the GUI or not.

### Conclusion

Before doing anything, I would like to hear NVAccess point of view (@seanbudd, @feerrenrut) as well as other members of the community (@josephsl, @leonardder)

michaelDCurran:
This pr is reverted by pr #14442 
A new pr based on this one, that also fixes the identified issue/s here, and also most likely incorporates pr #14432  would be welcome.

CyrilleB79:
@michaelDCurran it is unclear which solution from https://github.com/nvaccess/nvda/pull/14406#issuecomment-1338280470 NVAccess would like.

Solution 1 was fully implemented by #14406 + #14432. But you have reverted #14406 and closed #14432, which would let think that you prefer solution 2.

But on the other side you write:
> A new pr based on this one, that also fixes the identified issue/s here, and also most likely incorporates pr #14432 would be welcome.

That let me think that you agree with solution 1.

So, could you then react to https://github.com/nvaccess/nvda/pull/14406#issuecomment-1338280470 and indicate clearly what you are expecting? In other words, would you like WARNING, ERROR and CRITICAL to be available as NVDA log level or removed? And if you want to keep it, should these levels be available only with command line (as today) or also in the GUI?

Thanks.

michaelDCurran:
Firstly, reverting the pr should not be seen as a statement on which solution I liked. NV Access's policy is to revert a pr  if it directly causes an issue and a release containing the pr has not yet been made. What I wrote about making a new pr possibly incorporating #14406  and #14432  was more a general statement saying that there was a problem. We reverted. Please try again in a new pr. Sorry if I did not word this well.
Specifically considering the solutions you have provided, I would prefer solution 1, only list log levels from info down. We don't ever want users providing logs to us at level warning, error or critical as they are simply not useful as they drop too much information and therefore just waste time for the user and us.

PR #14489, opened 2022-12-29
Fix wts info class

LeonarddeR:
### Link to issue number:
None

### Summary of the issue:
The WTS_INFO_CLASS enum as defined in NVDA contained a WTSQuerySessionInformation value, which shouldn't be part of the actual enum according to [The docs](https://learn.microsoft.com/en-us/windows/win32/api/wtsapi32/ne-wtsapi32-wts_connectstate_class)
Also when calling WTSQuerySessionInformation, ctypes.pointer was used instead of ctypes.byref to pass values by reference.

None

Removed the entry and fixed the numbering. Use byref instead of pointer

### Testing strategy:
Tested something like this in a remote session:
`winAPI._wtsApi32.WTSQuerySessionInformation(winAPI._wtsApi32.WTS_CURRENT_SERVER_HANDLE, winAPI._wtsApi32.WTS_CURRENT_SESSION, 29, byref(ppBuffer), byref(pBytesReturned))`

None known

### Change log entries:
None needed

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/winAPI/_wtsApi32.py
+++ b/source/winAPI/_wtsApi32.py
@@ -89,9 +89,8 @@ class WTS_INFO_CLASS(IntEnum):
 	WTSConfigInfo = 26
 	WTSValidationInfo = 27  # Info Class value used to fetch Validation Information
 	# through the WTSQuerySessionInformation
-	WTSQuerySessionInformation = 28
-	WTSSessionAddressV4 = 29
-	WTSIsRemoteSession = 30
+	WTSSessionAddressV4 = 28
+	WTSIsRemoteSession = 29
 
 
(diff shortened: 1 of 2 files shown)


leonardder:
From the python docs:

> ctypes exports the byref function which is used to pass parameters by reference. The same effect can be achieved with the pointer function, although pointer does a lot more work since it constructs a real pointer object, so it is faster to use byref if you don't need the pointer object in Python itself:

seanbudd (approved):

PR #14543, opened 2023-01-15
Speech dictionaries: stop intra-dictionary and inter-dictionary cascading of replacements

### LINKED ISSUE #11823: Dictionary processing, as currently implemented, does not stop once the first match has been made and iterative processing produces wildly unexpected results
### Steps to reproduce:  This is not exactly applicable, but after discussions with @Qchristensen about dictionary processing, he produced this example:
Dictionary entries are processed in order, and do continue, but based on the current, not original, text.  So I just tried setting up the following entries:

change "dog" to "cat"
change "12345" to "54321"
change "432" to "dog"

Encountering the text 12345 now, the first rule is skipped, since there is no "dog" in this text.
The text includes 12345 so it becomes 54321.
The third rule is then evaluated and the 432 in the current text is changed to dog.

So NVDA reads: 

5dog1

If the text read "dog 12345", it would be read as:
cat 5dog1.

I'm not sure if that's documented anywhere, and given that rule order is meaningful, there is no current way to reorder rules (I wrote the 12345 to 54321 and 432 to dog rules, and then thought to try adding dog to cat first, but had to delete the other two and write all three from scratch in the order I wanted them).

### Expected behavior:

I would expect the outcome, with original input text of "dog 12345" to be NVDA speaking, "cat 54321," and if the original text is just 12345, the expected result would be NVDA speaking 54321.
Iterative processing where substituted text from a rule above gets passed through all following rules goes against any pattern matching processing of this sort I have dealt with in over 40 years.  When an initial match is made, the typical behavior is that match is acted upon and the matching stops, not that the substitution text for the first match is then passed along for processing by all further patterns.  The outcome of that, in many practical situations, is madness.  This is currently creating an issue when attempting to use regular expressions to have Roman Numerals read out as their component letters, as repeat substitution on substituted text for numerals IIII down to I, and VIII down to V, and IX all get parts of their substitution texts passed through additional patterns no matter how one tries to configure the pattern matching rule order.  Once the match is made, the substitution should be passed to the synthesizer and that's that.

XLTechie:
### Link to issue number:

Fixes #11823 

### Summary of the issue:

If an entry in a speech dictionary causes some text to be replaced, that replacement text was continuing to be processed, in effect as if it was new text, through the same and other dictionaries.
This could lead to all sorts of confusion and unexpected behavior for users, particularly the intra-dictionary reprocessing.

Examples can be found in the original issue, and an extensive discussion of the confusion and possible outcomes, along with arguments fore and against various fixes, can be found in [this mailing list discussion](https://nvda.groups.io/g/nvda/topic/96240115).

When an entry is found in a speech dictionary, a replacement will be made, and no further speech dictionary processing will be done for that entry or its replacement.

It will no longer be possible to replace text in one dictionary, and then have the replacement replaced by the same or another dictionary.

Inserted a conditional break statement in the processText function of speechDictHandler, causing it to leave the loop which continues looking for further matches.
To do this, and still return correctly, it was necessary to add a second variable to hold the result of replacements against which to compare.

I also added a comment and a docstring while I was at it.

This solution was proposed in an earlier form, by @tsiegel in a [post on the nvda-devel mailing list](https://groups.io/g/nvda-devel/message/46652), so I have listed him as a co-author.

### Testing strategy:

* Tested that normal entries in the default and temp dictionaries work.
* Tested that an identical entry in the default, voice, and temp dictionaries, still prefers the entry in temp in all cases.
* Tested that an entry in the temp dictionary, does not have its replacement text processed through the default dictionary.
* Tested that an entry in the default dictionary, does not have its replacement text processed through the temp dictionary.
* Tested that a replacement in the default dictionary, is properly overridden by an entry in the voice dictionary, whenever the voice dictionary is activated by use of its associated voice (per @josephsl's request below).
* Tested that the Default and temp dictionaries, do not further act on  the replacement output of a voice dictionary.

While intra-dictionary cascading has apparently been considered a bug for a long time, I can not tell whether inter-dictionary cascading has been.
In other words: if the temp dictionary generates some replacement text, was it a design choice that the replacement text would then be subject to replacement by, for example, a pattern in the voice dictionary?

I can see arguments either way, and the module has no documentation stating the original intention.

This PR, as it stands, disables all kinds of cascading, as the majority of users in the list discussion above appeared to prefer.

**That will be a breaking change for add-ons which rely on it, if any do.**

### Change log entries:

Changes, Bug fixes, For Developers

Speech dictionary processing no longer reprocesses replacement text through the same or other dictionaries, but stops after the first replacement is made, except that higher priority dictionaries still take precedence over lower (the current priority order is temporary, voice, default, built-in).

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/speechDictHandler/__init__.py
+++ b/source/speechDictHandler/__init__.py
@@ -115,7 +115,8 @@ class SpeechDict(list):
 def processText(text: str) -> str:
 	"""Calls dictionaries.sub on a text string for each defined dictionary, if dictionary processing is enabled.
 	"""
-	if not globalVars.speechDictionaryProcessing:
+	# If dictionary processing is turned off, or we have no dictionaries to process, return the text
+	if not globalVars.speechDictionaryProcessing or len(dictTypes) == 0:
 		return text
 	for type in dictTypes:
 		newText = dictionaries[type].sub(text)
--- a/source/speechDictHandler/__init__.py
+++ b/source/speechDictHandler/__init__.py
@@ -1,8 +1,8 @@
-#speechDictHandler.py
-#A part of NonVisual Desktop Access (NVDA)
-#Copyright (C) 2006-2023 NVDA Contributors <http://www.nvda-project.org/>
-#This file is covered by the GNU General Public License.
-#See the file COPYING for more details.
+# speechDictHandler.py
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2006-2023 NVDA Contributors <http://www.nvda-project.org/>
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
 
 import re
 import globalVars
@@ -112,6 +112,7 @@ class SpeechDict(list):
 				del self[index]
 		return text
 
+
 def processText(text: str) -> str:
 	"""Calls dictionaries.sub on a text string for each defined dictionary, if dictionary processing is enabled.
 	"""
@@ -124,6 +125,7 @@ def processText(text: str) -> str:
 		if newText != text:
 			break
 	return newText
+
 
 def initialize():
 	for type in dictTypes:
--- a/source/speechDictHandler/__init__.py
+++ b/source/speechDictHandler/__init__.py
@@ -1,6 +1,6 @@
 #speechDictHandler.py
 #A part of NonVisual Desktop Access (NVDA)
-#Copyright (C) 2006-2017 NVDA Contributors <http://www.nvda-project.org/>
+#Copyright (C) 2006-2023 NVDA Contributors <http://www.nvda-project.org/>
 #This file is covered by the GNU General Public License.
 #See the file COPYING for more details.
 
--- a/source/speechDictHandler/__init__.py
+++ b/source/speechDictHandler/__init__.py
@@ -112,7 +112,9 @@ class SpeechDict(list):
 				del self[index]
 		return text
 
-def processText(text):
+def processText(text: str) -> str:
+	"""Calls dictionaries.sub on a text string for each defined dictionary, if dictionary processing is enabled.
+	"""
 	if not globalVars.speechDictionaryProcessing:
 		return text
 	for type in dictTypes:
--- a/source/speechDictHandler/__init__.py
+++ b/source/speechDictHandler/__init__.py
@@ -116,8 +116,11 @@ def processText(text):
 	if not globalVars.speechDictionaryProcessing:
 		return text
 	for type in dictTypes:
-		text=dictionaries[type].sub(text)
-	return text
+		newText = dictionaries[type].sub(text)
+		# To stop intra-dictionary and inter-dictionary cascade replacements, break on successful replacement
+		if newText != text:
+			break
+	return newText
 
 def initialize():
 	for type in dictTypes:


XLTechie:
CC @josephsl @britechguy @tsiegel

josephsl:
Hi,

Few things to check:

* What happens when the voice dictionary gets involved, either due to an entry that can override the default or when synthesizer/voice changes?
* What happens if a given text must match more than one entry (example: replace "test one two three" with "test four five six" with one -> four, two -> five, three -> six)? I'm sure this PR can handle this as it will return as soon as the top-level dictionary on the stack will say something about the pattern and its replacement.

Thanks.

britechguy:
Joseph Lee wrote, in part:

   - What happens if a given text must match more than one entry (example:
   replace "test one two three" with "test four five six" with one -> four,
   two -> five, three -> six)? I'm sure this PR can handle this as it will
   return as soon as the top-level dictionary on the stack will say something
   about the pattern and its replacement.

I never thought about this before your having mentioned it, but even as
written I would think that "test one two three" would be the matching
pattern and "test four five six" would be the replacement.  Given what I
know (or think I know) the only way one would do this is via Regular
expression match using those literals but sans quotes.

I certainly would never think of having a dictionary entry for match one
and replace with four, another for two replace with five, and another for
three replace with six.  That would end up causing insane problems.  And
you'd never have any reason to put in a match and replace where what's
matched is replaced with itself.

Maybe I'm misunderstanding the concern, though.

Brian

XLTechie:
@josephsl wrote:

>     * What happens when the voice dictionary gets involved, either due to an entry that can override the default or when synthesizer/voice changes?

Added to test results.

>     * What happens if a given text must match more than one entry (example: replace "test one two three" with "test four five six" with one -> four, two -> five, three -> six)? I'm sure this PR can handle this as it will return as soon as the top-level dictionary on the stack will say something about the pattern and its replacement.

**Good catch!**
This result likely invalidates the whole idea.

I put the following entries in the following dictionaries:
DICTIONARY | PATTERN | REPLACEMENT
---------- | ------- | -----------
Default | tOne | tFour
Voice | tTwo | tFive
Temp | tThree | tSix

I used the phrase: "`Test tOne, tTwo, tThree`".

Current alpha's result was: "`Test tFour, tFive, tSix`".

The PR's result was: "`Test tOne, tTwo, tSix`".

XLTechie:
@britechguy Let's put aside @josephsl's example of a test phrase, and look at something more real-world.
👍 
Let's say you have in your default dictionary, the following:

PATTERN | REPLACEMENT
------- | -----------
cow | bovine
the moon | Luna

If you write the following two lines in notepad, and then do a say-all from the top of the file:
```
Don't have a cow, man.
It's just the moon.
```

The synth will say:

```
Don't have a bovine, man.
It's just Luna.
```

That is to be expected. However, if you have the phrase:

```
The cow jumped over the moon.
```

Your replacement will be:

```
The bovine jumped over the moon.
```

(Or `The cow jumped over Luna.`, if you had the entries in the opposite order.)

That is **not** what the average user would expect.
But because of the way speech is processed, that is what will happen under the "only process once" approach.

I should have seen that yesterday. Back to the drawing board on this one.

britechguy:
Luke,

I don't understand why the result you say you get is what you get.  I guess
this depends on what gets passed to the dictionary processor by NVDA.  For
the sentence, "The cow jumped over the moon," that after "cow" was
substituted by "bovine" that the remainder of the original sentence,
"jumped over the moon," would be parsed through the dictionary from the
beginning, with "moon" eventually being replaced by "Luna."

I clearly am not understanding what gets passed for parsing and, once a
replacement is made, that "the remainder" of a sentence is not re-parsed
from the beginning.  When I am thinking about "match and drop out" that is
at the "match regex, word, anywhere" level and that where things pick up is
immediately after the thing replaced as far as text position and that the
processing begins again at the start of the dictionary and keeps going
until something matches or the end of all entries is hit.

Brian

josephsl:
Hi, the text that the speech module was given to speak will be passed in, and in case of the example given, the entire line (when one presses up or down arrows in text fields where movement by line is supported, NVDA will expand the text info under the caret to include the line where the cursor is). Thanks.

britechguy:
Joseph,

I understand what gets passed to speak.  What I clearly don't get is how
the pre-processing of that by the dictionary works.  I think of a string,
the whole string, being looked at by the dictionary processor at the
word/phrase level.  If there are two words or phrases in a single string,
the match and replace process for the remainder of the string should not
stop just because a match and replace has occurred once already in the
string.  The entire original string must be parsed for matching and
replacing, and that matching and replacing search should always begin from
the beginning of the dictionary/dictionaries in use and stop, when a match
and replace is done, for a word or phrase, then continue for the remainder
of the original string.  You don't just stop when the first match and
replace occurs, but not all of the original string text has been parsed
yet.  And from the description given last night, the implication is that
the match/replace processing would stop whenever the first instance of a
match/replace occurs in a string, without regard to the fact that there
remains a part of the original string that still would need to be looked
at.  Until the entire original string has been examined, the process is not
over.

What I have been trying to say since the beginning is that if a
match/replace occurs, that replacement string should not continue to be
compared against all the dictionary entries that follow the one that was
used for the match/replace and, possibly, be match/replaced again,
repeatedly.  But that doesn't mean that if there was a 20 word original
string, and the match/replace was done on word 4, that words 5 through 20
should not be examined.  This should be an iterative process on the
original string until it's exhausted, but each instance of a match/replace
that's successful should be "final" and you continue processing the
remainder of the original string, starting with the beginning of the
dictionary/dictionaries, just like you had been doing up to the point where
the first match/replace was made.  There could be multiple match/replace-s
in a given original string.  Every word has the potential to be
matched/replaced.

Brian

tsiegel:
Luke.

I don't know how much difference it makes, but replacing your

Break

with the

Return

statement might solve this one.

XLTechie:
> I don't know how much difference it makes, but replacing your
> Break
> with the
> Return
> statement might solve this one.

Okay, @tsiegel, just to be thorough, I tested returning nothing (not expected to 
work), and returning the newText variable, which is the most closely matched to 
the method you proposed on nvda-devel.

Unsurprisingly, returning nothing broke NVDA almost entirely.

Returning the newText variable did the same as breaking the loop (as one would 
anticipate), in that if three entries (in separate dictionaries) should have 
effected separate words in a line of text, only the first (the one in the temp 
dictionary), actually changed the line.

I'm sorry, but this approach is insufficient to achieve the goals.

XLTechie:
@tsiegel wrote:
> the solution is simply to create a list of replaced words, and check said list before performing another replace. If word is in replacement list, the skip replace. If not, perform replace.

Simply, eh? Consider these two entries:
```
Pattern 1: SMS
Replacement 1: txt
Pattern 2: txt
Replacement 2: text
```

And this phrase:
```
Don't SMS me the txt of your speech.
```

Version after pattern 1:
```
Don't txt me the txt of your speech.
```
If we remembered replacements in pure text form, we would remember that "txt" was a replacement, and we would ignore it on the next scan through this string. Therefore, the next pass would result in:
```
Don't txt me the txt of your speech.
```
When what should happen, is that since the second "txt" was original content, not a replacement, it should be replaced by our second entry, resulting in:
```
Don't txt me the text of your speech.
```

XLTechie:
I'm pretty sure we either have to:

### Solution 1:

Mark each piece of replaced text with a tag of some sort (probably a GUID, followed by the number of characters in the replacement), and then skip over tag plus number of characters during the subsequent passes.

I haven't found a way to do that with the re library yet, although it will likely involve re.sub's callback function mechanism, which means the sub won't actually replace the text, just tell us how to in a Match object.

### Solution 2:

Following @tsiegel's list idea: keep a list of starting positions where matches were found, and replacement lengths. Then, if a new replacement would overlap that space, silently drop that replacement attempt.

That probably means we need to switch away from re.sub, and instead switch to match calls to check matchedness, and then replace in a second, conditional, step.

This second approach might be better, since it has the potential to not break regular expression look back matching, which the first one most likely would.

tsiegel:
A bit contrived, but I get the idea.

Fine then.

use a python dictionary to track both the word and the location in the 
string. (not character count, but word count), then you'd need to check 
both before replacing anything.

That's probably as good as it's going to get if you're not interested in 
all kinds of nonsense as discussed by the initial replacement solution.

britechguy:
I am unclear as to whether this issue ever resulted in a released code change or not.  It's a pretty direct spin-off of another issue I created, #11823, that remains open.

In one form or another, the attempt to continue through the dictionaries for "already replaced" text for subsequent possible repeat replacement by a later entry shouldn't happen.  And the starting point for continuing replacement should be at the beginning of whatever text comes immediately subsequent to that which has just been replaced, whether the total text being examined is a short phrase or a book-length treatise.

The replacement process should not be iterative and "first match" wins.

XLTechie:
@britechguy It did not. If you review the last few comments above, you can see 
that this is a non-trivial problem because of what NVDA considers a text chunk.

See my final comment for the two most likely paths we will have to follow to 
solve it. Either one involves a complete rewrite of the speech dictionary 
module, into something much more complex, and I have not had time for that.

XLTechie:
Oh, I think I was also waiting to see if we would implement a 
Filter extensionPoint for speech(). I know you probably don't know what that 
means, but if we did that, it would completely change the approach. Although not 
the underlying difficulty.

Adriani90:
cc: @CyrilleB79 fyi, maybe this is of interest to you.

PR #14585, opened 2023-01-25
Allow translators to specify row/column when reporting the end coordinate of a merged cell

### LINKED ISSUE #14581: Splitting translatable strings for merged cells makes translation more flexible
As far as I know, Western users are more accustomed to the reporting method of row xx column xx, but for Chinese users, on the contrary, Chinese users are more accustomed to the method of xx row xx column.
However, for vertical and horizontal merges, they share the same translatable string, which is "through %s"

NVDA reports the following:
* column XX through YY
* row XX through YY

The following is the Chinese translation:
* XX 列到 YY 列
* XX 行到 YY 行

In short, we need to append 'row' or 'column' after the translatable string "through %s".

Splitting translatable strings for merged cells makes translation more flexible
None
### Additional context
None

CyrilleB79:
### Link to issue number:
Fixes #14581 
### Summary of the issue:
For some translators (Chinese ones at least), in the case of merged cell it would be better to report "column 2 to column 3" instead of "column 2 to 3" as we do today.

Today, NVDA uses 2 translatable strings to report the coordinates of a merged cell:
* for rows: "row %s" and "through %s"
* for columns: "column %s" and  and "through %s"

Translators will now have two strings indicating the end coordinate of a merged cell, one for rows and one for columns.

I just have used `.format(...)` with named parameters. Defining strings with two parameters named differently allows to create 2 translatable strings instead of only one. As a side effect, it's also better for translators to have named parameters rather than just `%s` which is not explicit.

Note: using `pgettext` was another option, but with named parameters, it didn't turn out to be necessary.

### Testing strategy:

Manual test: Create po/pot file and check that we can translate separately the two strings indicating "row" or "column"
None
### Change log entries:
Not deserving an entry.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/speech/speech.py
+++ b/source/speech/speech.py
@@ -1,7 +1,7 @@
 # A part of NonVisual Desktop Access (NVDA)
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
-# Copyright (C) 2006-2022 NV Access Limited, Peter Vágner, Aleksey Sadovoy, Babbage B.V., Bill Dengler,
+# Copyright (C) 2006-2023 NV Access Limited, Peter Vágner, Aleksey Sadovoy, Babbage B.V., Bill Dengler,
 # Julien Cochuyt, Derek Riemer, Cyrille Bougot
 
 """High-level functions to speak information.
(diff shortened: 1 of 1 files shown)


amirmahdifard:
@CyrilleB79 your pull requests are good, hopefuly one day i see a pull request wich will add my suggestion to nvda from you: thanks

michaelDCurran (approved):

PR #14626, opened 2023-02-10
Makes the bug template friendlier for issue investigators

Adriani90:
### Link to issue number:
None
### Summary of the issue:
When investigating issues we often ask the same questions or request certain standard attachments. This process cost a lot of time when investigating issues and searching for solutions.
The users will get more insights on what is needed to investigate their issues efficiently.

I've added some new explanations in the attachments section of the bug template regarding NVDA logs, crash dumps and other attachments. I've also added two additional aspects:
• Question if users tested in other browsers in case the issue is related to a browser and request findings
• Request to provide a minimal test case as HTML file or a link to a codepen in case the issue is related to web design and DOM.

### Testing strategy:
Tested that it appears correctly in Github markdown

None
### Change log entries:
n/a

### Code Review Checklist:

  - description is up to date
  - change log entries
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [ ] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -30,6 +30,24 @@ You may additionally include an explanation.
 -->
 
 ### NVDA logs, crash dumps and other attachments:
+#### NVDA logs
+<!--
+In case NVDA is freezing, crashing or you encounter errors while using NVDA, we need the NVDA log file to investigate where the issue comes from. In the NVDA settings under "general settings" you can enable debugging . Make sure you set the logging level to "debug" before reproducing your issue.
+* After reproducing your issue with logging level set to "debug", press nvda key+f1 and save the log file as txt file. Then attach the log file to this issue.
+* The NVDA key can be the capslock key or the insert key or both.
+* Finally you can disable the logging level entirely in the NVDA general settings.
+In case NVDA is crashing completely, you need to restart NVDA or NVDA is restarting itself automatically, you find the old log file in the temp folder. Press Windows key+r, type "%temp%" and find the file called nvda_old.log. Save that as .txt file and attach that file to this issue.
+-->
+
+#### Crash dumps
+<!--
+If NVDA is crashing completely, restarting itself automatically or another application on your system is crashing while running NVDA, then a crash dump file will be generated. Please attach the crash dump files as follows:
+* NVDA crashed: Press Windows key+r, type "%temp%" and find the file called nvda_crash.dmp. Attach that file to this issue.
+* Another application chrashed while NVDA is running: press Windows key+r, type "%localappdata%" and navigate to the folder called "crash dumps". If you find a .dmp file for the application that crashed inside that folder, attach that file to this issue.
+-->
+
+#### Other attachments
+You can help us investigate your issue by adding other attachments such as test documents (i.e. for issues in Microsoft Office or other software), html files for browser issues, test applications etc.
 
 ### System configuration
 #### NVDA installed/portable/running from source:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -50,3 +68,8 @@ You may additionally include an explanation.
 #### If NVDA add-ons are disabled, is your problem still occurring?
 
 #### Does the issue still occur after you run the COM Registration Fixing Tool in NVDA's tools menu?
+
+#### in case your issue is related to a browser (i.e. Chrome), did you test the behavior of NVDA also in other browsers (i.e. Firefox or Microsoft Edge)? What are your findings in other browsers?
```
CyrilleB79, line 72: Use "e.g." instead "i.e."

```suggestion
#### in case your issue is related to a browser (e.g. Chrome), did you test the behavior of NVDA also in other browsers (e.g. Firefox or Microsoft Edge)? What are your findings in other browsers?
```

Adriani90: I actually really meant "for example" using i.e. would gramatically limit the sentence to Chrome or Firefox or Edge. Here is an article on the correct use of i.e. and e.g.
https://www.dictionary.com/e/ie-vs-eg/

I.e. is more like "that means" and e.g. is "for example".

seanbudd:
the added information provided in the template is quite verbose. We try and explain the issue template in the linked document https://github.com/nvaccess/nvda/blob/master/devDocs/githubIssueTemplateExplanationAndExamples.md

The assumption is that if a user is not willing to read that document, they are unlikely to read and comply with large chunks of text inline within the template.

Adriani90:
However, a lot of users read the template though which would at least lower the amount of time spending with raising the same questions again and again. Von meinem iPhone gesendetAm 14.02.2023 um 00:36 schrieb Sean Budd ***@***.***>:﻿
the added information provided in the template is quite verbose. We try and explain the issue template in the linked document https://github.com/nvaccess/nvda/blob/master/devDocs/githubIssueTemplateExplanationAndExamples.md
The assumption is that if a user is not willing to read that document, they are unlikely to read and comply with large chunks of text inline within the template.

seanbudd:
I've been thinking about this and I think maybe a better solution is to make [githubIssueTemplateExplanationAndExamples](https://github.com/nvaccess/nvda/blob/master/devDocs/githubIssueTemplateExplanationAndExamples.md) more readable, and include some of the information added in this PR.

We should turn the guide into an issue reporting guide, and include what information to provide for the various types of issues we get. This would help submitters self-triage an issue.

seanbudd:
Sorry didn't mean to re-open @Adriani90, but what do you think of the proposal?

Adriani90:
I think it definitely makes sense to add some details to the guide, how ever in my experience many users don't read the guide but they fill the fields that are in the template and just delete the link to the guide.
For example in case of the COM registration fixing tool, many people still answer the question with "no" even they had an issue related to gaining focus, interface or anything else related. So we have to request them again to run the COM registration fixing tool. The same applies for NVDA running without addons.
But at least many users do fill those question,s dorrectly, don't know if they really read the guide though.

What I am trying to say, I think a combination of both 
1.  more details in the guide and 
2. some details related to test documents, testing different browsers, DOM and web minimal test cases in the template, would at least contribute to reducing noisy issues.

What if we take the last two questions from this PR to the template, and also the information regarding test documents?

The infos regarding log and crash dumps could be then part of the quide since these questions are actually already in the template but without details.

PR #14658, opened 2023-02-21
Import keyLabels only after translation is  initialized to make the labels displayed in the  user's language

### LINKED ISSUE #14657: NVDA 2023.1: Key labels are displayed in the system language instead of NVDA language set in preferences.
### Steps to reproduce:
1. Set  NVDA to the language different than your Windows language (in my case since my Windows is in English I'm going to use Polish for the rest of this example).
2. After restarting NVDA enable keyboard help and try pressing keys for which translation for your language provides localized labels e.g. space, left and right shift etc.
### Actual behavior:
Key labels are reported in the language of the system i.e. in my case 'space' is reported for the space bar key instead of the Polish equivalent 'spacja'.
### Expected behavior:
Key labels should be reported in the  language set in preferences similar to how this worked in NVDA 2022.4 and earlier.
None needed
Tested from sources, but this does not matter.
#### NVDA version:
Latest beta
#### Windows version:
Windows 10 version 21H2, but this does not depend on the version of Windows which is used.
None
Yes
This was introduced in PR #14528, so all versions after this PR have this bug.
Not relevant
Not relevant

## Additional details:
Upcoming PR

lukaszgo1:
Opened against beta, since this pull request fixes regression which ideally should not get into the release.

### Link to issue number:
Fixes #14657
### Summary of the issue:
In pull request #14528 `keyLabels` were imported pretty early during NVDA's startup. Since this module contains localized strings at the module level these are translated before `languageHandler` is initialized. This means that they're translated either to the default language of the system, or for locales for which Python's `locale.getdefaultlocale` fails they remain in English, disregarding the language set in preferences.
Key labels are once again presented in the language set in preferences.
1. `keyLabels` are imported lazily in the `configFlags`
2. I've added docstring to the `keyLabels` module explaining when it can be safely imported.
### Testing strategy:
On a English Windows set NVDA's language to Polish, ensured that key labels in the keyboard help and in the preferences are displayed in Polish.
While this pull request fixes this particular issue I'm of the opinion that we should not initialize a fake translation before parsing user's configured language. Since this would be bigger change, unsuitable for the beta branch at this stage, I'm going to open a separate issue to discuss and track this. Edit: I've created #14660
### Change log entries:
None needed - unreleased regression
### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/config/configFlags.py
+++ b/source/config/configFlags.py
@@ -17,7 +17,6 @@ from utils.displayString import (
 	DisplayStringStrEnum,
 	DisplayStringIntFlag,
 )
-from keyLabels import localizedKeyLabels
 
 
 @unique
@@ -36,6 +35,8 @@ class NVDAKey(DisplayStringIntFlag):
 	
 	@property
 	def _displayStringLabels(self):
+		# Imported lazily since this module is imported before gettext translation is installed.
+		from keyLabels import localizedKeyLabels
 		return {
 			NVDAKey.CAPS_LOCK: localizedKeyLabels['capslock'],
 			NVDAKey.NUMPAD_INSERT: localizedKeyLabels['numpadinsert'],
--- a/source/keyLabels.py
+++ b/source/keyLabels.py
@@ -4,6 +4,13 @@
 #See the file COPYING for more details.
 #Copyright (C) 2008-2016 NV Access Limited, Aleksey Sadovoy, Babbage B.v.
 
+"""Contains mapping from the Windows key names to their localized labels.
+
+As there are localizable strings at module level,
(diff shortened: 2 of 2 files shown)


lukaszgo1:
CC @CyrilleB79 You may want to review this PR.

seanbudd (approved):

CyrilleB79 (approved):
Thanks @lukaszgo1 for having noticed this issue and for your fix.

PR #14664, opened 2023-02-22
Fix-up of #14557: remove unwanted backslashes appearing in replacement expression of dictionary

### LINKED ISSUE #14663: Strange behavior in dictionaries if replacement contains space
### Steps to reproduce:

1. First, set symbol level of "\\" to have it reported (i.e., None);
2. open a dictionary and create something like "foo" replaced by "foo bar" (note the space in replacement);
3. read text: "test of foo".

### Actual behavior:

Voice says:
"test of foo\\bar"

### Expected behavior:

Voice should say:
"test of foo bar"
without any "\\".

Installed and running from source.

#### NVDA version:

NVDA alpha-27727,86e93666 (installed).

#### Windows version:

Windows 10 22H2 Build 19045.2604 64-bit.

Tested with Eloquence and Microsoft OneCore (Elsa) synthesizers.

Not tested.

2023.1beta1 seems to not have this issue.

Yes.

Not tested.

CyrilleB79:
### Link to issue number:
Fixes #14663
Fix-up of #14557
### Summary of the issue:

When reading text containing a non-regexp dictionary entry, if the replacement field contains some special characters (such as space, +, etc.), these characters are escaped with a backslash. Thus this backslash is heard in the text spoken by NVDA.

No unappropriated backslash will be spoken by NVDA when the replacement field of a dictionary entry contains space or other special character.

Use `theString.replace('\\', '\\\\')` instead of `re.escape(theString)`.

Indeed, it was well documented in [the doc of `re.escape`](https://docs.python.org/3.7/library/re.html):

> This function must not be used for the replacement string in sub() and subn(), only backslashes should be escaped.

### Testing strategy:
Manual tests:
* STR in #14663
* Tests in #14557
* STR in #14556

None
### Change log entries:
None: fixing unreleased bug.
### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/speechDictHandler/__init__.py
+++ b/source/speechDictHandler/__init__.py
@@ -44,7 +44,8 @@ class SpeechDictEntry:
 		if self.type == ENTRY_TYPE_REGEXP:
 			replacement = self.replacement
 		else:
-			replacement = re.escape(self.replacement)
+			# Escape the backslashes for non-regexp replacements
+			replacement = self.replacement.replace('\\', '\\\\')
 		return self.compiled.sub(replacement, text)
 
 class SpeechDict(list):


CyrilleB79:
System test failure is unrelated.

seanbudd (approved):

Issue #14641, opened 2023-02-15
Report "Press escape to close" on NVDA dialogs which the user needs to press escape to close


Qchristensen:
### Is your feature request related to a problem? Please describe.

There are several places where a user will press a keystroke which opens a dialog which may not be obvious to all users.  Examples are:

NVDA+f pressed twice opens a dialog with the formatting of the current caret location
NVDA+r runs OCR and presents the results in a buffer the user can navigate
NVDA+k pressed twice (from NVDA 2023.1) opens a dialog with the destination URL of the current link.

Where opening something like NVDA settings should (generally) be obvious that the user is in a dialog,, these examples may not be.

### Describe the solution you'd like

My suggestion is to append "press escape to close" to the title of these dialogs / the information which is read when the user activates the keystroke.

### Describe alternatives you've considered

As it is currently, I think most users work it out, although it may still be confusing at first.  I think the proposal wouldn't inconvenience anyone, as the information would be read last and could be skipped / ignored.

### Additional context
Suggested during a conversation with user Jared via email.

josephsl:
Hi, my question is: what if the dialog is invoked on a tablet or via a braille display gesture? Thanks.

Qchristensen:
I'm not sure how easy it is to check for an input gesture copying escape?  I guess looking for that could be one option.  Otherwise, I would expect users who have setup such a gesture should understand that they can press the gesture they have setup in lieu of escape.  Much like most users understand when a general instruction says "click the link" or "click the button" - that using NVDA you move focus to it and press enter.

I would also anticipate that this would likely not be the first occasion when such a user would encounter something they needed to press "escape" to close.  My thought is this information would be most useful to new users who are less likely to have setup (or had someone setup) a gesture overriding escape.

But if there is a way to check for an input gesture for escape, then sure, that would work too.

lukaszgo1:
In JAWS this has been solved by placing a link which says something like 'Press Escape to close this message' at the end of each virtual viewer window. As a bonus clicking this link also closes the virtual viewer which nicely solves the concern about tablets and Braille only users. I'm by no means saying we should copy competition, but that flow seems worth considering.
Having said that in the initial description it was stated that:
> Where opening something like NVDA settings should (generally) be obvious that the user is in a dialog,, these examples may not be.

What is so different about these windows versus the regular settings dialog? understanding where the user confusion comes from is also valuable in this case.

ChrisDuffley:
[...]

PR #14668, opened 2023-02-23
No longer use a single semicolon as separator between title and message with ui.browseableMessage

### LINKED ISSUE #14667: Report Link Destination partially broken for link titles with semicolons
### Steps to reproduce:

In a browser such as Firefox, enter the following into the address bar:

```html
data:text/html,<p>This is a test <a href="https://nvaccess.org/">link; to NV Access</a>.</p>
```

Navigate to the link, and press Ctrl+k (the report link destination key) twice.

### Actual behavior:

NVDA reports the title of the Window shown as:
```
Destination of: link
```

And the content of the window as:
```
to NV Access;
https://nvaccess.org/
```

### Expected behavior:

NVDA reports the title of the Window shown as:
```
Destination of: link; to NV Access;
```

And the content of the window as:
```
https://nvaccess.org/
```

### The reason:

The string (message and title) sent to `message.html`'s javascript from `ui.browseableMessage`, is naively split on a single semicolon. This could have happened before the link destination feature was added, for any `ui.browseableMessage` with a semicolon in its title.

Noticed while working on #14641.

Fix coming shortly.

N/A

Portable

#### NVDA version:

2023.1 beta 1

#### Windows version:

10 Home

N/A

Not tested.

Also true for alpha and source builds.

N/A

Yes

XLTechie:
### Link to issue number:

Fixes #14667 

### Summary of the issue:

If `ui.browseableMessage` was given a title containing a semicolon (;), the title would be broken at the semicolon, and the rest of the title would be rendered before (as part of) the message in the HTML portion of the window.
This had the potential to result in unexpected outcomes, if the message content was HTML; not to mention being confusing for the user.

While this was noticed in relation to the Report link destination feature introduced by #14583, it did not result from that feature, and could have occurred prior to its introduction. It does however effect it significantly, as the title of the link the destination of which is being identified, is presented in the title of a ui.browseableMessage, if the command is called twice, or the alternate version of the command is mapped to a key.

If a link name contains a semicolon, portions of that name will no longer appear along with the link address, when reporting link destination in its window version.

Changed the regex separator string in `ui.browseableMessage`, and in the `message.html` file, to one which should hopefully be much less likely to appear in the wild. Instead of a single semicolon ("`;`"), the string is now "`__NVDA:split-here__`".

Also removed the blank line from the top of `message.html`. I know of no purpose for having a blank line there.

### Testing strategy:

* Performed the STR in #14667, and confirmed that the title no longer breaks.
* Tested some other cases of `ui.browseableMessage`, such as NVDA+f, to make sure they act as expected.

### Change log entries:

This was introduced in beta. If fixed in beta, no user level change should need to be listed.

For Developers
Titles in `ui.browseableMessage` calls, can now include semicolons without side effects.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/message.html
+++ b/source/message.html
@@ -1,4 +1,3 @@
-
 <!doctype html>
 <HTML style="width : 350; height: 300">
 <HEAD>
@@ -15,7 +14,7 @@ function escapeKeyPress(e) {
 function windowOnLoad() {
 	// #5875: string.prototype.split strips the tail when a limit is supplied,
 	// so use a regexp instead.
-	var args = window.dialogArguments.match(/^(.*?);([\s\S]*)$/);
+	var args = window.dialogArguments.match(/^(.*?)__NVDA:split-here__([\s\S]*)$/);
 	// args[0] is the whole string.
 	if (args && args.length == 3){
 		document.title= args[1];
--- a/source/ui.py
+++ b/source/ui.py
@@ -88,6 +88,7 @@ def browseableMessage(message: str, title: Optional[str] = None, isHtml: bool = 
 	@param title: The title for the message.
 	@param isHtml: Whether the message is html
 	"""
+	splitWith: str = "__NVDA:split-here__"  # Unambiguous regex splitter for javascript in message.html, #14667
 	if _isSecureDesktop():
 		import wx  # Late import to prevent circular dependency.
 		wx.CallAfter(_warnBrowsableMessageNotAvailableOnSecureScreens, title)
@@ -103,7 +104,7 @@ def browseableMessage(message: str, title: Optional[str] = None, isHtml: bool = 
 		title = _("NVDA Message")
 	if not isHtml:
 		message = f"<pre>{escape(message)}</pre>"
-	dialogString = f"{title};{message}"
+	dialogString = f"{title}{splitWith}{message}"
 	dialogArguements = automation.VARIANT( dialogString )
 	gui.mainFrame.prePopup() 
 	windll.mshtml.ShowHTMLDialogEx( 


XLTechie:
@seanbudd Two things:

1. I have targeted beta with this, because the feature which highlighted this problem was introduced there, and because:
2. Technically, I think this is API breaking. While highly unlikely, it is remotely possible that somebody somewhere was relying on a semicolon in the title of a ui.browseableMessage, to break the title and put part of it in the message itself. While that usage seems absurd and unintended, it's possible. Therefore, if NV Access considers this API breaking, we can only get it in beta, or wait until 2024.1 to fix this.

XLTechie:
I'm reasonably sure none of these failed system tests result from this PR's changes.

CyrilleB79:
@XLTechie two points:
1. Using a very unlikely string as a separator, the code would not be waterproof to an hypothetical situation with a link containing the same string, e.g. reading NVDA's code itself where each line or variable/token would be a link. Have you investigated an alternative solution with encapsulation of each argument, e.g. with double quotes, and escaping or doubling the encapsulation character in the encapsulated content?
2. I do not think that this PR is considered API breaking change. Fixing obviously bugged functions (thus changing their result) has never been considered API-breaking and it does not make sense to me.

XLTechie:
@CyrilleB79
>  1. Using a very unlikely string as a separator, the code would not be waterproof to an hypothetical situation with a link containing the same string, e.g.

It was far less waterproof with a simple semicolon.

>     reading NVDA's code itself where each line or variable/token would be a link.

Do you know any real world example of that that would apply here? It is a string 
used inside a variable, which is then passed through the mshtml library to 
a javascript function.
Does anything really do HTML annotations of the contents of strings?

> Have you investigated an alternative solution with encapsulation of each
>     argument, e.g. with double quotes, and escaping or doubling the encapsulation character in the encapsulated content?

We would need a Python function which encodes (escapes) HTML within double 
quotes, and a javascript function which agrees with it on the other side to 
decode it back into a string which can be split unambiguously.
I am not a js expert, but there probably is such a couple of functions already.

The best way, would be to pass some kind of object (or a dict) through the 
pointer that gets dereferenced later for the javascript. That way we would have 
individual properties to work with.

I have done a bit of reading up on this very old function, and have yet to come 
up with a way to do it, but since you object to the method I have used, I will 
keep trying.

A possible alternate solution I have just thought of, is to prepend the 
title string with the length of the title string, then add the message.
Split it with substring() on the javascript side.

>  2. I do not think that this PR is considered API braking change. Fixing obviously bugged functions (thus changing their result) has never been considered API-breaking and it does not make sense to me.

I tend to agree with you. However, I'm not NV Access. It is their consideration 
to make. It was not strictly a bug, just a bad implementation.

seanbudd (approved):

XLTechie:
@CyrilleB79, in response to your comments, I wrote:
> I am attempting now to do it with JSON, as I would like something that allows individual properties without splitting.

And the result, after much banging head against wall, is to learn that mshtml is running in compatibility mode, and therefore does not support the JSON object.
Thus, while it is very possible to pass a JSON string into message.html, `JSON.parse` does not exist to do anything with it.

Unless Microsoft is going to change mshtml as they remove IE from people's machines, we are left with few options but to either include a JSON library in message.html, or use another approach. I suspect NV Access would not want a third party JSON library included.

I have looked at passing other types than string through `dialogArguments`. While this seems possible (some of the automation languages do it), either the monkeypatching we do to the types we pass to COM interfaces doesn't allow this, or those languages are performing some magic that we don't, because I haven't found a way to make it happen.

The last two ideas I've had, are base64 encoding some kind of object and passing that; or the numeric offset method I described earlier. I would like to discover a robust way to pass an arbitrary number of isolated strings to the javascript, so that we have options for implementing features like those under discussion in #14641.

I will continue to play with this as I have time. Suggestions appreciated.

Issue #7770, opened 2017-11-17
GUI: About dialog is rendered in English


josephsl:
Hi,

A critical showstopper found: when opening About dialog in languages other than English, content is displayed in English.

STR:

1. Switch to a language other than English (NVDA menu/Preferences/General).
2. Restart NVDA.
3. Open About dialog.

Expected: content is displayed in the new language.
Actual: English is displayed.
Severity: critical.
Affects: NVDA 2017.4 RC, master, next snapshots.

Thanks.

k-kolev1985:
Yes, I can definitely reproduce this. But why it happened is a mystery to me...

josephsl:
Hi,

After testing various binary snapshot builds, it appears Python 2.7.14 upgrade from the build server is the likely culprit.

Thanks.

zstanecic:
well, how we know that,

do you have an evidence?


W dniu 17.11.2017 o 15:20, Joseph Lee pisze:

florianionascu7:
Hello, I also confirm this bug. I hope it'll be fixed in 2017.4

josephsl:
Hi, the last master that did work was 14539, and that was with Python 2.7.13. Starting from 14549, it started using Python 2.7.14. At the same time, next snapshots began showing this problem, with the last next snapshot with no problems being 14536. This leads me to believe that something that was committed around October 18th might be to blame, and we cannot leave out Python 2.7.14. Thanks.

leonardder:
[**@josephsl**](https://github.com/josephsl) commented on [17 nov. 2017 15:20 CET](https://github.com/nvaccess/nvda/issues/7770#issuecomment-345255871 "2017-11-17T14:20:11Z - Replied by Github Reply Comments"):
> After testing various binary snapshot builds, it appears Python 2.7.14 upgrade from the build server is the likely culprit.

I tend to disagree with this finding. I"m running Python 2.7.13 on my local system, and when running from source, the about dialog is also showing up in English. Note that the title of the dialog is properly translated.

josephsl:
Hi,

In that case, Python is not to blame.

Further testing shows that the new likely culprit might be #7649, as prior to that, there was no problem with this dialog.

CC @MichaelDCurran

leonardder:
Comfirmed.

I did a git checkout 1fcc7fe, and the issue did occur. After that, i did a git checkout head~, and the translation showed up correctly. Assuming this is the right way to test this, of course.

leonardder:
[...]

PR #14677, opened 2023-02-27
Don't instal gettext and pgettext before initializing languageHandler

lukaszgo1:
### Link to issue number:
Related to #14660

### Summary of the issue:
When starting NVDA installs a 'fake' `gettext` and `pgettext` functions which translates to the language set as the current Windows locale. Later in the startup process `languageHandler` overwrites these installed translation functions with a new one which translate strings to the language chosen by the user (either in the configuration or from the CLI). Unfortunately if there is a module with translatable strings at its top level and it is imported before `languageHandler` is initialized these strings are either translated to the default system language, or for locales for which Python's `locale.getdefaultlocale` fails they remain in English. Similar problems were reported in #7770 and in #14657. Both of these are fixed, but the main problem is that bugs like this are difficult to identify before the release, since most contributors either use both NVDA and Windows in English, or at least their NVDA is in the same language as their operating system.
This should not have any user visible impact.
When starting NVDA or `nvda_slave` `gettext` and `pgettext` are no longer installed before initializing `languageHandler`.
### Testing strategy:
Made sure NVDA starts - this require a longer period of testing in Alpha anyway.
None known
### Change log entries:
None needed - this has no impact for add-ons developers since add-ons were initialized after `languageHandler`.
### Code Review Checklist:

  - description is up to date
  - change log entries
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/nvda.pyw
+++ b/source/nvda.pyw
@@ -15,13 +15,11 @@ import os
 
 import typing
 
-import builtins
 import globalVars
 import ctypes
 from ctypes import wintypes
 import monkeyPatches
 import NVDAState
-from languageHandler import makePgettext
 
 
 monkeyPatches.applyMonkeyPatches()
@@ -61,26 +59,7 @@ globalVars.appDir = appDir
 globalVars.appPid = os.getpid()
 
 
-import locale
-import gettext
-
-try:
-	trans = gettext.translation(
-		'nvda',
-		localedir=os.path.join(globalVars.appDir, 'locale'),
-		languages=[locale.getdefaultlocale()[0]]
-	)
-	trans.install()
-	# Install our pgettext function.
-	builtins.pgettext = makePgettext(trans)
-except:
-	gettext.install('nvda')
-	# Install a no-translation pgettext function
-	builtins.pgettext = lambda context, message: message
-
-import time
 import argparse
-import globalVars
 import config
 import logHandler
 from logHandler import log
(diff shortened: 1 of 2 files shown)


CyrilleB79:
What about keeping these functions as a fallback but logging an error instead (`log.error`)? This would allow to avoid NVDA not starting at all in case such error is encountered.

Of course, the log should be delayed in some way because the log handler may not be initialized when the fake translation (or rather fallback translation) is called.

lukaszgo1:
> What about keeping these functions as a fallback but logging an error instead (`log.error`)? This would allow to avoid NVDA not starting at all in case such error is encountered.
> 
> Of course, the log should be delayed in some way because the log handler may not be initialized when the fake translation (or rather fallback translation) is called.



This would add quite a lot of complexity for very little benefit IMHO.
I see three scenarios worth considering for this work:
1. When implementing a new feature developer mistakenly imports a module with localized variables at the top level before initializing translations - in that case NVDA failing to start rather than logging an error is a good thing.
2. In some scenario which we haven't tested before merging this PR NVDA imports a module with translatable variables - if it fails in Alpha NVDA failing to start is an acceptable outcome and users tent to ignore errors in the log anyway.
3. For stable NVDA errors in the log aren't visible for most users, - while NVDA failing to start is not ideal we would at least know that something has failed.

CyrilleB79:
@lukaszgo1 your reasoning makes sense and, at the end, I agree with you.

seanbudd (approved):

PR #13045, opened 2021-11-09
Driver for Tivomatic Caiku Albatross 46/80 displays

burmancomp:
### Summary of the issue:
No support for Tivomatic Caiku Albatross 46/80 braille display models.
### Description of how this pull request fixes the issue:
General

Different to many other displays, Albatross models do not wait for the driver to send a query of their presence to the device. Instead, the device continuously sends initialization packets until driver requests to them quit, as the connection has been established.

These displays also expect regular data packets, with frequency of at least one every 2 seconds. Otherwise they fall back to "wait for connection" state and start sending the initialization packets until the driver sends a quit packet.

Similarly, if the user enters the device internal menu, after exiting, initialization packets are continuously sent until driver sends quit packet.

Init packets are also sent when device is powered off and then on.

Display init packets consist of two bytes: the first one is \xff which tells that this is an init packet. The second one is a settings byte which contains display settings like length of display and number of status cells. The most meaningful setting is the length of display. Other settings contained by settings byte can be regarded as notes to screenreader, and it is screenreader or driver job to use them when applicable. For example, there are no separate status cells in the device but if screenreader supports using status cells, it can be notified to use them by settings byte.

The settings byte can be anything between \x00 and \xff. Thus it could be the same as init byte.

It is possible that settings byte may have same value with any display buttons. Because init packets may be sent also during session (user exits display internal menu for example), it is essential to know if byte is button press or part of init packet.

When display is in "wait for connection" state, it sends init packets continuously. As such, there may be hundreds of bytes to handle. There are several rx buffers between device and driver which seems to cause situation that all data cannot be read with one read operation. It cannot be known when all data has been read. This is the case with init packets but also with key combinations.

There are other devices with same PID&VID. When automatic braille display detection is used, other displays with same PID&VID are tried before Albatross. Those drivers try to send queries to the port to detect their own displays. These queries may cause Albatross to send unexpected init packets which in turn could disturb this driver - it could get inappropriate settings byte. This is tried to prevent by resetting I/O buffers so that strange packets would be discarded.

If however, there are still strange init packets, Albatross should be manually selected from the display list.

[...]

PR #14686, opened 2023-03-01
Fixup #14531: Make Albatross again the last driver in detection order

LeonarddeR:
## Link to issue number:
None

### Summary of the issue:
As part of #14531, I ordered display order in bdDetect alphabetically. However, it seems that querying Albatross first has a major negative impact on detecting other displays using the same Serial to USB converter (see #13045)

Handy Tech displays such as the Modular Evolution instantly auto detect again.

Restored the order, Albatross is last in the dictionary again. I'm pretty unhappy with how these displays work, but it looks like it's a design thing in the displays itself.

### Testing strategy:
Tested that Handy Tech displays are detected immediately.

No changes to the albatross driver. @burmancomp  Would it be possible to provide a follow up for #13045 that decreases MAX_INIT_RETRIES or changes SLEEP_TIMEOUT to a lower value? This driver takes around 8 seconds to find out whether a display is connected or not if it doesn't receive the necessary packets. I assume as the display sends lots of these packets initially, it should be possible to find out whether initialisation worked after at most 5 attempts.

### Change log entries:
None needed

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/bdDetect.py
+++ b/source/bdDetect.py
@@ -545,11 +545,6 @@ def initialize():
 	scanForDevices.register(_Detector._bgScanBluetooth)
 
 	# Add devices
-	# albatross
-	addUsbDevices("albatross", KEY_SERIAL, {
-		"VID_0403&PID_6001",  # Caiku Albatross 46/80
-	})
-
 	# alva
 	addUsbDevices("alva", KEY_HID, {
 		"VID_0798&PID_0640",  # BC640
@@ -798,6 +793,11 @@ def initialize():
 		"seikantk",
 		isSeikaBluetoothDeviceMatch
 	)
+
+	# albatross
+	addUsbDevices("albatross", KEY_SERIAL, {
+		"VID_0403&PID_6001",  # Caiku Albatross 46/80
+	})
 
 
 def terminate():


burmancomp:
In the worst case it takes time to detect Albatross:

- device can be connected with 19200 and 9600 baud rate, driver does not know which one is used so it has to try with both starting 19200 which is the default
- although Albatross send continuously its initialization packet when there is no connection, it waits approximately 2 seconds before sending packets if there was working connection. This means that when driver or NVDA are restarted there must be enough time/retries so that device has surely started to send initialization packets
- I/O buffer reset seems to be necessary for proper operation (this might not be needed if Albatross were first in the list but it would require careful testing), and one reset seems not to be enough (maybe because there are several buffers between NVDA and device).

From constants.py:

`MAX_INIT_RETRIES = 20
"""
How many times port initialization or opening is tried.
This large value is required for braille to start.
This should not effect other devices with the same PID and VID,
because this device is detected as last one.
"""
`

I think it's better to detect Albatross after other devices with same PID and VID.

leonardder:
Thanks for the quick clarification. Let's go ahead with this then and restore the old behavior.

michaelDCurran (approved):

PR #14691, opened 2023-03-02
Fixed a string that was not translatable.

CyrilleB79:
### Link to issue number:
[This thread](https://groups.io/g/nvda-translations/topic/problem_with_has_details/97336868?p=,,,100,0,0,0::recentpostdate/sticky,,,100,2,0,97336868,previd%3D1677758598895856486,nextid%3D1639270081052421299&previd=1677758598895856486&nextid=1639270081052421299) on the translators' mailing list.

### Summary of the issue:

The string "has %s" is not translatable whereas it should.

This is due to the fact that the formatting has been included in the gettext (underscore function)'s argument.

For translators: the translation of the string "has %s" should now be working.

First call gettext `_` (underscore) function, and then format the result.
### Testing strategy:
Manual test:
These strings have already been translated. Thus, test with the appveyor build or from source that translation is working:
- [x] in the example of the mailing list
- [x] in the [example](https://mdn.github.io/html-examples/aria-annotations/) of PR #14480 (paragraph "Comments")
- [x] in the [example](https://codepen.io/reefturner/pen/poKLQyj) of PR #14507.

None
### Change log entries:
Not needed for this small fix.
### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/speech/speech.py
+++ b/source/speech/speech.py
@@ -1821,7 +1821,7 @@ def getPropertiesSpeech(  # noqa: C901
 				textList.append(
 					# Translators: Speaks when there are further details/annotations that can be fetched manually.
 					# %s specifies the type of details (e.g. "comment, suggestion, details")
-					_("has %s" % roleString)
+					_("has %s") % roleString
 				)
 		else:
 			textList.append(


XLTechie:
Why not change that to a .format call while you're at it?
My understanding was that we're trying to remove % substitutions wherever we 
find them (all over the place).

CyrilleB79:
> Why not change that to a .format call while you're at it? My understanding was that we're trying to remove % substitutions wherever we find them (all over the place).

It seems indeed that the trend is more to `.format(...)` for translatable strings and f-string for other string formatting, especially for new code. But this does not seem a strong requirement from NVAccess and I have seen situations where `%` was still used in recent PRs.

Regarding the current PR, keeping the string as is allows translators not to have to re-translate it.

Let's wait for the feedback from NVAccess' review.

seanbudd (approved):

Issue #11169, opened 2020-05-15
Infrequent crash in Firefox


MarcoZehe:
### Steps to reproduce:

There are no steps to reproduce. Sometimes, when working in Firefox, NVDA just falls silent, and braille stops outputting something. This is different from the RPC hang @jcsteh and I have been chasing for two years now. This is just NVDA suddenly going away. So there are no exact steps. But this is in the old log.

```
WARNING - watchdog._watcher (13:40:21.007) - watchdog (16560):
Core frozen in stack:
  File "nvda.pyw", line 215, in <module>
  File "core.pyc", line 545, in main
  File "wx\core.pyc", line 2134, in MainLoop
  File "gui\__init__.pyc", line 1052, in Notify
  File "core.pyc", line 515, in run
  File "queueHandler.pyc", line 83, in pumpAll
  File "queueHandler.pyc", line 50, in flushQueue
  File "eventHandler.pyc", line 64, in _queueEventCallback
  File "eventHandler.pyc", line 208, in executeEvent
  File "eventHandler.pyc", line 241, in doPreGainFocus
  File "browseMode.pyc", line 1255, in event_treeInterceptor_gainFocus
  File "browseMode.pyc", line 56, in reportPassThrough
  File "nvwave.pyc", line 396, in playWaveFile
  File "nvwave.pyc", line 311, in stop
  File "nvwave.pyc", line 283, in _idleUnbuffered
```

### Actual behavior:

Crash.

### Expected behavior:

No crash.

#### NVDA installed/portable/running from source:

Installed.

#### NVDA version:

version alpha-20206,dba33431

#### Windows version:

10 19H2.

#### Name and version of other software in use when reproducing the issue:

Firefox Nightly 78, or Beta 77.

#### Other information about your system:

Focus Braille Display 5th generation connected.

#### Does the issue still occur after restarting your computer?

Yes.

#### Have you tried any other versions of NVDA? If so, please report their behaviors.

2020.1, same result after some time.

#### If addons are disabled, is your problem still occuring?

Yes.

#### Did you try to run the COM registry fixing tool in NVDA menu / tools?

Yes, but since all is working normally until the crash, nothing changed.

Info for @jcsteh , this also happened before the very last COM fix you put into Nightly this week. So it is not a side effect of that. The log suggests rather that something makes NVDA freeze.

jcsteh:
Looks like NVDA froze trying to play audio. That's pretty weird and
[...]

Issue #10413, opened 2019-10-22
nvwave.playWaveFile not fully async


leonardder:
nvwave.playWaveFile has a keword to play the wave asynchronously. However, only the playback of the wave is async. A complete call of the function still takes around 10 miliseconds for the calling threat to return.

Code snippet:

```
import nvwave, time

def waveTest():
	curTime = time.time()
	nvwave.playWaveFile("waves\\start.wav")
	return time.time() - curTime

waveTest()
```

Adriani90:
cc: @jcsteh

Adriani90:
What is left to be done on this issue now that #14697 is merged?

jcsteh:
The wave file is still opened and its header read on the calling thread. That should usually be very fast, but if i/o is slow for whatever reason, it might not be. Fixing that would require moving most of the rest of the function into the background thread.

Issue #1409, opened 2011-03-09
Ability to adjust volume of sounds


nvaccessAuto:
**Reported by KevanGC on 2011-03-09 06:12**
Hi,

I've noticed that when you're browsing the web and you switch between Browse and Focus mode, NVDA plays these sound-effects. The sound-effects are very loud and sometimes they get annoying. It would be cool if you could add a feature for users to adjust the volumes of sound effects and progress-bar tones?
**Blocking #892, #2978, #3100, #4590**

ctoth:
This is still giving a terrible user experience for the default case in 2016.
I often turn down my synthesizer volume when I need to turn up my system volume to hear something quiet. If then an NVDA UI sound plays, it is actively painful.

Please stop hurting my ears.

Laughingthunder:
I agree with this. I think there should definitely be an option to tie 
the UI sounds to the speech volume.

jcsteh:
I provided details on the problems with the current patch in these [two](https://github.com/nvaccess/nvda/issues/1409#issuecomment-155288375) [comments](https://github.com/nvaccess/nvda/issues/1409#issuecomment-155288376). I don't have time to work on this right now, but I'm very happy to discuss if someone can propose a solution which deals with the problems I outlined.

bhavyashah:
Merely introducing another option in the Voice Settings dialog or the Synthesizer Settings dialog to configure both the sounds and tones of NVDA combined would suffice, in my opinion. An average user would not differentiate between NVDA's sounds, tones and beeps, and thus rolling all of their customizability into one control should not be restrictive at all. Splitting this proposed option into sounds and tones, on the other hand, may seem unnecessary clutter and an overkill for some. Upon reading the discussion on this ticket though, I reckon I agree that merging this proposal into the Voice Volume option itself would be counter-intuitive. Could this ticket, if possible, please also be prioritized since there seems enough demand for this facility viewing the number of other tickets it blocks or at least assigned a priority?

feerrenrut:
Its hard to prioritise this above the current core projects. The demand is clear, and this seems to be an important problem to solve. On the other hand, at this stage, it is not clear what the preferred UX or implementation would be. This makes the implementation cost high. I'll set this as a P3 for now. If we can become more certain about the UX and try to simplify the issue and thus the cost of implementing this then we may be able to make it a P2.

fernando-jose-silva:
I would just like to put my point of distress.
[...]

PR #14697, opened 2023-03-06
Support for audio output using WASAPI

### LINKED ISSUE #5096: Move NVWave to C/C++
**Reported by camlorn on 2015-05-18 18:51**
#4877, the speech refactor, is going to need some sort of new, higher-level mixer.  This mixer will need to b implemented in C.  The existing NVWave code is sufficient but is currently Python, necessitating a great deal of complexity and copying to be reused in this manner.

I propose moving the NVWave API into C, using a background thread for all audio processing which is communicated with via a threadsafe queue and a command structure.  The background thread allows large delays when closing devices and, as discussed on IRC, the ability to insert blocks of zeros for Windows 8 audio ducking (something I cannot personally test).
I propose three changes to the interface of the API which should be discussed before I begin coding.
First, we should fix the bit depth at 16 bits.  8 is nearly useless and WaveOut does not provide straightforward ways to query for higher bit depths, meaning that most code should already be using 16-bit audio anyway.
Second, we should replace blocking on the last buffer with blocking for a configurable number of milliseconds.
Third, we should add and encourage code to call a finalize method. 
I propose [Powercores](http://github.com/camlorn/powercores.git) for the threadsafe queue.  This is written by me, released under the Unlicense, and currently header only.  It comes with tests and has been used in Libaudioverse for at least 6 months.
Internally, I want to mix in chunks.  If we wish to block around 100 milliseconds, we can split the audio into 100 MS pieces internally, maintain exactly two buffers, fill both, and block until one of them becomes available for new data.  This will allow lowering the buffer size in synths, as the API will refuse to block until it has at least 100 MS of audio (i.e. 64-sample chunks are fine, we'll append them together in C).  This should result in increased callback accuracy in future.  Because we are using a push-based API, blocking at slightly higher MS thresholds is acceptible.
Device enumeration stays as-is in Python.  This is probably about 300-500 lines of C/C++.  I plan to write the patch.
Let's talk about where I'm either being stupid or need to explain further.  I'm sure I've missed something.

### LINKED ISSUE #10185: Python 3 versions of NVDA produce a scratch in the speech when finishing the end of a line
### Steps to reproduce:
Open a dokument
Press NVDA+A for "say all"
### Actual behavior:
You will find that NVDA scratches sometimes. Mostly at the end of a line, either a line you did with enter, or the automatic line change if one long line is too big for the screen.
### Expected behavior:
NVDA shouldn't scratch a single time
portable
#### NVDA version:
alpha-18574,56f5fcd2
#### Windows version:
7 professional sp1 64 bit
doesn't matter here
yes
No. The latest stable obviously doesn't do that, since it's still running on python 2. However, I didn't try it with the threshold or any other version there may be on a different branch running python 3.

### LINKED ISSUE #11061: Texts with multiple line spacings are voiced with NVDA + down arrow and voices crack
### Steps to reproduce:
1. Create text as follows in the notepad.
```
test




test 123 123


test 111




test   ooooooo




test one



test two





teeest


wauw







test




test 123 123


test 111




test   ooooooo




test one



test two





teeest


wauw






test




test 123 123


test 111




test   ooooooo




test one



test two





teeest


wauw

```
2. Go to the beginning of the text.
3. Press the NVDA + down arrow.

### Actual behavior:

Voices crack in some places.

installed
#### NVDA version:

2019.3.1

#### Windows version:

windows 10 pro, 64 bit, 18363.778

Yes
Tested with versions after 2019.3.1.
Sounds are cracking.
Yes
Yes

### LINKED ISSUE #11615: nvWave: Use Windows Audio Session API

jcsteh:
### Link to issue number:
Fixes #11615. Fixes #5096. Fixes #10185. Fixes #11061. Partially addresses #10413 and #1409. May address #11169.

### Summary of the issue:
NVDA's existing audio output code (nvwave) is largely very old and uses WinMM, a very old legacy Windows audio API. It is also written in pure Python, contains quite a few threading locks necessitated by WinMM, and parts of it have become rather difficult to reason about. There are several known stability and audio glitching issues that are difficult to solve with the existing code.

At the very least, this fixes audio glitches at the end of some utterances as described in #10185 and #11061.

I haven't noticed a significant improvement in responsiveness on my system, but my system is also very powerful. It's hard to know whether the stability issues (e.g. #11169) are fixed or not. Time will tell as I run with this more.

1. The bulk of the WASAPI implementation is written in C++. The WASAPI interfaces are easy to access in C++ and difficult to access in Python. In addition, this allows for the best possible performance, given that we regularly and continually stream audio data.
2. The WinMM code fired callbacks by waiting for the previous chunk to finish playing before sending the next chunk, which could result in buffer underruns (glitches) if callbacks were close together (#10185 and #11061). In contrast, the WASAPI code uses the audio playback clock to fire callbacks independent of data buffering, eliminating glitches caused by callbacks.
3. The WinMM WavePlayer class is renamed to WinmmWavePlayer. The WASAPI version is called WasapiWavePlayer. Rather than having a common base class, this relies on duck-typing. I figured it didn't make sense to have a base class given that WasapiWavePlayer will likely replace WinmmWavePlayer altogether at some point.
4. WavePlayer is set to one of these two classes during initialisation based on a new advanced configuration setting. WASAPI defaults to disabled.
5. WasapiWavePlayer.feed can take a ctypes pointer and size instead of a Python bytes object. This avoids the overhead of additional memory copying and Python objects in cases where we are given a direct pointer to memory anyway, which is true for most (if not all) speech synthesisers.
6. For compatibility, WinmmWavePlayer.feed supports a ctypes pointer as well, but it just converts it to a Python bytes object.
7. eSpeak and oneCore have been updated to pass a ctypes pointer to WavePlayer.feed.
8. When playWaveFile is used asynchronously, it now feeds audio on the background thread, rather than calling feed on the current thread. This is necessary because the WASAPI code blocks once the buffer (400 ms) is full, rather than having variable sized buffers. Even with the WinMM code, playWaveFile code could block for a short time (#10413). This should improve that also.
9. WasapiWavePlayer supports associating a stream with a specific audio session, which allows that session to be separately configurable in the system Volume Mixer. NVDA tones and wave files have been split into a separate "NVDA sounds" session. WinmmWavePlayer has a new setSessionVolume method that can be used to set the volume of a session. This at least partially addresses #1409.

### Testing strategy:
- Given that this deals with realtime audio output, there isn't a way to unit test or system test this.
- I've been running with this code for several weeks as my main screen reader.
- I've tested with both eSpeak and OneCore.
- I've tested various situations requiring indexing with both eSpeak and OneCore, including say all, indentation tones and focusing dialogs (which relies on indexing due to multiple utterances).
- I've tested switching default devices, disconnecting the default device and disconnecting a preferred device.
- I've tested configuring NVDA to use a specific (non-default) device, exiting NVDA, disconnecting that device and starting NVDA. NVDA falls back to the default device as it should.
- I've tested independently setting the volume of NVDA speech and sounds in the system Volume Mixer.
- I have not tested with operating systems other than Windows 11 21h2, as I don't have any other versions of Windows set up to test with.
- I also haven't tested with any other speech synthesisers, as I don't have any.

- This needs a lot more testing across various computers, Windows versions, audio devices, speech synthesisers and configurations. It is enabled by default to gather maximum feedback, but can be easily disabled if it causes problems.
- Calling tones.beep with a length longer than 399 ms will block, again due to the different buffering strategy. I don't think this will matter in practice, as I doubt beeps that long are used in practice.
- If a synthesiser sends very small chunks of audio, playback might glitch. I'm not sure if any synths actually do this, so I'm not sure if this needs to be addressed or not. I do have some thoughts on how we could handle it though.
- If a preferred device is specified and it is disconnected, playback falls back to the default device. The WinMM code had some logic to switch back to the preferred device if it is reconnected. I haven't implemented this for WASAPI; it will remain on the default device. I think it's possible to implement this, but it increases complexity and I'm not sure how useful it is, especially since the WinMM code only did this on idle anyway.

### Change log entries:
New features
NVDA now outputs audio via the Windows Audio Session API (WASAPI), which may improve the responsiveness, performance and stability of NVDA speech and sounds. This can be disabled in Advanced settings if audio problems are encountered.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/config/configSpec.py
+++ b/source/config/configSpec.py
@@ -56,6 +56,7 @@ schemaVersion = integer(min=0, default={latestSchemaVersion})
 # Audio settings
 [audio]
 	audioDuckingMode = integer(default=0)
+	wasapi = boolean(default=false)
 
 # Braille settings
 [braille]
--- a/source/config/configSpec.py
+++ b/source/config/configSpec.py
@@ -56,7 +56,7 @@ schemaVersion = integer(min=0, default={latestSchemaVersion})
 # Audio settings
 [audio]
 	audioDuckingMode = integer(default=0)
-	wasapi = boolean(default=false)
+	wasapi = boolean(default=true)
 
 # Braille settings
 [braille]
--- a/source/nvwave.py
+++ b/source/nvwave.py
@@ -693,8 +693,15 @@ def playWaveFile(
 	)
 
 	def play():
+		global fileWavePlayer
 		fileWavePlayer.feed(f.readframes(f.getnframes()))
 		fileWavePlayer.idle()
+		# #11169: Files might not be played that often. Leaving the device open
+		# until the next file is played really shouldn't be a problem regardless of
+		# how long we wait, but closing the device seems to hang occasionally.
+		# There's no benefit to keeping it open - we're going to create a new
+		# player for the next file anyway - so just destroy it now.
+		fileWavePlayer = None
 
 	if asynchronous:
 		if fileWavePlayerThread is not None:
@@ -897,7 +904,7 @@ class WasapiWavePlayer(garbageHandler.TrackedObject):
 			yield devId, name
 
 	@staticmethod
-	def _deviceNameToId(name):
+	def _deviceNameToId(name, fallbackToDefault=True):
 		if name == WAVE_MAPPER:
 			return ""
 		for devId, devName in WasapiWavePlayer._getDevices():
@@ -907,6 +914,8 @@ class WasapiWavePlayer(garbageHandler.TrackedObject):
 				return devId
 		# Check if this is the WinMM sound mapper device, which means default.
 		if name == next(_getOutputDevices())[1]:
+			return ""
+		if fallbackToDefault:
 			return ""
 		raise LookupError
 
--- a/source/nvwave.py
+++ b/source/nvwave.py
@@ -21,6 +21,9 @@ from ctypes import (
 	create_unicode_buffer,
 	sizeof,
 	byref,
+	c_void_p,
+	CFUNCTYPE,
+	string_at,
 )
 from ctypes.wintypes import (
 	HANDLE,
@@ -31,7 +34,10 @@ from ctypes.wintypes import (
 	UINT,
 	LPUINT
 )
+from comtypes import HRESULT, BSTR
+from comtypes.hresult import S_OK
 import atexit
+import weakref
 import garbageHandler
 import winKernel
 import wave
@@ -39,6 +45,7 @@ import config
 from logHandler import log
 import os.path
 import extensionPoints
+import NVDAHelper
 
 
 __all__ = (
@@ -140,7 +147,7 @@ def _isDebugForNvWave():
 	return config.conf["debugLog"]["nvwave"]
 
 
-class WavePlayer(garbageHandler.TrackedObject):
+class WinmmWavePlayer(garbageHandler.TrackedObject):
 	"""Synchronously play a stream of audio.
 	To use, construct an instance and feed it waveform audio using L{feed}.
 	Keeps device open until it is either not available, or WavePlayer is explicitly closed / deleted.
@@ -337,7 +344,8 @@ class WavePlayer(garbageHandler.TrackedObject):
 
 	def feed(
 			self,
-			data: bytes,
+			data: typing.Union[bytes, c_void_p],
+			size: typing.Optional[int] = None,
 			onDone: typing.Optional[typing.Callable] = None
 	) -> None:
 		"""Feed a chunk of audio data to be played.
@@ -348,9 +356,13 @@ class WavePlayer(garbageHandler.TrackedObject):
 		This allows for uninterrupted playback as long as a new chunk is fed before
 		the previous chunk has finished playing.
 		@param data: Waveform audio in the format specified when this instance was constructed.
+		@param size: The size of the data in bytes if data is a ctypes pointer.
+			If data is a Python bytes object, size should be None.
 		@param onDone: Function to call when this chunk has finished playing.
 		@raise WindowsError: If there was an error playing the audio.
 		"""
+		if size is not None:
+			data = string_at(data, size)
 		if not self._minBufferSize:
 			self._feedUnbuffered_handleErrors(data, onDone=onDone)
 			return
@@ -583,6 +595,8 @@ class WavePlayer(garbageHandler.TrackedObject):
 			return False
 
 
+WavePlayer = WinmmWavePlayer
+
 def _getOutputDevices():
 	"""Generator, returning device ID and device Name in device ID order.
 		@note: Depending on number of devices being fetched, this may take some time (~3ms)
@@ -677,17 +691,21 @@ def playWaveFile(
 		outputDevice=config.conf["speech"]["outputDevice"],
 		wantDucking=False
 	)
-	fileWavePlayer.feed(f.readframes(f.getnframes()))
+
+	def play():
+		fileWavePlayer.feed(f.readframes(f.getnframes()))
+		fileWavePlayer.idle()
+
 	if asynchronous:
 		if fileWavePlayerThread is not None:
 			fileWavePlayerThread.join()
 		fileWavePlayerThread = threading.Thread(
 			name=f"{__name__}.playWaveFile({os.path.basename(fileName)})",
-			target=fileWavePlayer.idle
+			target=play
 		)
 		fileWavePlayerThread.start()
 	else:
-		fileWavePlayer.idle()
+		play()
 
 # When exiting, ensure fileWavePlayer is deleted before modules get cleaned up.
 # Otherwise, WavePlayer.__del__ will fail with an exception.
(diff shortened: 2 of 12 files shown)


source/config/configSpec.py
```diff
@@ -56,6 +56,7 @@
 # Audio settings
 [audio]
 	audioDuckingMode = integer(default=0)
+	wasapi = boolean(default=false)
```
CyrilleB79, line 59: The "audio" section is saved in profiles. However it seems to me that this options is only relevant for the base profile since it takes effect after restart.

Would it be worth to put this option in another or a new section (e.g. global settings / startup settings)?

Note that there is probably already some confusion on this point regarding some advanced settings; I am specifically thinking to "Registration for UI Automation events and property changes" option; but there may be others in adv settings.

codeofdusk, line 59: NV Access will probably request a ternary setting in the config (auto/wasapi/wave out), even if it's only exposed as a checkbox in the UI during development. The "auto" setting will be treated like "wave out" until the default is changed, at which point "auto" becomes "wasapi".

leonardder:
@jcsteh or @seanbudd could you please provide a signed try build for this, so I can test on my coroporate Dell system?

jcsteh:
I actually have no idea what the policy or procedure is for generating signed try builds these days.

seanbudd:
@leonardder a try build is building here: https://ci.appveyor.com/project/NVAccess/nvda/builds/46590616/artifacts

leonardder:
Here are my first findings. Note that I have WASAPI enabled on secure screens as well.

1. Overall performance definitely feels a bit more rapid, especially when typing. Note that the try build doesn't have your change regarding caret movement detection, as far as I know.

2. When selecting an USB soundcard and disconnecting that card, the behavior is really snappy. It instantly switches back to the default card and switches back to the USB card when reconnecting it.

3. I restarted my system twice and both times, I had no speech on the logon screen. Synth was set to no speech and I had to choose it explicitly. I Have seen this with winm incidentally, but certainly not as prevalent as now.

jcsteh:
> I restarted my system twice and both times, I had no speech on the logon screen. Synth was set to no speech and I had to choose it explicitly.

Oof. That's going to be hellish to debug. I guess the Windows audio service or the audio device isn't initialised at that point? I don't really know what we're supposed to do in that case. I would've thought the Windows Accessibility manager thing wouldn't try to start a screen reader until audio was up, but I guess there's no such restriction.

jcsteh:
I did just discover and fix a bug where if you have a specific output device configured (not default, AKA Microsoft Sound Mapper) and you start NVDA without that device connected, nvwave would barf and end up setting the synth to no speech. I don't think that's the bug you're seeing, since I'm guessing you have the device set to default on secure screens, but still, it's an important fix.

zstanecic:
Hi,

Is there a try build for this?

codeofdusk:
I've been running this PR for several weeks. So far so good!

cary-rowen:
@zstanecic 
You can always get the latest build of this PR here:
https://ci.appveyor.com/project/NVAccess/nvda/builds/46595783/artifacts

seanbudd:
A try build with the latest commit has been built here: https://ci.appveyor.com/project/NVAccess/nvda/builds/46648246/artifacts

leonardder:
I tried this build and three out of three times, NVDA started talking on the logon screen. Furthermore, I actually don't see why WASAPI would be unable to find devices while WinM would be able to, since WinM is just a WASAPI client as you said.

I'd say, lets bring this to 2023.2 as is and then try to make it default in 2023.3!

jcsteh:
Is it possible you accidentally configured a specific (non-default) device on the logon screen? As per https://github.com/nvaccess/nvda/pull/14697#issuecomment-1482652541, I did have to fix a bug with that if said device was not connected.

source/config/configSpec.py
jcsteh: I agree this could be a point of confusion. That said, it seems overkill to create an entirely new config section for a setting which will eventually be removed. (I don't think it makes sense to keep both audio backends long-term.) It also doesn't fit so well in any of the existing global sections. Nevertheless, I'm of course happy to change this if the reviewer has a specific recommendation.

jcsteh: I'm not really clear on when we should be using a feature flag versus a boolean. Some advanced settings (even some added recently) are simple booleans, while others are feature flags. Again, I'm happy to implement whatever is recommended during review.

leonardder:
I have `outputDevice = Microsoft Sound Mapper`. in nvda.ini on the logon screen. Not sure whether you consider that default enough ;)

jcsteh:
That sounds pretty defaulty to me. No idea what went wrong then. :(

amirsol81:
My test results are also quite encouraging! Hopefully we can test it as part of an upcoming alpha release for better trialing.

Qchristensen (approved):
User Guide looks good.  There is an extra space betrween "This option" and "enables audio output".  Although I don't think that will make it through to the HTML anyway?

Qchristensen (approved):
Perfect, thanks Jamie!

source/nvwave.py
```diff
@@ -140,7 +149,29 @@ def _isDebugForNvWave():
...
+	session. The name is shown in the system Volume Mixer.
+	"""
+	guid: GUID
+	name: str
+
+
+#: The audio session to use by default.
+defaultSession = AudioSession(
+	GUID("{C302B781-00AF-4ECC-ACB7-7DF16AF7D55E}"),
+	"NVDA"
```
jcsteh, line 164: Ideally, we'd call this "NVDA speech", since that's mostly what it will get used for. However, I didn't do that because it's possible add-ons output non-speech sound without using playWaveFile or tones (or otherwise setting a session). If this isn't a "real" concern or if the benefits of calling it "NVDA speech" outweigh the potential losses, I can change this.

dpy013:
> @jcsteh or @seanbudd could you please provide a signed try build for this, so I can test on my HP Battle 66 Notebook?

jcsteh:
After discussion with @michaelDCurran, it has been decided to enable this by default to get maximum feedback on alpha.

dpy013:
> > @jcsteh or @seanbudd could you please provide a signed try build for this, so I can test on my HP Battle 66 Notebook?
Thank you for your work
The morning build is running great on my HP Battle 66 laptop.

seanbudd:
@dpy013 - there is no testable build available. The last build had significant test failures.

zstanecic:
Hi,

What’s up with this?

michaelDCurran (approved):

zstanecic:
I hope that build will not fail when this is going to be merged.

Now it fails for some ugly reason.

michaelDCurran (changes requested):
NVDA in this PR fails to run / install on Appveyor due to Appveyor having no audio devices.
```
CRITICAL - __main__ (06:18:57.381) - MainThread (1032):
core failure
  File "core.pyc", line 515, in main
  File "nvwave.pyc", line 982, in initialize
  File "nvwave.pyc", line 823, in __init__
  File "nvwave.pyc", line 850, in open
  File "_ctypes/callproc.c", line 935, in GetResult
OSError: [WinError -2147023728] Element not found
```

leonardder:
I think this can be reproduced locally by disabling the Windows Audio Service.

zstanecic:
Oops,continuous-integration/appveyor/pr — AppVeyor was unable to build non-mergeable pull request

zstanecic:
It really now works! I like how it's responsive.

michaelDCurran (approved):

ahicks92:
Late to this party and not able to easily review NVDA code anyway, but with respect to logon screens I'm pretty sure that WaveOut is implemented on top of Wasapi these days and would expect that Wasapi would be working if WaveOut is.

years and years ago when I maintained Unspoken it was using Wasapi on logon screens without issue for quite a while, until people got annoyed enough that it ended up going through NVWave.  But that was only because people wanted it to follow audio devices, which almost no one implements right on Wasapi (Windows 10 and later can do it automatically, but it was much harder in 2015 or so).  That said, if you guys find out you want to use a third-party implementation given all the Wasapi edge cases, miniaudio is public domain and I can attest to it's stability, as it has been used by System Fault (an audiogame) for a while.

Someone is going to need to test on weird channel configurations.  In particular, one place Wasapi likes to be fragile is around surround sound setups, requesting/using exclusive devices, etc etc.  Time will tell on that, I imagine.

Figured dropping some context might be useful given my background.

jcsteh:
@ahicks92, you noted that WASAPI in Windows 10 can auto follow audio devices. Do you have any references or pointers for that? I ended up having to implement that manually and it was annoying. It'd be nice if it could be done more cleanly, at least at some point in the future when we no longer support anything less than Windows 10.

ahicks92:
It's available in Windows 10 version 1607 and later.  See: https://learn.microsoft.com/en-us/windows/win32/coreaudio/automatic-stream-routing

Miniaudio does it via listening to events instead, which works on older versions of Windows, and I'd not be surprised if they know about some bug or something I don't.  Honestly Wasapi implementations are stupidly tricky and I regret having ever tried to write one myself (it didn't hurt that I was a worse programmer then, but even so...).  I don't know that Miniaudio's API is compatible with NVDA and obviously this work is already done, but reading their code mightn't be the worst idea: https://github.com/mackron/miniaudio

Admittedly I haven't heavily tested Miniaudio's device switching because it's something I haven't cared a lot about.  I've occasionally gotten reports of it not working, so idk if it's actually perfect.  One problem with Wasapi that comes up is that devices sometimes flat out lie.  Steel Series headsets for example say that their minimum period is 1 sample.  So in the general it's kind of hard to tell sometimes if a bug is because (1) your implementation is broken, (2) the device is buggy and WaveOut had some sort of patch over that, or (3) WaveOut is so conservative about what it does that it just papers over things.  Event-based switching being buggy vs. miniaudio being buggy vs. drivers that are doing bad things? You decide...

And if you don't know about the device events, I'm referring to these: https://learn.microsoft.com/en-us/windows/win32/coreaudio/device-events

cary-rowen:
Will this PR break API compatibility?
I have a problem with "tonysEnhancements" add-on, here is the log snippet:
```
  File "nvwave.pyc", line 718, in playWaveFile
  File "nvwave.pyc", line 823, in __init__
  File "C:\Users\cary\AppData\Roaming\nvda\addons\tonysEnhancements\globalPlugins\tonysEnhancements\__init__.py", line 618, in preWaveOpen
    winmm.waveOutSetVolume(selfself._waveout, volume2)
AttributeError: 'WasapiWavePlayer' object has no attribute '_waveout'
```

zstanecic:
Hi,

Tony should make adjustment to his code. He prefers win mm, instead of trying to check  interfaces.

cary-rowen:
Thanks I also created an Issue for Tony.

zstanecic:
But, when it comes to wasapi, it is a good move.

I confirm that there are more good sides, than bad ones of this move.

I am rugging now the latest alpha with the code, and RHVoice synthesizer.

jcsteh:
This change isn't API breaking in the sense that it doesn't change anything in the public API. However, it will certainly break anything that monkey patched or relied upon private implementation details. It looks like this add-on does that, so I'm not surprised it broke. It will need to be updated.

btman16:
Hello,

Running NVDA Alpha 28166, and using WASAPI, my system does not go to sleep automatically even though this is set in Power Options.

When using NVDA without WASAPI this works correctly.

For me, I rely on my system being able to automatically go into sleep regularly and other users might as well, so I wanted to bring this issue up early, as I could see there being some concern about this and it is a concern I have as well.

Other than this issue, however, WASAPI support is working well for me.
The issue also occurs with a fresh copy of NVDA, with no add-ons.

Thank you very much for your time and I look forward to any input you may have.

Sincerely,

Brandon Tyson

Lo-lo78:
Hi James,
surely you have already thought about how to do it.
Wasapi should be in shared mode, so you can use applications like Reaper.
Greetings!

Adriani90:
@josephsl it seems using WASAPI in combination with the sound splitter addon the sounds for switching to focus or browse mode are not working anymore. I guess this needs an update on the add-on's side.

josephsl:
Hi, any add-on that patches nvwave module contents must be edited to detect this change, and in case of Sound Splitter, try asking Luke Davis as I’m no longer maintaining Sound Splitter. Thanks.

jcsteh:
NVDA WASAPI is already using shared mode.

ruifontes:
Shouldn't a change like this only occur in version year.1?

It will impact a lot of add-ons...


Best regards,

Rui Fontes
NVDA portuguese team


Às 21:53 de 03/05/2023, Adriani90 escreveu:

jcsteh:
As per the Developer Guide:

> The NVDA Add-on API includes all NVDA internals, except symbols that are prefixed with an underscore.

By that definition, this change is not API breaking. It does not make changes to public (non-underscore prefixed) symbols. The fact that add-ons have chosen to touch private symbols (prefixed with an underscore) is their choice and they are thus subject to breakage outside of .1 releases.

That said, pragmatically speaking, I acknowledge this has an impact on users, not just add-on authors. So, I guess we need to decide whether the benefits of landing this sooner (responsiveness, volume adjustment, etc.) are outweighed by the risk of add-on breakage.

jcsteh:
If it is decided that add-on breakage is too much of a problem, an alternative to reverting this could be to disable it by default. That way, users who don't depend on these add-ons can still benefit if they choose to.

cary-rowen:
@jcsteh 
> This change isn't API breaking in the sense that it doesn't change anything in the public API. However, it will certainly break anything that monkey patched or relied upon private implementation details. It looks like this add-on does that, so I'm not surprised it broke. It will need to be updated.


I very much agree with what @jcsteh said, I think addon authors need to update these addons.

It is more important that the option for this PR should remain checked by default, so that more users can test it and find potential problems.

Thanks

XLTechie:
Yes, I'm hoping to have time next week to spend figuring out how to fix that, 
and any other side effects of WASAPI..

@Adriani90
> it seems using WASAPI in combination with the sound splitter addon the sounds for switching to focus or browse mode are not working anymore. I
> guess this needs an update on the add-on's side.

gregjozk:
Hello,

what about start sound of NVDA? when I tested it (yes, without any
addon), the start piano sound has not been played. Stop sound is
played when exiting NVDA. When I start NVDA, I hear only a woarning
sound.
however, thanks for your good work.

Jožef

2023-05-04 4:03 GMT+02.00, Luke Davis ***@***.***>:

MarcoZehe:
Must be a particular add-on that must be causing problems. I can definitely hear the NVDA starting sound when NVDA is starting.

seanbudd:
@btman16 @gregjozk -
This PR was merged last week into NVDA alpha.
Commenting on closed PRs and issues is lost easily, we cannot easily track comments in closed issues/PRs.

At this stage, issues with NVDA are alpha issues, not issues with this PR.
If you discover issues on alpha, please report them as a new issue.
For broader discussion on WASAPI and NVDA audio output, create a discussion or new issues.
I am locking the discussion on this PR.

PR #14701, opened 2023-03-07
Improve responsiveness of input and focus by pumping immediately instead of after a delay.

jcsteh:
### Link to issue number:
None.

### Summary of the issue:
Currently, everything in NVDA, including responding to keyboard input and focus changes, goes through the core pump. The core pump is always run with a slight delay. Assuming nothing else is already queued, any input or focus change will not be processed for a minimum of 10 ms. In reality, it's potentially longer due to the system timer resolution and the fact that timers are low priority messages. We should be aiming for the fastest possible response to user input and focus changes.

NVDA will respond slightly faster to commands and focus changes.

1. The core pump has been refactored. Rather than using NonReEntrantTimer, it now uses wx.Timer directly and guards re-entry itself for better control. There is a main thread method (request) which manages scheduling/executing the pump to make state management easier.
2. requestPump has a new immediate argument which specifies whether the pump should happen as soon as possible or after the delay. Delayed pumps are still useful in most cases for rate limiting and filtering, so this defaults to False.
3. queueHandler.queueFunction has an `_immediate` argument which is passed to requestPump. It is `_` prefixed to mitigate conflicts with keyword arguments intended for queued functions.
4. inputCore, scriptHandler, focus changes and live text output all request an immediate core pump.

### Testing strategy:
I used "time since input" logging to measure the differences in times before and after the change in various situations:

- Report time (NVDA+f12) and input help took ~10 ms before the patch, ~0 ms after.
- Navigating in the NVDA menu took > 20 ms before the change, ~10 ms after.
- Moving the caret in Firefox took ~35 ms before the change, ~20 ms after.

This can be quite variable, so it's only a rough guide, but the results are pretty suggestive.

None.

### Change log entries:
Bug fixes
NVDA now responds slightly faster to commands and focus changes.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


jcsteh:
I realise 10 ms is a pretty small difference, but even that potentially matters when it comes to feeling snappy, especially since there are other sources of delay we can't control (cross-process communication, application delays, minimum audio latency, speech synth processing, etc.). That said, given that we are talking about such a small delta, it's hard to be sure this makes a practical difference other than looking at the numbers.

burmancomp:
There is cumulative effect when thinking all users daily use.

michaelDCurran (approved):
This looks good to me.
Several system tests that generally don't fail, did fail on the first run, but then succeeded on the second. We'll want to keep an eye on this.

Adriani90:
I can confirm that this improves performance in MS Word significantly, I tested with a 164 pages document with many large tables and images and including converted text from through OCR which sometimes usually slowed down NVDA. Now it works really fluently even with UIA disabled. Still there is a noticeable difference when reporting pages in document formating settings is enabled or disabled.

PR #14743, opened 2023-03-25
Revert "Settings Ring: Lazily load available setting values the first time they're needed"

### LINKED ISSUE #14741: Error sounds heard when navigating up in the synth settings ring
### Steps to reproduce:
1. Switch to OneCore.
2. Navigate to the voice option with the synth settings ring, then move upwards until you reach the last voice.
### Actual behavior:

An error sound is heard when navigating up past the last voice. This doesn't seem to happen when navigating down.
### Expected behavior:

No error, last voice should be repeated like it was previously.
Log fragment: error executing script: <bound method GlobalCommands.script_increaseSynthSetting of <globalCommands.GlobalCommands object at 0x0813C710>> with gesture 'ctrl+shift+NVDA+up arrow'
  File "scriptHandler.pyc", line 295, in executeScript
  File "globalCommands.pyc", line 292, in script_increaseSynthSetting
  File "synthSettingsRing.pyc", line 122, in increase
  File "synthSettingsRing.pyc", line 19, in increase
  File "synthSettingsRing.pyc", line 55, in _set_value
IndexError: list index out of range
Installed
#### NVDA version:
alpha-27892,42dbabcc
#### Windows version:
Windows 11, build 25324.1000.
N/A
N/A
Yes.
Yes. This seems to have started occurring when #14704 was merged.
Yes.
Yes.

LeonarddeR:
When implementing #14704, I overlooked that initialising a StringSynthSetting actually requires the values to be known at initialisation time because the number of values is saved on the instance. So, a lazy init this way is simply not possible or we have to change things dramatically. Let's revert this to how it was before.

Reverts nvaccess/nvda#14704
fixes #14741

### DIFF
--- a/source/synthSettingsRing.py
+++ b/source/synthSettingsRing.py
@@ -37,12 +37,10 @@ class SynthSetting(baseObject.AutoPropertyObject):
 	def _get_reportValue(self):
 		return self._getReportValue(self.value)
 
-
 class StringSynthSetting(SynthSetting):
-
-	def _get__values(self):
-		self._values = list(getattr(self.synth, f"available{self.setting.id.capitalize()}s").values())
-		return self._values
+	def __init__(self,synth,setting):
+		self._values=list(getattr(synth,"available%ss"%setting.id.capitalize()).values())
+		super(StringSynthSetting,self).__init__(synth,setting,0,len(self._values)-1)
 
 	def _get_value(self):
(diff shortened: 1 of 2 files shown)


leonardder:
Ugh, I wasn't entirely awake it seems. There should be an easier method to fix this and keep the lazy loading behaviour.

PR #14806, opened 2023-04-07
Fixed a bug causing wrong deprecation warnings or errors.

CyrilleB79:
### Link to issue number:
None
Fix-up of #14350.
### Summary of the issue:

Running NVDA 2023.1 (and older), with add-ons and/or scratchpad enabled:

I get the following deprecation warnings:
* When opening start menu:
  ```
  WARNING - appModuleHandler._warnDeprecatedAliasAppModule (15:49:46.531) - MainThread (19688):
  Importing appModules.searchapp is deprecated, instead import appModules.searchui.
  ```
* When switching to Notepad++:
  ```
  WARNING - appModuleHandler._warnDeprecatedAliasAppModule (15:50:26.621) - MainThread (19688):
  Importing appModules.notepad++ is deprecated, instead import appModules.notepadPlusPlus.
  ```

And if I force errors on deprecated code by modifying `NVDAState._allowDeprecatedAPI` to return False, I get errors instead.

This is not expected since these warnings/errors come from NVDA's core code which should not (and actually does not) contain  deprecated code.

No more wrong deprecation warnings/errors due to NVDA's core appModules when opening start menu or switching to Notepad++

`appModules.__path__` contains the paths where NVDA has to search appModules. It contains first the folder in the scratchpad, then the folders in add-on containing appModules and finally the core appModule folder. Thus we need to look at the last element of `appModules.__path__`, not the first as it was done.

### Testing strategy:
* Test with add-ons and scratchpad that no deprecation warning nor error appear anymore.
* Tested that the following call in the console still causes the warning to be logged (upon first call):
`from appModules import searchapp`

Note: This was probably missed during #13366 test, maybe because it was tested without add-on nor scratchpad.

### Change log entries:
Bug fixes (or Changes for developers? Where should bugfixes impacting only dev be listed?)

`NVDA will no longer log inaccurate warnings or errors about deprecated
appModules. (#14806)`

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/appModuleHandler.py
+++ b/source/appModuleHandler.py
@@ -1,7 +1,7 @@
 # -*- coding: UTF-8 -*-
 # A part of NonVisual Desktop Access (NVDA)
 # Copyright (C) 2006-2023 NV Access Limited, Peter Vágner, Aleksey Sadovoy, Patrick Zajda, Joseph Lee,
-# Babbage B.V., Mozilla Corporation, Julien Cochuyt, Leonard de Ruijter
+# Babbage B.V., Mozilla Corporation, Julien Cochuyt, Leonard de Ruijter, Cyrille Bougot
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
 
@@ -48,7 +48,7 @@ from systemUtils import getCurrentProcessLogonSessionId, getProcessLogonSessionI
 
 # Dictionary of processID:appModule pairs used to hold the currently running modules
 runningTable: Dict[int, AppModule] = {}
-_CORE_APP_MODULES_PATH: os.PathLike = appModules.__path__[0]
+_CORE_APP_MODULES_PATH: os.PathLike = appModules.__path__[-1]
 _getAppModuleLock=threading.RLock()
 #: Notifies when another application is taking foreground.
 #: This allows components to react upon application switches.
--- a/source/appModuleHandler.py
+++ b/source/appModuleHandler.py
@@ -48,7 +48,7 @@ from systemUtils import getCurrentProcessLogonSessionId, getProcessLogonSessionI
 
 # Dictionary of processID:appModule pairs used to hold the currently running modules
 runningTable: Dict[int, AppModule] = {}
-_CORE_APP_MODULES_PATH: os.PathLike = appModules.__path__[-1]
+_CORE_APP_MODULES_PATH: os.PathLike = os.path.dirname(appModules.__spec__.origin)
 _getAppModuleLock=threading.RLock()
 #: Notifies when another application is taking foreground.
 #: This allows components to react upon application switches.
--- a/user_docs/en/changes.t2t
+++ b/user_docs/en/changes.t2t
@@ -84,6 +84,7 @@ Similarly, the new ``getCompletionRoutine`` method allows you to convert a pytho
 - ``offsets.OffsetsTextInfo._get_boundingRects`` should now always return ``List[locationHelper.rectLTWH]`` as expected for a subclass of ``textInfos.TextInfo``. (#12424)
 - ``highlight-color`` is now a format field attribute. (#14610)
 - NVDA should more accurately determine if a logged message is coming from NVDA core. (#14812)
+- NVDA will no longer log inaccurate warnings or errors about deprecated appModules. (#14806)
 -
 
 === Deprecations ===


lukaszgo1 (changes requested):
Thanks for this PR @CyrilleB79 ! In fact this regressed in #14350  as follows:
- Before #14350 at the first import of `appModuleHandler` no additional packages were added to the path of `appModules`, so first index indeed represented the path of modules bundled in NVDA
- Now code in scratchpad and in add-ons is added to the path *before* `appModuleHandler` is first imported, which means that first item in the path represents either the code in the scratchpad, or the first loaded add-on with appModules.


Given that relying on the order of imports turned out to be fragile I'd suggest to remove `_CORE_APP_MODULES_PATH`, and just check if the module is a part of NVDA's core using `logHandler.isPathExternalToNVDA`, which should be way more reliable. What do you think?
As an aside, I'm not sure if `logHandler` is the best place for `isPathExternalToNVDA`, but that is an issue for a different discussion perhaps.

CyrilleB79:
> Thanks for this PR @CyrilleB79 ! In fact this regressed in #14350 as follows:
> 
>     * Before [Streamline importing of addon directories for python packages #14350](https://github.com/nvaccess/nvda/pull/14350) at the first import of `appModuleHandler` no additional packages were added to the path of `appModules`, so first index indeed represented the path of modules bundled in NVDA
> 
>     * Now code in scratchpad and in add-ons is added to the path _before_ `appModuleHandler` is first imported, which means that first item in the path represents either the code in the scratchpad, or the first loaded add-on with appModules.


Thanks @lukaszgo1 for the thorough investigation and the details. I have updated the initial description to link with the correct PR.


> Given that relying on the order of imports turned out to be fragile I'd suggest to remove `_CORE_APP_MODULES_PATH`, and just check if the module is a part of NVDA's core using `logHandler.isPathExternalToNVDA`, which should be way more reliable. What do you think? As an aside, I'm not sure if `logHandler` is the best place for `isPathExternalToNVDA`, but that is an issue for a different discussion perhaps.

Well, I am not sure that this function is more robust. On my system the log messages emitted by core files are prefixed by "external" when run from source...
I just realize that on my system `sys.path[0]` is set to `'C:\\Users\\Cyrille\\AppData\\Roaming\\nvda\\addons\\nvdaChatGPT\\globalPlugins\\nvdaChatGPT\\site-packages'`. (Cc @mo29cg FYI)
There are various add-ons in the wild playing with `sys.path` and we can see that sometimes, they forget to restore it. Probably that's the source of this wrong prefix.

On the contrary appModule import order is much less likely to be modified and I do not expect add-ons modifying `appModules.__path__`.

Of course, a third solution (if any) retrieving the path of `appModules` folder more directly could still be an improvement.

XLTechie:
Suggested change to changelog wording:

> NVDA will not log anymore inappropriate warnings or errors for deprecated appModules.

NVDA will no longer log inaccurate warnings or errors about deprecated 
appModules.

Brian1Gaff:
Does this also hold true for add ons that give these messages? By which I 
mean if I see blah blag  command deprecated, use   wotsit instead, is that 
coming from nvda core code even though its generated in a line in an add on, 
or is this nvda actually saying this won't work in the next version of the 
compiler?
 Take a look at Sound schemes add on for some examples in the log.
 Brian

-- 
***@***.***
in the display name field.
----- Original Message ----- 
From: "Cyrille Bougot" ***@***.***>
Cc: "Subscribed" ***@***.***>
Sent: Friday, April 07, 2023 3:36 PM
Subject: [nvaccess/nvda] Fixed a bug causing wrong deprecation warnings or 
errors. (PR #14806)

CyrilleB79:
@XLTechie I had the feeling that my wording did not sound very well actually; many thanks for the English wording fix.

CyrilleB79:
> Does this also hold true for add ons that give these messages? By which I mean if I see blah blag command deprecated, use wotsit instead, is that coming from nvda core code even though its generated in a line in an add on, or is this nvda actually saying this won't work in the next version of the compiler? Take a look at Sound schemes add on for some examples in the log. Brian

Unless the add-on imports appModules from NVDA's core, it's unrelated. I do not know nor use Sound schemes add-on but IMO it's very unlikely that this PR changes anything.

lukaszgo1:
> Well, I am not sure that this function is more robust. On my system the log messages emitted by core files are prefixed by "external" when run from source... I just realize that on my system `sys.path[0]` is set to `'C:\\Users\\Cyrille\\AppData\\Roaming\\nvda\\addons\\nvdaChatGPT\\globalPlugins\\nvdaChatGPT\\site-packages'`. (Cc @mo29cg FYI) There are various add-ons in the wild playing with `sys.path` and we can see that sometimes, they forget to restore it. Probably that's the source of this wrong prefix.


That's pretty unfortunate - I've assumed that even if someone appends to `sys.path` they would clean up afterwards, but there is very little we can do about this. I'll look into how we  can make `logHandler.isPathExternalToNVDA` more resilient against changes like this.

> 
> Of course, a third solution (if any) retrieving the path of `appModules` folder more directly could still be an improvement.


This can be easily done with `os.path.dirname(appModules.__spec__.origin)` (this works well for binary copies, but please test the source copy if you'll decide to implement this).

lukaszgo1:
> > Well, I am not sure that this function is more robust. On my system the log messages emitted by core files are prefixed by "external" when run from source... I just realize that on my system `sys.path[0]` is set to `'C:\\Users\\Cyrille\\AppData\\Roaming\\nvda\\addons\\nvdaChatGPT\\globalPlugins\\nvdaChatGPT\\site-packages'`. (Cc @mo29cg FYI) There are various add-ons in the wild playing with `sys.path` and we can see that sometimes, they forget to restore it. Probably that's the source of this wrong prefix.
> 
> That's pretty unfortunate - I've assumed that even if someone appends to `sys.path` they would clean up afterwards, but there is very little we can do about this. I'll look into how we can make `logHandler.isPathExternalToNVDA` more resilient against changes like this.

This is now implemented in #14812

seanbudd:
Labelling this as blocked by #14812  and marking as draft

lukaszgo1:
I'm not sure why this is blocked by #14812  unless the solution for retrieving real path of core app modules, which I've proposed in [this comment](https://github.com/nvaccess/nvda/pull/14806#issuecomment-1500865437) is considered inappropriate for some reason.

seanbudd:
Apologies @lukaszgo1 I misunderstood your comment.
Am I wrong in thinking that this should still be marked as a draft until a safer solution is implemented?

CyrilleB79:
Thanks @lukaszgo1, I have modified the code with your suggestion for more robustness. Your suggestion gives the path on installed version as well as on a version running from source.

@seanbudd this PR is ready again. I have tested again as described in the initial description.

seanbudd (approved):

seanbudd:
Thanks @CyrilleB79

PR #14816, opened 2023-04-10
Fix open of Contributors and License Help menu options on Windows 11 

### LINKED ISSUE #14725: Contributors file do not open
### Steps to reproduce:
1 - Open NVDA menu;
2 - Go to Help submenu;
3 - Choose Contributors and press Enter.

### Actual behavior:
Is displayed a message informing that "Bloco de Notas", the portuguese name of notepad, is not a valid file name
or an error stating that file C:\SESSION:BnVnonDQhUCUNfDwgxtw+QE6QzpcUHJvZ3JhbSBGaWxlcyAoeDg2KVxOVkRBXGRvY3VtZW50YXRpb25cY29udHJpYnV0b3JzLnR4dAAAAQUAAAAB3FvMAgAAAAAAAA==.txt was not found.

### Expected behavior:
The contributors.txt should be open in notepad.exe

The only reference to notepad is the same that happen when opening the notepad app and is:
DEBUGWARNING - RPC process 300 (Notepad.exe) (16:19:29.306) - Dummy-6716 (11240):
Thread 12908, build\x86_64\remote\ia2Support.cpp, IA2Support_inProcess_initialize, 169:
Not installing IA2 support as  process is suspendable

Only happens in installed versions.

#### NVDA version:
Only tested with NVDA 2022.4, 2023.1 RC and last alpha-27830,d2d4bfce (2023.2.0.27830)

#### Windows version:
It seems that only occurs on Windows 11 22h2 build greater 22621

NA
NA
Yes.

NA
Yes.

Yes.

XLTechie:
### Link to issue number:

Fixes #14725 

### Summary of the issue:

Under Windows 11, attempting to open the NVDA Help menu items "Contributors" or "License", results in Notepad trying to open nonsense filenames while using installed, normal user level, copies of NVDA

Through testing, this has been narrowed to a bug in `os.startfile` in Python 3.7 as invoked on Windows 11, though the actual bug may be in a lower level library. The bug does not seem to occur when using the commandline "`start`" command.
It also does not occur in portable or admin level copies of NVDA.

Restores the ability to open the Contributors List or License file from the NVDA Help menu.

Through testing, I and @britechguy were able to narrow the source of the errors, and determine that attempting to open the file for editing, rather than by trusting the Windows open file association for .txt files, would open these files in Notepad correctly.

Taken from the text of the commit:
* Added the `_displayTextFileWorkaround` function to the `systemUtils` module.
    - Called it wherever `os.startfile` was opening a .txt file in `gui/__init__.py`.
    - The function uses `os.startfile`'s "`edit`" operation to open the text file, which still uses Notepad, at least on most systems.

Additionally:
* gui.MainFrame was importing systemUtils late, in __init__, and onRunCOMRegistrationFixesCommand.
    - Moved to a single import in top section.

### Testing strategy:

* Manual testing with 2023.1's Python console, that calling `os.startfile` as described, opens these files correctly under Windows 11. While calling it the standard way from the Python console, results in nonsense filenames being delivered to Notepad, and the subsequent failure of the task.
* Using a try build of this PR, tested opening these help menu items on Windows 10, with expected success.
* Tested that installing a try build of this PR on Windows 11, could open the files from the menu, and could open a copy of `copying.txt` at the root of the system drive, using the Python console.
* A portable copy generated from this PR, placed in the root directory of a USB stick, could open its included text files from the Help menu, and could open a copy of `copying.txt` placed at the root of the USB drive.

Additionally, to confirm that consolidating and moving the imports of `systemUtils` didn't break anything, I tested that NVDA launched, and the COM Registration Fixing Tool still works, under Windows 10, which is sufficient.

Tested on:
* Windows 11: 22H2, build 22621.1555
* Windows 10: 22H2, build 19045.2846

* It is not necessary to use `operation="edit"` under Windows 10. Since this seemed (1) rather harmless, and (2) a low impact bug, I didn't think it was necessary to test for Windows version before deciding which call structure to use for the new function.
* It is conceivable, that opening for editing will have side effects on some systems. However, since it seems reasonable to presume that a user who has made editing text files open in some other program than Notepad is quite experienced, I thought the negative potential here was rather low to non-existent.
* When we migrate to a newer Python version, the necessity of this workaround should be reevaluated, as hopefully this will be fixed by a newer library.

### Change log entries:
Bug fixes
In Windows 11, it is (once again) possible to open the Contributors and License items on the NVDA Help menu.

### Code Review Checklist:
  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [X] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/__init__.py
+++ b/source/gui/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: UTF-8 -*-
 # A part of NonVisual Desktop Access (NVDA)
 # Copyright (C) 2006-2023 NV Access Limited, Peter Vágner, Aleksey Sadovoy, Mesar Hameed, Joseph Lee,
-# Thomas Stivers, Babbage B.V., Accessolutions, Julien Cochuyt, Cyrille Bougot
+# Thomas Stivers, Babbage B.V., Accessolutions, Julien Cochuyt, Cyrille Bougot, Luke Davis
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
 
@@ -11,6 +11,7 @@ import threading
 import ctypes
 import wx
 import wx.adv
+
 import globalVars
 import tones
 import ui
@@ -21,6 +22,7 @@ import versionInfo
 import speech
 import queueHandler
 import core
+import systemUtils
 from .message import (
 	# messageBox is accessed through `gui.messageBox` as opposed to `gui.message.messageBox` throughout NVDA,
 	# be cautious when removing
@@ -87,7 +89,6 @@ class MainFrame(wx.Frame):
 		#: @type: list of L{NVDAObject}
 		self.prevFocusAncestors = None
 		# If NVDA has the uiAccess privilege, it can always set the foreground window.
-		import systemUtils
 		if not systemUtils.hasUiAccess():
 			# This makes Windows return to the previous foreground window and also seems to allow NVDA to be brought to the foreground.
 			self.Show()
@@ -371,7 +372,6 @@ class MainFrame(wx.Frame):
 			_("Please wait while NVDA tries to fix your system's COM registrations.")
 		)
 		try:
-			import systemUtils
 			systemUtils.execElevated(config.SLAVE_FILENAME, ["fixCOMRegistrations"])
 		except:
 			log.error("Could not execute fixCOMRegistrations command",exc_info=True) 
@@ -484,10 +484,18 @@ class SysTrayIcon(wx.adv.TaskBarIcon):
 			self.Bind(wx.EVT_MENU, lambda evt: os.startfile("http://www.nvda-project.org/"), item)
 			# Translators: The label for the menu item to view NVDA License document.
 			item = menu_help.Append(wx.ID_ANY, _("L&icense"))
-			self.Bind(wx.EVT_MENU, lambda evt: os.startfile(getDocFilePath("copying.txt", False)), item)
+			self.Bind(
+				wx.EVT_MENU,
+				lambda evt: systemUtils._displayTextFileWorkaround(getDocFilePath("copying.txt", False)),
+				item
+			)
 			# Translators: The label for the menu item to view NVDA Contributors list document.
 			item = menu_help.Append(wx.ID_ANY, _("C&ontributors"))
-			self.Bind(wx.EVT_MENU, lambda evt: os.startfile(getDocFilePath("contributors.txt", False)), item)
+			self.Bind(
+				wx.EVT_MENU,
+				lambda evt: systemUtils._displayTextFileWorkaround(getDocFilePath("contributors.txt", False)),
+				item
+			)
 			# Translators: The label for the menu item to open NVDA Welcome Dialog.
 			item = menu_help.Append(wx.ID_ANY, _("We&lcome dialog..."))
 			self.Bind(wx.EVT_MENU, lambda evt: WelcomeDialog.run(), item)
--- a/source/systemUtils.py
+++ b/source/systemUtils.py
@@ -1,6 +1,6 @@
 # -*- coding: UTF-8 -*-
 # A part of NonVisual Desktop Access (NVDA)
-# Copyright (C) 2020-2022 NV Access Limited, Łukasz Golonka
+# Copyright (C) 2020-2023 NV Access Limited, Łukasz Golonka, Luke Davis
 # This file may be used under the terms of the GNU General Public License, version 2 or later.
 # For more details see: https://www.gnu.org/licenses/gpl-2.0.html
 
@@ -17,6 +17,7 @@ import shellapi
 import winUser
 import functools
 import shlobj
+from os import startfile
 
 
 @functools.lru_cache(maxsize=1)
@@ -179,3 +180,12 @@ def _isSecureDesktop() -> bool:
 	 - SecureMode and SecureScreens
 	"""
 	return _getDesktopName() == "Winlogon"
+
+
+def _displayTextFileWorkaround(file: str) -> None:
+	# os.startfile does not currently (NVDA 2023.1, Python 3.7) work reliably to open .txt files in Notepadunder
+	# Windows 11, if relying on the default behavior (i.e. `operation="open"`). (#14725)
+	# Using `operation="edit"`, however, has the desired effect--opening the text file in Notepad. (#14816)
+	# Since this may be a bug in Python 3.7's os.startfile, or the underlying Win32 function, it may be
+	# possible to deprecate this workaround after a Python upgrade.
+	startfile(file, operation="edit")
--- a/source/systemUtils.py
+++ b/source/systemUtils.py
@@ -183,7 +183,7 @@ def _isSecureDesktop() -> bool:
 
 
 def _displayTextFileWorkaround(file: str) -> None:
-	# os.startfile does not currently (NVDA 2023.1, Python 3.7) work reliably to open .txt files in Notepadunder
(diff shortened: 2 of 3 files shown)


XLTechie:
If @britechguy or someone else with Windows 11 could test [this try build](https://ci.appveyor.com/api/buildjobs/0clweo9ma4r46i7v/artifacts/output%2Fnvda_snapshot_pr14816-28058%2Cb5633854.exe), which you must install rather than running as temporary or portable, it would be appreciated.

source/gui/__init__.py
```diff
@@ -484,10 +484,22 @@ def __init__(self, frame: MainFrame):
 			self.Bind(wx.EVT_MENU, lambda evt: os.startfile("http://www.nvda-project.org/"), item)
 			# Translators: The label for the menu item to view NVDA License document.
 			item = menu_help.Append(wx.ID_ANY, _("L&icense"))
-			self.Bind(wx.EVT_MENU, lambda evt: os.startfile(getDocFilePath("copying.txt", False)), item)
+			# Using operation="edit" to os.startfile below, to work around a Windows 11 related
+			# bug in Python 3.7's os.startfile. (#14725) Test reverting to original behavior after Python upgrade.
```
lukaszgo1, line 488: Can this comment and the logic responsible for opening text files be moved to a separate function in `systemUtils`? This ensures that both code and  the accompanying comment is not duplicated, and provides add-on developers with a tested way for opening text files under all versions of Windows.

Brian1Gaff:
Good idea, what are we saying then, that running it twice ruins the fix in 
older windows?
 Brian

-- 
***@***.***
in the display name field.
----- Original Message -----

source/gui/__init__.py
XLTechie: It certainly can be. However I didn't do that, though I considered it, for the following reasons:

1. This is hopefully a temporary condition that will be resolved by upgrading Python later this year. If so, the need for such a specific function will go away, and we will revert this. If instead we create a dedicated function for text files only, we'll likely be stuck with it for a long time, even if internally we change its behavior to be identical to `os.startfile`'s intended use once this bug is resolved.
2. If add-ons start using such a function assuming it is "the preferred way", which opens for editing as a workaround instead of opening for opening, it increases the chance of unexpected behavior. Better for add-on devs to call the edit operation directly, so they know their doing something slightly unusual, even if likely harmless, and can stop doing it when we revert this. It draws attention to the fact that this **is a workaround**.
3. Opening all other file types will still be done with `os.startfile`. I question the value in fragmenting the file opening modalities.

All that said, I will follow whatever @seanbudd  or @michaelDCurran  think is best practice.

seanbudd: > If instead we create a dedicated function for text files only, we'll likely be stuck with it for a long time, even if internally we change its behavior to be identical to os.startfile's intended use once this bug is resolved.

I don't think we would be stuck with it, I would encourage using an underscored function for this, so we can remove it when `os.startfile` is fixed. We would also not be upgrading python outside of an API changing release, so being stuck with a helper function is a non-issue.
Using a private function also prevents the issue raised with add-on misuse.

I have been unable to find a bug reported against Python or Windows related to this. 
Without an external bug to track, I don't think we can assume this will be fixed in a future release.
A helper function makes tracking and using this workaround easier in the future.

XLTechie: @seanbudd I was mainly thinking in terms of the public function @lukaszgo1 was asking for, so that add-on devs could use it. A private function makes more sense to me than a public one does.

XLTechie:
@seanbudd PR description updated; new function created; a couple little unrelated odds and ends fixed. Ready for re-review.

XLTechie:
@lukaszgo1 Regarding your request to put this in a public function so that add-ons can use it:

It turns out, according to [this mailing list post](https://nvda.groups.io/g/nvda/message/105816), that it would not have worked anyway. Evidently, the `operation="edit"` fix, only works for files under "`Program Files`" directories.

Brian1Gaff:
This seems strange to me. If a routine is being called one might expect it 
to be used by all such file open attempts using the same function, unless 
there is another, similar one we do not know about, or course.
 Brian

-- 
***@***.***
in the display name field.
----- Original Message ----- 
From: "Luke Davis" ***@***.***>
Cc: "Brian Gaff" ***@***.***>; "Comment" 
***@***.***>
Sent: Tuesday, April 11, 2023 9:06 AM
Subject: Re: [nvaccess/nvda] Fix open of Contributors and License Help menu 
options on Windows 11 (PR #14816)

source/systemUtils.py
```diff
@@ -89,8 +90,7 @@ class TokenOrigin(ctypes.Structure):
 
 
 def getProcessLogonSessionId(processHandle: int) -> int:
-	"""
-	Retrieves the ID of the logon session that created the process that the given processHandle belongs to.
+	"""Retrieves the ID of the logon session that created the process that the given processHandle belongs to.
```
seanbudd, line 93: this diff should probably be removed

source/systemUtils.py
```diff
@@ -166,9 +166,7 @@ def _getDesktopName() -> str:
 
 
 def _isSecureDesktop() -> bool:
-	"""
-	When NVDA is running on a secure screen,
-	it is running on the secure desktop.
+	"""	When NVDA is running on a secure screen, it is running on the secure desktop.
```
seanbudd, line 169: Fix lint
```suggestion
	"""
	When NVDA is running on a secure screen, it is running on the secure desktop.
```

seanbudd (commented):
I've noticed more discussion in the [mailing list thread](https://nvda.groups.io/g/nvda/topic/new_version_of_tesseractocr/98159402)

Does this fix still work correctly? What if NVDA is running as a portable copy outside program files?

>Copy file "copying.txt" to "c:\"
>and in the console, execute the command "os.startfile(r"C:\copying.txt", operation="edit").
>The error occurs.
>Now copy the same file to "c:\Program files (x86)" and run the command:
os.startfile(r"C:\Program Files (x86)\copying.txt", operation="edit").

source/systemUtils.py
XLTechie: Can I ask what the lint error is in these?

My understanding was that we preferred the opening line of docstring on same line as docstring quotes method, which is listed as optional for [Multi-line Docstrings in PEP-257](https://peps.python.org/pep-0257/#multi-line-docstrings). We have nothing in tests/lint/readme.md either way, though I thought we used to have it written up somewhere that this was the NVDA choice. That was why I changed those. I'm happy to change them back, but I would like to understand why I'm doing it.

Examples of that style in core: `gui.AddonsDialog._instance`, `inputCore.InputGesture`, `inputCore.InputGesture.reportExtra`. Just a few of many.

seanbudd: The lint issue is a tab prefacing the start of the string.
I think we are agnostic as to whether you start the text on the same line or not, but I think we more commonly do it on a new line.

XLTechie:
@seanbudd The tests you requested are pending. The original problem didn't exist for portable copies as far as anyone reported, so we hadn't tested that exhaustively.
@britechguy is running those tests and I will update the PR description and re-request review when he gets back to me. Also testing a rather unexpected usage of installed copies, just to be sure.

XLTechie:
@seanbudd GTG

seanbudd (approved):

seanbudd:
Thanks @XLTechie

PR #14878, opened 2023-04-27
NVDAObject.presentationType: groupings with position information but no name and no description should still be treated as content I.e. announced in focus ancestry.

michaelDCurran:
### Link to issue number:
None

### Summary of the issue:
NVDA filteres controls with a role of groupping out of being presented in the focus ancestry if they have no name or description. However, there is a trend with some newer Microsoft products to add position info to these unlabeled groupings (E.g. grouping 1 of 5) and it is requested that NvDA announce these as they provide very deliberate topology to the user.
 
NVDA will now report unlabeled groupings that have useful position information, such as in recent versions of Microsoft Office 365 menus.
 
NVDA.presentationType now treats groupings specifically, only changing them to layout if they have no name, no description, and no position information.

### Testing strategy:
* Open a recent build of Microsoft Word 365.
* Type a spelling mistake.
* Open the context menu on the mistake, and arrow up and down the menu. Notice that the suggestions are in "grouping 1 of 2" and the Add to Dictionary and Ignore items are in "grouping 2 of 2".
Note: Some builds of word 365 have started placing a whatespace character in the nameof unlabeld groupings to force NVDA to get this behaviour without this pr. but this was temporary. thus when testing this, strip whitespace from UIA NVDAObject's name property first.
 
### Change log entries:
New features
Changes
Bug fixes
NVDA will now report unlabeled groupings that have useful position information, such as in recent versions of Microsoft Office 365 menus.
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


seanbudd (approved):

codeofdusk:
Note that, when running with this PR, UWP WhatsApp now says "grouping 1 of 5" before the archived chats item and "grouping 2 of 5" before the first chat in the chat list, but there don't appear to be any sane third, forth, or fifth items.

rperez030:
I'm finding that this new way of announcing groupings is significantly slowing down navigation for me, particularly in Office menus. The problem is that, context menus in particular, contain a number of groups that only contain 1 element, so it becomes meaningless. I plan to raise the issue internally at Microsoft, but from the NVDA side, I wonder if the new behavior can be made configurable.

michaelDCurran:
Thanks for the feedback. Raising this internally with Microsoft would be 
useful, yes.
I don't think having an option would be useful here as I doubt anyone 
would be motivated enough to turn it on.

rperez030:
> I don't think having an option would be useful here as I doubt anyone would be motivated enough to turn it on.

Agree. Maybe we can set it to on by default, or we can label it as ignore groups that do not contain name or description so that users can turn it on using app profiles. In other words, set the new behavior as default, but allow users to optout for now. It is a fairly new approach in Office, and it may need some time to stabilize.

PR #11214, opened 2020-05-27
Selective UIA event handling

LeonarddeR:
### Link to issue number:
Closes #11209 
fixes #11077
Improves #11002
Follow up of #8742 

### Summary of the issue:
NVDA is listening globally to many UIA elements. This causes a major performance hit when many events are received (#11109)

### Description of how this pull request fixes the issue:
1. This pr introduces an easier method for event registration by means of EventHandlerGroups. This feature was introduced in IUIAutomation6. I wrote a FakeEventHandlerGroup class to mimic this behavior on older versions of UIA. In short, this allows us to request for automation/ notification and property events in one call instead of three on newer versions of UIA.
2. For some events, we still keep listening globally. However, for many of them, I switched to local event registration, with that I mean only to objects of interest. Currently, these are the focus object and its ancestors.
3. When changing focus, we register for events fired at the focus object and its ancestors. This includes all focus changes, including these for non native UIA elements. This includes focus changes for IAccessible objects as well.

### Testing performed:
* @codeofdusk confirmed that #11002 has improved by this
* I myself can confirm that the Nuget Package manager as well as other aspects of Visual Studio are much snappier.

### Known issues with pull request:
* @josephsl noted that announcements for modern input are somewhat broken, i.e. the announcement that the emoji panel is closed. I haven't yet looked into this. I'd like to have a basic review about the prototype and its performance improvements first.

### Change log entry:
* Bug fixes
    + Major performance improvements in Microsoft Visual Studio and other UIAutomation based applications. (#11077, #11209)

PR #14888, opened 2023-04-30
Avoid UIA event flooding in NvDA by  moving UIA event handler COM objects into c++

michaelDCurran:
### Summary of the issue:
If NVDA is flooded with UI Automation events, such as textChange events from Windows consoles, NVDA may seem to freeze, and or UIA focus tracking completely stops.

NVDA should remain responsive when being flooded with many UI Automation events, E.g. large amounts of text being written to a Windows console.
 
Implement a new UI Automation event handler COM object in c++ which avoids calling into Python, and instead stores the events, and requests NvDA's main thread to later wake and read off the events when it can. This COM object also coalesces duplicate automation and propertyChange events for the same element, so that NVDA is no longer flooded with duplicate events, E.g. textChange events in Windows consoles.
The path for a UI Automation event is now as follows:
* In an MTA thread, the new c++ UIA event handler COM object receives the event via its handle*Event method from UI Automation core.
* The event is stored in a list of events.
* If the event type supports coalescing (automation and propertyChange events), An existing duplicate event is looked up in a map, and if one exists, the existing event is removed from the list, and the map entry is then pointed to the newest event.
* If there was no existing event, a new entry is added to the map for this event.
* In other words: the events are stored in the list in order they were received, but any duplicate events are removed, leaving only the latest ones. Ecentually an ordered dictionary.
* If this event type does not support coalescing (E.g. focusChange event), NVDA is requested to wake and flush the event queue on its main thread as soon as it can.
* If this event does support coalescing, and this is the first event since the last flush, then NVDA is requested to wake and flush the event queue in around 30 ms, giving time for more events to possibly be received and coalesced.
* When NvDA wakes on its main thread, It requests the rate limited UIA event handler to flush its event list, emitting all the stored events out to our original Python UIA event handler COM object for normal handling.
These changes mean that UIA core in the MTA is never blocked waiting for the Python GIL, and from its perspective, event handling is instantanious. Thus UIA core should never feel the need to kill off events, including focus change events.

This pr also provides a couple of extra optimizations for UIA handling in Python:
- UIAHandler events: avoid needlessly creating an NVDAObject when the event is for the focus object.
- UIA event handlers: pass windowHandle into NvDAObject constructor if we already have it.
 
### Testing strategy:
* Ensure that NVDA is set to use UIA for Windows consoles, but diffing (not notification events).
* Open a cmd Windows console.
* In the NVDA repo, run the command: `git -P log -1000`. to majorly flood the console with text.
* Keep silencing speech with control until all text is written, probably about 15 seconds or so. And note that NVDA remains responsive, and focus change events are no longer missing.
 
UIA notification events are currently not coalesced in any way. Though I feel that Windows Terminal provides more 'responsible' rate of notification events, as compared to textchange events. If this is a problem, then perhaps we could coalesce on activityID, though I'm not sure if that is the same value for all console text events. Do you know / have an opinion @codeOfDusk?

### Change log entries:
New features
Changes
Bug fixes
- NVDA should remain responsive when being flooded with many UI Automation events, E.g. large amounts 
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


codeofdusk:
This is fantastic to see!

Yes, I believe the activity ID is the same for all notification events, but also CCing @carlos-zamora to be sure.

Another pain point with event flooding is Visual Studio (`devenv`), especially when navigating project directory structures and using the debugger. Hoping these optimizations will help there as well!

codeofdusk:
Also, just to clarify: the diffing/notifications setting only affects Windows Terminal (`wt.exe`), not Windows Console (`conhost.exe`), though in SV2+ `wt` is now the default terminal for most users.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
```diff
@@ -0,0 +1,227 @@
...
+#include <comutil.h>
+#include <common/log.h>
+#include "eventRecord.h"
+#include "rateLimitedEventHandler.h"
+
+template<EventRecordConstraints EventRecordClass, typename... EventRecordArgTypes>
+HRESULT RateLimitedEventHandler::queueEvent(EventRecordArgTypes&&... args) {
+	LOG_DEBUG(L"RateLimitedUIAEventHandler::queueEvent called");
+	bool needsFlush = false;
+	const unsigned int flushTimeMS = (EventRecordClass::isCoalesceable)?30:0;
```
jcsteh, line 34: The flush time is tricky. On one hand, we want rate limiting to be useful. On the other hand, this could impact perceived performance. In particular, property change events on the focus can be relevant to immediate user interaction; e.g. value changes on a focused volume slider. The response is likely to be slower than 30 ms because of timer resolution, additional time to queue the UIA event to NVDA core, etc.

I think the gains outweigh the losses here, but this is something we should at least consider for future optimisation. Perhaps we don't coalesce certain property changes on the focus or something like that?

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
```diff
@@ -0,0 +1,227 @@
...
+		auto& record = std::get<EventRecordClass>(recordVar);
+		if constexpr(EventRecordClass::isCoalesceable) {
+			LOG_DEBUG(L"RateLimitedUIAEventHandler::queueEvent: Is a coalesceable event");
+			auto coalescingKey = record.generateCoalescingKey();
+			auto existingKeyIter = m_eventRecordsByKey.find(coalescingKey);
+			if(existingKeyIter != m_eventRecordsByKey.end()) {
+				LOG_DEBUG(L"RateLimitedUIAEventHandler::queueEvent: found existing event with same key"); 
+				auto& [existingRecordVarIter,existingCoalesceCount] = existingKeyIter->second;
+				LOG_DEBUG(L"RateLimitedUIAEventHandler::queueEvent: updating key and count to "<<(existingCoalesceCount+1));
+				existingKeyIter->second = {recordVarIter, existingCoalesceCount + 1};
```
jcsteh, line 52: Are you absolutely certain these vector iterators will point at the same thing always? I would have thought that removing other entries (as can happen below) would invalidate any iterators after that position in the vector such that subsequent iterators would now point to the wrong entry or even be out of bounds.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
```diff
@@ -0,0 +1,227 @@
...
+		auto count = i.second.second;
+		if(count > 1) {
+			return acc + count;
+		}
+		return acc;
+	});
+	if(count > 0) {
+		Beep(440 + (count*10), 40);
+	}
+	*/
```
jcsteh, line 224: I assume this was for debugging to figure out how many events were coalesced? I guess this might be useful in future, but if you're going to keep this, please add a comment above it explaining what it does and what it's useful for.

nvdaHelper/local/UIAEventLimiter/utils.h
```diff
@@ -0,0 +1,33 @@
...
+#include <ranges>
+#include <variant>
+#include <uiautomation.h>
+#include <comutil.h>
+
+std::vector<int> SafeArrayToVector(SAFEARRAY* pSafeArray);
+
+std::vector<int> getRuntimeIDFromElement(IUIAutomationElement* pElement);
+
+template <typename T, typename V, std::size_t... indexes>
```
jcsteh, line 27: I think I understand what supports_alternative  is for, but given how cryptic some of this stuff appears at first glance, please add a brief comment explaining what this does.

source/UIAHandler/__init__.py
```diff
@@ -1376,3 +1405,14 @@ def terminate():
...
 	return config.conf["debugLog"]["UIA"]
+
+
+WM_NVDA_UIA_FLUSH = winUser.registerWindowMessage("WM_NVDA_UIA_FLUSH")
+
+
+def eventLimiterWindowProc(msg, wParam, lParam):
+	if msg == WM_NVDA_UIA_FLUSH:
+		if lParam == 0:
+			NVDAHelper.localLib.rateLimitedUIAEventHandler_flush(wParam)
```
jcsteh, line 1416: On one hand, doing this in the main thread simplifies things, since threading issues can be tricky to debug. On the other hand, this means we're now doing things like IsUIAWindow on the main thread instead of a background thread. I guess IAccessibleHandler was already doing this, but I wanted to check whether this is something you considered and your reasons for not believing it to be a concern.

jcsteh (changes requested):
Very nice overall!

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran: I think it would be worth then only doing coalescing for automation events, and not propertyChange events in this case. Not hard to change. Noting that the main motivation for this PR was textchange automation events. I think I should also add an extra variable to track if a flush message has been posted due to a non-coalesced event, and not post any more until the next flush. Currently if there are a whole bunch of non-coalesced events in a row, we'll keep flooding NVDA with pointless flush messages.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran: To clarify, they are list iterators, not vector iterators. According to https://en.cppreference.com/w/cpp/container/list
"Adding, removing and moving the elements within the list or across several lists does not invalidate the iterators or references. An iterator is invalidated only when the corresponding element is deleted."

source/UIAHandler/__init__.py
michaelDCurran: We could set up a window in a dedicated background thread. But if isUIAWindow did hang, processing of all UIA events would also pause, as they all went via that thread. So I don't think this is much better than it being on the main thread. It is still worth separately considering making isUIAWindow execute on a background thread, but  perhaps something a little bit more involved where we have a thread per UI thread, so that if the UI in one thread freezes, it does not block another one. I think that might be separate to this pr.

michaelDCurran:
@jcsteh  re your comments around documentation, I'll go through and document much of it shortly.

beqabeqa473:
A regression was found in this PR.

Try to open voice settings panel with NVDA+crtl+v
NVDA stops responding on my machine.
There is no such behaviour when running latest master branch

On 5/2/23, Michael Curran ***@***.***> wrote:
> @michaelDCurran commented on this pull request.
>
>
>
>> @@ -1376,3 +1405,14 @@ def terminate():
>
>  def _isDebug():
>  	return config.conf["debugLog"]["UIA"]
> +
> +
> +WM_NVDA_UIA_FLUSH = winUser.registerWindowMessage("WM_NVDA_UIA_FLUSH")
> +
> +
> +def eventLimiterWindowProc(msg, wParam, lParam):
> +	if msg == WM_NVDA_UIA_FLUSH:
> +		if lParam == 0:
> +			NVDAHelper.localLib.rateLimitedUIAEventHandler_flush(wParam)
>
> We could set up a window in a dedicated background thread. But if
> isUIAWindow did hang, processing of all UIA events would also pause, as they
> all went via that thread. So I don't think this is much better than it being
> on the main thread. It is still worth separately considering making
> isUIAWindow execute on a background thread, but  perhaps something a little
> bit more involved where we have a thread per UI thread, so that if the UI in
> one thread freezes, it does not block another one. I think that might be
> separate to this pr.
>
> --
> https://github.com/nvaccess/nvda/pull/14888#discussion_r1182095079
>
> Message ID: ***@***.***>


-- 
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473

beqabeqa473:
@michaelDCurran Overal performance of NVDA dialogs are very bad, this is shown in voice, braille panels, also in the category list

codeofdusk:
@beqabeqa473 Cannot repro this in a self-signed build

beqabeqa473:
I am running NVDA from source and can reproduce it every time i run it.
I am now trying to bysect it.

On 5/2/23, Bill Dengler ***@***.***> wrote:
> @beqabeqa473 Cannot repro this in a self-signed build
>
> --
> https://github.com/nvaccess/nvda/pull/14888#issuecomment-1530912903
>
> Message ID: ***@***.***>


-- 
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473

beqabeqa473:
So, yes, #de5332f119ac0b6673df0deb08a0e90ff55fd6d7 breaks NVDA settings dialog on my machine

michaelDCurran:
I can reproduce this with UIA registration set to global.
I'll look into this.
Possibly something about uIA events coming from NVDA's main thread which 
I have not considered clearly enough.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
jcsteh: Doing this only for automation events seems reasonable for this first PR if you prefer. However, I could see situations where property change event flooding could be a real issue. Task Manager is a good example. In that case, I think it does make sense to coalesce property change events except for the focus.

Yeah, the pointless flush messages thing is definitely worth fixing. I hadn't considered that.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
jcsteh: Oh, so they are. My bad. Still, I'm glad that's clear up for the record. :)

source/UIAHandler/__init__.py
jcsteh: There is probably lower hanging fruit with isUIAWindow anyway. For example, I wonder whether we should increase the cache time for UIA windows. It's also easier to manage hangs with the watchdog if they're on the main thread.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran: A couple of points:
1. PropertyChange events are only coalesced if their runtimeIDs match, and they are the same property (property ID). I.e. two name property changes on the same element are coalesced. I'm not too sure that this pr will really help task manager as I'm thinking its many properties across many list items.
2. Selective UIA registration is used in NVDA by default on Windows 11 now. thus there are hardly any property changes on elements other than the focus or its ancestors. I'd love to have this on by default in Windows 10 also, but apparently there are some bugs in Task list / emoji picker which stop us from doing that. @josephsl  @codeofdusk  ?
3. I've now added a commit that stops propertyChange events from being coalesced. and also a commit which stops the pointless flush messages.

codeofdusk:
@michaelDCurran commit b94f4a1 gives me a linker error:

```
nvdaHelperLocal.def : error LNK2001: unresolved external symbol rateLimitedUIAEventHandler_flush                        
build\x86\local\nvdaHelperLocal.lib : fatal error LNK1120: 1 unresolved externals
```

michaelDCurran:
Yep. Just forgot to commit that change.
Pushed now.

source/core.py
```diff
@@ -477,6 +477,9 @@ def _doLoseFocus():
 			log.exception("Lose focus error")
 
 
+messageWindow = None
```
leonardder, line 480: Is this still necessary?

codeofdusk:
@michaelDCurran I've been running this PR for the past couple of weeks without issues.

michaelDCurran:
@codeOfDusk great to know. But just checking: you do mean the latest 
commit on the pr right? as I did quite significantly rewrite it towards 
the end.

codeofdusk:
Yes – I've been pulling down updates.

bramd:
I am currently running a build with this PR as well as #14928 and #14924 since yesterday. Performance in Terminal has improved a lot and the review cursor doesn't "stick" to the line with caret anymore after receiving lots of text changes.

However, when I run something that generates lots of output such as a build on a Linux machine (think `./configure` and `make`) in the background and I browse in Edge, I notice some delay in the Edge UIA notifications such as "Loading". Previously these notifications were dropped completely, so this is certainly an improvement. However, it would be nice if UIA events that are not relevant for the foreground window could be discarded immediately without causing a delay.

That being said, this PR is definitely an improvement and I hope we can get this merged soon.
If I can test anything specific or provide certain logs I would be glad to do so.

codeofdusk:
@michaelDCurran I'm seeing one problem with this PR: caret tracking in UWP edit fields is very unreliable:

1. Open the start menu.
2. In the search field, type "abc".
3. press left arrow twice.

Expected: NVDA says "b".
Actual: NVDA says "c". It appears that the position is off-by-one.

bramd:
> @michaelDCurran I'm seeing one problem with this PR: caret tracking in UWP edit fields is very unreliable:
> 
> Expected: NVDA says "b". Actual: NVDA says "c". It appears that the position is off-by-one.


I can't reproduce that here, but as noted earlier in the thread I run a build that combines a few PRs. Running on Windows 11 22h2.

codeofdusk:
I especially notice this issue in UWP WhatsApp.

michaelDCurran:
Looks like in certain UIA implementations, uIA textSelectionChange event (which we map to caret) is fired before the underlying model reflects the new caret position when fetched. In other words, in these scenarios the caret event is useless as it is too early.
So far I am only seeing this in XAML implementations: Windows Mail's address fields, File Explorer address bar, Windows Settings search.
@codeofdusk  Is the problematic Watsapp text field also XAML (UIA framework ID of 'XAML')?
If so, then we can just set the caretMovementDetectionUsesEvents property to false on these NVDAObjects, and the problem is avoided. I'll do this in this pr if you can confirm  your issue is also with XAML.

codeofdusk:
@michaelDCurran Yes, the frameworkId is `xaml`.

cary-rowen:
Yes Unigram is also XAML

codeofdusk:
@michaelDCurran It appears that disabling `caretMovementDetectionUsesEvents` on `XAML` elements does indeed fix the issue.

source/core.py
michaelDCurran: No longer needed. I've reverted this.

nvdaHelper/local/UIAEventLimiter/eventRecord.h
```diff
@@ -0,0 +1,107 @@
...
+#include <atlcomcli.h>
+#include <UIAutomation.h>
+#include "utils.h"
+
+// The following structs are used to represent UI Automation event params.
+// A part from holding the params,
+// each struct also contains a method for generating a comparison key
+// which is used to detect and remove duplicate events.
+// The key is made up of the element's runtime ID,
+// plus any extra event params that make the vent unique,
```
seanbudd, line 28: Fix minor typos

```suggestion
// Apart from holding the params,
// each struct also contains a method for generating a comparison key
// which is used to detect and remove duplicate events.
// The key is made up of the element's runtime ID,
// plus any extra event params that make the event unique,
```

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2158,6 +2168,16 @@ def event_UIA_dropTargetEffect(self):
 			ui.message(dropTargetEffect)
 
 
+class Xaml(UIA):
+	""" a UIA element exposed by the XAML framework."""
+
+	# XAML fires UIA textSelectionChange events before the caret position change is reflected
+	# in the related UIA text pattern.
+	# This means NVDA cannot rely on textSelectionChange (caret) events in XAML
+	# to detect if the caret has moved, As it occurs too early.
```
seanbudd, line 2177: Fix typo (caps)

```suggestion
	# to detect if the caret has moved, as it occurs too early.
```

source/UIAHandler/__init__.py
```diff
@@ -526,17 +536,20 @@ def MTAThreadFunc(self):
 			else:
 				break
 		self.clientObject.RemoveAllEventHandlers()
+		del self.localEventHandlerGroup
+		del self.localEventHandlerGroupWithTextChanges
+		del self.globalEventHandlerGroup
 
-	def _registerGlobalEventHandlers(self):
-		self.clientObject.AddFocusChangedEventHandler(self.baseCacheRequest, self)
+	def _registerGlobalEventHandlers(self, handler):
```
seanbudd, line 543: could handler get a type annotation?

source/UIAHandler/__init__.py
```diff
@@ -563,24 +576,24 @@ def _registerGlobalEventHandlers(self):
...
 			self.globalEventHandlerGroup.AddActiveTextPositionChangedEventHandler(
 				UIA.TreeScope_Subtree,
 				self.baseCacheRequest,
-				self
+				handler
 			)
 		self.addEventHandlerGroup(self.rootElement, self.globalEventHandlerGroup)
 
-	def _createLocalEventHandlerGroup(self):
+	def _createLocalEventHandlerGroup(self, handler):
```
seanbudd, line 596: could handler get a type annotation?

source/UIAHandler/__init__.py
```diff
@@ -909,49 +929,56 @@ def IUIAutomationPropertyChangedEventHandler_HandlePropertyChangedEvent(self,sen
...
 		import NVDAObjects.UIA
 		if (
 			isinstance(focus, NVDAObjects.UIA.UIA)
 			and self.clientObject.compareElements(focus.UIAElement, sender)
 		):
 			if _isDebug():
-				log.debug("propertyChange event is for focus")
-			pass
+				log.debug(
+					"propertyChange event is for  focus. "
```
seanbudd, line 941: Fix spacing

```suggestion
					"propertyChange event is for focus. "
```

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.h
```diff
@@ -0,0 +1,85 @@
...
+#include <mutex>
+#include <uiautomation.h>
+#include "eventRecord.h"
+
+
+// @brief a class that listens for various UI Automation events, 
+// stores them in in an internal queue (removing any duplicates),
+// and sends them onto an existing UI Automation event handler in a separate thread.
+// This ensures that UI Automation core is never blocked while sending events to this class.
+class RateLimitedEventHandler: public IUIAutomationEventHandler, public IUIAutomationFocusChangedEventHandler, public IUIAutomationPropertyChangedEventHandler, public IUIAutomationNotificationEventHandler, public IUIAutomationActiveTextPositionChangedEventHandler {
```
seanbudd, line 27: this might be worth splitting into new lines for readability

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include <comutil.h>
+#include <common/log.h>
+#include "eventRecord.h"
+#include "rateLimitedEventHandler.h"
+
+template<EventRecordConstraints EventRecordClass, typename... EventRecordArgTypes>
+HRESULT RateLimitedEventHandler::queueEvent(EventRecordArgTypes&&... args) {
+	LOG_DEBUG(L"RateLimitedUIAEventHandler::queueEvent called");
+	bool needsFlush = false;
+	{ std::lock_guard lock(m_mtx);
```
seanbudd, line 34: should this be on a new line?

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// IUIAutomationPropertyChangedEventHandler method
+HRESULT STDMETHODCALLTYPE RateLimitedEventHandler::HandlePropertyChangedEvent(IUIAutomationElement* sender, PROPERTYID propertyID, VARIANT newValue) {
+	LOG_DEBUG(L"RateLimitedUIAEventHandler::HandlePropertyChangedEvent called");
+	return queueEvent<PropertyChangedEventRecord_t>(sender, propertyID, newValue);
+}
+
+// IUIAutomationNotificationEventHandler method
+HRESULT STDMETHODCALLTYPE RateLimitedEventHandler::HandleNotificationEvent(IUIAutomationElement* sender, NotificationKind notificationKind, NotificationProcessing notificationProcessing, BSTR displayString, BSTR activityID) {
```
seanbudd, line 216: why not take a `NotificationEventRecord_t` here? same with the other `RateLimitedEventHandler`, why not take their equivalent records as parameters?

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
```diff
@@ -0,0 +1,243 @@
...
+	return m_pExistingActiveTextPositionChangedEventHandler->HandleActiveTextPositionChangedEvent(record.sender, record.range);
+}
+
+RateLimitedEventHandler::~RateLimitedEventHandler() {
+	LOG_DEBUG(L"RateLimitedUIAEventHandler::~RateLimitedUIAEventHandler called");
+}
+
+RateLimitedEventHandler::RateLimitedEventHandler(IUnknown* pExistingHandler):
+	m_pExistingAutomationEventHandler(pExistingHandler), m_pExistingFocusChangedEventHandler(pExistingHandler), m_pExistingPropertyChangedEventHandler(pExistingHandler), m_pExistingNotificationEventHandler(pExistingHandler), m_pExistingActiveTextPositionChangedEventHandler(pExistingHandler),
+	m_flusherThread([this](std::stop_token st){ this->flusherThreadFunc(st); })
```
seanbudd, line 145: could this be linted? it's hard to visually parse

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran: I can... Though I personally like this style specifically for scoped locks as it makes it very clear the scope is for that purpose. If it is confusing we can add a new line.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran: Is this comment referring to the RateLimitedEventHandler constructor, or the m_flusherThread variable? Apologies, github is saying lines 143 to 145, but the comment appears directly below the m_flusherThread line... Could be an accessibility issue with the github interface.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
jcsteh: FWIW, here's another style I've seen to deal with this:

```
{ // Scope for lock
  std::lock_guard lock(m_mtx);
```

This makes it clear what the block is for without potentially making it look like a misplaced brace.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran: These methods are declared by existing UI Automation event handler interfaces. E.g. IUIAutomationNotificationEventHandler. I can't change what parameters they take.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.h
michaelDCurran: Like this?
```suggestion
class RateLimitedEventHandler:
	public IUIAutomationEventHandler,
	public IUIAutomationFocusChangedEventHandler,
	public IUIAutomationPropertyChangedEventHandler,
	public IUIAutomationNotificationEventHandler,
	public IUIAutomationActiveTextPositionChangedEventHandler
{
```

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
seanbudd: No - this is my mistake.
I was referring to line 144.

```suggestion
RateLimitedEventHandler::RateLimitedEventHandler(IUnknown* pExistingHandler):
	m_pExistingAutomationEventHandler(pExistingHandler),
	m_pExistingFocusChangedEventHandler(pExistingHandler), 
	m_pExistingPropertyChangedEventHandler(pExistingHandler),
	m_pExistingNotificationEventHandler(pExistingHandler), 
	m_pExistingActiveTextPositionChangedEventHandler(pExistingHandler),
	m_flusherThread([this](std::stop_token st){ this->flusherThreadFunc(st); })
```

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran: Yep, I'm personally happy with that style.

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.h
seanbudd: yes, thanks

nvdaHelper/local/UIAEventLimiter/rateLimitedEventHandler.cpp
michaelDCurran:
```suggestion
	{ // scoped lock
		std::lock_guard lock(m_mtx);
```

seanbudd (approved):

codeofdusk:
Going to test now with the latest commit, but as of b4d9f17250a5cdd6c82ec1624a0bbad5efced4c7 pressing backspace in XAML edit controls (such as UWP WhatsApp) does not announce deleted text.

CC @simon818.

codeofdusk:
@michaelDCurran Even with the latest commit, deletions with backspace are still not reported in these controls.

brunoprietog:
@codeofdusk This also happens to me sometimes and I solve it by restarting NVDA. I don't think this PR is the cause

codeofdusk:
On my system, I can always reproduce this issue with this PR and never without it. It is not transient.

michaelDCurran:
I can reproduce this as well.
Really not sure how to fix this at the moment.
Essentially UIA text range objects from XAML mutate with backspace, thus 
comparing a text range copied from before backspace with a text range 
fetched after backspace compare equal.
This sort of worked before as detecting caret events caused the 
_hasCaretMoved code to think the caret had moved.
We might need to come up with a hybrid approach where we don't use the 
event to break out of the loop, but, we do return true for hasCaretMoved 
if an event was detected.
But I'm not quite sure how to abstract this cleanly so that it can be 
only done for XAML.
So simply put, in hasCaretMoved, we would again check if a caret event 
is pending, but rather than breaking out of the loop, we would just set 
a caretEventDetected variable to True. Then when we return after the 
loop, we return caretEventDetected,newInfo, rather than False,newInfo.
I don't know how bad it would be to do this generically, if caret event 
detection is disabled.

leonardder:
> Essentially UIA text range objects from XAML mutate with backspace, thus comparing a text range copied from before backspace with a text range fetched after backspace compare equal. 

I assume you meant "don't mutate with backspace"?

I think I understand what you're aiming at, and as I had a bit of playing with this code in the past, I'm happy to give this a try.

jcsteh:
Wouldn't doing that just cause us to always wait the maximum time for backspacing in these fields, since we won't break out of the loop?

leonardder:
Note that there is a specific _backspaceScriptHelper function that deals with backspacing. I guess we can override that to temporarily enable caretMovementDetectionUsesEvents when backspacing so _hasCaretMoved will rely on events for backspacing only.

michaelDCurran:
It would yes.
As noted above in this pr, XAML fires caret events before the caret 
actually moves. So although we can perhaps use the event to detect "if" 
the caret moved, we can't use it to actually know "when" it moved.

leonardder:
Ah of course, so even if we'd rely on events for backspacing only, NVDA will likely still stay silent as the event might already fire before backspace deleted something.

michaelDCurran:
@codeofdusk   With Leonard's change is this pr now functioning okay for you?

codeofdusk:
It sure is! Thanks @leonardder.

seanbudd (approved):

codeofdusk:
@michaelDCurran This PR has severely broken caret movement in UWP OneNote making the app almost completely unusable.

michaelDCurran:
@codeofdusk would you mind providing clear steps to reproduce please? I 
am not sure Which Onenote you mean / how to get it?

The Onenote I have installed on my machine by default or through Office 
365 seems to work completely fine for me with this pr.

However, I also installed Onenote from the Microsoft Store, which seems 
to be called Onenote for Windows 10 in my start menu I think. That one 
hardly works for me at all (I can't seem to move the focus anywhere 
meaningful) with this pr or otherwise.

brunoprietog:
Anyway, Onenote for Windows 10 will be deprecated

<https://support.microsoft.com/en-us/office/what-is-happening-to-onenote-for-windows-10-2b453bfe-66bc-4ab2-9118-01e7eb54d2d6>

codeofdusk:
@michaelDCurran Yes, this is with OneNote for Windows 10.

The problems were immediately apparent for me when I just edited text in a page (create or browse to a notebook, then create or browse to a new page), but if you're having trouble I can try and write replicable steps to reproduce.

user_docs/en/changes.t2t
```diff
@@ -61,6 +61,7 @@ It is also updated with UIA enabled, when typing text and braille is tethered to
 - If the audio output device is set to something other than the default and that device becomes available again after being unavailable, NVDA will now switch back to the configured device instead of continuing to use the default device. (#15759, @jcsteh)
 - It is not possible anymore to overwrite NVDA's Python console history. (#15792, @CyrilleB79)
 - NVDA now resumes audio if the configuration of the output device changes or another application releases exclusive control of the device. (#15758, #15775, @jcsteh)
+- NVDA should remain responsive when being flooded with many UI Automation events, E.g. large amounts of text in a terminal. (#14888)
```
seanbudd, line 64:
```suggestion
- NVDA should remain responsive when being flooded with many UI Automation events, e.g. large amounts of text in a terminal. (#14888)
```

seanbudd (approved):

codeofdusk:
Given that a merge of this appears imminent and no substantial changes appear to have been made, could this please be put behind a feature flag (even if on by default) so it can be disabled if needed in some scenarios?

michaelDCurran:
@codeofdusk  as I mentioned previously, I am unable to reproduce the issue with 1note. This is a significant change to NvDA, which needs as much testing as possible, but unfortunately would be very tricky to be put behind a feature flag. I'm lead to believe by yourself and others that the benefits this pr brings (such as stability in terminals) is significant.
Of course if we get too close to a beta and there are still significant issues, we would revert the whole thing per our policy on these things.
But I do plan to merge this by the end of the week, unless I am able to reproduce the issue. Any clear steps you can provide would help greatly. Also, there has been a suggestion that this particular variant of 1Note may be discontinued. Is this your understanding?

beqabeqa473:
I vote for merging it as soon as possible, because it becomes really
hard to use nvda with UIA and benefits it could give are much larger
than one particular app.

On 11/22/23, Michael Curran ***@***.***> wrote:
> @codeofdusk  as I mentioned previously, I am unable to reproduce the issue
> with 1note. This is a significant change to NvDA, which needs as much
> testing as possible, but unfortunately would be very tricky to be put behind
> a feature flag. I'm lead to believe by yourself and others that the benefits
> this pr brings (such as stability in terminals) is significant.
> Of course if we get too close to a beta and there are still significant
> issues, we would revert the whole thing per our policy on these things.
> But I do plan to merge this by the end of the week, unless I am able to
> reproduce the issue. Any clear steps you can provide would help greatly.
> Also, there has been a suggestion that this particular variant of 1Note may
> be discontinued. Is this your understanding?
>
> --
> https://github.com/nvaccess/nvda/pull/14888#issuecomment-1822198705
>
> Message ID: ***@***.***>


-- 
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473

leonardder:
Merging it sooner than later will also allow more people to jump on the train of tracking down the underlying issues, if any.

brunoprietog:
The Onenote for Windows 10 UWP version will be discontinued in the near future, as I mentioned in the comment above. The link I posted is a Microsoft blog post where they announce it.

That said, I still feel it's a shame they are discontinuing it. The UWP version is much lighter and I prefer it over the other version, I feel it faster, responsive and simpler to use. but anyway NVDA has problems using it without this PR, for example with table navigation, where the commands don't work at all, and I think this is a problem that occurs in both Onenote anyway.

I'm also anxious to see this PR merged, but if there are a lot of people using Onenote for Windows 10 and the problem is real, I imagine you'll have to suggest them to move to the other version, which is included by default in the conventional Microsoft Office desktop installation. Otherwise, it could be quite annoying. For the same reason, personally, I don't think it's a bad idea to move this to the NVDA advanced settings so that users who wish to continue using Onenote for Windows 10 can do so until it loses support.

leonardder:
@michaelDCurran I see potential to make this an advanced setting, I can possibly implement that later today if you're interested.

michaelDCurran:
@leonardder  please feel free to have a go if you have the time. But I would ask that the feature be on by default, so that it gets the most testing possible.

leonardder:
There you go with #15814.

michaelDCurran:
@codeofdusk It is now possible to disable this code with a feature flag. Can you please confirm that when running NVDA with this feature flag disabled (remember to also restart NVDA) that 1Note works correctly for you?
Thank you to @leonardder  for working on the feature flag.

zstanecic:
Hi @michaelDCurran 
All works as you described:
Feature flag works, One note works in a described situation.

codeofdusk:
Interesting, I can no longer reproduce the UWP OneNote issues. Thanks @leonardder and @michaelDCurran for adding the flag though in case I run into it again. I'm going to leave it enabled for now and feel confident with this being merged!

zstanecic:
I have built this now from source and make sure that this works again.

Adriani90:
Now that this is merged, is #11214 for selectively register for UIA events still needed at all in NVDA?

PR #14940, opened 2023-05-21
Fix for the left and right double angle bracket readout, when the symbol level is set to none

### LINKED ISSUE #14939: Text reading with symbols: «тўлиқ«тўлиқ» left pointing angle bracket and right point angle bracket are read when symbol level is set to none
### Steps to reproduce:
1. find the text with the symbols described in the issue title, 
2. 2. just read it
3. example text: «тўлиқ»
### Actual behavior:

the symbols are read out, regardless of the symbol level. This symbol level should be overwritten by more appropriate one in the symbols.dic, because it now conflicts with cldr definition
### Expected behavior:

These symbols should not be read when the symbol level is set to none.
Not applicable
any NVDA version from the above mentioned
#### NVDA version:
Not applicable, issue is reproducible on all versions downwards and upwards
#### Windows version:
Not applicable, reproducible system wide.
Any text editor, web browser, and so on
Not applicable
yes
yes, Behaviour is the same.
Yes
Yes, and it is not applicable

zstanecic:
### Link to issue number:
Fixes #14939 
### Summary of the issue:
When the symbol level is set to none, left double angle bracket and right double angle bracket are read, which should not be the case.
The left double angle bracket and right double angle bracket symbols are no longer read, when the symbol level is set to none and speech synthesizers will properly react on these symbols.
This needed just the change of the english symbols.dic file. It should propagate all locales and languages. It overwrites erroneous definition in dldr.dic for all locales.
### Testing strategy:
Tested on the latest alpha with my local modified symbol files.
tried the following sentence with the ukrainian tts:
Процвітання. Слово «процвітати» дуже позитивне, оскільки уособлює собою зростання, розвиток та рух на краще.
None known.
### Change log entries:
New features
Changes
Bug fixes
For Developers
Bug fixes
### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/locale/en/symbols.dic
+++ b/source/locale/en/symbols.dic
@@ -124,8 +124,8 @@ _	line	most
 ◆	black diamond	some
 §	section	all
 °	degrees	some
-«	double left pointing angle bracket	none
-»	double right pointing angle bracket	none
+«	double left pointing angle bracket	most	always
+»	double right pointing angle bracket	most	always
 µ	micro	some
 ⁰	superscript 0	some
 ¹	superscript 1	some


CyrilleB79:
This PR changes two things: the symbol level and the preserve parameter (from never to always).

I understand the change of level; actually in French where these quoting marks are used, we have already raised the level to "most".

Regarding the "preserve" field changed from "never" to "always", why did you do it? If you have seen a change of behaviour due to this, please explain it in the test section indicating an example of sentence read and the synthesizer(s) and language(s) used to note this change of behaviour.

At last, you write:
"It overwrites erroneous definition in dldr.dic for all locales."
I do not know what this erroneous definition is. Anyway, this PR does not overrite CLDR since the symbols were already present in the English `symbols.dic` file and thus CLDR symbols were already overwritten before this PR.

zstanecic:
Hi Cyrille,

Let me explain these things here first.

Erroneous cldr thing is that: although in the default English symbols you had a none symbol level, and the same in cldr, it starts to pronounce these quote signs where it should not.

Regarding the speech synthesizers, the always parameter was included to send these symbols to synth, for the sake of the naturalness in reading, because some synthesizers know how to interpret these, and if not passed to them, even the user dictionary processing fails. This cannot be fixed in speech synthesizer itself, but in NVDA. Jaws doesn’t suffer from this problem.

To explain what it overwrites in cldr: 

It overwrites levels, which changes the first thing, it stops synthesizer from reading the symbol out in the text.

So, the first thing, it is the level raising what I want to standardize across all locales, to have the naturalness of the spoken text.

This change is very important as it fixes also the way in which speech synthesizer read the text. For example, RHVoice for Ukrainian starts to read properly accented text when this parameter is changed.

If we start to read left pointing angle bracket as the words, and then the text, and these closing brackets, we get actually misread text.

zstanecic:
Note, i have also updated the description with the synths used: the RHVoice is in question, and to add, maybe synthesizers that i don't know of.

CyrilleB79 (approved):
Thanks @zstanecic for this PR.

I have tested in French; these characters are actually interpreted by the synth "Microsoft Speech API version 5" and "Windows OneCore" (tested with Hortense). So I guess that these characters are interpreted in many languages by MS synths. So setting the "preserve" field to "always" definitely seems the good choice.
The choice of level "most" corresponds to English quotation marks (") so it is coherent.

seanbudd (approved):

PR #15054, opened 2023-06-23
Add-on store - Clarify window title and list label

### LINKED ISSUE #15042: Modify the add-on store title bar text
### Steps to reproduce:
* Open the add-on store
* Press NVDA+T
### Actual behavior:
NVDA reports:
"Add-on Store - All Installed add-ons"

### Expected behavior:
NVDA should report:
"Add-on Store - Installed add-ons (All)"

Indeed, the "Installed add-ons" information is more important than the "All" filter value.

Also, in English, by chance, "All Installed add-ons" sounds like "All installed add-ons" (note the difference in case). That means that concatenating a filter value and a tab name, by chance, allows to create a grammatically correct single expression (if you ignore the case). But in translations it does not sound so good as discussed in [this thread](https://groups.io/g/nvda-translations/topic/new_strings_for_store_addon/99566662?p=,,,100,0,0,0::recentpostdate/sticky,,,100,2,0,99566662,previd%3D1687330883456295628,nextid%3D1651299684347517459&previd=1687330883456295628&nextid=1651299684347517459) on the translators mailing list.

IMO it's safer (regarding potential translation issues) to concatenate unrelated values explicitly, i.e. adding parenthesis or dash in between, rather than trying to solve this problem with a translatable compound string such as `"{channel} {tabName}"`.

N/A
portable
#### NVDA version:
Version: alpha-28477,50ef7d38 (2023.2.0.28477)
#### Windows version:
10
N/A
N/A
Yes
N/A
N/A
N/A

CyrilleB79:
### Link to issue number:
Fixes #15042
### Summary of the issue:
The title of add-on store window is formatted such as "Add-on Store - All Installed add-ons"
After the dash, the string is a concatenation of the channel and the type of add-ons in the list.
* In English,it makes something almost grammatically correct to build a single expression (if we ignore case). But in other languages, this concatenation without any separator may produce stranger results that sound bad.
* Also, it has been noted that the type of add-on listed ("Available add-ons" in this example) is much more important than the channel; thus it should come first.

* Reworked the window title to create a title such as "Add-on Store - Installed add-ons (All)"
* While at it, also removed the channel in the list name as pointed in https://github.com/nvaccess/nvda/issues/15042#issuecomment-1603600419, that is, the list's name is now "Available add-ons" instead of "All Available add-ons" so that the list name is less verbose.

Implemented changes described below.

Note: I have used parenthesis for the channel, but I may have used a second dash as a separator instead, i.e. "Add-on store - Available add-ons - All". Let me know if you prefer this second version.
### Testing strategy:
Manually checked the window title and the list name.
### Change log entries:
Not needed on unreleased code
### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/_addonStoreGui/controls/storeDialog.py
+++ b/source/gui/_addonStoreGui/controls/storeDialog.py
@@ -274,11 +274,11 @@ class AddonStoreDialog(SettingsDialog):
 
 	@property
 	def _titleText(self) -> str:
-		return f"{self.title} - {self._listLabelText}"
+		return f"{self.title} - {self._statusFilterKey.displayString} ({self._channelFilterKey.displayString})"
 
 	@property
 	def _listLabelText(self) -> str:
-		return f"{self._channelFilterKey.displayString} {self._statusFilterKey.displayString}"
+		return f"{self._statusFilterKey.displayString}"
 
 	def _setListLabels(self):
 		self.listLabel.SetLabelText(self._listLabelText)


XLTechie:
I might prefer:

Available add-ons (showing all)

Available add-ons (showing stable)

Though if that is not liked, I am fine with how @CyrilleB79 has done it, and 
prefer the parenthetical expression to hyphenation.

CyrilleB79:
> I might prefer: Available add-ons (showing all) Available add-ons (showing stable) Though if that is not liked, I am fine with how @CyrilleB79 has done it, and prefer the parenthetical expression to hyphenation.

If we do not change the casing for channel names (that begin with a capital), we would rather have "Available add-ons (showing All)" with "All" beginning with a capital letter.

Personally however, I prefer the current wording "Available add-ons (All)" to @XLTechie's suggestion "Available add-ons (showing All)". Adding "showing" is just more verbose IMO.

For a similar GUI example in MS Outlook, when you open an e-mail, the window title is "This is the subject - Message (HTML)" and not something more verbose such as "This is the subject - Message (format is HTML)".

It's just my opinion though. And I will implement what NVAccess and the community prefer. Let's wait for other feedback if any.

cary-rowen:
@CyrilleB79 wrote:
> Personally however, I prefer the current wording "Available add-ons (All)" to  @XLTechie 's suggestion "Available add-ons (showing All)". Adding "showing" is just more verbose IMO.

i totally agree

seanbudd (approved):

PR #15129, opened 2023-07-11
Fix missing first SAPI5 utterance after silence

### LINKED ISSUE #15082: Losing speech when using SAPI5 if speech has not occurred for some time
### Steps to reproduce:
1. Run an NVDA snapshot that uses WASAPI (the latest Alpha exhibitted this behavior the last time I tried)
2. Switch to SAPI5 and use NVDA normally.
### Actual behavior:

NVDA should speak normally, but after waiting about 30 seconds to a minute without taking any action, I seem to lose speech for a few seconds, and it recovers when I continue to interact with the system. This only seems to happen when using WASAPI.
### Expected behavior:

Just like when using NVDA without WASAPI, the expected result is that no matter how much time has past since I've pressed a key that I wouldn't lose speech.
This was running NVDA from the launcher.
#### NVDA version:
This happened with alpha-28507,3f89b600, but I have not tried it with the snapshot released on June 30.
#### Windows version:
Windows 11 21H2 (10.0.22000). I am not sure if other Windows versions are affected, as I have not had a chance to extensively test it on my other systems not running 11.
The only other installed items of note are that I'm using some other SAPI5 voices, but I think this also would affect the Microsoft voices.
Yes.
NVDA 2023.1 does not exhibit this behavior as it does not support WASAPI. Additionally, with add-ons disabled the issue still shows up.
Yes.
Yes.

LeonarddeR:
### Link to issue number:
Fixes #15082 

### Summary of the issue:
The first utterance after around 10 seconds of silence with a SAPI5 synth is not spoken.

Issue does no longer occur.

Since #14759 we are modifying some window messages in our process. This includes setting the SMTO_ABORTIFHUNG flag. I was able to pinpoint that setting this flag to a message of type WM_APP caused this particular issue with SAPI5. Therefore I excluded this range of messages (WM_APP through 0xBFFF).

### Testing strategy:
Tested that NVDA with SAPI5 no longer stays silent when pressing down arrow on the desktop after more than 10 seconds of silence.

None known.

### Change log entries:
None needed.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/nvdaHelper/local/nvdaHelperLocal.cpp
+++ b/nvdaHelper/local/nvdaHelperLocal.cpp
@@ -97,8 +97,11 @@ LRESULT cancellableSendMessageTimeout(HWND hwnd, UINT Msg, WPARAM wParam, LPARAM
 		SetLastError(ERROR_CANCELLED);
 		return 0;
 	}
-
-	fuFlags |= SMTO_ABORTIFHUNG;
+	if (Msg < WM_APP || Msg > 0xBFFF) {
+		// Message in the range 0x8000 through 0xBFFF are available for applications to use as private messages.
+		// Setting the SMTO_ABORTIFHUNG for these messages is known to cause problems, for example with SAPI5 (#15082)
+		fuFlags |= SMTO_ABORTIFHUNG;
+	}
 	fuFlags &= ~SMTO_NOTIMEOUTIFNOTHUNG;
 
 	if (uTimeout > 10000) {


jcsteh (approved):
Bizarre. I guess Windows thinks the SAPI5 thread is hung even though it's still processing messages? How is that possible? 😕

I can't see any problem with this. At worst, we won't return immediately for messages in the WM_APP range if an app hangs, but they'll still be cancellable by watchdog.

michaelDCurran (approved):
Thanks :)

michaelDCurran:
I can confirm this is no longer a problem on my machine.

Issue #15115, opened 2023-07-10
braille: update display when tethered to review and review cursor does not follow caret


burmancomp:
### Is your feature request related to a problem? Please describe.
Sometimes it would be useful if braille display content would be updated automatically although braille is tethered to review and review cursor does not follow caret. This might help when using terminal software like Teraterm or Putty in non-graphical Linux server environment.
### Describe the solution you'd like
Advantage or disadvantage is at the moment that user action is needed to update display. Is it advantage or disadvantage depends on case. Maybe there should be a setting but this should be discussed.
### Describe alternatives you've considered
Slightly modifying handleCaretMove in review.py.
### Additional context
When review cursor does not follow caret, moving it to next/previous character may give strange results. However this should not depend on braille display updating so I suppose that it should be a separate issue. Solution for that might have something to do with wideToStrOffsets in textUtils.py.

burmancomp:
@leonardder do you have thoughts how this should be implemented?

burmancomp:
Because task is to update display automatically when current navigator object content changes, implementation would seem to be quite simple, especially if it is acceptable that current behavior changes.

A setting would be possible to keep current behavior but it would likely require much more code than updating braille line automatically. Generally I prefer to implement a setting because then user can decide but it is normal that display is updated, so in this case a setting sounds somewhat redundant or at least I am somewhat unsure.

To summarize: a setting or behavior change?

leonardder:
I'm afraid I don't understand the issue. Could you provide a real life scenario along with what you'd expect and what happens instead?

burmancomp:
> I'm afraid I don't understand the issue. Could you provide a real life scenario along with what you'd expect and what happens instead?

When you want to monitor what happens in given line without caret movement disturbing. Considering especially non-graphical user interface (like Linux server environment command line (shell like bash)) which can be used remotely with terminal software like Putty or Teraterm.

Screen is explored with review cursor. Braille tethered to review. When exploring screen it is useful that cursor movement does not move review position (review does not follow caret) but if review does not follow caret and braille is tethered to review then display content is not updated automatically.

For example there might be "dialog" with:

[ ] this is "check box" which could be checked with spacebar or sometimes with enter

But system cursor is somewhere - and when using arrow keys to change active item in dialog and if review follows caret, press of arrow key moves review location to cursor which is somewhere else. As said, if review does not follow caret, then display is not updated automatically. If above check box would be active and spacebar would change its state it would require extra step to detect state change.

[...]

PR #15134, opened 2023-07-12
Update braille when tethered to review but review cursor does not follow caret

burmancomp:
### Link to issue number:
15115
### Summary of the issue:
I think use case is mainly with terminal software like Putty or Teraterm when remotely managing Linux systems from command line using non-graphical user interface (typically shell like bash).

Caret is not always "real cursor", and following caret may be useless or harmful because it may lead to wrong location in window. Content of terminal window is often explored with review cursor.

When using review cursor, so that it does not follow caret, gives better possibilities to review window content because key presses or other things which trigger cursor movement do not affect on review position. However, although content which braille display is currently showing would change for example due to key press, braille display is not updated automatically. Automatic update would facilitate detection of changes and thus remote management.
When braille is explicitly tethered to review, and review cursor does not follow system caret, braille display content is updated if:

- focus object is navigator object and
- content which should be shown on braille display changes or updates due to caret movement.

At the moment I think a setting for this would be somewhat redundant. Currently there is no changes in user guide because it does not seem to define braille behavior when braille is tethered to review and review cursor does not follow system caret.
Implementation is in BrailleHandler.handleCaretMove function in braille.py.
### Testing strategy:
Tested with Putty and Albatross 80 model.
none
### Change log entries:
New features
Changes
When braille is explicitly tethered to review and review cursor does not follow system caret, braille display content is updated if:

- focus object is navigator object and
- content which should be shown on braille display changes or updates due to caret movement.
Bug fixes
For Developers

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/braille.py
+++ b/source/braille.py
@@ -2468,6 +2468,16 @@ class BrailleHandler(baseObject.AutoPropertyObject):
 		if shouldAutoTether:
 			self.setTether(TetherTo.FOCUS.value, auto=True)
 		if self._tether != TetherTo.FOCUS.value:
+			# However, braille display content is updated in case where:
+	# braille is tethered to review, review cursor does not follow system caret,
+			# and navigator object is focus object or its ancestor.
+			if (
+				not config.conf["reviewCursor"]["followCaret"]
+				and config.conf["braille"]["tetherTo"] == TetherTo.REVIEW.value
+			):
+				navigatorObject: NVDAObject = api.getNavigatorObject()
+				if navigatorObject == api.getFocusObject() or navigatorObject in api.getFocusAncestors():
+					self.handleUpdate(navigatorObject)
 			return
 		region = self.mainBuffer.regions[-1] if self.mainBuffer.regions else None
 		if region and region.obj==obj:
--- a/source/braille.py
+++ b/source/braille.py
@@ -2469,7 +2469,7 @@ class BrailleHandler(baseObject.AutoPropertyObject):
 			self.setTether(TetherTo.FOCUS.value, auto=True)
 		if self._tether != TetherTo.FOCUS.value:
 			# However, braille display content is updated in case where:
-	# braille is tethered to review, review cursor does not follow system caret,
+			# braille is tethered to review, review cursor does not follow system caret,
 			# and navigator object is focus object or its ancestor.
 			if (
 				not config.conf["reviewCursor"]["followCaret"]
--- a/source/braille.py
+++ b/source/braille.py
@@ -2468,16 +2468,12 @@ class BrailleHandler(baseObject.AutoPropertyObject):
 		if shouldAutoTether:
 			self.setTether(TetherTo.FOCUS.value, auto=True)
 		if self._tether != TetherTo.FOCUS.value:
-			# However, braille display content is updated in case where:
+			# Braille display content is updated in case where:
 			# braille is tethered to review, review cursor does not follow system caret,
-			# and navigator object is focus object or its ancestor.
-			if (
-				not config.conf["reviewCursor"]["followCaret"]
-				and config.conf["braille"]["tetherTo"] == TetherTo.REVIEW.value
-			):
-				navigatorObject: NVDAObject = api.getNavigatorObject()
-				if navigatorObject == api.getFocusObject() or navigatorObject in api.getFocusAncestors():
-					self.handleUpdate(navigatorObject)
+			# and focus object is navigator object.
+			if not config.conf["reviewCursor"]["followCaret"]:
+				if obj == api.getNavigatorObject():
+					self.handleUpdate(obj)
 			return
 		region = self.mainBuffer.regions[-1] if self.mainBuffer.regions else None
 		if region and region.obj==obj:
--- a/source/review.py
+++ b/source/review.py
@@ -1,5 +1,5 @@
 # A part of NonVisual Desktop Access (NVDA)
-# Copyright (C) 2013-2022 NV Access Limited
+# Copyright (C) 2013-2023 NV Access Limited, Burman's Computer and Education Ltd.
 # This file may be used under the terms of the GNU General Public License, version 2 or later.
 # For more details see: https://www.gnu.org/licenses/gpl-2.0.html
 
@@ -10,14 +10,19 @@ from typing import (
 
 import api
 from baseObject import ScriptableObject
+import braille
 import winUser
 from logHandler import log
 from NVDAObjects import NVDAObject, NVDAObjectTextInfo
 from NVDAObjects.window import Window
-from treeInterceptorHandler import DocumentTreeInterceptor
+from treeInterceptorHandler import (
+	DocumentTreeInterceptor,
+	TreeInterceptor,
+)
 from displayModel import DisplayModelTextInfo
 import textInfos
 import config
+from config.configFlags import TetherTo
 
 def getObjectPosition(obj):
 	"""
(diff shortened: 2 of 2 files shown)


leonardder:
I'm honestly not sure about this feature. I'd say that if someone has review follows system caret disabled and they yet want their display to update for caret updates, they should just enable review cursor follows system caret option at least temporarily.

burmancomp:
> I'm honestly not sure about this feature. I'd say that if someone has review follows system caret disabled and they yet want their display to update for caret updates, they should just enable review cursor follows system caret option at least temporarily.

This is useful when one wants to monitor changes in given line caused or updated by caret move which occurs somewhere else within navigator object or its descendants.

This solution has its lack because it depends on caret move but it is more than nothing.

source/braille.py
```diff
@@ -2468,6 +2468,16 @@ def handleCaretMove(
 		if shouldAutoTether:
 			self.setTether(TetherTo.FOCUS.value, auto=True)
 		if self._tether != TetherTo.FOCUS.value:
+			# However, braille display content is updated in case where:
```
leonardder, line 2471:
```suggestion
			# Braille display content is updated in case where:
```

source/braille.py
```diff
@@ -2468,6 +2468,16 @@ def handleCaretMove(
 		if shouldAutoTether:
 			self.setTether(TetherTo.FOCUS.value, auto=True)
 		if self._tether != TetherTo.FOCUS.value:
+			# However, braille display content is updated in case where:
+			# braille is tethered to review, review cursor does not follow system caret,
+			# and navigator object is focus object or its ancestor.
+			if (
+				not config.conf["reviewCursor"]["followCaret"]
+				and config.conf["braille"]["tetherTo"] == TetherTo.REVIEW.value
```
leonardder, line 2476: This one is not necessary.
The if check above will already ensure that tethering is set to review, i.e. it is not focus and when tethering is set to automatically, this block won't apply as tethering switched to focus above.
```suggestion
```

source/braille.py
```diff
@@ -2468,6 +2468,16 @@ def handleCaretMove(
 		if shouldAutoTether:
 			self.setTether(TetherTo.FOCUS.value, auto=True)
 		if self._tether != TetherTo.FOCUS.value:
+			# However, braille display content is updated in case where:
+			# braille is tethered to review, review cursor does not follow system caret,
+			# and navigator object is focus object or its ancestor.
```
leonardder, line 2473: I think it is enough to check whether obj is the navigator object.

source/braille.py
```diff
@@ -2468,6 +2468,16 @@ def handleCaretMove(
...
 		if self._tether != TetherTo.FOCUS.value:
+			# However, braille display content is updated in case where:
+			# braille is tethered to review, review cursor does not follow system caret,
+			# and navigator object is focus object or its ancestor.
+			if (
+				not config.conf["reviewCursor"]["followCaret"]
+				and config.conf["braille"]["tetherTo"] == TetherTo.REVIEW.value
+			):
+				navigatorObject: NVDAObject = api.getNavigatorObject()
+				if navigatorObject == api.getFocusObject() or navigatorObject in api.getFocusAncestors():
```
leonardder, line 2479:
```suggestion
				if navigatorObject == obj:
```

source/braille.py
```diff
@@ -2468,6 +2468,16 @@ def handleCaretMove(
...
+					self.handleUpdate(navigatorObject)
```
leonardder, line 2480:
```suggestion
					self.handleUpdate(obj)
```

LeonarddeR (changes requested):
Nice one. I don't think this needs a config option, indeed.
I think that after my suggestions are applied, the behavior will still be the same and the code is somewhat cleaner.cleaner.

burmancomp:
What do you think should there be something in user guide?

burmancomp:
@leonardder why logic could not be in braille module pumpAll function? Does it use too much resources there for example? I ask because then display would be updated also when there is no caret movement. Then for example, selecting check box "[ ]" with spacebar would be updated on braille display to "[x]" automatically although there is no cursor movement.

leonardder:
> @leonardder do we agree no setting for this?

Don't think that's necessary no.



> What do you think should there be something in user guide?

An entry in the changelog's changes section is enough.

> @leonardder why logic could not be in braille module pumpAll function? Does it use too much resources there for example? I ask because then display would be updated also when there is no caret movement. Then for example, selecting check box "[ ]" with spacebar would be updated on braille display to "[x]" automatically although there is no cursor movement.

Using pumpAll would mean an update is performed way too often.
Now the issue you raise about check boxes not being updated, I assume you mean in Browse mode when using the review cursor? As when in object review, I'm pretty sure an update check box should reflect its state when updated also when braille is tethered to review and review mode is set to object.

burmancomp:
Referring to my [previous comment](https://github.com/nvaccess/nvda/pull/15134#issuecomment-1634044826) here are some additional examples.

Use gestures to these:

- set braille tethered to automatically, select text in editor, and disable/enable "braille show selection" with gesture
- when braille shows cursor, disable/enable it with gesture.

As you should note your changes (disabling/enabling showing selection or braille cursor) are not automatically updated to braille line.

I think that reason is same as with "check boxes" in Linux using Putty.

But when pumpAll contains code (code in my previous comment is not optimized) these changes (disabling/enabling showing selection or braille cursor) are updated to display automatically as well as "Linux check boxes" state.

leonardder:
> Use gestures to these:
> 
>     * set braille tethered to automatically, select text in editor, and disable/enable "braille show selection" with gesture
> 
>     * when braille shows cursor, disable/enable it with gesture.
> 
> 
> As you should note your changes (disabling/enabling showing selection or braille cursor) are not automatically updated to braille line.

I think this is a separate issue. could you please open a new issue for this?

leonardder:
Using something like that in pumpAll, while it sounds effective, is is really way to heavy. It will try to update objects way to often.

I think the cases you mention are pretty valid though. I think you should be able to fix them pretty easily.

1. In NVDAObjects.behaviours, in the LiveText class, in the _reportNewLines method, add braille.handler.handleUpdate(self) to the end of the method.
2. In NVDAObjects.Uia.winConsoleUIA, in the _NotificationsBasedWinTerminalUIA class, in the event_UIA_notification method, add braille.handler.handleUpdate(self) to the end of the method.
Pretty sure you will have covered all the terminal cases for braille. I think you can even avoid the current implementation in this pr, at least for terminals, however I think it still makes sense to keep it because it deals with caret updates.

burmancomp:
> > As you should note your changes (disabling/enabling showing selection or braille cursor) are not automatically updated to braille line.
> 
> I think this is a separate issue. could you please open a new issue for this?

I think that these and what I have encountered with Putty in Linux are both special cases of same issue.

Do you think that issue #15115 should be closed as well as this pull request and then open a new issue?

burmancomp:
> Using something like that in pumpAll, while it sounds effective, is is really way to heavy. It will try to update objects way to often.

It is not important to do it in pumpAll but it should be done somewhere not too often but however frequently enough.

> I think the cases you mention are pretty valid though. I think you should be able to fix them pretty easily.

Thank you.

However, I think that whole problem (some display updates, which should be done, are not done) should be solved not just given special case.

leonardder:
> Do you think that issue #15115 should be closed as well as this pull request and then open a new issue?

No, I think this pr is fine as is. We really should call handleUpdate when tethered to review and the caret is updated on an object.

burmancomp:
As to pumpAll code it was not optimal and it did not handle all cases (at least browse mode was not properly handled). However, I am quite sure that number of redundant updates could be notably reduced with code optimization because most of time all needed updates are done.

Maybe one approach could be to check in pumpAll if update is already done during this cycle.

I think it is important that:

- whole problem (some lacking display updates) should be solved.

It is not essential if required logic is in pumpAll or elsewhere as long as it is somewhere.

Unfortunately I am somewhat thinking that this pull request should be modified to more generic before it is ready for review.

What do you think:

- is it possible to optimize pumpAll code so that it is not too heavy? I have local version which seems to somewhat handle all display update problems I have encountered during this process. Local PumpAll alternative has 10 code lines at this moment and 1 in BrailleHandler.handleUpdate. It checks if update is already done. Hopefully effects on cpu usage would be ass small as possible.
- if it is not possible with pumpAll where logic should be located and whatkind it should be?

burmancomp:
Replaced with pull request #15163.

Issue #3276, opened 2013-06-10
Text on the braille display doesn't update after using Undo in Word


nvaccessAuto:
**Reported by falinn.onda on 2013-06-10 17:16**
Hi,

Assuming you do the following:
1. open new word 2010 document.
2. Type a text. for example "1234"
3. press undo CTRL+Z once.

 Expected behavior: braille displays "123" as the current document content after undo.
Actual behavior: Braille displays "1234"

I am using Focus 40 blue via Bluetooth.

bhavyashah:
Could Braille experts please test and report whether these Braille synchronization issues (particularly with MS Word's Undo and Redo, Ctrl+Backspace and Ctrl+Delete) still occur?

Adriani90:
@falinn.onda is this issue still reproducible?
cc: @bramd

dkager:
I can still reproduce this with Word 2016, non-UIA. It isn't specific to a particular braille display.

Adriani90:
@dkager how is this now with NVDA 2023.1 RC1 or NVDA 2022.4?

burmancomp:
I tried locally to add `braille.handler.handleUpdate` to source\NVDAObjects\iaccessible\winword.py and there to `event_caret` function.

It worked almost. If there is for example: "12345", and then ctrl-z is pressed 5 times, there is not necessarily blank display. If pressing and holding ctrl-z it seems to work or if pressing ctrl-z with long delay between presses, it seems to work also. But if something between these, it may not work.

@leonardder what do you think about this? Is my approach wrong?

leonardder:
@burmancomp is this with or without UIA enabled in Word?

burmancomp:
Do you mean this setting: "Use UI Automation to access Microsoft Word document controls"? It is currently " Default (Where suitable)".

burmancomp:
In blank document typing q then ctrl-z. Quite often q remains in braille line. If there are more characters like qwerty, others y, t ... seems to be removed correctly. It seems that `event_caret` would not be always executed.

burmancomp:
It may work better with UIA set to always, and when NVDAObjects\UIA\wordDocument.py also includes event_caret.

I have understood that UIA does not work properly in ms word before Windows 11. I have not Windows 11 installed.

[...]

PR #15163, opened 2023-07-18
Update braille independently on caret

### LINKED ISSUE #15115: braille: update display when tethered to review and review cursor does not follow caret

burmancomp:
### Link to issue number:
Fixes #15115
### Summary of the issue:
This extends pull request #15134 and includes pull request #15156. Collaborated with @leonardder.
Braille line is updated in terminal window when there is text change.

In addition braille line is updated in any application when:

- braille cursor is turned on or off with gesture
- if there is selected text in edit control or document, changing "show selection" state with gesture updates braille line when otherwise supported
Terminal window updates are handled with:

- NVDAObjects.behaviors.Terminal.event_textChange and
- event_UIA_notification in winConsoleUIA._NotificationsBasedWinTerminalUIA.

Objects are passed to BrailleHandler.handleUpdate function.

When braille show selection state is changed with gesture, BrailleHandler.initialDisplay function is executed.

When braille cursor is toggled with gesture, BrailleHandler._updateDisplay function is executed.
### Testing strategy:
Tested with Albatross 80 and little with Optelec Alva BC680.
none known
### Change log entries:
New features
Changes
- When the text in a terminal changes without updating the caret, the text on a braille display will now properly update when positioned on a changed line. This includes situations where braille is tethered to review.

Bug fixes
- The braille cursor and selection indicators will now always be updated correctly after showing or hiding respective indicators with a gesture.

For Developers
- braille.handler.handleUpdate and braille.handler.handleReviewMove have been changed in order not to update instantly. Before this change, when either of these methods was called very often, this would drain many resources. These methods now queue an update at the end of every core cycle instead. They should also be thread safe, making it possible to call them from background threads.

Deprecations:
- braille.BrailleHandler.handlePendingCaretUpdate is now deprecated with no public replacement.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


leonardder:
While I like this pr as a proof of concept, I see several reasons why this doesn't fit for NVDA Core, as it is merely a workaround and obfuscator of other bugs that should have our attention instead:

1. Most notably, it has a polling/timer based approach instead of event driven.
2. A timer on the main thread like this may severely impact NVDA's performance in objects that are slow to update, such as Excel, some more complex UIA implementations, etc. A timer on another thread will probably result in issues with thread safety.

Again, I really like your investment in this, but I'd rather see #15134 being merged first and then have smaller follow ups like #15156, but even the latter is already to broad in scope. The braille subsystem of NVDA is pretty complex and hard to oversee and it is very, very easy to break things

burmancomp:
Thank you @leonardder for your comments.

>     1. Most notably, it has a polling/timer based approach instead of event driven.

What kind of event driven approach you are thinking?

>     2. A timer on the main thread like this may severely impact NVDA's performance in objects that are slow to update, such as Excel, some more complex UIA implementations, etc. A timer on another thread will probably result in issues with thread safety.

Currently it is run once in 0.5 seconds, and it tries to avoid redundant updates. There is also _cursorBlinkTimer polling in 0.25 seconds cycles as default in main thread. Is an additional timer a problem or its polling cycle or what it is doing? 

I tried also with timer in its own thread but in browser there is _ctypes.COMError when switching from browse mode to focus mode.

This pull request should only do something with timer when:

- current object is terminal window (and before update it is checked that braille content has been changed)
- state of show selection is changed
- braille cursor is toggled.

I suppose that most of time only above checks are done, and this is once in 0.5 seconds.

> Again, I really like your investment in this, but I'd rather see #15134 being merged first and then have smaller follow ups like #15156, but even the latter is already to broad in scope and I decided to close it. The braille subsystem of NVDA is pretty complex and hard to oversee and it is very, very easy to break things

There may be problems but I suppose they are met in testing or maybe they can shown with calculations.

Could you tell whatkind of tests should be done?

It is not important that this implementation would be merged but these things are important.

When and if for example font attributes are presented in braille switching that state on and off is important. What I try to say is that these things should be solved anyway so why not earlier than later?

leonardder:
> What kind of event driven approach you are thinking?

There is already the textChange event fired on terminal objects when the text changes. That event should be fired when anything changes in the text, including check boxes in terminals. If not, that's a bug that needs to be looked into first.
Currently, no braille updates are performed when updating terminal text, so that would be a good starting point.

codeofdusk:
> There is already the textChange event fired on terminal objects when the text changes.

Well technically, yes on `LiveText` objects, but not when using UIA notifications in wt as text changes are disabled in that case to improve performance.

burmancomp:
Using `textChange` event on terminal object seems to work for me using Putty in non-graphical Linux. Maybe there should be more appropriate description in "user facing changes" as to terminal windows?

codeofdusk:
Yes, because MinTTY uses `DisplayModelLiveText`.

leonardder:
> > There is already the textChange event fired on terminal objects when the text changes.
> 
> Well technically, yes on `LiveText` objects, but not when using UIA notifications in wt as text changes are disabled in that case to improve performance.

The events are disabled for UIA performance reasons right, not for perf reasons in NVDA itself? Basically the UIA notifications are an alternative approach to tell us that text has changed. Therefore I wonder what would happen if we make the UIA notification implementation call event_textChange internally.

source/NVDAObjects/behaviors.py
```diff
@@ -475,6 +476,13 @@ def _get_caretMovementDetectionUsesEvents(self):
 		prompt to be read when quickly deleting text."""
 		return False
 
+	def event_textChange(self) -> None:
+		"""Fired when the text changes.
+		@note: Updates also braille.
+		"""
+		super().event_textChange()
+		braille.handler.handleUpdate(self)
```
leonardder, line 484: Note the `@note: It is safe to call this directly from threads other than the main thread.` on LiveText. I think this might be problematic for the legacy winconsole implementation as another thread will try to update the braille handler.

source/braille.py
```diff
@@ -2686,6 +2703,17 @@ def _ackTimeoutResetter(self, param: int):
 			self.display._awaitingAck = False
 			self._writeCellsInBackground()
 
+	def updateShowSelection(self) -> None:
```
leonardder, line 2706: I think this should go in a separate pr as this is a different issue altogether.

source/NVDAObjects/behaviors.py
burmancomp: There is a lock in use. After adding it as far as I know I have not encountered race condition.

source/braille.py
burmancomp: It can be regarded as another aspect of issue: braille line is not updated although it should be.

But if separate issue and pull request are needed, no problem.

source/NVDAObjects/behaviors.py
leonardder: I don't think a lock is a good idea here. If a method (event-texctChange) requires thread safety, the implementation should always be thread safe. I would personally remove the lock and go for an alternative for the time being, see my next comment.

leonardder, line 484:
```suggestion
		# braille.handler.handleUpdate must be called from the main thread.
		queueHandler.queueFunction(
			queueHandler.eventQueue,
			braille.handler.handleUpdate,
			self,
			_immediate=True
		)
```

burmancomp: Thank you. I have removed lock, changed code as you suggested and tested. It seems to work.

source/braille.py
leonardder: I can't speak for others, but I'd prefer this to be handled separately. I don't think the new function is needed either, you can probably just call the initialDisplay method.

source/braille.py
```diff
@@ -1205,7 +1205,12 @@ def _addTextWithFields(self, info, formatConfig, isSelection=False):
 				elif cmd == "controlEnd":
 					# Exiting a controlField should break a run of clickables
 					inClickable=False
-					field = ctrlFields.pop()
+					# Index error may occur with Teraterm in non-graphical Linux.
```
leonardder, line 1208: This change doesn't fix a bug but rather obfuscates it. There might be something wrong in DisplayModelTextInfo that causes it (i.e. i assume Teraterm uses the display model).

source/braille.py
```diff
@@ -1569,6 +1574,7 @@ def bufferPosToRegionPos(self, bufferPos):
 		raise LookupError("No such position")
 
 	def regionPosToBufferPos(self, region, pos, allowNearest=False):
+		start: int = 0
```
leonardder, line 1577: Ig you want proper annotations, please annotate the RegionWithPositions class properly. See https://www.geeksforgeeks.org/typing-namedtuple-improved-namedtuples/

Above `def _get_regionsWithPositions(self):`, add:
```
	#: typing information for autoproperty _get_regionsWithPositions(
	regionsWithPositions: Generator[RegionWithPositions, None, None]
```

LeonarddeR (changes requested):
So if this works okay for terminals now, please add a braille.handler.update(self) call to event_UIA_notification in winConsoleUIA._NotificationsBasedWinTerminalUIA.

source/braille.py
burmancomp: I am about to revert this. New issue some day if needed.

source/braille.py
burmancomp: I suppose there was sometimes "UnboundLocalError" as to start variable. I guess for some reason there was no data in `self.regionsWithPositions` so start might be referenced before assignment. Reverting this.

burmancomp:
> So if this works okay for terminals now, please add a braille.handler.update(self) call to event_UIA_notification in winConsoleUIA._NotificationsBasedWinTerminalUIA.

I suppose you mean braille.handler.handleUpdate(self).

I do not know where to test this.

source/braille.py
leonardder: Ugh, now you're saying that, the code in this method indeed assumes that start is set, so I guess it is fine to leave this as is. Good catch.

source/globalCommands.py
```diff
@@ -3300,6 +3300,7 @@ def script_braille_toggleShowCursor(self, gesture):
 			state = _("Braille cursor on")
 			config.conf["braille"]["showCursor"]=True
 		ui.message(state)
+		braille.handler.update()
```
leonardder, line 3303: Could you explain in a code comment why this is necessary? I'm also a bit concerned by the fact that you're updating the braille handler after executing ui.message, which should print a message to the braille display as well.

LeonarddeR (approved):
I think this looks good code wise.

source/globalCommands.py
burmancomp: I have not noticed problems with ui.message and braille.handler.update.

But unfortunately there is another problem: at least with Albatross when I type "dir/s" in NVDA directory whic contains source directory for example, result is that it takes long time when all text is brailled (34842 File(s) and 11846 Dir(s)).

There should be for example a queue with suitable length to avoid the situation that all thousands of handleUpdates should be performed.

Do you have approach you prefer?

leonardder: Sure, I actually expected something like that. I think you can merge in the changes in braille.py for #15156 to compensate for this delay, though we might also need to move the progress bar logic out of handleUpdate.

burmancomp:
@seanbudd what product decision is needed? With newest commit result of "dir/s" in root of NVDA distribution source seems to be quite similar with this version which could be expected with NVDA (it takes about 25 seconds in my system with Alva BC680).

There is an exception, Albatross. It takes about 10 - 15 seconds longer but Albatross driver has write queue (to try to ensure that braille line would update correctly). And "dir/s" in NVDA distribution root is not daily use.

seanbudd (approved):
@michaelDCurran - can you confirm that you're happy with the user facing changes?

source/braille.py
```diff
@@ -2042,6 +2044,9 @@ def __init__(self):
 		self._cursorPos = None
 		self._cursorBlinkUp = True
 		self._cells = []
+		# Queue of objects to use in L{handleUpdate}. Length 1 keeps
+		# recent object and discards others.
```
leonardder, line 2048: What is the point of having a queue? Why not simply updating an attribute to a new object?

michaelDCurran (dismissed):

LeonarddeR (changes requested):
I have some concerns about this approach. Please hold back from merging until these are addressed.

leonardder:
Rather than only reviewing the user facing changes, I would also like to see a substantive review from @michaelDCurran.

One of the findings in this pr is that braille.handler.update is too expensive. Therefore, instead of braille.handler.update being called directly, we now queue an object for update and execute the actual update in a core pump. However, this does not really fix the underlying issue, i.e. handleUpdate is still pretty expensive. I would rather see this issue being addressed at the core of it. I implemented a proof of concept for a revamp of handleUpdate in https://github.com/nvaccess/nvda/pull/15156 and suggested in https://github.com/nvaccess/nvda/pull/15163#discussion_r1277569200 that it could be used. Honestly, I'm sorry now that I closed that pr.

burmancomp:
If there is text change event in terminal window, there should be only at most one `handler.handleUpdate()` execution in core cycle.

burmancomp:
It seems that my previous comment was inaccurate, sorry. Could we agree that comparing to current implementation this pull request causes at most one additional `handler.handleUpdate` execution when terminal window is active and there is text change event?

leonardder:
It is no problem at all if you call handler.handleUpdate multiple times, as long as handler.handleUpdate is changed in such a way that it compensates for this by outsource the heavier work to the core pump.

burmancomp:
@leonardder,

Could we solve this combining #15156 and this pull request. I have tested this initially by merging #15156 to local branch which contained code of this pull request. For now there was very little to do - nearly merge couple of commits. There may be some problems but maybe they are small.

I suppose merging would work also as opposite - mergin this pull request code to #15156 code.

Then if desired/appropriate this pull request could be closed and a new pull request opened with combined code.

What do you think about this?

leonardder:
I'd be very grateful if you'd merge in #15156. Basically it would indeed 
be a merge, discard the changes except for braille.py and then remove 
the handleUpdate queue, calling braille.handler.handleUpdate instead.

burmancomp:
Problem I have found until now is that if braille is tethered to review (and review follows caret) display does not always show actual content. What should I try to change?

burmancomp:
It seems to come from #15156. I added `region.pendingCaretUpdate=True` to `handleReviewMode`, and it seems to help.

I could push that combined code so that you could review if it is acceptable for you.

leonardder:
That's fine with me, go ahead please.

source/NVDAObjects/UIA/winConsoleUIA.py
```diff
@@ -471,6 +472,7 @@ def event_UIA_notification(
 		# Do not announce output from background terminals.
 		if self.appModule != api.getFocusObject().appModule:
 			return
+		braille.handler.handleUpdate.append(self)
```
leonardder, line 475:
```suggestion
		braille.handler.handleUpdate(self)
```

LeonarddeR (approved):
Thanks. If this works as expected for you, I'd like to go with this approach. Just one thing.

source/NVDAObjects/UIA/winConsoleUIA.py
burmancomp: Thanks.

Adriani90:
Will the combined code touch some how also #3095 and #3276?

burmancomp:
> Will the combined code touch some how also #3095 and #3276?

No, as far as I understand.

leonardder:
This doesn't fix #3276 yet, but I think we can easily start calling handler.hanldeUpdate in more places, since it is both thread safeand fairly passive now. Should be something for a follow up though.

leonardder:
I had a look into the unit tests. They fail because there's no core pump running during tests, and handleUpdate no longer updates braille instantly.

leonardder:
@burmancomp could you please add the following to tests/unit/__init__py, below `braille.filter_displaySize.register(getFakeCellCount)`?

```
_original_handleReviewMove = braille.handler.handleReviewMove


def _patched_handleReviewMove(shouldAutoTether=True):
	"""
	For braille unit tests that rely on updating the review cursor, ensure that we simulate a core cycle
	by executing BrailleHandler._handlePendingUpdate after Braillehandler.handleReviewMove
	"""
	_original_handleReviewMove(shouldAutoTether)
	braille.handler._handlePendingUpdate()


braille.handler.handleReviewMove = _patched_handleReviewMove
```

LeonarddeR (approved):

leonardder:
I'd like to suggest the following changelog entries:

Changes:
- When the text in a terminal changes and braille is tethered to review, the text on a braille display will now properly update if the review cursor is positioned on a changed line. (#15115)
Bug fixes
- The braille cursor and selection indicators will now always be updated correctly after showing or hiding respective indicators with a gesture.
Changes for developers
- braille.handler.handleUpdate and braille.handler.handleReviewMove have been changed in order not to update instantly. Before this change, when either of these methods was called very often, this would drain many resources. These methods now queue an update at the end of every core cycle instead. They should also be thread safe, making it possible to call them from background threads.
Deprecations:
- braille.BrailleHandler.handlePendingCaretUpdate is now deprecated with no public replacement.

leonardder:
Good catch. SOmething like:;

- When the text in a terminal changes without updating the caret, the text on a braille display will now properly update when positioned on a changed line. This includes situations where braille is tethered to review.

seanbudd (approved):
Thanks @burmancomp

burmancomp:
> Thanks @burmancomp

And thanks @leonardder as well.

PR #15349, opened 2023-08-31
Fixes to address hangs caused by WASAPI idle checks

### LINKED ISSUE #15150: Intermitent crashing on latest alphas

jcsteh:
### Link to issue number:
Fixes #15150.

### Summary of the issue:
When WASAPI is enabled, NVDA sometimes hangs and cannot be recovered, especially when switching audio devices. The freezes all occur in `WasapiWavePlayer._idleCheck` when calling `wasPlay_idle`.

Fixed the hangs.

1. If getting the playback position fails with an error, playback has probably been interrupted; e.g. because the device disconnected. Treat this as if playback has finished so we don't wait forever and so that we fire any pending callbacks.
2. Never treat a stream as idle while we're feeding audio to it. Aside from this not making any sense, we might otherwise try to sync and stop it in the main thread, which could cause hangs or crashes.

### Testing strategy:
I'm unable to reproduce this, so I asked several people to test try builds in #15150. They report that this addresses the problem for them.
I also verified (using beeps in the code, now removed) that streams are still correctly treated as idle when they should be.

None.

### Change log entries:
Bug fixes
When WASAPI is enabled, NVDA no longer hangs occasionally, especially when switching audio devices.

### Code Review Checklist:

  - description is up to date
  - change log entries
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
- [x] Documentation:
  - User Documentation
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/nvdaHelper/local/wasapi.cpp
+++ b/nvdaHelper/local/wasapi.cpp
@@ -413,7 +413,10 @@ UINT64 WasapiPlayer::getPlayPos() {
 	UINT64 pos;
 	HRESULT hr = clock->GetPosition(&pos, nullptr);
 	if (FAILED(hr)) {
-		return 0;
+		// If we get an error, playback has probably been interrupted; e.g. because
+		// the device disconnected. Treat this as if playback has finished so we
+		// don't wait forever and so that we fire any pending callbacks.
+		return sentMs;
 	}
 	return pos * 1000 / clockFreq;
 }
--- a/source/nvwave.py
+++ b/source/nvwave.py
@@ -883,6 +883,8 @@ class WasapiWavePlayer(garbageHandler.TrackedObject):
 		if self._audioDucker:
 			self._audioDucker.enable()
 		feedId = c_uint() if onDone else None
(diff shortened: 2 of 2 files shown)


seanbudd (approved):
Thanks @jcsteh .

I'm not going to use the change log entry as this fixes an unreleased issue (kinda)

PR #15583, opened 2023-10-04
Sapi4: Only send prosody commands when there is prosody in the sequence

### LINKED ISSUE #15580: SAPI 4: impossible to adjust any speech parameters.
### Steps to reproduce:
1. Install SAPI 4 along with a synthesizer. Instructions are available in #12152.
2. Go to the voice settings, and try to adjust a parameter like rate, pitch, volume, etc.
### Actual behavior:

Adjusting rate, pitch and volume doesn't work.
### Expected behavior:

Adjusting speech rate, pitch and volume should work.
Installed.
#### NVDA version:
alpha-29490,2adb77cb.
#### Windows version:
Windows 11, Version 23H2 (OS Build 25951.1010).
SAPI 4, American English TruVoice, however this doesn't seem to be specific to any particular synthesizer.
N/A.
Yes.
Yes, 2023.3beta4 works normally.
Yes.
Yes.

LeonarddeR:
### Link to issue number:
Fixes #15580 
Follow up of #15502 

### Summary of the issue:
In #15502, I fixed some regressions regarding capital pitch changes introduced in #15271. This meant that we'd add prosody commands to every speech sequence to support a specific version of IBM TTS that seemed to need that. However, it introduced a regression in that it was no longer possible to change speech parameters using the gui.

It is again possible to change SAPI4 parameters using the GUI.

Only add blank prosody commands when there is any prosody in the sequence. IF not, don't do anything.

### Additional context
A prosody command in SAPI4 (like pitch) behaves as the pitch would really be changed with PitchSet. In other words, when you'd fetch the pitch from the synth driver during a long sequence that contains a PitchCommand, the driver would return the Pitch as reflected by the PitchCommand. This meant that, when we'd set the pitch using the GUI, the pitch would also be affected by the PitchCommand added to the speech sequence. This caused conflicts. Thus, we should only add prosody commands when necessary.

### Testing strategy:
Tested str from #15580 and #15500

None nkown

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/synthDrivers/sapi4.py
+++ b/source/synthDrivers/sapi4.py
@@ -134,11 +134,15 @@ class SynthDriver(SynthDriver):
 		# #15500: Some SAPI4 voices reset all prosody when they receive any prosody command,
 		# whereas other voices never undo prosody changes when a sequence is interrupted.
 		# Add all default values to the start and end of the sequence,
-		# but avoid duplicating the first command, if any.
-		prosodyToAdd = [
-			c() for c in self.supportedCommands
+		# but avoid duplicating the first command, if any,
+		# And only add the defaults when there is a prosody command in the sequence.
+		supportedProsody = [
+			c for c in self.supportedCommands
 			if issubclass(c, BaseProsodyCommand)
 		]
+		prosodyToAdd = []
+		if any(type(i) in supportedProsody for i in unprocessedSequence):
+			prosodyToAdd.extend(c() for c in supportedProsody)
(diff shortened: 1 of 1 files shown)


seanbudd (approved):
Thanks @leonardder

PR #15702, opened 2023-10-28
Do not offer External channel in tabs where it is not relevant

CyrilleB79:
### Link to issue number:
Closes #15039
### Summary of the issue:
External appears in channel filter list even in tabs where it is not relevant, i.e. in Available add-ons and Updatable add-ons.
Remove these External channel from the channel filter list in these tabs.
See code.
### Testing strategy:
Manual testing.
None

### Change log
None for such a small fix.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/_addonStoreGui/controls/storeDialog.py
+++ b/source/gui/_addonStoreGui/controls/storeDialog.py
@@ -158,7 +158,6 @@ class AddonStoreDialog(SettingsDialog):
 			wxCtrlClass=wx.Choice,
 			choices=list(c.displayString for c in _channelFilters),
 		))
-		gv.dbg = self.channelFilterCtrl
 		self.channelFilterCtrl.Bind(wx.EVT_CHOICE, self.onChannelFilterChange, self.channelFilterCtrl)
 		self.bindHelpEvent("AddonStoreFilterChannel", self.channelFilterCtrl)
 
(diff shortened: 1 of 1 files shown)


lukaszgo1:
> External appears in channel filter list even in tabs where it is not relevant, i.e. in Available add-ons and Updatable add-ons.

Why you consider 'external' to be irrelevant in the 'updateable' tab? As far as I can tell add-ons in that tab can be either installed from the store, or installed from the add-on bundle, but with a more recent version present in the store.###

CyrilleB79:
> Why you consider 'external' to be irrelevant in the 'updateable' tab? As far as I can tell add-ons in that tab can be either installed from the store, or installed from the add-on bundle, but with a more recent version present in the store.

*Because the channel visible in the Updatable add-ons tab is not the channel of the installed add-on but the channel through which a new version can be offered from the store.

E.g. say you have installed Speech logger 23.3.08 externally. Checking the updatable add-ons:
* if you select stable channel, you are offered the stable version from the store, i.e. 23.3.09
* if you select dev channel, you are offered the dev version from the store, i.e. 23.3.109  sélectionné
* if you select external channel, you are not offered anything since the store does not offer anything through external channel by definition

To summarize, the column "Channel" in this tab should be understood as "Available channel", not as "Installed channel".

seanbudd (approved):
Thanks @CyrilleB79

PR #15748, opened 2023-11-06
Edge with UIA: allow activating  buttons in browse mode by walking ancestors until an action succeeds

### LINKED ISSUE #14612: NVDA fails to interact with some web controls on Google pages if using UIA for Microsoft Edge
### Steps to reproduce:
1. From NVDA's advanced settings, set the option "Use UIA with Microsoft Edge and other Chromium based browsers when available" to "Yes".
2. Open Microsoft Edge.
3. Go to youtube.com.
4. Go to the toggle button "Guide" and press ENTER to activate it.

### Actual behavior:
Nothing happens.

### Expected behavior:
The guide panel on the left should appear/disappear, depending on its status and NVDA should report "checked"/"Not checked" as a status for the focused "Guide" toggle button.

[NVDA_MSEdge_UIA.log](https://github.com/nvaccess/nvda/files/10664718/NVDA_MSEdge_UIA.log)

Installed.

#### NVDA version:
NVDA version: alpha-27627,79ad4f07.

#### Windows version:
Operating system: Windows 11 Pro version 22H2 (build 22621.1105), 64-bit, in Bulgarian with all locale settings set to "Bulgarian".

Microsoft Edge version: 109.0.1518.78, in bulgarian.

Yes, it does.

Yes - NVDA 2022.4 and the result is the same.

Yes, it is.

I haven't done that recently, but previously I have for some other diagnostics and the problem was occurring before that and after it, so this is not relevant.

michaelDCurran:
### Link to issue number:
Fixes #14612

### Summary of the issue:
when accessing Microsoft Edge with UI Automation, NVDA is unable to activate most buttons and controls in browse mode.
This is because NvDA only tries to invoke/select/toggle the deepest element at the browse mode caret position, which may be a text leaf, or span within the interactive control, and not the control itself.

Microsoft Edge via UI Automation: NVDA can now activate buttons and other controls in browse mode.

* UIA NVDAObject's doAction method: if the element cannot be invoked or selected or toggled, then NotImplementedError is raised.
* UIA BrowseMode document's _activateNVDAObject method: Rather than just trying to activate the given object with its doAction method, instead start at the given object and move up its ancestors, until a call to doAction succeeds. I.e. activates the given object or its closest ancestor that supports performing actions.
 
### Testing strategy:
* Ensure NVDA's use UI Automation to access Edge / chromium is set to yes.
* Open Microsoft Edge and navigate to https://www.youtube.com/
* Tab to the Guide button and press enter, and ensure that it activates.

None known.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -2020,10 +2020,10 @@ class UIA(Window):
 			if self.UIAInvokePattern:
 				self.UIAInvokePattern.Invoke()
 				return
-			elif self.UIATogglePattern:
+			if self.UIATogglePattern:
 				self.UIATogglePattern.toggle()
 				return
-			elif self.UIASelectionItemPattern:
+			if self.UIASelectionItemPattern:
 				self.UIASelectionItemPattern.select()
 				return
 		raise NotImplementedError
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -2019,11 +2019,13 @@ class UIA(Window):
 		if index==0:
 			if self.UIAInvokePattern:
 				self.UIAInvokePattern.Invoke()
+				return
 			elif self.UIATogglePattern:
 				self.UIATogglePattern.toggle()
+				return
 			elif self.UIASelectionItemPattern:
 				self.UIASelectionItemPattern.select()
-			return
+				return
 		raise NotImplementedError
 
 	def _get_hasFocus(self):
--- a/source/UIAHandler/browseMode.py
+++ b/source/UIAHandler/browseMode.py
@@ -501,10 +501,13 @@ class UIABrowseModeDocument(UIADocumentWithTableNavigation,browseMode.BrowseMode
 		raise NotImplementedError
 
 	def _activateNVDAObject(self,obj):
-		try:
-			obj.doAction()
-		except NotImplementedError:
-			pass
+		while obj in self:
+			try:
+				obj.doAction()
+				return
+			except NotImplementedError:
+				pass
+			obj = obj.parent
 
 	def _get_isAlive(self):
 		if not winUser.isWindow(self.rootNVDAObject.windowHandle):
--- a/source/UIAHandler/browseMode.py
+++ b/source/UIAHandler/browseMode.py
@@ -500,15 +500,6 @@ class UIABrowseModeDocument(UIADocumentWithTableNavigation,browseMode.BrowseMode
 			return UIAControlQuicknavIterator(nodeType,self,pos,condition,direction)
 		raise NotImplementedError
 
-	def _activateNVDAObject(self,obj):
-		while obj in self:
-			try:
-				obj.doAction()
-				return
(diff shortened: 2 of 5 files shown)


jcsteh:
Unless I'm missing something, this is not going to work well for non-semantic clickable controls. Obviously, those are authoring error, but there are plenty of them. This might also result in a click potentially getting routed to a far away, irrelevant ancestor that happens to have a click handler for event listening purposes.

With gecko_ia2, when walking ancestors, if doAction fails, NVDA simulates a mouse click if the ancestor has a relevant location and considers that a success. This does occasionally fail and click the wrong place, since location might not be perfect due to bugs, pop-ups obscuring things, etc. That's why Chrome and Firefox implemented click ancestor, but I guess we can't access that with UIA, so click simulation is probably the best we can do.

michaelDCurran:
@jcsteh  I've now moved that gecko_ia2 logic into the base browse mode _activateNVDAObject. I can't see any reason why all implementations can't use that then if we are happy with it.

source/NVDAObjects/UIA/__init__.py
```diff
@@ -2019,11 +2019,13 @@ def doAction(self,index=None):
 		if index==0:
 			if self.UIAInvokePattern:
 				self.UIAInvokePattern.Invoke()
+				return
 			elif self.UIATogglePattern:
 				self.UIATogglePattern.toggle()
+				return
 			elif self.UIASelectionItemPattern:
 				self.UIASelectionItemPattern.select()
-			return
```
jcsteh, line 2026: Why this change? I don't think there's a functional difference, but having elif + return is redundant. If you want to keep the early returns in each branch, I'd change the elifs to ifs.

jcsteh (commented):
"Happy with it" is probably a stretch. As I noted, it can result in clicking the wrong thing in some cases. However, I think this is probably the most pragmatic solution given the prevalence of non-semantic clickables with ancestor click handlers.

Could this be a problem for non-web contexts; Microsoft Word, etc.?

source/NVDAObjects/UIA/__init__.py
michaelDCurran: Previously it would raise NotImplementedError only if actionIndex was not 0. Now it also raises NotImplementedError if neither invoking, toggling or selecting is possible. At least, that was what I was trying to go for...

michaelDCurran:
@jcsteh  Some quick tests with MS Word seems to suggest things are working okay. It is possibly now more accurate when jumping to references.

source/NVDAObjects/UIA/__init__.py
jcsteh: Ah gotcha. Sorry, missed that subtlety. In that case, please change the elifs to ifs and we're good here.

jcsteh (approved):

seanbudd (approved):
Please fix the failing lint check before merging

PR #15754, opened 2023-11-07
WasapiWavePlayer: Don't explicitly remove the instance from the instances map in the destructor.

### LINKED ISSUE #15752: NVDA freezes when using wasapi
cc @jcsteh 
### Steps to reproduce:
There is a way of reproducing this when using Newfon or RHVoice synth drivers
install one of the listed addons.
enter the following code in NVDA console
import synthDriverHandler; [synthDriverHandler.setSynth(synthDriverHandler.getSynth().name) for i in range(20)]
### Actual behavior:
NVDA is freezing and logging lots of errors.
Player is becoming None and idle cannot be called
### Expected behavior:
NVDA should be stable.
Running from source, but reproduceable on 2023.3 Release version
#### NVDA version:
Master
#### Windows version:
Windows 10 22h2
Yes
Yes, 2023.3 behaves the same
Yes, but easily reproduceable with Newfon or RHVoice synth
Yes

jcsteh:
### Link to issue number:
Fixes #15752.

### Summary of the issue:
Some users sometimes see errors like this in their logs, accompanied by long freezes:

```
ERROR - stderr (15:18:17.518) - MainThread (29504):
Exception ignored in:
ERROR - stderr (15:18:17.529) - MainThread (29504):
<function WasapiWavePlayer.__del__ at 0x0467A6B8>
ERROR - stderr (15:18:17.950) - MainThread (29504):
ERROR - stderr (15:18:18.392) - MainThread (29504):
  File "nvwave.pyc", line 843, in __del__
ERROR - stderr (15:18:18.830) - MainThread (29504):
  File "weakref.pyc", line 145, in __delitem__
ERROR - stderr (15:18:18.843) - MainThread (29504):
KeyError
ERROR - stderr (15:18:18.855) - MainThread (29504):
:
ERROR - stderr (15:18:19.283) - MainThread (29504):
80057872
```

NVDA no longer sometimes freezes when speaking a large amount of text.

Previously, WasapiWavePlayer's `__del__` method removed itself from the `_instances` map. However, a WeakValueDictionary actually removes the reference itself when the object dies. A weakref callback can run before `__del__` in some cases, which would mean that the item was already removed, resulting in this KeyError.

To fix this, we just don't explicitly remove the item and rely entirely on WeakValueDictionary to do this.

### Testing strategy:
I can't reproduce this, but @beqabeqa473 confirmed that this change seems to fix the issue.

None.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/nvwave.py
+++ b/source/nvwave.py
@@ -840,7 +840,11 @@ class WasapiWavePlayer(garbageHandler.TrackedObject):
 			return
 		if self._player:
 			NVDAHelper.localLib.wasPlay_destroy(self._player)
-			del self._instances[self._player]
+			# Because _instances is a WeakValueDictionary, it will remove the
+			# reference to this instance by itself. We don't need to do it explicitly
+			# here. Furthermore, doing it explicitly might cause an exception because
+			# a weakref callback can run before __del__ in some cases, which would mean
+			# it has already been removed from _instances.
 			self._player = None
 
 	def open(self):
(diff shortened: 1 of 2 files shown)


seanbudd (approved):
Thanks @jcsteh

Issue #7747, opened 2017-11-12
Flexible ring of settings


Adriani90:
Steps to reproduce:
1. Certain users don't ned certain settings very often but some settings are of high importance for them
2. Some important settings would make the user experience much better if they could be changed directly out of the ring.

Expected behavior:
The user should have the flexibility to add or remove settings from the ring.

Actual behavior:
The ring of settings has not changed since many years although it is one of the most significant advantage of NVDA compared to other screen readers.

System configuration:
NVDA version:
all

NVDA Installed or portable:
both

Other information:

Windows version:
all

Name and version of other software in use when reproducing the issue:
all

Other questions:

Does the issue still occur after restarting your PC?
yes

Have you tried any other versions of NVDA?
yes, all

ehollig:
Could this be accomplished with configuration profiles? For example, you could have one profile announce font size and font attributes while another profile had these disabled. Then, with implementation of #4209, you could press a keystroke to switch between the manual activated profiles. Leaning to a close in favor of #4209

Adriani90:
Hello,

 

this would not solve the problem since you would have to define which font and which color should be activated for which profile. This is far to complicated for an usual user since you have to do lots of settings for each profile. There are to many keystrokes on the market at this point. Narator has hundreeds, jaws as well and now NVDA as well. I think the best way is to adjust the settings ring.

 

Or is it not possible from a programming perspective?

 

 

Best

Adriani

 

Von: Ethan Holliger [mailto:notifications@github.com] 
Gesendet: Montag, 13. November 2017 19:48
An: nvaccess/nvda <nvda@noreply.github.com>
Cc: Adriani90 <adriani.botez@googlemail.com>; Author <author@noreply.github.com>
Betreff: Re: [nvaccess/nvda] Flexible ring of settings (#7747)

 

Could this be accomplished with configuration profiles? For example, you could have one profile announce font size and font attributes while another profile had these disabled. Then, with implementation of #4209 <https://github.com/nvaccess/nvda/issues/4209> , you could press a keystroke to switch between the manual activated profiles. Leaning to a close in favor of #4209 <https://github.com/nvaccess/nvda/issues/4209>

Adriani90:
Slightly edited

derekriemer:
[...]

PR #15804, opened 2023-11-19
Add an "on-demand" speech mode

### LINKED ISSUE #481: Mute automatic speech feedback
**Reported by tvalusek on 2009-11-26 08:38**
Some screenreaders for mobile devices have an extremely useful feature for partially-sighting users - an option to mute automatic feedback issued during normal device use. However, all commands involving a dedicated "screenreader" key operate as usual, including voice feedback. That means one can silently navigate menus, and when there is an item with incomplete title shown, a screenreader command for reading current item can be used without re-enabling speech.

I'd like to see similar behavior in NVDA - a mode with no spoken feedback on ordinary key presses (including command keys such as TAB, Enter, Space etc.), but all NVDA commands involving NVDA key operating as usual. This could be useful for Braille users too. Virtual buffers should be automatically engaged in form mode when auto-feedback is muted, so that they don't interfere with common control of e.g. web browsers, PDF readers etc.

CyrilleB79:
### Link to issue number:
Closes #481
### Summary of the issue:
In some situations, e.g. meetings or when an alternative information channel (braille, visual with magnification), the user does not want NVDA to provide speech feedback on actions such as moving the cursor or focus. But they still want to have NVDA speaking on specific request.

Add a new speech "on demand" mode in the NVDA+S toggle command. When speech on demand is selected, NVDA does not provide a speech feedback for actions, but still speak on explicit request.

The criterion to determine if a command should speak in "on demand" mode is:
* if the command is mainly to perform an action and the speech is only a feedback after this action, NVDA does not speak. Said otherwise, this command could have a usefulness when used blindly without any speech (or braille or visual) feedback
* if the goal command is mainly to provide information without performing any other action, NVDA speaks.

Note: although say all commands move the cursor, they are speaking in the "on demand"  mode since there is no point running them if speech is totally off.

- Added a new speech mode
- Added a new "speakOnDemand" parameter for the script decorator to indicate which script should speak in this mode.
- In `documentBase.py`, do not use script decorator as per https://github.com/nvaccess/nvda/pull/13435#discussion_r820276688

### Notes for reviewers
- Regarding UX, was it a good solution to add a fourth speech mode? There is already an add-on to remove the beep mode from the speech mode toggle command...
- For documentation, what should be the correct name to use? "speech on demand", "speak on demand", etc. Also, speak or talk? @XLTechie would you mind review documentation (change log, user guide, command description)?
- Do I need to include an API change item when extending the speech mode enum?

### Testing strategy:
Manual tests of the modified scripts to support "on demand" mode:
* All the say all scripts (caret, review, foreground window, row and column)
* Most of the other modified scripts in global commands (all those having a default shortcut)
* Word with and without UIA, Excel, poedit
* Tested miranda32 and vipmud with a fake test renaming an executable since I have not these applications.

Foobar not tested.

Tested other scripts to check that they do not speak in "on demand" mode.

None
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/IAccessible/winword.py
+++ b/source/NVDAObjects/IAccessible/winword.py
@@ -275,6 +275,12 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 			speech.speakTextInfo(info, reason=controlTypes.OutputReason.FOCUS)
 			braille.handler.handleCaretMove(self)
 
+	@script(
+		# Translators: a description for a script
+		description=_("Reports the text of the comment where the System caret is located."),
+		gesture="kb:NVDA+alt+c",
+		speakOnDemand=True,
+	)
 	def script_reportCurrentComment(self,gesture):
 		info=self.makeTextInfo(textInfos.POSITION_CARET)
 		info.expand(textInfos.UNIT_CHARACTER)
@@ -294,8 +300,6 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 						return
 		# Translators: a message when there is no comment to report in Microsoft Word
 		ui.message(_("No comments"))
-	# Translators: a description for a script
-	script_reportCurrentComment.__doc__=_("Reports the text of the comment where the System caret is located.")
 
 	def _moveInTable(self,row=True,forward=True):
 		info=self.makeTextInfo(textInfos.POSITION_CARET)
@@ -423,7 +427,6 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 		"kb:alt+end":"caret_moveByCell",
 		"kb:alt+pageUp":"caret_moveByCell",
 		"kb:alt+pageDown":"caret_moveByCell",
-		"kb:NVDA+alt+c":"reportCurrentComment",
 	}
 
 
--- a/source/NVDAObjects/IAccessible/winword.py
+++ b/source/NVDAObjects/IAccessible/winword.py
@@ -277,7 +277,7 @@ class WordDocument(IAccessible, EditableTextWithoutAutoSelectDetection, winWordW
 
 	@script(
 		# Translators: a description for a script
-		description=_("Reports the text of the comment where the System caret is located."),
+		description=_("Reports the text of the comment where the system caret is located."),
 		gesture="kb:NVDA+alt+c",
 		speakOnDemand=True,
 	)
--- a/source/NVDAObjects/UIA/wordDocument.py
+++ b/source/NVDAObjects/UIA/wordDocument.py
@@ -600,7 +600,8 @@ class WordDocument(UIADocumentWithTableNavigation,WordDocumentNode,WordDocumentB
 	@script(
 		gesture="kb:NVDA+alt+c",
 		# Translators: a description for a script that reports the comment at the caret.
-		description=_("Reports the text of the comment where the System caret is located.")
+		description=_("Reports the text of the comment where the System caret is located."),
+		speakOnDemand=True,
 	)
 	def script_reportCurrentComment(self,gesture):
 		caretInfo=self.makeTextInfo(textInfos.POSITION_CARET)
--- a/source/NVDAObjects/UIA/wordDocument.py
+++ b/source/NVDAObjects/UIA/wordDocument.py
@@ -600,7 +600,7 @@ class WordDocument(UIADocumentWithTableNavigation,WordDocumentNode,WordDocumentB
 	@script(
 		gesture="kb:NVDA+alt+c",
 		# Translators: a description for a script that reports the comment at the caret.
-		description=_("Reports the text of the comment where the System caret is located."),
+		description=_("Reports the text of the comment where the system caret is located."),
 		speakOnDemand=True,
 	)
 	def script_reportCurrentComment(self,gesture):
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -185,7 +185,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing three times will spell the line using character descriptions."
 		),
 		category=SCRCAT_SYSTEMCARET,
-		gestures=("kb(desktop):NVDA+upArrow", "kb(laptop):NVDA+l")
+		gestures=("kb(desktop):NVDA+upArrow", "kb(laptop):NVDA+l"),
+		speakOnDemand=True,
 	)
 	def script_reportCurrentLine(self,gesture):
 		obj=api.getFocusObject()
@@ -259,7 +260,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing three times spells it using character descriptions."
 		),
 		category=SCRCAT_SYSTEMCARET,
-		gestures=("kb(desktop):NVDA+shift+upArrow", "kb(laptop):NVDA+shift+s")
+		gestures=("kb(desktop):NVDA+shift+upArrow", "kb(laptop):NVDA+shift+s"),
+		speakOnDemand=True,
 	)
 	def script_reportCurrentSelection(self,gesture):
 		obj=api.getFocusObject()
@@ -285,7 +287,8 @@ class GlobalCommands(ScriptableObject):
 		# Translators: Input help mode message for report date and time command.
 		description=_("If pressed once, reports the current time. If pressed twice, reports the current date"),
 		category=SCRCAT_SYSTEM,
-		gesture="kb:NVDA+f12"
+		gesture="kb:NVDA+f12",
+		speakOnDemand=True,
 	)
 	def script_dateTime(self,gesture):
 		if scriptHandler.getLastScriptRepeatCount()==0:
@@ -1108,7 +1111,8 @@ class GlobalCommands(ScriptableObject):
 			"and pressing three times Copies name and value of this object to the clipboard"
 		),
 		category=SCRCAT_OBJECTNAVIGATION,
-		gestures=("kb:NVDA+numpad5", "kb(laptop):NVDA+shift+o")
+		gestures=("kb:NVDA+numpad5", "kb(laptop):NVDA+shift+o"),
+		speakOnDemand=True,
 	)
 	def script_navigatorObject_current(self, gesture: inputCore.InputGesture):
 		curObject=api.getNavigatorObject()
@@ -1173,6 +1177,7 @@ class GlobalCommands(ScriptableObject):
 			"or location of the navigator object if there is no text under review cursor."
 		),
 		category=SCRCAT_OBJECTNAVIGATION,
+		speakOnDemand=True,
 	)
 	def script_reportReviewCursorLocation(self, gesture):
 		self._reportLocationText((api.getReviewPosition(), api.getNavigatorObject()))
@@ -1183,6 +1188,7 @@ class GlobalCommands(ScriptableObject):
 			"Reports information about the location of the current navigator object."
 		),
 		category=SCRCAT_OBJECTNAVIGATION,
+		speakOnDemand=True,
 	)
 	def script_reportCurrentNavigatorObjectLocation(self, gesture):
 		self._reportLocationText((api.getNavigatorObject(),))
@@ -1195,6 +1201,7 @@ class GlobalCommands(ScriptableObject):
 			"or location of the currently focused object if there is no caret."
 		),
 		category=SCRCAT_SYSTEMCARET,
+		speakOnDemand=True,
 	)
 	def script_reportCaretLocation(self, gesture):
 		self._reportLocationText(
@@ -1208,6 +1215,7 @@ class GlobalCommands(ScriptableObject):
 			"Reports information about the location of the currently focused object."
 		),
 		category=SCRCAT_FOCUS,
+		speakOnDemand=True,
 	)
 	def script_reportFocusObjectLocation(self, gesture):
 		self._reportLocationText((api.getFocusObject(),))
@@ -1219,7 +1227,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing twice may provide further detail."
 		),
 		category=SCRCAT_OBJECTNAVIGATION,
-		gestures=("kb:NVDA+shift+numpadDelete", "kb(laptop):NVDA+shift+delete")
+		gestures=("kb:NVDA+shift+numpadDelete", "kb(laptop):NVDA+shift+delete"),
+		speakOnDemand=True,
 	)
 	def script_navigatorObject_currentDimensions(self, gesture):
 		if scriptHandler.getLastScriptRepeatCount() == 0:
@@ -1236,7 +1245,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing twice may provide further detail."
 		),
 		category=SCRCAT_SYSTEMCARET,
-		gestures=("kb:NVDA+numpadDelete", "kb(laptop):NVDA+delete")
+		gestures=("kb:NVDA+numpadDelete", "kb(laptop):NVDA+delete"),
+		speakOnDemand=True,
 	)
 	def script_caretPos_currentDimensions(self, gesture):
 		if scriptHandler.getLastScriptRepeatCount() == 0:
@@ -1554,7 +1564,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing three times will spell the line using character descriptions."
 		),
 		category=SCRCAT_TEXTREVIEW,
-		gestures=("kb:numpad8", "kb(laptop):NVDA+shift+.")
+		gestures=("kb:numpad8", "kb(laptop):NVDA+shift+."),
+		speakOnDemand=True,
 	)
 	def script_review_currentLine(self, gesture: inputCore.InputGesture):
 		info=api.getReviewPosition().copy()
@@ -1741,7 +1752,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing three times spells the word using character descriptions"
 		),
 		category=SCRCAT_TEXTREVIEW,
-		gestures=("kb:numpad5", "kb(laptop):NVDA+control+.", "ts(text):hoverUp")
+		gestures=("kb:numpad5", "kb(laptop):NVDA+control+.", "ts(text):hoverUp"),
+		speakOnDemand=True,
 	)
 	def script_review_currentWord(self, gesture: inputCore.InputGesture):
 		info=api.getReviewPosition().copy()
@@ -1871,7 +1883,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing three times reports the numeric value of the character in decimal and hexadecimal"
 		),
 		category=SCRCAT_TEXTREVIEW,
-		gestures=("kb:numpad2", "kb(laptop):NVDA+.")
+		gestures=("kb:numpad2", "kb(laptop):NVDA+."),
+		speakOnDemand=True,
 	)
 	def script_review_currentCharacter(self, gesture: inputCore.InputGesture):
 		info=api.getReviewPosition().copy()
@@ -1982,7 +1995,8 @@ class GlobalCommands(ScriptableObject):
 	@script(
 		# Translators: Input help mode message for Review Current Symbol command.
 		description=_("Reports the symbol where the review cursor is positioned. Pressed twice, shows the symbol and the text used to speak it in browse mode"),
-		category=SCRCAT_TEXTREVIEW
+		category=SCRCAT_TEXTREVIEW,
+		speakOnDemand=True,
 	)
 	def script_review_currentSymbol(self,gesture):
 		info=api.getReviewPosition().copy()
@@ -2265,7 +2279,8 @@ class GlobalCommands(ScriptableObject):
 	@script(
 		# Translators: Input help mode message for report formatting command.
 		description=_("Reports formatting info for the current review cursor position."),
-		category=SCRCAT_TEXTREVIEW
+		category=SCRCAT_TEXTREVIEW,
+		speakOnDemand=True,
 	)
 	def script_reportFormattingAtReview(self, gesture):
 		self._reportFormattingHelper(api.getReviewPosition(), False)
@@ -2285,7 +2300,8 @@ class GlobalCommands(ScriptableObject):
 			" If pressed twice, presents the information in browse mode"
 		),
 		category=SCRCAT_TEXTREVIEW,
-		gesture="kb:NVDA+shift+f"
+		gesture="kb:NVDA+shift+f",
+		speakOnDemand=True,
 	)
 	def script_reportFormatting(self, gesture):
 		repeats = scriptHandler.getLastScriptRepeatCount()
@@ -2297,7 +2313,8 @@ class GlobalCommands(ScriptableObject):
 	@script(
 		# Translators: Input help mode message for report formatting at caret command.
 		description=_("Reports formatting info for the text under the caret."),
-		category=SCRCAT_SYSTEMCARET
+		category=SCRCAT_SYSTEMCARET,
+		speakOnDemand=True,
 	)
 	def script_reportFormattingAtCaret(self, gesture):
 		self._reportFormattingHelper(self._getTIAtCaret(True), False)
@@ -2317,7 +2334,8 @@ class GlobalCommands(ScriptableObject):
 			" If pressed twice, presents the information in browse mode"
 		),
 		category=SCRCAT_SYSTEMCARET,
-		gesture="kb:NVDA+f"
+		gesture="kb:NVDA+f",
+		speakOnDemand=True,
 	)
 	def script_reportOrShowFormattingAtCaret(self, gesture):
 		repeats = scriptHandler.getLastScriptRepeatCount()
@@ -2381,6 +2399,7 @@ class GlobalCommands(ScriptableObject):
 			"Report summary of any annotation details at the system caret."
 		),
 		category=SCRCAT_SYSTEMCARET,
+		speakOnDemand=True,
 	)
 	def script_reportDetailsSummary(self, gesture: inputCore.InputGesture):
 		"""Report the annotation details summary for the single character under the caret or the object with
@@ -2448,7 +2467,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing three times spells it using character descriptions."
 		),
 		category=SCRCAT_FOCUS,
-		gesture="kb:NVDA+tab"
+		gesture="kb:NVDA+tab",
+		speakOnDemand=True,
 	)
 	def script_reportCurrentFocus(self, gesture: inputCore.InputGesture):
 		focusObject=api.getFocusObject()
@@ -2527,6 +2547,7 @@ class GlobalCommands(ScriptableObject):
 			"Reads the current application status bar."
 		),
 		category=SCRCAT_FOCUS,
+		speakOnDemand=True,
 	)
 	def script_readStatusLine(self, gesture):
 		text = self._getStatusBarText()
@@ -2544,6 +2565,7 @@ class GlobalCommands(ScriptableObject):
 			"Spells the current application status bar."
 		),
 		category=SCRCAT_FOCUS,
+		speakOnDemand=True,
 	)
 	def script_spellStatusLine(self, gesture):
 		text = self._getStatusBarText()
@@ -2597,7 +2619,8 @@ class GlobalCommands(ScriptableObject):
 			"If pressed three times, copies the status bar to the clipboard"
 		),
 		category=SCRCAT_FOCUS,
-		gestures=("kb(desktop):NVDA+end", "kb(laptop):NVDA+shift+end")
+		gestures=("kb(desktop):NVDA+end", "kb(laptop):NVDA+shift+end"),
+		speakOnDemand=True,
 	)
 	def script_reportStatusLine(self, gesture):
 		text = self._getStatusBarText()
@@ -2619,6 +2642,7 @@ class GlobalCommands(ScriptableObject):
 		),
 		category=SCRCAT_FOCUS,
 		gestures=("kb:shift+numpad2", "kb(laptop):NVDA+control+shift+."),
+		speakOnDemand=True,
 	)
 	def script_reportFocusObjectAccelerator(self, gesture: inputCore.InputGesture) -> None:
 		obj = api.getFocusObject()
@@ -2706,7 +2730,7 @@ class GlobalCommands(ScriptableObject):
 		# Translators: Input help mode message for read foreground object command (usually the foreground window).
 		description=_("Reads all controls in the active window"),
 		category=SCRCAT_FOCUS,
-		gesture="kb:NVDA+b"
+		gesture="kb:NVDA+b",
 	)
 	def script_speakForeground(self,gesture):
 		obj=api.getForegroundObject()
@@ -2901,7 +2925,8 @@ class GlobalCommands(ScriptableObject):
 		# Translators: Input help mode message for report battery status command.
 		description=_("Reports battery status and time remaining if AC is not plugged in"),
 		category=SCRCAT_SYSTEM,
-		gesture="kb:NVDA+shift+b"
+		gesture="kb:NVDA+shift+b",
+		speakOnDemand=True,
 	)
 	def script_say_battery_status(self, gesture: inputCore.InputGesture) -> None:
 		reportCurrentBatteryStatus()
@@ -2926,7 +2951,8 @@ class GlobalCommands(ScriptableObject):
 			"Speaks the filename of the active application along with the name of the currently loaded appModule"
 		),
 		category=SCRCAT_TOOLS,
-		gesture="kb:NVDA+control+f1"
+		gesture="kb:NVDA+control+f1",
+		speakOnDemand=True,
 	)
 	def script_reportAppModuleInfo(self,gesture):
 		focus=api.getFocusObject()
@@ -3118,7 +3144,8 @@ class GlobalCommands(ScriptableObject):
 	@script(
 		# Translators: Input help mode message for the report current configuration profile command.
 		description=_("Reports the name of the current NVDA configuration profile"),
-		category=SCRCAT_CONFIG
+		category=SCRCAT_CONFIG,
+		speakOnDemand=True,
 	)
 	def script_reportActiveConfigurationProfile(self, gesture):
 		activeProfileName = config.conf.profiles[-1].name
@@ -3409,7 +3436,8 @@ class GlobalCommands(ScriptableObject):
 			"Pressing three times spells it using character descriptions."
 		),
 		category=SCRCAT_SYSTEM,
-		gesture="kb:NVDA+c"
+		gesture="kb:NVDA+c",
+		speakOnDemand=True,
 	)
 	def script_reportClipboardText(self,gesture):
 		try:
@@ -3803,7 +3831,8 @@ class GlobalCommands(ScriptableObject):
 			"If pressed twice, shows the URL in a window for easier review."
 		),
 		gesture="kb:NVDA+k",
-		category=SCRCAT_TOOLS
+		category=SCRCAT_TOOLS,
+		speakOnDemand=True,
 	)
 	def script_reportLinkDestination(
 			self, gesture: inputCore.InputGesture, forceBrowseable: bool = False
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -2161,7 +2161,8 @@ class GlobalCommands(ScriptableObject):
 			" moving the review cursor as it goes"
 		),
 		category=SCRCAT_TEXTREVIEW,
-		gestures=("kb:numpadPlus", "kb(laptop):NVDA+shift+a", "ts(text):3finger_flickDown")
+		gestures=("kb:numpadPlus", "kb(laptop):NVDA+shift+a", "ts(text):3finger_flickDown"),
+		speakOnDemand=True,
 	)
 	def script_review_sayAll(self, gesture: inputCore.InputGesture):
 		# This script is available on the lock screen via getSafeScripts
@@ -2172,7 +2173,8 @@ class GlobalCommands(ScriptableObject):
 		# Translators: Input help mode message for say all with system caret command.
 		description=_("Reads from the system caret up to the end of the text, moving the caret as it goes"),
 		category=SCRCAT_SYSTEMCARET,
-		gestures=("kb(desktop):NVDA+downArrow", "kb(laptop):NVDA+a")
+		gestures=("kb(desktop):NVDA+downArrow", "kb(laptop):NVDA+a"),
+		speakOnDemand=True,
 	)
 	def script_sayAll(self, gesture: inputCore.InputGesture):
 		sayAll.SayAllHandler.readText(sayAll.CURSOR.CARET)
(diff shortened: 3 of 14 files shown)


Adriani90:
Great work on this! Really highly appreciated. Does include also application specific commands like changing formating in Ms Office etc.? These should be reported as well in my view.
Also such commands like changing nvda settings might be worthy to consider.
The beep mode by the way is very useful, especially when you work with 2 sessions of Nvda one on a local and one on a virtual machine, especially when it is not allowed to use the remote addon due to restrictions via  security policies.
Naming this mode speech on demand totally makes sense, because it is consistent with other speech modes already in place.

Adriani90:
In my view say all could be implemented on a later stage, if the users demand it. It should not prevent this PR from being merged.
But how should we handle reporting commands introduced by addons such as Windows magnifier addon? It would maybe make sense to include all commands which contain the nvda key  except for the ones used for moving system focus, carret, review cursor, virtual cursor and mouse cursor. This includes also commands for navigation such as object navigation and review, jumping to last rows or columns in tables etc. or such commands for navigations introduced by addons such as golden cursor or objnav.
So maybe there could be a condition be implemented where NVDA looks for the movements of these cursors and then speech is not fired. Might be more efficient than the command by command implementation.

leonardder:
>     * Regarding UX, was it a good solution to add a fourth speech mode? There is already an add-on to remove the beep mode from the speech mode toggle command...

Could the speech modes be made opt out, at least the beeps and on demant modes? That would make the no beeps mode add-on obsolete, but I think that would be the best solution.

CyrilleB79:
Thanks @leonardder  and @Adriani90 

Answering to all:

> Does include also application specific commands like changing formating in Ms Office etc.? These should be reported as well in my view.

If you are speaking about Office commands such as toggle bold (control+U), toggle italic (control+I), etc., they are not reported in "on demand" mode. Indeed these scripts perform an action. In the "on demand" mode however, only the script whose goal is to report an information are speaking. The scripts that perform an action indeed do not speak. The use case is the following. You are in a meeting so you do not want NVDA to speak when you write notes in Word. Thus NVDA does not speak when you type or when you change formatting (bold, etc.) so that you can perform these actions without being bothered by NVDA speech. If you explicitly want to know font formatting, you still can request it with NVDA+F. The same way, the time command (NVDA+F12) speaks because you have no reason to press this command without any speech feedback.
To summarize, the commands that speak in the "on demand" mode are those that would be useless when speech is muted and with no braille.


> Also such commands like changing nvda settings might be worthy to consider.

For the same reason as before, they are not included, because you may want to toggle a parameter without being bothered by a speech feedback.

> The beep mode by the way is very useful, especially when you work with 2 sessions of Nvda one on a local and one on a virtual machine, especially when it is not allowed to use the remote addon due to restrictions via security policies.

I do not contest the usefulness of the beep mode for some users. I was just worrying of the increasing number of speech mode, when the vast majority of users will use only two of them: normal speech and either off or "on demand".

> Naming this mode speech on demand totally makes sense, because it is consistent with other speech modes already in place.

OK thanks for the positive feedback.

> In my view say all could be implemented on a later stage, if the users demand it. It should not prevent this PR from being merged.

Agree with this plan if NV Access also agrees since this PR is already an improvement. Still, I will try to investigate the say all issue in the next days.

> But how should we handle reporting commands introduced by addons such as Windows magnifier addon?

In the specific case of Windows Magnifier add-on, all the commands speak after an associated action, e.g. move something, change a setting, etc. Thus they should be muted in the "on demand" mode.

More generally however, the add-on authors may use the "speakOnDemand" parameter of the script decorator to indicate if their scripts should speak when speech mode is "on demand".

> It would maybe make sense to include all commands which contain the nvda key except for the ones used for moving system focus, carret, review cursor, virtual cursor and mouse cursor. This includes also commands for navigation such as object navigation and review, jumping to last rows or columns in tables etc. or such commands for navigations introduced by addons such as golden cursor or objnav. So maybe there could be a condition be implemented where NVDA looks for the movements of these cursors and then speech is not fired. Might be more efficient than the command by command implementation.

The use or not of NVDA key is and should not be a criterion since keys could be reassigned.

More generally, you seem to have an idea of something more verbose than I regarding the "speech on demand" mode. The mode as I have implemented it seems to match more or less the one in Jaws; thus, I imagine (since it is not so new), that the use case has been confirmed among users. If there is a need though, we may imagine different levels of verbosity for the "on demand" mode. To define a different verbosity, please provide a concrete use case as well as a precise criterion to determine if a given command (script) should speak or not.



> > ```
> > * Regarding UX, was it a good solution to add a fourth speech mode? There is already an add-on to remove the beep mode from the speech mode toggle command...
> > ```
> 
> Could the speech modes be made opt out, at least the beeps and on demant modes? That would make the no beeps mode add-on obsolete, but I think that would be the best solution.

Technically, all can be done. I am asking from a UX point of view. Is it better to add a new GUI control so that people can choose which speech mode is available in NVDA+S command? Or is it just cluttering the GUI with a parameter considered obscure for most users? In the latter case, the "no beep speech mode" add-on could be improved to provide this feature for those who need it (cc @ABuffEr).

XLTechie:
Thanks for this work @CyrilleB79.
I will review the specifics on Monday or Tuesday, depending on time.

I would like to say that I agree with @leonardder. Some checkboxes in the speech 
category to hide beeps mode, and to hide on-demand speech mode, both unchecked 
by default.

Doing that, makes including these in speech modes to be reasonable. Which is 
where I think they should be.

Given how many people desired the no beeps mode add-on over the years, making a 
checkbox for this finally does seem warranted, though I personally love beeps 
mode.

CyrilleB79:
I have just added say all scripts and input help in the "on demand" mode and updated the dev doc. And I have updated the initial description according to these last changes and also included relevant information from our discussion.

Regarding a configuration to filter the speech modes available in NVDA+S loop, I prefer to wait for NV Access review before implementing something. I am not totally convinced that it's the best way to handle the "too many speech modes" issue.
As an alternative, we can also merge this PR as is and think to the number of speech modes in NVDA+S while it is tested in alpha since a solution can easily be merged later in alpha cycle (or even after 2024.1) so that we get more feedback.

ABuffEr:
Hi,
about UX, I agree to have two checkbox, like "Include beeps in speech modes" and "Include on-demand in speech modes"; about checked status, maybe it'll be better both checked, to warn users about all actually available modes.
Anyway, I'll be absolutely happy to pension NoBeepsSpeechMode 😁

source/scriptHandler.py
```diff
@@ -349,6 +358,7 @@ def script(
 	@param allowInSleepMode: Whether this script should run when NVDA is in sleep mode.
 	@param resumeSayAllMode: The say all mode that should be resumed when active before executing this script.
 	One of the C{sayAll.CURSOR_*} constants.
+	@param speakOnDemand: Either this script should speak when NVDA speech mode is "on demand"
```
leonardder, line 361: Could you make this sphinx style while at it?
```suggestion
	:param speakOnDemand: Either this script should speak when NVDA speech mode is "on demand"
```

source/speech/speech.py
```diff
@@ -82,13 +82,15 @@ class SpeechMode(IntEnum):
...
 	talk = 2
+	onDemand = 3
 
 
 @dataclass
 class SpeechState:
 	beenCanceled = True
 	isPaused = False
-	#: How speech should be handled; one of SpeechMode.off, SpeechMode.beeps or SpeechMode.talk.
+	#: How speech should be handled; one of SpeechMode.off, SpeechMode.beeps, SpeechMode.talk. or
```
leonardder, line 92: I think it is obvious that this should be one of the enum values:
```suggestion
	#: How speech should be handled.
```

leonardder:
I would personally go for a checkable list, similar to the NVDA modifier keys list.

lukaszgo1:
I have created #15806 to discuss configurability of speech modes switching.

Adriani90:
> we may imagine different levels of verbosity for the "on demand" mode.

I think this might indeed make things to complicated. My use case was not really well developped yet in my mind, I think your approach suits better as it is. I guess if users need more commands to speak in on demand mode, then this can be implemented later on.

> I would personally go for a checkable list, similar to the NVDA modifier keys list.

I like this idea as well. The less we have to tab through dialogs, the better. :-) However, there is a risk with these checkboxes: imagine a new user disables by accident the speech mode speak and presses nvda+s afterwards, NVDA would change to one of the less verbose or no verbose modes. There is no way of enabling the speak mode again unless you know the menu structure by hard and you know how many times you have to tab around to include speech mode speak again into the nvda+s command.

Another better alternative in my view would be to let both speech mode speak and speech mode off as they are, and include also a speech mode category in the synth settings ring which contains all 4 speech modes. Then, in case you enable speech mode on demand, beeps or off from the synth settings ring, pressing nvda+s switches to speech mode speak again. So speech mode speak has always priority when pressing nvda+s.

Adriani90:
Ofcourse my proposal would mean we have to enhance the synth settings ring to a speech and synth settings ring but this would improve things anyway.

source/scriptHandler.py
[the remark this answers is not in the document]
CyrilleB79: I had erroneously replace "whether" by "either"...

Adriani90:
I can imagine this checkable list could directly enable or disable the speech mode itself instead of its inclusion in a command. So if I check the checkbox on demand, then the on demand mode is enabled, then I can also change to another mode via the synth settings ring, but when I press nvda+s speech mode speak is resumed.

CyrilleB79:
@lukaszgo1  and @leonardder, thanks for the review. I have applied most of your suggestions in 72ca5b4af2b9eb73d05f2dd1746d5f2bf5ac4fa2 and resolved the corresponding comments.

CyrilleB79:
> @lukaszgo1 and @leonardder, thanks for the review. I have applied most of your suggestions in [72ca5b4](https://github.com/nvaccess/nvda/commit/72ca5b4af2b9eb73d05f2dd1746d5f2bf5ac4fa2) and resolved the corresponding comments.

Actually some review action have also been pushed in 4a35f3110c1d9a68a970eb0c314ca64057294c92.

XLTechie:
@Adriani90 wrote:

> Another better alternative in my view would be to let both speech mode speak and speech mode off as they are, and include also a speech mode category in the
> synth settings ring which contains all 4 speech modes. Then, in case you enable speech mode on demand, beeps or off from the synth settings ring, pressing
> nvda+s switches to speech mode speak again. So speech mode speak has always priority when pressing nvda+s.

I agree that this is overly complex--not in ability to use, but in time to get 
to the options on the rotor.

> there is a risk with these checkboxes: imagine a new user disables by accident the speech mode speak and presses nvda+s afterwards

I am having trouble understanding the use case for making regular speech 
uncheckable (or even having it on the list).
These are all "alternate" speech modes. But "speech" is the primary mode.

Beeps: many users don't understand or have use for it.
On-demand: only useful in certain circumstances.
Off: with the first two available, I personally will not want "off" cluttering 
up my speech modes selector, as it has no use to me then.

But disabling speech? If your only use for NVDA is for while on-demand is 
active, which seems a rare use case at best, nothing stops you from NVDA+s'ing 
as soon as you start, or having it set that way in your default profile.

Or is this a lack of imagination on my part?

Adriani90:
> I agree that this is overly complex--not in ability to use, but in time to get
to the options on the rotor.

This could be solved if a checkable list is implemented which lets you add or remove categories from the settings ring.

> I am having trouble understanding the use case for making regular speech
uncheckable (or even having it on the list).

Good point. Not having it in the list would definitely be the easiest solution.
After thinking more about it, I think the two checkboxes proposed by @XLTechie would be the safest way sofar with lower implementation efforts.

CyrilleB79:
@Adriani90, @XLTechie:
Now that #15806 is open, I'd prefer focus on the core of this PR, i.e. the speech "on demand" mode. And discuss the number and availability of modes in NVDA+S in #15806.

Regarding the synth ring selector, today it is restricted to synthesizer parameters. If you want to discuss the generalization of the parameters available in the ring, please do it in #7747. Thanks.

source/scriptHandler.py
```diff
@@ -324,6 +324,13 @@ def clearLastScript():
 	_lastScriptCount = 0
 
 
+def getCurrentScript():
```
seanbudd, line 327: can this please get a return type added

source/NVDAObjects/IAccessible/winword.py
```diff
@@ -275,6 +275,12 @@ def script_caret_moveByCell(self,gesture):
 			speech.speakTextInfo(info, reason=controlTypes.OutputReason.FOCUS)
 			braille.handler.handleCaretMove(self)
 
+	@script(
+		# Translators: a description for a script
+		description=_("Reports the text of the comment where the System caret is located."),
```
seanbudd, line 280:
```suggestion
		description=_("Reports the text of the comment where the system caret is located."),
```

source/NVDAObjects/UIA/wordDocument.py
```diff
@@ -600,7 +600,8 @@ def _caretMoveBySentenceHelper(self, gesture, direction):
 	@script(
 		gesture="kb:NVDA+alt+c",
 		# Translators: a description for a script that reports the comment at the caret.
-		description=_("Reports the text of the comment where the System caret is located.")
+		description=_("Reports the text of the comment where the System caret is located."),
```
seanbudd, line 603:
```suggestion
		description=_("Reports the text of the comment where the system caret is located."),
```

user_docs/en/changes.t2t
```diff
@@ -10,6 +10,7 @@ What's New in NVDA
 - Added support for Bluetooth Low Energy HID Braille displays. (#15470)
 - The Add-on Store now supports installing add-ons in bulk by selecting multiple add-ons. (#15350)
 - From the add-on store, it's possible to open a dedicated webpage to see or provide feedback about the selected add-on. (#15576, @nvdaes)
+- A new "speech on demand" mode has been added. When speech is on demand, NVDA does not speak automatically (e.g. when moving the cursor) but still speaks when calling commands whose goal is explicitly to report something (e.g. report window title). (#481, @CyrilleB79)
```
seanbudd, line 13:
```suggestion
- A new "speech on demand" mode has been added.
When speech is on demand, NVDA does not speak automatically (e.g. when moving the cursor) but still speaks when calling commands whose goal is explicitly to report something (e.g. report window title). (#481, @CyrilleB79)
```

user_docs/en/changes.t2t
```diff
@@ -94,6 +95,10 @@ Setting ``numCells`` is still supported for single line braille displays and ``n
   - wil to ``5e9be7b2d2fe3834a7107f430f7d4c0631f69833``. (#15695)
   -
 - When the configuration file ``nvda.ini`` is corrupted, a backup copy is saved before it is reinitialized. (#15779, @CyrilleB79)
+- When defining a script with the script decorator, the 'speakOnDemand' boolean argument can be specified to control if a script should speak while in "on demand" mode. (#481, @CyrilleB79)
```
seanbudd, line 98:
```suggestion
- When defining a script with the script decorator, the ``speakOnDemand`` boolean argument can be specified to control if a script should speak while in "on demand" mode. (#481, @CyrilleB79)
```

user_docs/en/userGuide.t2t
```diff
@@ -504,7 +504,7 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Stop speech | Control | control | 2-finger tap | Instantly stops speaking |
 | Pause Speech | shift | shift | none | Instantly pauses speech. Pressing it again will continue speaking where it left off (if pausing is supported by the current synthesizer) |
 | NVDA Menu | NVDA+n | NVDA+n | 2-finger double-tap | Pops up the NVDA menu to allow you to access preferences, tools, help, etc. |
-| Toggle Speech Mode | NVDA+s | NVDA+s | none | Toggles speech mode between speech, beeps and off. |
+| Toggle Speech Mode | NVDA+s | NVDA+s | none | Toggles speech mode between speech, speech on demand, beeps and off. |
```
seanbudd, line 507:
```suggestion
| Toggle Speech Mode | ``NVDA+s`` | ``NVDA+s`` | none | Toggles speech mode between speech, speech on demand, beeps and off. |
```

seanbudd, line 507: I think it would be worth adding a new user guide section "Speech mode" to explain each of these modes. This will also be useful when creating #15806

source/speech/speech.py
```diff
@@ -950,6 +951,18 @@ def speak(  # noqa: C901
 		return
 	if _speechState.isPaused:
 		cancelSpeech()
+	if _speechState.speechMode == SpeechMode.onDemand:
+		
+		import inputCore
+		from scriptHandler import getCurrentScript
+		from .sayAll import SayAllHandler
+		script = getCurrentScript()
+		if not (
```
seanbudd, line 960: is it worth considering speaking high priority speech (e.g. speechPriority.NOW) and speech from `ui.message` in this mode?

seanbudd (commented):
>Regarding UX, was it a good solution to add a fourth speech mode? There is already an add-on to remove the beep mode from the speech mode toggle command...

As it has been said, I think we can work this out in #15806

> For documentation, what should be the correct name to use? "speech on demand", "speak on demand", etc. Also, speak or talk? @XLTechie would you mind review documentation (change log, user guide, command description)?

I'll wait for @Qchristensen thoughts here, I think either is fine

>Do I need to include an API change item when extending the speech mode enum?

Not needed, no

codeofdusk:
what should be the correct name to use? "speech on demand", "speak on demand", etc.

I think this new mode should be called "on demand", without "speech" or "speak" etc. Then, <kbd>nvda</kbd>+<kbd>s</kbd> reports "speech mode on demand".

source/speech/speech.py
XLTechie: @seanbudd Are you suggesting a checkbox in the UI to include high priority messages, or are you suggesting including them always?
In the general case, I think the answer to your question is "no". One of the purposes for this mode, is to prevent the screen reader from blurting things out at inconvenient times. Imagine you're doing some recording through a microphone, and suddenly an item that NVDA considers high priority pops up? If I'm using this feature, I've already decided that nothing the computer could possibly say, is more important than the reason I want it being quiet. And yet, I still want the ability to quickly use screen review or similar commands.

Yes, it means that I might miss a low battery notification or such, but if that's important to me there's probably a Windows sound I can rely upon to give me the idea that something critical is happening, without speech suddenly coming out of nowhere.

Although for the magnification user, perhaps the needs are different? (@lukaszgo1)

user_docs/en/changes.t2t
XLTechie, line 13:
```suggestion
- A new "on-demand" speech mode has been added. When speech is on-demand, NVDA does not speak automatically (e.g. when moving the cursor) but still speaks when calling commands whose goal is explicitly to report something (e.g. report window title). (#481, @CyrilleB79)
```

user_docs/en/userGuide.t2t
XLTechie: @seanbudd can this be done in a follow-up?

XLTechie (approved):
Other than some question as to whether it should be "on-demand" instead of "on demand" (See https://editorninja.com/is-on-demand-hyphenated/), and a comment I already left about calling it "on-demand mode" (or "on demand mode") vs "speak on-demand" or other options, the changelog and user guide changes look fine to me.

user_docs/en/userGuide.t2t
CyrilleB79: I agree that a more detailed description in the User Guide would be nice.

@seanbudd, @Qchristensen  or any other reviewer, where would you see this more detailed description in the User Guide? It's a bit tricky since speech mode is not a configuration parameter, i.e. it's not saved when restarting NVDA.

source/speech/speech.py
CyrilleB79: > is it worth considering speaking high priority speech (e.g. speechPriority.NOW) and speech from ui.message in this mode?

Not at all. This mode means that NVDA is not speaking unless when calling commands that only report information, i.e.
perform no other action.
* High priority speech can occur at any time while the user press no command at all, e.g. alert event. This would defeat the goal of this PR. E.g. I am in a meeting and want to hear no speech feedback unless when I ask one explicitely.
* `ui.message` is used in many more script that do not belong to "on demand" mode, e.g. toggle a setting (e.g. audio ducking), perform mouse action (e.g. left click), etc.

user_docs/en/userGuide.t2t
XLTechie: > any other reviewer, where would you see this more detailed description in the User Guide? It's a bit tricky since speech mode is not a configuration parameter, i.e. it's not saved when restarting NVDA.


Either in Chapter `4. Using NVDA`, specifically as a new section `4.8. Speech Modes`; or
in chapter 12, subchapter `12.2. miscellaneous Settings`, specifically in a new section `12.2.4. Speech Modes`.

user_docs/en/changes.t2t
CyrilleB79: Fixed comments of @seanbudd and @XLTechie  on this line in 65dbc8ac6df72b1b73bcecb131d1fc20a96f127d.

source/scriptHandler.py
CyrilleB79: Fixed in 4a8a7db9ad5a670f9eb4a62e3392e3cd36a1d050

source/globalCommands.py
```diff
@@ -2008,18 +2023,20 @@ def script_review_currentSymbol(self,gesture):
 	@script(
 		description=_(
 			# Translators: Input help mode message for toggle speech mode command.
-			"Toggles between the speech modes of off, beep and talk. "
+			"Toggles between the speech modes of off, beep, talk and on-demand. "
 			"When set to off NVDA will not speak anything. "
 			"If beeps then NVDA will simply beep each time it its supposed to speak something. "
 			"If talk then NVDA will just speak normally."
+			"If on-demand then NVDA will only speak when calling commands whose goal is to speak something, but not"
```
XLTechie, line 2030: I think this may be too long for comfort in input help. Perhaps:

```suggestion
			"If on-demand, NVDA will only speak when commands which have no purpose other than to provide feedback are used. "
```

source/globalCommands.py
```diff
@@ -2008,18 +2023,20 @@ def script_review_currentSymbol(self,gesture):
 	@script(
 		description=_(
 			# Translators: Input help mode message for toggle speech mode command.
-			"Toggles between the speech modes of off, beep and talk. "
+			"Toggles between the speech modes of off, beep, talk and on-demand. "
 			"When set to off NVDA will not speak anything. "
 			"If beeps then NVDA will simply beep each time it its supposed to speak something. "
 			"If talk then NVDA will just speak normally."
+			"If on-demand then NVDA will only speak when calling commands whose goal is to speak something, but not"
+			" when performing an action such as moving the focus or the cursor."
```
XLTechie, line 2031: Of course, if you accept my former suggestion, this line should go away.

However, more broadly, as it is likely some version of #15806 will be implemented before this leaves alpha, it may be better to shorten this entire long message to something like: "Cycles through the available speech modes."

```suggestion
```

XLTechie (changes requested):

codeofdusk:
> Other than some question as to whether it should be "on-demand" instead of "on demand"

"on demand"

source/speech/speech.py
lukaszgo1: I agree with what been said by both @CyrilleB79  and @XLTechie  here. Also to be extra clear I am not a magnification user, so my opinion's in that regard are mostly brain dumps on how someone in that situation may want to use NVDA, and should not be considered anything conclusive.

CyrilleB79:
In the end, I have added a paragraph in section 4 of the User Guide. Not in section 12 since section 12 describes the GUI (menus and dialogs) whereas this option is not configurable through dialogs.

Since the description of each mode is now in the User Guide, I have shortened the input help which was way too long: I have removed the descriptions of each speech mode from input help. @XLTechie, I have taken into account your review on the input help to write the paragraph in the User Guide.

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,20 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Report clipboard text | NVDA+c | Reports the Text on the clipboard if there is any. |
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+There are four speech modes available.
+* Talk: NVDA will just speak normally. This is the default speech mode.
+* On-demand: NVDA will only speak when calling commands whose goal is to speak an information (e.g. report the title of the window); but it will not speak in reaction to actions such as moving the focus or the cursor.
+* Off: NVDA will not speak anything. 
+* Beeps: NVDA will replace normal speech with quiet beeps.
```
seanbudd, line 527: I believe these lists also need a terminator. Is there a reason `*` was used over `-`?

```suggestion
* Beeps: NVDA will replace normal speech with quiet beeps.
*
```

seanbudd, line 527: I believe these lists also need a terminator. Is there a reason `*` was used over `-`?

```suggestion
* Beeps: NVDA will replace normal speech with quiet beeps.
*
```

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,20 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
...
 
+++ Speech modes ++[SpeechModes]
+
+There are four speech modes available.
+* Talk: NVDA will just speak normally. This is the default speech mode.
+* On-demand: NVDA will only speak when calling commands whose goal is to speak an information (e.g. report the title of the window); but it will not speak in reaction to actions such as moving the focus or the cursor.
+* Off: NVDA will not speak anything. 
+* Beeps: NVDA will replace normal speech with quiet beeps.
+
+A gesture allows to cycle through the various speech modes:
```
seanbudd, line 529:
```suggestion
A gesture allows cycling through the various speech modes:
```

seanbudd (approved):
Thanks @CyrilleB79

source/globalCommands.py
```diff
@@ -2008,18 +2023,15 @@ def script_review_currentSymbol(self,gesture):
 	@script(
 		description=_(
 			# Translators: Input help mode message for toggle speech mode command.
```
XLTechie, line 2025: Shouldn't this now be "cycle speech modes command", or "Switch to next speech mode command"? It is no longer a toggle.

```suggestion
			# Translators: Input help mode message for switch to next speech mode command.
```

XLTechie (changes requested):
I'm pretty sure this command should no longer be called a toggle. You can't toggle between multiple options, not in the normal usage of that word (N.B. I am aware of three position toggle switches, but that's not how people normally understand it, I think).

source/globalCommands.py
```diff
@@ -2008,18 +2023,15 @@ def script_review_currentSymbol(self,gesture):
...
-			"Toggles between the speech modes of off, beep and talk. "
-			"When set to off NVDA will not speak anything. "
-			"If beeps then NVDA will simply beep each time it its supposed to speak something. "
-			"If talk then NVDA will just speak normally."
+			"Toggles between the speech modes of off, beep, talk and on-demand."
 		),
 		category=SCRCAT_SPEECH,
 		gesture="kb:NVDA+s"
 	)
 	def script_speechMode(self,gesture):
```
XLTechie, line 2031: Can this be linted?

user_docs/en/changes.t2t
```diff
@@ -10,6 +10,8 @@ What's New in NVDA
 - Added support for Bluetooth Low Energy HID Braille displays. (#15470)
 - The Add-on Store now supports installing add-ons in bulk by selecting multiple add-ons. (#15350)
 - From the add-on store, it's possible to open a dedicated webpage to see or provide feedback about the selected add-on. (#15576, @nvdaes)
+- A new "on-demand" speech mode has been added.
+When speech is on-demand, NVDA does not speak automatically (e.g. when moving the cursor) but still speaks when calling commands whose goal is explicitly to report something (e.g. report window title). (#481, @CyrilleB79)
```
XLTechie, line 14: Should this be referencing the issue number or the PR number?

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,20 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Report clipboard text | NVDA+c | Reports the Text on the clipboard if there is any. |
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+There are four speech modes available.
+* Talk: NVDA will just speak normally. This is the default speech mode.
+* On-demand: NVDA will only speak when calling commands whose goal is to speak an information (e.g. report the title of the window); but it will not speak in reaction to actions such as moving the focus or the cursor.
```
XLTechie, line 525: Since I suspect you will change this to a hyphenated list, I have used that in the below to make it easy to apply my suggestion if you otherwise like it.

I have removed "an information" because "information" is plural in English. I have replaced "calling" (which is often understood in technical circles but may not be how users think of command use), with a more user friendly term.
Lastly, I have added some further clarification about what this mode does.

```suggestion
- On-demand: NVDA will only speak when you use commands which speak information (e.g. report the title of the window); but it will not speak in reaction to actions such as moving the focus or the cursor; and will not speak any screen changes, notifications, etc.
```

user_docs/en/userGuide.t2t
XLTechie, line 527: I know "quiet" was my word, but it was a mistake.
Also, I have added some explanatory paragraphs about beeps and on-demand modes, since people are always asking what beeps mode is good for.
Lastly, I have used a hyphen for the list marker, and closed the list, since I think you will likely replace the stars used on the other items.

```suggestion
- Beeps: NVDA will replace normal speech with short beeps.
-

Beeps mode may be useful when some very verbose output is scrolling in a terminal window, but you don't care what it is, only that it is continuing to scroll; or in other circumstances when the fact that there is output is more relevant than the output itself.

The on-demand mode may be valuable when you don't need constant feedback about what is happening on screen or on the computer, but you periodically need to check particular things using review commands, etc.
Examples include while recording audio, when using screen magnification, or as an alternative to beeps mode.
```

user_docs/en/userGuide.t2t
```diff
@@ -4049,7 +4062,7 @@ Please see the display's documentation for descriptions of where these keys can
 | Route to braille cell | ``routing`` |
 | Report text formatting under braille cell | ``secondary routing`` |
 | Toggle the way context information is presented in braille | ``attribute1+attribute3`` |
-| Toggles between the speech modes of off, beep and talk | ``attribute2+attribute4`` |
+| Toggles between the speech modes of off, beep, speech and on-demand | ``attribute2+attribute4`` |
```
XLTechie, line 4065:
```suggestion
| Toggles between the speech modes of off, beeps, talk and on-demand | ``attribute2+attribute4`` |

```

XLTechie (changes requested):

wmhn1872265132:
It is recommended that the table introducing the Toggles the speech mode shortcut keys in the user guide be simplified into three columns: Name, Key, and Description to enhance the reading experience.

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,20 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Report clipboard text | NVDA+c | Reports the Text on the clipboard if there is any. |
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+There are four speech modes available.
+* Talk: NVDA will just speak normally. This is the default speech mode.
```
Qchristensen, line 524: Small thing, but I don't like using "just" here.  Someone will be reading the user guide because they don't know how things work and "just", to me, implies it is something they should have already known (there are articles on this such as https://www.excitant.co.uk/just-is-a-dangerous-word/ ).  Perhaps:

Talk: The Default speech mode, NVDA speaks in reaction to actions such as moving the focus, notifications, or issuing commands.

Also, for consistency, should this group be in the same order as the command itself and the list on line 4065 (last change in this PR)

Qchristensen (changes requested):
Looks good.  I made one small suggestion

user_docs/en/userGuide.t2t
XLTechie: I agree with @Qchristensen here. I had removed "just" in my original suggestion on the input help text for this, but it slipped back in to the user guide version. I figured @CyrilleB79 really wanted it there, so I didn't comment further.

```suggestion
- Talk: NVDA will speak normally in reaction to screen changes, notifications, and actions such as moving the focus, or issuing commands.
```

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,20 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Report clipboard text | NVDA+c | Reports the Text on the clipboard if there is any. |
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+There are four speech modes available.
+* Talk: NVDA will just speak normally. This is the default speech mode.
+* On-demand: NVDA will only speak when calling commands whose goal is to speak an information (e.g. report the title of the window); but it will not speak in reaction to actions such as moving the focus or the cursor.
+* Off: NVDA will not speak anything.
```
XLTechie, line 526:
```suggestion
- Off: NVDA will not speak anything, though unlike sleep mode it will act on commands. 
```

user_docs/en/userGuide.t2t
CyrilleB79: Thanks for the formatting fix. No reason to use * instead of -. I'll change it.

Adriani90:
Can you change the sentence for on-demand mode in something like this to make it more readable for people who need easy language?

>- On-demand: NVDA will speak only when pressing a key command with a reporting function (e.g. report the title of the window)

> - It will not speak in reaction to actions such as moving the focus or copying content to clipboard.

 

 

 

Von: Cyrille Bougot ***@***.***> 
Gesendet: Donnerstag, 23. November 2023 21:38
Betreff: Re: [nvaccess/nvda] Add a speech "on demand" mode (PR #15804)

 

@CyrilleB79 commented on this pull request.

  _____  

In user_docs/en/userGuide.t2t <https://github.com/nvaccess/nvda/pull/15804#discussion_r1403711191> :

> @@ -519,6 +518,20 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Report clipboard text | NVDA+c | Reports the Text on the clipboard if there is any. |
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+There are four speech modes available.
+* Talk: NVDA will just speak normally. This is the default speech mode.
+* On-demand: NVDA will only speak when calling commands whose goal is to speak an information (e.g. report the title of the window); but it will not speak in reaction to actions such as moving the focus or the cursor.
+* Off: NVDA will not speak anything. 
+* Beeps: NVDA will replace normal speech with quiet beeps.

Thanks for the formatting fix. No reason to use * instead of -. I'll change it.

user_docs/en/changes.t2t
CyrilleB79: I have put the issue number as it is usually done. If you see a good reason to use the PR number here instead, please let it know.
In any case, someone who want to find the PR will find the reference in the issue.

user_docs/en/userGuide.t2t
CyrilleB79: I'll take this comment into account. But I want to keep the idea that the goal of the command is to report information. Writing "a command which speaks information" is too broad in my view.

user_docs/en/userGuide.t2t
CyrilleB79: Thanks. I'll include it and add one more example: during a meeting or a call.

user_docs/en/userGuide.t2t
CyrilleB79: @XLTechie, I had missed this from your feedback, but keeping "just" was not intended. Thanks for having insisted on this point.

CyrilleB79:
Thanks @XLTechie, @Qchristensen, @Adriani90, @wmhn1872265132 and @seanbudd for all these valuable ideas and comments to improve the documentation.

I have tried to take into account the ideas from everyone which were very valuable.

Let me know if some point is still missing.

user_docs/en/userGuide.t2t
CyrilleB79: @Qchristensen wrote:
> Also, for consistency, should this group be in the same order as the command itself and the list on line 4065 (last change in this PR)

Yes for consistency, provided "Talk" is in first position since it is the normal mode.

seanbudd (approved):

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,26 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Report clipboard text | NVDA+c | Reports the Text on the clipboard if there is any. |
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+There are four speech modes available.
```
XLTechie, line 523: Since this is the first time speech modes have been discussed at all, jumping right into it without some introduction feels a bit bumpy. Perhaps something like this?

```suggestion
The speech mode, governs how screen content, notifications, responses to commands, and other output, is spoken during operation of NVDA.
The default mode is "talk", and provides for the kind of speech output that is usually expected when using a screen reader.
However, under certain circumstances, or when running particular programs, you may find one of the other speech modes valuable.

The four available speech modes are:
```

XLTechie (approved):
One final comment for your possible consideration, but either way LGTM!
Thanks for going back and forth with me and others over the details.

user_docs/en/userGuide.t2t
CyrilleB79: Very nice and relevant suggestion indeed @XLTechie. Many thanks to you for all your review work and suggestions on this PR.

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,30 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 | Report clipboard text | NVDA+c | Reports the Text on the clipboard if there is any. |
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+The speech mode, governs how screen content, notifications, responses to commands, and other output, is spoken during operation of NVDA.
+The default mode is "talk", and provides for the kind of speech output that is usually expected when using a screen reader.
+However, under certain circumstances, or when running particular programs, you may find one of the other speech modes valuable.
```
seanbudd, line 525: Grammar fix

```suggestion
The speech mode governs how screen content, notifications, responses to commands, and other output is spoken during operation of NVDA.
The default mode is "talk", which outputs speech in situations that are expected when using a screen reader.
However, under certain circumstances, or when running particular programs, you may find one of the other speech modes valuable.
```

user_docs/en/userGuide.t2t
```diff
@@ -519,6 +518,30 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
...
 %kc:endInclude
 
+++ Speech modes ++[SpeechModes]
+
+The speech mode, governs how screen content, notifications, responses to commands, and other output, is spoken during operation of NVDA.
+The default mode is "talk", and provides for the kind of speech output that is usually expected when using a screen reader.
+However, under certain circumstances, or when running particular programs, you may find one of the other speech modes valuable.
+
+The four available speech modes are:
+- Talk: The Default speech mode, NVDA will speak normally in reaction to screen changes, notifications, and actions such as moving the focus, or issuing commands.
```
seanbudd, line 528:
```suggestion
- Talk (Default): NVDA will speak normally in reaction to screen changes, notifications, and actions such as moving the focus, or issuing commands.
```

Qchristensen (approved):
Look good, great work everyone!

PR #12773, opened 2021-08-24
Linting updates

trypsynth:
### Link to issue number:
None
### Summary of the issue:
I know we've been trying to cut back on linting issues. I tried to tackle this a few months back, but didn't really know what I was doing. Let me know how it looks now.
### Description of how this pull request fixes the issue:
I linted a few files (buildVersion.py, XMLFormatting.py, wincon.py, and a few others). I've also created a new branch (lintingUpdates), if people wish to use it for future changes such as these.
### Testing strategy:
NVDA runs from source with no issues.
### Known issues with pull request:
None
### Change log entries:
B=None
### Code Review Checklist:
- [x] Pull Request description is up to date.
- [x] Unit tests.
- [x] System (end to end) tests.
- [x] Manual testing.
- [x] User Documentation.
- [x] Change log entry.
- [x] Context sensitive help for GUI changes.
- [x] UX of all users considered:
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

PR #15823, opened 2023-11-24
UIA handler: work directly with UIA interfaces introduced in Windows 8 and later

josephsl:
### Link to issue number:
None

### Summary of the issue:
NVDA checks for Windows 7/8 UIA interfaces at startup when the minimum version is Windows 8.1.

None

At startup, NVDA's UIA handler thread will instantiate Widnows 8 UIA interface direclty as opposed tr doing it behind a try/except block. This also removes the check for UIA8 flag when instantiating later UIA interfaces. As part of this work, UIA handler main file was linted and some functions marked as too complex (according to Flake8).

### Testing strategy:
Manual testing: making sure UIA works as expected (at least on Windows 11).

None

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/UIAHandler/__init__.py
+++ b/source/UIAHandler/__init__.py
@@ -1,5 +1,5 @@
 # A part of NonVisual Desktop Access (NVDA)
-# Copyright (C) 2008-2022 NV Access Limited, Joseph Lee, Babbage B.V., Leonard de Ruijter, Bill Dengler
+# Copyright (C) 2008-2023 NV Access Limited, Joseph Lee, Babbage B.V., Leonard de Ruijter, Bill Dengler
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
 
--- a/source/UIAHandler/__init__.py
+++ b/source/UIAHandler/__init__.py
@@ -450,12 +450,7 @@ class UIAHandler(COMObject):
 	def MTAThreadFunc(self):
 		try:
 			oledll.ole32.CoInitializeEx(None,comtypes.COINIT_MULTITHREADED) 
-			isUIA8=False
-			try:
-				self.clientObject=CoCreateInstance(CUIAutomation8._reg_clsid_,interface=IUIAutomation,clsctx=CLSCTX_INPROC_SERVER)
-				isUIA8=True
-			except (COMError,WindowsError,NameError):
-				self.clientObject=CoCreateInstance(CUIAutomation._reg_clsid_,interface=IUIAutomation,clsctx=CLSCTX_INPROC_SERVER)
+			self.clientObject=CoCreateInstance(CUIAutomation8._reg_clsid_,interface=IUIAutomation,clsctx=CLSCTX_INPROC_SERVER)
 			# #7345: Instruct UIA to never map MSAA winEvents to UIA propertyChange events.
 			# These events are not needed by NVDA, and they can cause the UI Automation client library to become unresponsive if an application firing winEvents has a slow message pump. 
 			pfm=self.clientObject.proxyFactoryMapping
--- a/source/UIAHandler/__init__.py
+++ b/source/UIAHandler/__init__.py
@@ -479,20 +479,19 @@ class UIAHandler(COMObject):
 					# Therefore remove the entry and re-insert it.
 					pfm.removeEntry(index)
 					pfm.insertEntry(index,e)
-			if isUIA8:
-				# #8009: use appropriate interface based on highest supported interface.
-				# #8338: made easier by traversing interfaces supported on Windows 8 and later in reverse.
-				for interface in reversed(CUIAutomation8._com_interfaces_):
-					try:
-						self.clientObject=self.clientObject.QueryInterface(interface)
-						break
-					except COMError:
-						pass
-				# Windows 10 RS5 provides new performance features for UI Automation including event coalescing and connection recovery. 
-				# Enable all of these where available.
-				if isinstance(self.clientObject,IUIAutomation6):
-					self.clientObject.CoalesceEvents=CoalesceEventsOptions_Enabled
-					self.clientObject.ConnectionRecoveryBehavior=ConnectionRecoveryBehaviorOptions_Enabled
+			# #8009: use appropriate interface based on highest supported interface.
+			# #8338: made easier by traversing interfaces supported on Windows 8 and later in reverse.
+			for interface in reversed(CUIAutomation8._com_interfaces_):
+				try:
+					self.clientObject=self.clientObject.QueryInterface(interface)
+					break
+				except COMError:
+					pass
+			# Windows 10 RS5 provides new performance features for UI Automation including event coalescing and connection recovery. 
+			# Enable all of these where available.
+			if isinstance(self.clientObject,IUIAutomation6):
+				self.clientObject.CoalesceEvents=CoalesceEventsOptions_Enabled
+				self.clientObject.ConnectionRecoveryBehavior=ConnectionRecoveryBehaviorOptions_Enabled
 			log.info("UIAutomation: %s"%self.clientObject.__class__.__mro__[1].__name__)
 			self.windowTreeWalker=self.clientObject.createTreeWalker(self.clientObject.CreateNotCondition(self.clientObject.CreatePropertyCondition(UIA_NativeWindowHandlePropertyId,0)))
 			self.windowCacheRequest=self.clientObject.CreateCacheRequest()
--- a/source/UIAHandler/__init__.py
+++ b/source/UIAHandler/__init__.py
@@ -481,17 +481,18 @@ class UIAHandler(COMObject):
 					pfm.insertEntry(index,e)
 			# #8009: use appropriate interface based on highest supported interface.
 			# #8338: made easier by traversing interfaces supported on Windows 8 and later in reverse.
-			for interface in reversed(CUIAutomation8._com_interfaces_):
+			for interface in reversed(UIA.CUIAutomation8._com_interfaces_):
 				try:
-					self.clientObject=self.clientObject.QueryInterface(interface)
+					self.clientObject = self.clientObject.QueryInterface(interface)
 					break
 				except COMError:
 					pass
-			# Windows 10 RS5 provides new performance features for UI Automation including event coalescing and connection recovery. 
+			# Windows 10 RS5 provides new performance features for UI Automation
+			# including event coalescing and connection recovery.
 			# Enable all of these where available.
-			if isinstance(self.clientObject,IUIAutomation6):
-				self.clientObject.CoalesceEvents=CoalesceEventsOptions_Enabled
-				self.clientObject.ConnectionRecoveryBehavior=ConnectionRecoveryBehaviorOptions_Enabled
+			if isinstance(self.clientObject, UIA.IUIAutomation6):
+				self.clientObject.CoalesceEvents = UIA.CoalesceEventsOptions_Enabled
+				self.clientObject.ConnectionRecoveryBehavior = UIA.ConnectionRecoveryBehaviorOptions_Enabled
 			log.info(f"UIAutomation: {self.clientObject.__class__.__mro__[1].__name__}")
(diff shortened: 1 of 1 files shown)


source/UIAHandler/__init__.py
```diff
@@ -68,15 +68,14 @@
...
-
+HorizontalTextAlignment_Left = 0
+HorizontalTextAlignment_Centered = 1
+HorizontalTextAlignment_Right = 2
+HorizontalTextAlignment_Justified = 3
 
 
 # The name of the WDAG (Windows Defender Application Guard) process
-WDAG_PROCESS_NAME=u'hvsirdpclient'
+WDAG_PROCESS_NAME = u'hvsirdpclient'
```
seanbudd, line 78:
```suggestion
WDAG_PROCESS_NAME = "hvsirdpclient"
```

source/UIAHandler/__init__.py
```diff
@@ -68,15 +68,14 @@
...
 
-HorizontalTextAlignment_Left=0
-HorizontalTextAlignment_Centered=1
-HorizontalTextAlignment_Right=2
-HorizontalTextAlignment_Justified=3
-
+HorizontalTextAlignment_Left = 0
+HorizontalTextAlignment_Centered = 1
+HorizontalTextAlignment_Right = 2
+HorizontalTextAlignment_Justified = 3
```
seanbudd, line 74: If you are touching this part of code, please make it in an enum

source/UIAHandler/__init__.py
```diff
@@ -441,78 +440,84 @@ def terminate(self):
...
+			# including event coalescing and connection recovery.
+			# Enable all of these where available.
+			if isinstance(self.clientObject, UIA.IUIAutomation6):
+				self.clientObject.CoalesceEvents = UIA.CoalesceEventsOptions_Enabled
+				self.clientObject.ConnectionRecoveryBehavior = UIA.ConnectionRecoveryBehaviorOptions_Enabled
+			log.info(f"UIAutomation: {self.clientObject.__class__.__mro__[1].__name__}")
+			self.windowTreeWalker = self.clientObject.createTreeWalker(
+				self.clientObject.CreateNotCondition(
+					self.clientObject.CreatePropertyCondition(
+						UIA.UIA_NativeWindowHandlePropertyId, 0
```
seanbudd, line 499:
```suggestion
						UIA.UIA_NativeWindowHandlePropertyId,
						0
```

source/UIAHandler/__init__.py
```diff
@@ -689,7 +694,10 @@ def func():
 			self._localEventHandlerGroupElements.remove(element)
 		self.MTAThreadQueue.put_nowait(func)
 
-	def IUIAutomationEventHandler_HandleAutomationEvent(self,sender,eventID):
+	# C901: 'IUIAutomationEventHandler_HandleAutomationEvent' is too complex
+	# Note: when working on IUIAutomationEventHandler_HandleAutomationEvent, look for opportunities to simplify
+	# and move logic out into smaller helper functions.
+	def IUIAutomationEventHandler_HandleAutomationEvent(self, sender, eventID):  # noqa: C901
```
seanbudd, line 700: please add typing here

source/UIAHandler/__init__.py
```diff
@@ -770,30 +778,34 @@ def IUIAutomationEventHandler_HandleAutomationEvent(self,sender,eventID):
...
+		eventHandler.queueEvent(NVDAEventName, obj)
 
 	# The last UIAElement that received a UIA focus event
-	# This is updated no matter if this is a native element, the window is UIA blacklisted by NVDA, or  the element is proxied from MSAA 
-	lastFocusedUIAElement=None
-
-	def IUIAutomationFocusChangedEventHandler_HandleFocusChangedEvent(self,sender):
+	# This is updated no matter if this is a native element, the window is UIA blacklisted by NVDA,
+	# or  the element is proxied from MSAA
+	lastFocusedUIAElement = None
```
seanbudd, line 803: please add typing here

source/UIAHandler/__init__.py
```diff
@@ -770,30 +778,34 @@ def IUIAutomationEventHandler_HandleAutomationEvent(self,sender,eventID):
...
+
+	# C901: 'IUIAutomationFocusChangedEventHandler_HandleFocusChangedEvent' is too complex
+	# Note: when working on IUIAutomationFocusChangedEventHandler_HandleFocusChangedEvent,
+	# look for opportunities to simplify and move logic out into smaller helper functions.
+	def IUIAutomationFocusChangedEventHandler_HandleFocusChangedEvent(self, sender):  # noqa: C901
```
seanbudd, line 808: please add typing here``

source/UIAHandler/__init__.py
```diff
@@ -873,18 +885,27 @@ def IUIAutomationFocusChangedEventHandler_HandleFocusChangedEvent(self,sender):
...
 				f"for obj {obj} "
 			)
-		eventHandler.queueEvent("gainFocus",obj)
+		eventHandler.queueEvent("gainFocus", obj)
 
-	def IUIAutomationPropertyChangedEventHandler_HandlePropertyChangedEvent(self,sender,propertyId,newValue):
+	# C901: 'IUIAutomationPropertyChangedEventHandler_HandlePropertyChangedEvent' is too complex
+	# Note: when working on IUIAutomationPropertyChangedEventHandler_HandlePropertyChangedEvent,
+	# look for opportunities to simplify and move logic out into smaller helper functions.
+	def IUIAutomationPropertyChangedEventHandler_HandlePropertyChangedEvent(  # noqa: C901
```
seanbudd, line 893: please add typing here

source/UIAHandler/__init__.py
```diff
@@ -414,19 +413,19 @@ def getUIAElementDebugString(self, element: UIA.IUIAutomationElement) -> str:
...
+		super(UIAHandler, self).__init__()
 		self.globalEventHandlerGroup = None
 		self.localEventHandlerGroup = None
 		self.localEventHandlerGroupWithTextChanges = None
 		self._localEventHandlerGroupElements = set()
-		self.MTAThreadInitEvent=threading.Event()
+		self.MTAThreadInitEvent = threading.Event()
 		self.MTAThreadQueue = Queue()
-		self.MTAThreadInitException=None
+		self.MTAThreadInitException = None
```
seanbudd, line 423: please add typing here

source/UIAHandler/__init__.py
josephsl: This will break NVDAObjects.UIA text alignment code, so I would be careful with this one.

josephsl: I can add a "getattr" function at the end in case we do find add-ons using UIAHandler.Horizontal* attributes (and perhaps a good test case for structured pattern matching). Will add an entry to changelog file.

source/UIAHandler/__init__.py
josephsl: This is an exception, and a Stack Overflow post says we can't use typing to annotate exceptions (at this time) unless NVDA code base has found a way to do so which I may not be aware.

source/UIAHandler/__init__.py
```diff
@@ -441,78 +444,85 @@ def terminate(self):
...
-			isUIA8=False
-			try:
-				self.clientObject=CoCreateInstance(CUIAutomation8._reg_clsid_,interface=IUIAutomation,clsctx=CLSCTX_INPROC_SERVER)
-				isUIA8=True
-			except (COMError,WindowsError,NameError):
-				self.clientObject=CoCreateInstance(CUIAutomation._reg_clsid_,interface=IUIAutomation,clsctx=CLSCTX_INPROC_SERVER)
+			oledll.ole32.CoInitializeEx(None, comtypes.COINIT_MULTITHREADED)
+			self.clientObject = CoCreateInstance(
+				UIA.CUIAutomation8._reg_clsid_,
+				interface=UIA.IUIAutomation,
```
leonardder, line 458:
```suggestion
				interface=UIA.CUIAutomation8._com_interfaces_[0],
```

source/UIAHandler/__init__.py
```diff
@@ -1390,5 +1440,23 @@ def terminate():
...
 		handler = None
 
+
 def _isDebug():
 	return config.conf["debugLog"]["UIA"]
+
+
+def __getattr__(attrName: str) -> Any:
+	"""Module level `__getattr__` used to preserve backward compatibility."""
+	if attrName == "HorizontalTextAlignment_Left" and NVDAState._allowDeprecatedAPI():
```
leonardder, line 1450: I would personally do the if check for deprecated API first and then use a match statement for the constants.

LeonarddeR (changes requested):

source/UIAHandler/__init__.py
```diff
@@ -441,78 +444,85 @@ def terminate(self):
 			)
 		)
 		self.MTAThreadQueue.put_nowait(None)
-		#Wait for the MTA thread to die (while still message pumping)
-		if windll.user32.MsgWaitForMultipleObjects(1,byref(MTAThreadHandle),False,200,0)!=0:
+		# Wait for the MTA thread to die (while still message pumping)
+		if windll.user32.MsgWaitForMultipleObjects(1, byref(MTAThreadHandle), False, 200, 0) != 0:
```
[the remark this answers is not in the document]
josephsl: Done - this is one of those good use cases of assignment expression (:=).

source/UIAHandler/__init__.py
josephsl: This will start with IUIA2 (Windows 8), and since minimum Windows version is 8.1, we can actually start with IUIA3 (index = 1). Wil add a comment to that effect.

source/UIAHandler/__init__.py
josephsl: Agreed - this makes the code more readable.

lukaszgo1 (changes requested):
I'm not sure if something has changed during the time when  I was not following NVDA's development last year, but in the past we used not to accept mass linting changes like this (for one example see concluding comments of #12773). Currently when reviewing this code it is not as easy as it should be to say which part of the change is strictly necessary to address the main problem and which is purely code reformat. In fact only around 10% of the patch here has any relation to using Windows 8.1 interfaces. Has it been considered to split this PR into separate ones for linting, conversion into an enumeration and one for the main problem? At the minimum it would be good not to squash merge this, but split into the 3 (or more) coherent commits, so that blaming is not negatively affected. It would also be nice to decide and document if changes like this should be even done in the first place. My initial intuition is that they should not be, since we have discussion about more or less auto reformatting in progress, many other projects explicitly state that mass linting like this should not be smuggled together with unrelated code changes, and in general they make reviewing much more time consuming. This is not to say that I do think consistent code style is unimportant, just that it should be applied in an organized way and that code patches whose main point is not linting, should not consist of purely linting changes in almost their entirety.

Brian1Gaff:
Does this mean that an older piece of software which works in Windows8 or later almost by accident, may then become unusable?
 Brian

-- 
***@***.***
in the display name field.
----- Original Message -----

CyrilleB79:
> My initial intuition is that they should not be, since we have discussion about more or less auto reformatting in progress, many other projects explicitly state that mass linting like this should not be smuggled together with unrelated code changes, and in general they make reviewing much more time consuming. This is not to say that I do think consistent code style is unimportant, just that it should be applied in an organized way and that code patches whose main point is not linting, should not consist of purely linting changes in almost their entirety.

I have not read the code of this PR. But I totally second what @lukaszgo1 writes. If we go through code reformatting, it would be nice to have it all in a single independent commit that we can exclude with `git blame`.

josephsl:
Hi,

My thinking is that 2024.1 development milestone is a good opportunity to lint the surrounding code and the module. But if people believe that splitting the module-level lint is better, then I'm willing to remove lint commits from this PR.

Thanks.

seanbudd:
Yes, please do split out the lint where its not relevant to this fix, it makes it difficult to tell what the code changes are in this PR.

LeonarddeR (approved):

seanbudd (approved):
Thanks @josephsl

PR #15930, opened 2023-12-17
Translations from Crowdin

michaelDCurran:
Translations exported from Crowdin as a zip file and then unpacked into NVDA repository.
Note that all po files have changed due to:
* Any custom copyright / other comments in the header has been stripped.
* Crowdin metadata comments have been added
* location comments added for every string (our old translation system used to strip these)

Subsequent merges should not be anywhere near as large. Size mostly due to the re-adding of the location comments for each string.


michaelDCurran:
The only two translations that really have been updated are hr and fi. And I have checked that at least one of our new strings ("native selection mode") were in each and have been translated.

PR #15936, opened 2023-12-18
Honour choice when discarding the installation of an incompatible add-on

CyrilleB79:
### Link to issue number:
Issue probably introduced in #15623.
### Summary of the issue:
When installing an incompatible add-on (e.g. add-on last tested with NVDA 2023.3) from Windows Explorer, we get a warning asking to confirm the installlation. If I answer No, the installation process still continues.
It's because the function to display the dialog box now returns a 2-tuple which always evals to True.

Answering No in the incompatibility warning dialog will now stop the installation process as expected.
Only check the first element of the tuple to know the choice made in the dialog.
### Testing strategy:
Manually tested running from source before and after this change.
### Change log entry
Not needed; unreleased bug.
None
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/addonGui.py
+++ b/source/gui/addonGui.py
@@ -150,7 +150,8 @@ def installAddon(parentWindow: wx.Window, addonPath: str) -> bool:  # noqa: C901
 		_showAddonRequiresNVDAUpdateDialog(parentWindow, bundle._addonGuiModel)
 		return False  # Exit early, addon does not have required support
 	elif bundle.canOverrideCompatibility:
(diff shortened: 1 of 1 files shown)


seanbudd (approved):
Thanks @CyrilleB79

PR #15966, opened 2023-12-26
Fix logic error making it impossible to uninstall multiple add-ons

lukaszgo1:
This should go into beta, as it fixes a bug introduced in 2024.1 development cycle. For the same reason I do not include a change log entry,as it is unnecessary. cc @CyrilleB79 
### Link to issue number:
None
### Summary of the issue:
When trying to uninstall multiple add-ons nothing was happening, and according to the log none of the selected add-ons was relevant for remove action.
It is now possible to uninstall multiple add-ons in the Add-ons Store.
The condition which was checking if the given add-on can be removed is inverted, so that the removal is possible when add-on can indeed be removed.
### Testing strategy:
Selected two add-ons in the store, executed 'remove' from the context menu, made sure the removal process works.
None known
### Code Review Checklist:

- [X] Documentation:
  - Change log entry
  - User Documentation
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/addonStoreGui/viewModels/store.py
+++ b/source/gui/addonStoreGui/viewModels/store.py
@@ -282,7 +282,7 @@ class AddonStoreVM:
 		shouldRemove = True
 		shouldRememberChoice = False
 		for aVM in listItemVMs:
-			if aVM.canUseRemoveAction():
+			if not aVM.canUseRemoveAction():
 				log.debug(f"Skipping {aVM.Id} ({aVM.status}) as it is not relevant for remove action")
 			else:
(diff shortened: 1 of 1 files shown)


seanbudd (approved):
Thanks @lukaszgo1

CyrilleB79:
Thanks @lukaszgo1 for fixing this error introduced by my PR.

I am quite sure to have tested this scenario, but probably not re-tested after doing some changes while reviewing the PR.

PR #15968, opened 2023-12-26
Update translations

michaelDCurran:


lukaszgo1:
@michaelDCurran This PR introduces `source/locale/en/LC_MESSAGES/nvda.po`. I'm flagging this as a concern, because even though I cannot tell if it causes any real issues, usually the base language has no translation file.

michaelDCurran:
thanks for pointing this out. There is no negative side affect here, 
except for including some extra data in the repository that will never 
be used.

`source/locale/en/LC_MESSAGES/nvda.po` should be removed from beta and 
added to `.gitignore` so that it is not committed in future.

lukaszgo1:
Do we understand why it got created by Crowdin in the first place?

Issue #15854, opened 2023-11-29
Bug: clicking on the label + checkbox, NVDA is announcing inverted values of checkbox.


franzitaz:
### Steps to reproduce:
1. Open NVDA, version 2023.3
2. Open this link for test and screenshots: [NVDA testing input checkbox + label](https://codepen.io/franzitaz/pen/bGzxwqE)

### Actual behavior:

When we click on the **label**, NVDA is announcing inverted values of checkbox.
**Ps:** if we add a **::before pseudo-element** to label, NVDA announces the expected values of checkbox.

### Expected behavior:

After clicking on the **label**, the expected is NVDA announces the correct value of checkbox when is checked or not.

#### NVDA installed/portable/running from source:
Installed

---
#### NVDA version:
2023.3

---
#### Windows version:
Version	10.0.19045 Compilation 19045

---
#### Does the issue still occur after restarting your computer?
Yes

---
#### Have you tried any other versions of NVDA? If so, please report their behaviors.
No, I didn't find on NVDA website a previous version to test.

---
#### Screenshots:
![example 1](https://github.com/nvaccess/nvda/assets/11142312/c0bf8335-667f-4c5a-aa57-9aa7d535e9c9)
![example 2-before pesudo-element](https://github.com/nvaccess/nvda/assets/11142312/25e82c5e-1871-4534-819d-4ffb3510fe9a)

lukaszgo1:
Please mention in which web browsers (mentioning exact version numbers)  this can be reproduced.

Adriani90:
I think this is only reproducible in Chrome and Edge, because in Firefox NVDA does not announce the change of the state of the checkboxes at all in focus mode or when using the mouse.
@jcsteh what is the reason for Firefox not anouncing the state change at all?

Using the "before" or "after" atribute seems to lead to failure of success criterion 1.3.1 in WCAG:
https://www.w3.org/TR/WCAG20-TECHS/F87.html

I think supporting this correctly needs changes on both, screenreader and browsers side.
I personally would like to have consistent support for this because it could help in dynamic situations such as video conferencing where an event could be announced when focusing a name (e.g. "name" is speaking, where "is speaking" is inserted via the after atribute).

jcsteh:
Is this failing in browse mode, focus mode or both?

In Firefox, the check box does get reported for me when I mouse click the label. I do intermittently see the state reported incorrectly, but the presence of ::before doesn't seem to make a difference.

[...]

Issue #12067, opened 2021-02-12
Inconsistent presentation of `aria-invalid` with screen layout disabled


lukaszgo1:
Discovered as part of #11906 
### Steps to reproduce:
1. In either Firefox or Chrome open `data:text/html,<p>Big <span aria-invalid="spelling">caat</span> meos</p>`
2. Navigate through the text with screen layout both enabled and disabled.
### Actual behavior:
With screen layout disabled the document is presented as follows:
```
Big
spelling error caat meos
```

### Expected behavior:
Spelling error should not be presented in the separate line regardless of the screen layout being enabled or disabled. Alternatively spelling error should be presented in a separate line though I don't like this very much.
#### NVDA installed/portable/running from source:
From sources
#### NVDA version:
Latest master as of 12-th of February.
#### Windows version:
Windows 7 localy and whatever version of Windows 10 is currently in use on AppVeyor
#### Name and version of other software in use when reproducing the issue:
Firefox, Chrome
#### Does the issue still occur after restarting your computer?
Yes
#### Have you tried any other versions of NVDA? If so, please report their behaviors.
The same problem can be reproduced in NVDA 2019.2
#### If addons are disabled, is your problem still occuring?
Yes
#### Did you try to run the COM registry fixing tool in NVDA menu / tools?
No

JulienCochuyt:
@lukaszgo1 wrote:
> ### Expected behavior:
> Spelling error should not be presented in the separate line regardless of the screen layout being enabled or disabled.

Isn't it precisely what screen layout is all about?
Please do correct me if I'm wrong, but the way I understand it:
 - screen layout on: Present as much as possible on one line what fits on one line on screen
 - screen layout off: Split on logical boundaries, to allow for more granular review and interaction.

The spelling error being presented on a line on its own, it is then easier to determine where it begins and where it ends.

lukaszgo1:
> The spelling error being presented on a line on its own, it is then easier to determine where it begins and where it ends.


I was looking at this from the perspective of a word  processor user where spelling errors are just a part of the text  but I'm happy either way as long as this is consistent. I've mentioned what you've proposed as an alternative implementation in my first comment i.e.
> Alternatively spelling error should be presented in a separate line...

JulienCochuyt:
Sorry for the misunderstanding.
Do you complain about aria-invalid on a web page not presented the exact same way as a spelling error in a word processor depending on the "screen layout" setting?
[...]

PR #11906, opened 2020-12-05
Disable screen layout by default

OzancanKaratas:
### Link to issue number:
Fixes #11716
### Summary of the issue:
A new NVDA user may not work with web pages.
### Description of how this pull request fixes the issue:
This pull request will disable the screen layout, making it easier to use the arrow keys on web pages.
### Testing performed:
Start NVDA in default settings and open a web page.
### Known issues with pull request:
Problems may occur with some Office programs that use IAccessible2. Please test Word while testing this pull request.
### Change log entry:

- Changes
    - NVDA will no longer use the screen layout feature by default.

PR #15979, opened 2023-12-28
Adjusted the default value for two settings

### LINKED ISSUE #11716: Set screen layout mode off by default
For an in-depth discussion, see: #11187

Essentially, screen layout can cause important information to be missed by users, the need for an understanding of screen layout is more specialized.

Make screen layout off by default in new config.

Adriani90:
### Link to issue number:
Fixes #11716.
Fixes an issue where mouse users could not hear the control when entering it by default.
### Summary of the issue:
1. Screen layout was enabled by default, making browse mode navigation in browsers ineeficient when using up and down or ctrl+up and ctrl+down arrow keys. Moreover, screen layout can cause important information to be missed by users who use these navigation paterns when browsing a website in a browser. Screen layout is not uased by default in other screen readers like Jaws, so users changing the screen reader will have difficulties to adapt until they understand the setting can be changed. Problems also occur when teaching users because this setting makes NVDA behavior confusing for new users.
This change has already been proposed in #11906 and has also been approved by NV Access but the author has closed that PR in the middle of the process.
2. Users using the mouse could not hear element role when entering it. More and more blind users adapt to mouse navigation, i.e. the level of education in this regard is higher. Thus, we have to make sure a minimum standard reporting is guaranteed when using mouse navigation for blind users as well. This has been requested from many users per Email, at least in Germany and Romania.
1. NVDA will display websites in browse mode like other screen readers by default.
2. NVDA will report controls by default when entering them with the mouse.

Changed the default vlaue in config spec to false for screen layout and true for report object controls when entering them with the mouse.

### Testing strategy:
Tested that screen layout is disabled by default and report object controls when entering them withe mouse is enabled by default.

There are still some issues with screen layout off such as #12067 and for reporting object controls when entering them such as #15854. However, since the majority of people in the community rather use both settings contrary to the default value, this should not block this PR from being merged. The issues are not that heavy as to block this change. Moreover, changing the default values for these settings allows for broader testing through the community.
Additionally, system tests will probably fail due to #12067 but this can be solved in a subsequent PR.
Users who never changed the old default value of these settings will face a change and they have to adjust the settings manually if they don't want it, but this is documented in the change log of the NVDA version in which this PR will be merged.
The default values are not documented in the user guide, but this was also the case before this PR. That piece of documentation will be done in a subsequent PR.

### Code Review Checklist:
- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -4,7 +4,7 @@
 # Derek Riemer, Babbage B.V., Davy Kager, Ethan Holliger, Bill Dengler, Thomas Stivers,
 # Julien Cochuyt, Peter Vágner, Cyrille Bougot, Mesar Hameed, Łukasz Golonka, Aaron Cannon,
 # Adriani90, André-Abush Clause, Dawid Pieper, Heiko Folkerts, Takuya Nishimoto, Thomas Stivers,
-# jakubl7545, mltony, Rob Meredith, Burman's Computer and Education Ltd.
+# jakubl7545, mltony, Rob Meredith, Burman's Computer and Education Ltd, hwf1324.
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
 import logging
@@ -1649,6 +1649,7 @@ class VoiceSettingsPanel(AutoSettingsMixin, SettingsPanel):
 		self.speechModesList.Checked = [
 			mIndex for mIndex in range(len(self._allSpeechModes)) if mIndex not in excludedModes
 		]
+		self.speechModesList.Bind(wx.EVT_CHECKLISTBOX, self._onSpeechModesListChange)
 		self.speechModesList.Select(0)
 
 	def _appendDelayedCharacterDescriptions(self, settingsSizerHelper: guiHelper.BoxSizerHelper) -> None:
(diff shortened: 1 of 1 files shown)


lukaszgo1:
> ### Known issues with pull request:
> Additionally, system tests will probably fail due to #12067 but this can be solved in a subsequent PR. 


This has to be either solved now, or the failing test has to be adapted into the new behavior (with a note explaining that this is a temporary measure of course), as otherwise each build after this PR gets merged will fail in CI.

> Users who never changed the old default value of these settings will face a change and they have to adjust the settings manually if they don't want it, but this is documented in the change log of the NVDA version in which this PR will be merged. The default values are not documented in the user guide, but this was also the case before this PR. That piece of documentation will be done in a subsequent PR.


As requested in #11906 it would be ideal, at least for the screen layout case, to add a config migration step, which ensures that for old configs screen layout remains enabled. It is always better not to change users settings where possible and in this case this is more than doable.

Adriani90:
Blocked by #12067.

PR #15984, opened 2023-12-29
Adjusted the default value for screen layout setting and for report object controls when entering them with the mouse

### LINKED ISSUE #11716: Set screen layout mode off by default

Adriani90:
### Link to issue number:
Fixes #11716.
Fixes an issue where mouse users could not hear the control when entering it by default.
### Summary of the issue:
1. Screen layout was enabled by default, making browse mode navigation in browsers ineeficient when using up and down or ctrl+up and ctrl+down arrow keys. Moreover, screen layout can cause important information to be missed by users who use these navigation paterns when browsing a website in a browser. Screen layout is not uased by default in other screen readers like Jaws, so users changing the screen reader will have difficulties to adapt until they understand the setting can be changed. Problems also occur when teaching users because this setting makes NVDA behavior confusing for new users.
This change has already been proposed in #11906 and has also been approved by NV Access but the author has closed that PR in the middle of the process.
2. Users using the mouse could not hear element role when entering it. More and more blind users adapt to mouse navigation, i.e. the level of education in this regard is higher. Thus, we have to make sure a minimum standard reporting is guaranteed when using mouse navigation for blind users as well. This has been requested from many users per Email, at least in Germany and Romania.
1. NVDA will display websites in browse mode like other screen readers by default. For users that that have screenlayout set explicitely to enabled, this setting will be preserved to the enabled value. The new default value will apply only for new installations with clean config files or for the default profiles where this setting has never been explicitely changed before.
2. NVDA will report controls by default when entering them with the mouse.

* In the configspec, Changed the default value to false for screen layout and true for report object controls when entering them with the mouse.
### Testing strategy:
1. Ran from source and tested that screen layout is disabled by default and report object controls when entering them withe mouse is enabled by default.
2. Created an installer binary with scons launcher and applied that on my current local NVDA installation where screen layout was enabled and checked that the setting still remains enabled.
2.1. Created a new Profile with the same state for Screen layout as the default profile and checked that the state is not changed after the profile is running. Checked the profile.ini file and made sure it is empty, so no explicit change of the setting was performed
2.2. Disabled screen layout in the default profile and created a new profile, checked that the disabled screen layout setting still remains as explicitely seet in the default profile.

* There is no upgrade step for this setting implemented because there is no easy way to do that. The default profiles with untouched .ini files cannot be distinguished between old or new. The config would have to be redesigned to allow for a propper upgrade step and this is too much effort for this small change. We expect users will be able to adapt quite easily to the new default behavior. The advice to change it manually back to the old behavior, if prefered, is documented in the change log.
* There are still some issues with screen layout off such as #12067 and for reporting object controls when entering them such as #15854. However, since the majority of people in the community rather use both settings contrary to the default value, this should not block this PR from being merged. The issues are not that heavy as to block this change. Moreover, changing the default values for these settings allows for broader testing through the community.
* Additionally, system tests will probably fail due to #12067 but this can be solved in a subsequent PR. For a fix, the NVDA helper needs to be visited and the field nodes need to be redefined to prevent NVDA from building new child objects when spelling errors are identified in the virtual document.
* The default values are not documented in the user guide, but this was also the case before this PR. That piece of documentation I plan to do in a subsequent PR for the whole user guide at once.

### Code Review Checklist:
- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


Adriani90:
I had to create a new PR based on a specific branch, I forgot that I am tracking the master branch and it automatically merges PRs from NV Access into my forg while discarting my own changes.

Adriani90:
Blocked by #12067.

See also https://github.com/nvaccess/nvda/pull/15979#issuecomment-1871499304

seanbudd:
Hi @Adriani90 - this is not necessarily blocked by #12067, rather these tests should be fixed as part of an attempt on this issue.
Marking this as a draft until that work, and the smooth config update steps outlined in https://github.com/nvaccess/nvda/pull/15979#issuecomment-1871499304 are included.

Adriani90:
@seanbudd sorry for solving #12067 I don't have enough technical skills to do it, this needs to be changed in the NVDA helper and the handling of spellir errors as fieldnodes which triggers the line break in browse mode. I tried to understad some stuff there to solve it but I don't have enough time to digg deep into that at this stage.
I don't also know how to make system tests ignore this when building, and I think it makes sense to do it yourself at NV Access if the decision is to do it now. So your help in this regard is very appreciated.

@lukaszgo1 as for the config migration step, this is a general process that didn't really exist in the NVDA development yet. When a setting's default value has been changed, it has always been changed for alpha users first, and came into stable release at a later stage. But in any case, people that were on stable got this change suddenly and there was no way to migrate yourself manually when you want. This was the case when the default setting for cancellable speech has been changed, but also for the WASAPI setting, the automatic move focus to focusable elements and other settings that changed their default values.

You said in https://github.com/nvaccess/nvda/pull/15979#issuecomment-1871499304:
> it would be ideal, at least for the screen layout case, to add a config migration step, which ensures that for old configs screen layout remains enabled. It is always better not to change users settings where possible and in this case this is more than doable.

This step is not as trivial as you might think. How do you define old configuration in this regard? I don't see a way to define old configurations. The user will get this setting after an update, so it is the decision of the user to do it or not and the information about the new default value is documented in the change log.
We might say old configuration means people did not install NVDA, but updated an older version, so this default change should only apply for installations, not for updates of NVDA. But this seems too over complicated in my opinion and I don't know of any setting that changed its default value in the past which had this migration step.
And how about portable NVDA versions? They are not installed at all and yet they might contain old configurations, so after an update of a portable version there is no mechanism to detect whether it has a new configuration or not.

lukaszgo1:
@Adriani90 This has been originally requested in https://github.com/nvaccess/nvda/pull/11906#issuecomment-773885298 by Reef Turner, where they, quite clearly, outlined how this should be done in terms of coding. It seems you are completely unfamiliar with the config migrations and versioning in NVDA - I'd suggest to spent some time looking at this code before trying to invent the definitions of old configs, and all imaginary issues coming with it.
The examples of settings which changed  default values without config migration steps in the past should not be used as a precedent here since:
- The intent for these changes was not to be user visible i.e. they just improved the behavior. In practice this didn't work that well, (various issues with all of these options were identified), but these were issues, rather than subjective differences in behavior
- Most of these options had three states to cover the majority of users who  are fine with the NV Access recommended default, but also allowed to override it, in a way which provides assurance that user choice remains respected even when the default changes
- Here we cannot state that one setting is subjectively better - we must pick a default of course, but annoying existing users should be avoided

Adriani90:
Thanks @lukaszgo1 for the details, I will have a deeper look at that and try to come up with a solution.

@seanbudd I've looked further into the #12067 behavior and if I understand it correctly it seems in the NVDA helper field nodes produce child objects. Each object with html semantics such as button, link etc. is placed into a separate child object which triggers NVDA to display it on a new line when screen layout is disabled. This seems to happen also even for markdown formated text here on github. It seems to me that every change of a text property on the screen is detected as separate field node without checking whether it makes sense or not. i.e. ins or del property, spelling error, etc. they become separate child objects. When removing the spelling error property, the child object is merged with the previous one and that makes it work as expected.

I think the only way to solve this is to prevent NVDA to produce a separate child object when spelling error is detected in the virtual document.

Adriani90:
I added a config step for screen layout as requested. Should I move the entry from change log now? Or should it still stay there?

source/config/profileUpgradeSteps.py
```diff
@@ -376,3 +376,19 @@ def upgradeConfigFrom_10_to_11(profile: ConfigObj) -> None:
 			"hidBrailleStandard added to braille display auto detection excluded displays. "
 			f"List is now: {profile['braille']['auto']['excludedDisplays']}"
 		)
+
+
+def upgradeConfigFrom_11_to_12(profile: ConfigObj) -> None:
+	"""
+	use Screen layout where supported is disabled by default.
```
seanbudd, line 383: Just for easier parsing

```suggestion
	"Use Screen layout where supported" is disabled by default.
```

source/config/profileUpgradeSteps.py
```diff
@@ -376,3 +376,19 @@ def upgradeConfigFrom_10_to_11(profile: ConfigObj) -> None:
...
+	the old value.
+	Otherwise, the new default value <false>, will be used.
+	"""
+	try:
+		useScreenLayout: str = profile['virtualBuffers']['useScreenLayout']
+		del profile['virtualBuffers']['useScreenLayout']
+	except KeyError:
+		useScreenLayout = True
+	if configobj.validate.is_boolean(useScreenLayout):
+		profile['virtualBuffers']['useScreenLayout'] = "true"
```
seanbudd, line 394: please use double quotes consistently

```suggestion
		useScreenLayout: str = profile["virtualBuffers"]["useScreenLayout"]
		del profile["virtualBuffers"]["useScreenLayout"]
	except KeyError:
		useScreenLayout = True
	if configobj.validate.is_boolean(useScreenLayout):
		profile["virtualBuffers"]["useScreenLayout"] = "true"
```

user_docs/en/changes.t2t
```diff
@@ -11,6 +11,7 @@ What's New in NVDA
 
 
 == Changes ==
+- The default value for the setting "use screen layout (when supported)" in browse mode settings category changed to disabled (#11716) and the default behavior for the setting "report object when mouse enters it" in the mouse settings category changed to enabled.
```
seanbudd, line 14: I think these should be separate entries. Please also add relevant issue/PR numbers in the standard format (at the end of the line) and username credit if you'd like

seanbudd (commented):
Looks good generally just minor changes

CyrilleB79 (changes requested):
Such type of upgrade should not be done. This leads to situations where the configuration of a profile may change.

Example if this upgrade step is merged:
* Before migration: default profile has useScreenLayout=`False` (explicitly set) and firefox profile has not explicitly defined useScreenLayout. Thus, in Firefox, the value of default profile is used, i.e. `False`.
* After migration: in default profile useScreenLayout remains to `False` and in firefox profile, useScreenLayout is forced to `True` with the migration. As a result, in Firefox, `useScreenLayout` uses the value `True`.

As a user, I have explicitely set useScreenLayout to `False`. I expect this value to remain to `False`, no matter if I use a profile or not.

Thus, on contrary to what Reef wrote in https://github.com/nvaccess/nvda/pull/11906#issuecomment-773885298 and to what @lukaszgo1 wrote, I think that we should not upgrade the config. We can just change the default value and indicate clearly in the change log that the default value has change and that anybody who wants to turn back to the old behaviour should now set it explicitly.

Moreover, I do not think that distinguishing between an an upgrade and a new install is very clear from a user point of view. For example, I have never changed screen layout and thus have upgraded, keeping the same behaviour. Then I re-install NVDA from scratch, (e.g. on another machine) much later (e.g. in 2026). I will get a different experience with screen layout but will have forgotten that this parameter has changed default value only for new users two years before. It's better to change the default value for everyone and advertise it widely so that everybody discusses it at the same way and can help each other if needed.

If you still want to force the default value through a config upgrade, the current implementation is not the way to go in any case.

Note also that I would have other comments on this upgrade code, but since it should go away IMO, I have not commented.

cary-rowen:
@Adriani90 

Anyway, I think you should first change a more meaningful title for this pr, this really confuses me.

Adriani90:
You certainly get what is it about from the description. Let‘s not loose ourselves in such tiny details. Indeed the title says what the code should do at the end.Von meinem iPhone gesendetAm 16.01.2024 um 17:11 schrieb Rowen ***@***.***>:﻿
@Adriani90
Anyway, I think you should first change a more meaningful title for this pr, this really confuses me.

cary-rowen:
I don't think it's best practice to read through the description of each PR to understand what the author is trying to do.
As the PR author, it’s your responsibility to provide a more meaningful title rather than thinking it doesn’t matter.
While I don't necessarily think so, your response above may come across as stubborn and a bit arrogant.


Thanks

Adriani90:
Thanks for the flowers. Please if you have a suggestion, propose it and I can change. Anyway, if you want to review this you indeed should read the description, not only the title. Well, my intention is not to offend you, but I feel like you search for tiny details to squeeze others patience without seeing the work that others put in to this. It is not only this situation that you bring comments like this.So please, if this pr is gonna do things better through another title, propose one.And by the way I don‘t find any official source that prescribes how PR titles should look like and how they shouldn‘t. Von meinem iPhone gesendetAm 16.01.2024 um 17:28 schrieb Rowen ***@***.***>:﻿
I don't think it's best practice to read through the description of each PR to understand what the author is trying to do.
As the PR author, it’s your responsibility to provide a more meaningful title rather than thinking it doesn’t matter.
While I don't necessarily think so, your response above may come across as stubborn and a bit arrogant.
Thanks

CyrilleB79:
@Adriani90, I fully support what @cary-rowen writes.
Changing the title is not the most important thing to do but it will help people looking at the list of PRs if you give a quick and precise description of the two values being changed.

CyrilleB79:
By the way, in the first place, my opinion is that for clarity, even for small changes, 2 PRs should have been opened, since there is no link between these two features. It would have made discussion clearer. Since NV Access has accepted this PR as is, I have no objection though.

Just my two cents though; I hope you do not feel this comment offending.

Adriani90:
@CyrilleB79 about the splitting, yeah I was thinking about that as well in the first place but it was a small code change so I let it like this, but now with this discussion about the upgrade step for the screen layout I think I will split this if you guys need me to do that for a more efficient review.

Re: titles, I understand your point @CyrilleB79, but there are several other PRs with titles like "update button" or something like that, that doesn't say anything about the actual thing they improve or change, but there is no discussion about titles in those PRs. So if the general consensus is to provide a certain title style, then this should apply for all PRs not only for this one.
The question is how strict we want to be when reviewing things. But yeah, I will change the title accordingly and will have this in mind for the future.

CyrilleB79:
> @CyrilleB79 about the splitting, yeah I was thinking about that as well in the first place but it was a small code change so I let it like this, but now with this discussion about the upgrade step for the screen layout I think I will split this if you guys need me to do that for a more efficient review.

No, do not do so, unless NV Access asks otherwise. This PR is already under review, so I think it's better to continue with the current PR for both changes.

> Re: titles, I understand your point @CyrilleB79, but there are several other PRs with titles like "update button" or something like that, that doesn't say anything about the actual thing they improve or change, but there is no discussion about titles in those PRs. So if the general consensus is to provide a certain title style, then this should apply for all PRs not only for this one. The question is how strict we want to be when reviewing things.

My remark about the title was not to be considered as something mandatory to have the PR merged, but rather as a nice-to-have. It should be seen as a good practice tip. But I agree that some other PRs (or issues) also have titles that are unclear and that it has not been asked systematically to change them.

> But yeah, I will change the title accordingly and will have this in mind for the future.

Thanks for having changed the title for something more explicit.

cary-rowen:
Thanks to @Adriani90 for making the title look clearer.
Thanks to @CyrilleB79 for the further clarification.

I'm sorry if I offended you, but please note that every contribution deserves respect.

Adriani90:
@CyrilleB79 not sure I have done this correctly but it seems now it works on my side even with different profiles and the problem you describe does not occure. I have updated my testing steps.
So if this proves to be correct, then it was much simpler than I suspected. :))

source/config/configSpec.py
```diff
@@ -12,7 +12,7 @@
 #: provide an upgrade step (@see profileUpgradeSteps.py). An upgrade step does not need to be added when
 #: just adding a new element to (or removing from) the schema, only when old versions of the config
 #: (conforming to old schema versions) will not work correctly with the new schema.
-latestSchemaVersion = 11
+latestSchemaVersion = 12
```
CyrilleB79, line 15:
```suggestion
latestSchemaVersion = 11
```

source/config/profileUpgradeSteps.py
```diff
@@ -376,3 +376,19 @@ def upgradeConfigFrom_10_to_11(profile: ConfigObj) -> None:
...
+	or if the value has been explicitely set to <true> or <false> before upgrade,
+	preserve the old value.
+	Otherwise, the new default value <false>, will be used.
+	"""
+	try:
+		useScreenLayout: str = profile['virtualBuffers']['useScreenLayout']
+		isuseScreenLayoutMissing = False
+	except KeyError:
+		useScreenLayout = True
+		isuseScreenLayoutMissing = True
```
CyrilleB79, line 394: Just remove all the changes here:

```suggestion
```

CyrilleB79 (commented):
Actually, the config upgrade is now unneeded. Your "upgrade" function does not change anything in the config. So just remove it.

Adriani90:
Oh yeah :)) I tested this without the wrong upgrade step and indeed it was the same. :))
Hmm, so it means somehow we have to distinguish between an old installation where the user didn't touch this setting in the default profile vs. a new installation with this untouched setting but different default value.

I don't find a way how to specify this. Actually the upgrade function can look for the string or for boolean in the config and it can have the values from the configspec. but there is no further condition on how to make the difference between two untouched nvda.ini files with different default values. Right?
If you have any suggestions, or you can point me in the right direction, I would be happy to test out something else.
I think we might have to add a second shema to the configspec for old default values which might make this process easier. But this means quite a refactoring of the config I guess.

Otherwise I can add the notification about the change to the default value in the change log again.
@seanbudd what do you think?

CyrilleB79:
> Hmm, so it means somehow we have to distinguish between an old installation where the user didn't touch this setting in the default profile vs. a new installation with this untouched setting but different default value.
> 
> I don't find a way how to specify this. Actually the upgrade function can look for the string or for boolean in the config and it can have the values from the configspec. but there is no further condition on how to make the difference between two untouched nvda.ini files with different default values. Right? If you have any suggestions, or you can point me in the right direction, I would be happy to test out something else. I think we might have to add a second shema to the configspec for old default values which might make this process easier. But this means quite a refactoring of the config I guess.
> 
> Otherwise I can add the notification about the change to the default value in the change log again. @seanbudd what do you think?

My opinion is that we can just document the change in the change log. The assumption that people who have already installed NVDA want to stick with the config they have is not always true: some people may want to discover the new default setting and appreciate it, especially those who have switched from Jaws to NVDA recently or people who currently use more than one screen reader.

Anyway, if you still want to take into account the config as a whole (default + profiles) to decide how upgrades should be done, you should re-design the whole upgrade mechanism. I do not think it's worth doing it for this PR. I have no suggestion for it.

Note that if you remain without any upgrade mechanism instead, you need to adapt the initial description.

Regarding the known issues:
> Additionally, system tests will probably fail due to #12067 but this can be solved in a subsequent PR.

If we agree that #12067 should be fixed in a separate PR, the system tests that fail should be excluded and documented in #12067, so that the system tests do not report a known failure when this PR is merged.

> The default values are not documented in the user guide, but this was also the case before this PR. That piece of documentation will be done in a subsequent PR.

There is nothing preventing from doing it now while at it. Why not? It's also a good way to advertise the change for the default value.

Adriani90:
> There is nothing preventing from doing it now while at it. Why not? It's also a good way to advertise the change for the default value.

I am actually planing to do it for the whole user guide after this is merged but yeah I can do it in this PR first. I was just thinking to have all default values set before starting.

seanbudd:
Hi @Adriani90 - we cannot merge a PR with failing system tests, I'd encourage you to fix them in this PR

user_docs/en/changes.t2t
```diff
@@ -11,6 +11,10 @@ What's New in NVDA
 
 
 == Changes ==
+- The default value for the setting "use screen layout (when supported)" in browse mode settings category changed to disabled (#11716, @Adriani90)
```
seanbudd, line 14:
```suggestion
- The default value for the setting "use screen layout (when supported)" in the browse mode settings category was changed to disabled. (#11716, @Adriani90)
```

user_docs/en/changes.t2t
```diff
@@ -11,6 +11,10 @@ What's New in NVDA
 
 
 == Changes ==
+- The default value for the setting "use screen layout (when supported)" in browse mode settings category changed to disabled (#11716, @Adriani90)
+ - If you enabled the setting explicitely before, everything remains unchanged.
+ - If you never touched this setting but you still prefer screen layout when navigating websites, you can enable the setting again manually.
+- The default behavior for the setting "report object when mouse enters it" in the mouse settings category changed to enabled (@Adriani90).
```
seanbudd, line 17:
```suggestion
 - If you enabled the setting explicitly before, everything remains unchanged.
 - If you have disabled the setting explicitly before, you can re-enable this manually.
 -
- The default behavior for the setting "report object when mouse enters it" in the mouse settings category was changed to enabled. (#15984, @Adriani90).
-
```

user_docs/en/changes.t2t
```diff
@@ -11,6 +11,10 @@ What's New in NVDA
 
 
 == Changes ==
+- The default value for the setting "use screen layout (when supported)" in browse mode settings category changed to disabled (#11716, @Adriani90)
+ - If you enabled the setting explicitely before, everything remains unchanged.
+ - If you never touched this setting but you still prefer screen layout when navigating websites, you can enable the setting again manually.
```
CyrilleB79, line 16: These instructions are a bit complex. Could we instead just use instructions as the following one instead?

```suggestion
 - If you still prefer screen layout when navigating websites, make sure you have this setting enabled.
```

CyrilleB79:
@Adriani90, in #16127, the Chrome tests have been disabled due to intermittent failure. This intermittent failures are probably due to an unexpected behaviour on appVeyor, e.g. Chrome update, other window taking the foreground, etc.

Though, the tests that were failing in the current PR were failing due to default value being changed, not only to appVeyor surprises. Thus, I think that fixing them in this PR is still required.

First you need to revert #16127, for example without committing. Then run locally system tests and make them pass. Last, restore #16127.

To make tests pass, you have two choices:
1. either the result of the test is expected given the change of the default value; in this case, you need to update the references.
2. or the result is unexpected; in this case, either you fix it here or you should ignore individually the test and track the problem in a separate issue (e.g. #12067). In this case, the tracking issue should be mentioned.

seanbudd:
thanks for explaining the path forward Cyrille.

I expect us to restore chrome system tests later in this release cycle

codeofdusk:
Once again, I vote for this PR to be closed. Disabling screen layout is a philosophical shift, introduces inconsistencies in the product, reduces efficiency, and is confusing.

CyrilleB79:
@Adriani90, this PR is not going forward, maybe due to difficulty with tests and/or with some more discussion on the root issue.
Given this, I'd recommend that you separate from this PR the unrelated change regarding mouse echo and implement it in a dedicated separate PR. Indeed, it would allow to merge at least this uncontroversial part of your work.

@codeofdusk NV Access has accepted to change screen layout default value. If you want to make change their mind to NV Access and to the community, you'll need to provide more explicit arguments.
Why does disabling screen layout introduces inconsistencies in the product? Why does it reduces efficiency? Why is it confusing?
You have provided here only an opinion, no arguments at all, nor a link to these arguments.

Adriani90:
I will close this PR and split the work as advised.
@codeofdusk I don't really understand your argument. From a developer and a very advanced NVDA user perspective your point might make sense, but to be honnest for a novice user screen layout is not really helpful. Apart from that, there is no screen reader in the world except NVDA that enabled screen layout by default if both options exist.

CyrilleB79:
@Adriani90 my recommendation was rather to keep this PR open for default screen layout off, since there is already an history and a discussion on this topic in this PR, but remove from the current PR "report object controls when entering them with the mouse" option and put it in a separate PR.

If you can, please re-open so that we can keep a cleaner history. Thanks.

Adriani90:
It seems something went very wrong during the rebase process, I will create a new one.

CyrilleB79:
If you create a new PR, while at it, you can split the two independent features of this PR in two separate PRs.

PR #16071, opened 2024-01-20
Sound split

### LINKED ISSUE #12985: Audio settings with stereo headset (or speakers) - Send NVDA sounds on one side and the rest of Windows sounds on the other side
I have had this feature request in mind since many years. But I finally open this ticket thanks to another person [asking it](https://groups.io/g/nvda-fr/topic/son_de_nvda_seulement_d_un/86573343?p=,,,100,0,0,0::recentpostdate/sticky,,,100,2,0,86573343,previd=1635155623389494153,nextid=1631705229306738732&previd=1635155623389494153&nextid=1631705229306738732) on a French mailing list.

When participating to audio-conferences on the computer, it's often confusing to have NVDA's vocal feedback mixed up with the audio stream of the conference.

NVDA audio stream is produced by synthesizer output but also beeps (progress bars, indent reporting) as well as other sounds such as focus/browse mode and spelling mistakes detection.
With a stereo headset or stereo speakers, it would be handy to have an option allowing to split the audio stream in two parts:
* NVDA audio stream on one side, (e.g. left ear)
* the rest of the computer's audio stream on the other side (e.g. right ear). That includes Windows sounds and sounds of the applications, which may include:
  * the audio stream of an aundio conference (Skype, Zoom, Teams either withe desktop application or in the browser)
  * the audio stream of audio or video editing/processing software (e.g. mp3DirectCut)

* Maybe there is one or more applications that may already do this job; if yes, it would be good to reference it here.
* In this issue, we may also consider the opportunity to add this feature in an add-on.

But IMO it would be valuable to have this feature integrated in NVDA or in an add-on rather that an external application. It would allow many people to discover this feature even if they do not express the need clearly.
Also it would be a plus to have this feature available in a portable version of NVDA; I do not know if it is technically doable however.

### Additional context

Jaws 2022 will provide this feature (and that's what convinced me that this feature could be embedded in a screen reader). Below is a quote of its release notes ([What's new](https://support.freedomscientific.com/Downloads/PublicBeta)):

> Sound Splitter
> Earlier in JAWS and Fusion 2021, we introduced a way for users to lower the volume of JAWS speech independently of system audio to help make it easier to hear audio from other applications. Version 2022 takes this a step further by now allowing those using stereo headsets or speakers to route all JAWS or Fusion speech to one ear while routing audio from all other applications to the other ear. This helps make it easier to continue getting work done with JAWS while following along during a Zoom or Teams meeting, a Skype call, or a YouTube video.
> To split audio:
> 1. Press INSERT+SPACEBAR followed by V to activate the Volume layer.
> 2. Press B for Balance, and then do one of the following:
> ◦ Press LEFT ARROW to route JAWS Speech to the left and audio from all other applications to the right.
> ◦ Press RIGHT ARROW to route JAWS speech to the right and audio from all other applications to the left.
> ◦ Press UP ARROW to restore both JAWS speech and application audio to both channels.
> Note: Before attempting to split your audio, make sure your headset or speakers are properly oriented. For instance, if you route JAWS speech to the left and instead hear it in your right ear, your headset is most likely reversed.
> Once you split your JAWS speech and application audio, it remains in effect until you restart JAWS, or use the command to restore the audio balance.

I do not advocate for copying all features of Jaws in NVDA. But I think in this case that having this feature integrated in NVDA is a real plus.

mltony:
### Link to issue number:
Fixes #12985
### Summary of the issue:
Feature request: sound split. Splits system sound into two channels: NVDA speaks in one channel (e.g. left), while all other applications play their sound in the other channel (e.g. right).
* Added global command `NVDA+alt+s` that toggles sound split between off, NVDA on the left and NVDA on the right (default behavior).
* Added combo box on Audio panel in NVDA settings that also allows to switch between Sound split modes.
* Added list of checkboxes in Audio panel, that allows to change behavior of `NVDA+alt+s` command: it allows to select all modes that the global command will cycle through.

1. Added pycaw library as a dependency.
2. Created file `source\audio\soundSplit.py` where I implemented all logic. 
3. Contrary to what I said before, I managed to implement sound split without creating an extra monitor thread. It works like this:
    * When sound split is toggled, it uses `IAudioSessionEnumerator` to set volume in all currently active audio sessions.
    * Then it does `sessionManager.RegisterSessionNotification()` to create a callback that listens for any new audio sessions being created, an it executes the the same volume updating function upon creation.
    * On the next call or on shutdown we unregister the previous notification callback.
### Testing strategy:
* Performed thorough manual testing on Windows 10 and Windows 11.

* Sound split only works in wasapi mode. The reason for that is that when NVDA is using WinMM, adjusting its volume via core audio API doesn't work. In theory this can be worked around by still applying volume inside NVWave player; however since NVDA is migrating to wasapi, I thought this doesn't make much sense at this point.
* During shotdown we try to restore audio volume of all applications. However if an app is currently not playing any sounds, it doesn't have an active audio session and therefore there is no way to update its volume. As a result some applications might be playing sounds only in 1 channel long after NVDA is shut down. This is a known issue and there appears to be no work around.
* There is a very unlikely race condition that might occur when an application starts playing sound between the call `sessionManager.GetSessionEnumerator()` and `sessionManager.RegisterSessionNotification()`. In this case sound split volume will not be applied to that application. I don't see a way to fix this in code since this is rather a problem of COM API. However turning sound split off and on again is an easy and intuitive workaround in this case.
* Sound split won't work with applications playing sounds in mono mode or outputting to more than 2 channels (e.g. Dolby Surround). In practice I never encountered such applications.
* Please note, that sound split doesn't work as a mixer. For example, if an application is playing a stereo sound track while sound split is set to "NVDA on the left and applications on the right", then you will only hear the right channel of the sound track, while the left channel of the sound track will be muted. in order to allow NVDA to use the left channel.
* Sound split won't work with applications using legacy winMM API to play audio. However during the life span of sound split feature in Tony's enhancements add-on no users have reported of such applications.
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


source/audio.py
```diff
@@ -0,0 +1,210 @@
...
+			SoundSplitToggleMode.OFF_LEFT_RIGHT: _("Cycles through off, NVDA on the left and NVDA on the right"),
+		}
+	
+	def getPossibleStates(self) -> List[SoundSplitState]:
+		result = [SoundSplitState.OFF]
+		if 'LEFT' in self.name:
+			result.append(SoundSplitState.NVDA_LEFT)
+		if 'RIGHT' in self.name:
+			result.append(SoundSplitState.NVDA_RIGHT)
+		return result
```
lukaszgo1, line 83: Can't this method be simplified to:
```python
return [SoundSplitState.OFF, self]
```

source/audio.py
```diff
@@ -0,0 +1,210 @@
...
+			unregisterCallback(activeCallback)
+
+
+def getDefaultAudioDevice(kind: EDataFlow = EDataFlow.eRender) -> Optional[mmdeviceapi.IMMDevice]:
+	deviceEnumerator = comtypes.CoCreateInstance(
+		CLSID_MMDeviceEnumerator,
+		mmdeviceapi.IMMDeviceEnumerator,
+		comtypes.CLSCTX_INPROC_SERVER,
+	)
+	device = deviceEnumerator.GetDefaultAudioEndpoint(
```
lukaszgo1, line 122: This currently fails on machines (such as the CI node on AppVeyor) where there are no audio devices, with the following error:
```
  File "audio.pyc", line 98, in initialize
  File "audio.pyc", line 130, in getSessionManager
  File "audio.pyc", line 122, in getDefaultAudioDevice
_ctypes.COMError: (-2147023728, 'Element not found.', (None, None, None, 0, None))
```

source/audio.py
```diff
@@ -0,0 +1,210 @@
...
+	audioDevice = getDefaultAudioDevice()
+	if audioDevice is None:
+		raise RuntimeError("No default output audio device found!")
+	tmp = audioDevice.Activate(audiopolicy.IAudioSessionManager2._iid_, comtypes.CLSCTX_ALL, None)
+	sessionManager: audiopolicy.IAudioSessionManager2 = tmp.QueryInterface(audiopolicy.IAudioSessionManager2)
+	return sessionManager
+
+
+def applyToAllAudioSessions(
+		func: Callable[[audiopolicy.IAudioSessionControl2], NoReturn],
```
lukaszgo1, line 139: `NoReturn` is not a right type here. As per the [documentation](https://docs.python.org/3.11/library/typing.html#typing.NoReturn) it should be used for function which either should never be called , or for those which unconditionally raise an exception. This should just have `None` as a return annotation.

source/comInterfaces/coreAudio.bak/__init__.py
lukaszgo1, line 1: It looks like `coreAudio.bak` was committed by mistake, as it just duplicates the interface definition from `coreAudio`.

source/core.py
```diff
@@ -597,6 +597,9 @@ def main():
 	log.debug("Initializing tones")
 	import tones
 	tones.initialize()
+	log.debug("Initializing sound split")
+	import audio
+	audio.initialize()
```
lukaszgo1, line 602: `audio` should be reinitialized when user restores config both to the last saved, and to the factory defaults in `core.resetConfiguration`.

source/globalCommands.py
```diff
@@ -113,6 +113,9 @@
 #: Script category for document formatting commands.
 # Translators: The name of a category of NVDA commands.
 SCRCAT_DOCUMENTFORMATTING = _("Document formatting")
+#: Script category for audio streaming commands.
+# Translators: The name of a category of NVDA commands.
+SCRCAT_AUDIOSTREAMING = _("Audio streaming")
```
lukaszgo1, line 118: Isn't streaming too technical to be understandable? I'd certainly have a difficult time if I had to translate it to Polish. Perhaps something like 'audio managementt' will be better?

source/globalCommands.py
```diff
@@ -4382,6 +4386,18 @@ def script_cycleParagraphStyle(self, gesture: "inputCore.InputGesture") -> None:
...
+	@script(
+		description=_(
+			# Translators: Describes a command.
+			"Cycles through sound split modes",
+		),
+		category=SCRCAT_AUDIOSTREAMING,
+		gesture="kb:NVDA+alt+s",
+	)
+	def script_cycleSoundSplit(self, gesture: "inputCore.InputGesture") -> None:
+		import audio
```
lukaszgo1, line 4398: Can't this be imported at the module level? If there is a circular import making this difficult please add a comment stating that this is the case. In terms of performance this should not be noticeable, as `audio` is unconditionally imported in `core.main`.

user_docs/en/changes.t2t
```diff
@@ -7,6 +7,7 @@ What's New in NVDA
 
 == New Features ==
 - In Windows 11, NVDA will announce alerts from voice typing and suggested actions including the top suggestion when copying data such as phone numbers to the clipboard (Windows 11 2022 Update and later). (#16009, @josephsl)
+- Sound split - toggled by NVDA+Alt+S
```
lukaszgo1, line 10: Please add an issue number here and optionally credit yourself.

user_docs/en/userGuide.t2t
```diff
@@ -850,7 +850,7 @@ Here is a list of available commands:
 - Article
 - Grouping
 - Tab
--
+-2
```
lukaszgo1, line 853: This change looks unrelated.

source/audio.py
```diff
@@ -0,0 +1,210 @@
...
+import comInterfaces.coreAudio.audioclient as audioclient
+import comInterfaces.coreAudio.audiopolicy as audiopolicy
+import comInterfaces.coreAudio.mmdeviceapi as mmdeviceapi
+import comtypes
+import config
+from enum import IntEnum, unique
+import globalVars
+from logHandler import log
+import nvwave
+from typing import Tuple, Optional, Dict, List, Callable, NoReturn
```
lukaszgo1, line 21: Most of these should use the standard `list` and `tuple` which are now generic, or types from `collections.abc` which are preferred over stuff from `typing`. For `Optional` you can also use `foo | None` notation, though that is probably more subjective.

lukaszgo1 (changes requested):
> Sound split only works in wasapi mode.

While this is obviously fine I'd be nice to have a more technical explanation as to why in the PR description. In particular it is important to know if supporting this is impossible, too difficult to be worth doing or just a intentional design decision.

> multiple lint errors are coming from pycaw interface definitions. I hope an exemption can be made here, since we don't want to make potential updating of these files to require manual reformatting every time just to make linter happy.

These can easily be ignored in `tests\lint\flake8.ini`. Since we do the same for UIAutomation COM interface I don't think that will be problematic.

> During shotdown we try to restore audio volume of all applications. However if an app is currently not playing any sounds, it doesn't have an active audio session and therefore there is no way to update its volume. As a result some applications might be playing sounds only in 1 channel long after NVDA is shut down. This is a known issue and there appears to be no work around.

I have no experience with the audio stuff, but my initial intuition will be to register to a notification for destroying sessions and restore the original volume there. Is that possible?

cary-rowen:
Glad to see this PR has been drafted. But I sincerely hope you will consider the following request.

Sometimes, when we are watching a movie, we often want to appreciate the sound effects of this movie, but the sound of NVDA is relatively not so important.
I want to reduce the presence of NVDA sounds as much as possible, At this point it's a good idea to direct it to the left or right channel.
Obviously, I don't want to direct the sound of the film to the left or right, because that would definitely destroy the sound of the movie.
It would be nice to have an option that allows adjusting only the output  channel of NVDA.
But I'm not denying the validity of that pull request, just pointing out another dilemma I faced.

CyrilleB79:
@mltony, a big thank you for taking the time to integrate this feature in core! I think that it will be a real plus in NVDA's features.

I'll take time tomorrow and next week to test it, and will provide review comments (if needed), probably more on the UX point of view.

source/gui/settingsDialogs.py
```diff
@@ -2710,6 +2711,28 @@ def makeSettings(self, settingsSizer: wx.BoxSizer) -> None:
...
+			soundSplitLabelText,
+			wx.Choice,
+			choices=[mode.displayString for mode in audio.SoundSplitState]
+		)
+		self.bindHelpEvent("SelectSoundSplitMode", self.soundSplitComboBox)
+		index = config.conf["audio"]["soundSplitState"]
+		self.soundSplitComboBox.SetSelection(index)
+
+		# Translators: This is a label for the sound split Toggle Mode combo box in the Audio Settings dialog.
+		soundSplitToggleModeLabelText = _("NVDA+Alt+S command behavior:")
```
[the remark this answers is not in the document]
CyrilleB79: Also:

3. Question: With this PR, we would now be able to select a subset of values for a second cycle command, the first being speech mode. If NV Access accept this and if it becomes more frequent, would it be worth to dedicate a specific setting panel for such customization? E.g. command customization.

source/config/configSpec.py
```diff
@@ -54,6 +54,8 @@
 	WASAPI = featureFlag(optionsEnum="BoolFlag", behaviorOfDefault="enabled")
 	soundVolumeFollowsVoice = boolean(default=false)
 	soundVolume = integer(default=100, min=0, max=100)
+	soundSplitState = integer(default=0)
+	soundSplitToggleMode= integer(default=0)
```
CyrilleB79, line 58: Maybe it would be worth using a list, as done with `excludedSpeechModes`. It would be more easy to add new modes if needed in the future, e.g. https://github.com/nvaccess/nvda/pull/16071#issuecomment-1902237001.

CyrilleB79 (commented):
very nice and useful feature!

I have written a first set of comments, but will continue testing later.

Also a general note:
Implementing Sound Split for legacy audio mode is technically possible since you have done it in the add-on. But IMO, it's not worth doing it since NVDA moves to WASAPI.

CyrilleB79:
Also one small issue:
For applications playing a stereo signal, sound split keeps only one channel and puts NVDA in the other one. This can be heard for example listening to [Bohemian Rhapsody (Youtube)](https://www.youtube.com/watch?v=fJ9rUzIMcZQ) at time 3:15. With sound split enabled, we can hear well only one "Galelileo" in two, the other one being heard very far.
A better behaviour would be to have a mix (a sum) of left and right channel of the song on one side and NVDA on the other side.

If it is not possible for technical reasons or other, this request should not block this PR though, since the feature as developed in this PR is already more than useful, e.g. in meetings where stereo is of no use.
Though, this limitation should be documented, at least in this PR, and maybe also in the User Guide.

mltony:
@lukaszgo1,
1. Updated PR description.
2. I am looking at `tests\lint\flake8.ini` and I see `source/comInterfaces/*, ` appears to be  already excluded. So not sure why linter is still complaining.
3. In order to register a notification we'll need to keep NVDA process running. Since applications may play sound any time, this would mean we'd need to keep NVDA (or any other surrogate processs of it) to keep running indefinitely after we quit NVDA. That can be done in theory, but I believe this is not worth it. Moreover it might create some confusion, as we'll be running a process quietly  that adjusts audio volumes of all applications - that just doesn't sound like a good idea.

source/audio.py
mltony: I added try-catch clause downstream where I log this error.

source/gui/settingsDialogs.py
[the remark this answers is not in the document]
mltony: > 1. Please, do not use NVDA+Alt+S, but rather the name of the command (e.g. as seen in input gesture dialog); this way, the name of this control always remains valid and it is also more explicit.
Done
> 2. More generally, would it be interesting to convert this control to a checkable list box as done for the available speech mode selection? That may be done now or later in case [Sound split #16071 (comment)](https://github.com/nvaccess/nvda/pull/16071#issuecomment-1902237001) is accepted.
I like this idea - just implemented that. Also this allows to implement suggestion by @cary-rowen.

mltony:
> Also one small issue: For applications playing a stereo signal, sound split keeps only one channel and puts NVDA in the other one. This can be heard for example listening to [Bohemian Rhapsody (Youtube)](https://www.youtube.com/watch?v=fJ9rUzIMcZQ) at time 3:15. With sound split enabled, we can hear well only one "Galelileo" in two, the other one being heard very far. A better behaviour would be to have a mix (a sum) of left and right channel of the song on one side and NVDA on the other side.
> 
> If it is not possible for technical reasons or other, this request should not block this PR though, since the feature as developed in this PR is already more than useful, e.g. in meetings where stereo is of no use. Though, this limitation should be documented, at least in this PR, and maybe also in the User Guide.

I believe this is not possible - in coreAudio API I can only set volumes for left and right channels. What you are talking about would be mixer functionality I guess. I mentioned this both in this PR and documentation.

source/gui/settingsDialogs.py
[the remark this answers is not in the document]
mltony: I converted combobox to list of checkbox as you suggested. As for which panel to put it on, it seems to me that sound split modes and speech modes customizations should live on different panels because they are related to semantically different commands.

mltony:
@lukaszgo1 , @CyrilleB79, I addressed all your comments. Also @cary-rowen, I implemented your idea as well.

cary-rowen:
Great, thanks @mltony for the great work.
I'll try this out once the CI build is complete.

user_docs/en/userGuide.t2t
```diff
@@ -1892,6 +1913,28 @@ This slider allows you to set the volume of NVDA sounds and beeps.
...
+
+This option allows you to make NVDA sound output to be directed to either left or right channel and application volume to be directed to the other channel. This could be useful for users of headphones or stereo speakers.
+- Disabled sound split: both NVDA and other applications output sounds to both left and right channels.
+- NVDA on the left and applications on the right: NVDA will speak in the left channel, while other applications will play sounds in the right channel.
+- NVDA on the right and applications on the left: NVDA will speak in the right channel, while other applications will play sounds in the left channel.
+-
+
+This option is not available if you have started NVDA with [WASAPI disabled for audio output #WASAPI] in Advanced Settings.
+
+==== Customizing Sound split modes====[customizeSoundSplitModes]
```
CyrilleB79, line 1927: Uppercase, to match the suggestion made in the settings dialog file:

```suggestion
==== Customizing Sound split modes====[CustomizeSoundSplitModes]
```

source/gui/settingsDialogs.py
```diff
@@ -2743,6 +2808,23 @@ def _onSoundVolChange(self, event: wx.Event) -> None:
 			wasapi
 			and not self.soundVolFollowCheckBox.IsChecked()
 		)
+		self.soundSplitComboBox.Enable(wasapi)
+		self.soundSplitModesList.Enable(wasapi)
+
+	def isValid(self) -> bool:
+		enabledSoundSplitModes = self.soundSplitModesList.CheckedItems
+		if len(enabledSoundSplitModes) < 1:
+			log.debugWarning("No sound split modes enabled.")
```
CyrilleB79, line 2817: At least 2 modes are required: there is no point in having a cycling command cycling between only one mode:
```suggestion
		if len(enabledSoundSplitModes) < 2:
			log.debugWarning("Too few sound split modes enabled.")
```

source/gui/settingsDialogs.py
```diff
@@ -2743,6 +2808,23 @@ def _onSoundVolChange(self, event: wx.Event) -> None:
...
+		self.soundSplitComboBox.Enable(wasapi)
+		self.soundSplitModesList.Enable(wasapi)
+
+	def isValid(self) -> bool:
+		enabledSoundSplitModes = self.soundSplitModesList.CheckedItems
+		if len(enabledSoundSplitModes) < 1:
+			log.debugWarning("No sound split modes enabled.")
+			gui.messageBox(
+				# Translators: Message shown when no sound split modes are enabled.
+				_("At least one sound split mode has to be checked."),
```
CyrilleB79, line 2820: At least two modes for a cycling command:
```suggestion
				# Translators: Message shown when less than two sound split modes are enabled.
				_("At least two sound split mode have to be checked."),
```

CyrilleB79 (commented):
Here are some more review items, if you want to have a look.
I have to stop now; maybe there will be other ones tomorrow.

user_docs/en/userGuide.t2t
```diff
@@ -546,6 +546,27 @@ A gesture allows cycling through the various speech modes:
 
 If you only need to switch between a limited subset of speech modes, see [Modes available in the Cycle speech mode command #SpeechModesDisabling] for a way to disable unwanted modes.
 
+++ Sound split modes ++[SpeechModes]
```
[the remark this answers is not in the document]
mltony: Then perhaps I should remove the combobox from settings as well - if speech mode doesn't have a combo box then why sound split needs one?

source/gui/settingsDialogs.py
mltony: But it doesn't hurt at all to have only one mode.

source/gui/settingsDialogs.py
mltony: Again I see no point in forcing user to select the second mode if they don't want it. It is still going to cycle through 1 mode, nothing is going to break. Let's not add artificial restrictions on the user.

hwf1324:
You can add `comInterfaces/coreAudio` to the `Glob(exclude)` parameter of the `Clean` function in `comInterfaces_sconscript`. to exclude it during SCons cleanup.

user_docs/en/userGuide.t2t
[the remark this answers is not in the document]
CyrilleB79: > Then perhaps I should remove the combobox from settings as well - if speech mode doesn't have a combo box then why sound split needs one?

Usually, NVDA settings are reachable either through the GUI (settings dialog) or through gestures (assigned or unassigned). The features that are not modifiable through settings dialogs are the exception.

Having the feature controllable in the settings dialog allows:
* saving the config on exit and restoring it at NVDA startup
* using different profiles
* a better discoverability of the feature since its control is visible in the GUI; someone can use it even without knowing the shortcut NVDA+alt+S

IMO, this is suitable for sound split:
* maybe this is not very common but we cannot exclude that someone wants to have sound split always enabled, so also at NVDA startup
* profiles may be useful in the following cases:
  * VLC profile using: NVDA left and application in both channels
  * Teams / Skype / any conferencing application using : NVDA left and application right

source/config/configSpec.py
```diff
@@ -54,6 +54,9 @@
 	WASAPI = featureFlag(optionsEnum="BoolFlag", behaviorOfDefault="enabled")
 	soundVolumeFollowsVoice = boolean(default=false)
 	soundVolume = integer(default=100, min=0, max=100)
+	soundSplitState = integer(default=0)
+	# This is JSON-encoded list of integers.
```
lukaszgo1, line 58: Can't this use `int_list` from configobj?

source/gui/settingsDialogs.py
lukaszgo1: My opinion is  that in this case cycling by a single option is justified. I don't think comparing this with the speech modes cycling makes sense, since the only way to change the mode is by a keyboard command, or restart of NVDA. For sound split you can have a default preferred mode and a second one which you need rarely, but switching to it has to be done fast. In that case you use the cycle command as a single way toggle and enable whatever mode you use as your main one from the GUI afterwards.

user_docs/en/userGuide.t2t
```diff
@@ -546,6 +546,27 @@ A gesture allows cycling through the various speech modes:
...
+| Cycle Sound Split Mode | ``NVDA+alt+s`` | Cycles between speech modes. |
+%kc:endInclude
+
+By default this command will cycle between the following modes:
+- Disabled sound split: both NVDA and other applications output sounds to both left and right channels.
+- NVDA on the left and applications on the right: NVDA will speak in the left channel, while other applications will play sounds in the right channel.
+- NVDA on the right and applications on the left: NVDA will speak in the right channel, while other applications will play sounds in the left channel.
+-
+
+There are more advanced sound split modes availabel in NVDA settings; for more information please see [Customizing Sound split modes #CustomizeSoundSplitModes] section.
```
lukaszgo1, line 566: availabel -> available
```suggestion
There are more advanced sound split modes available in NVDA settings; for more information please see [Customizing Sound split modes #CustomizeSoundSplitModes] section.
```

source/config/configSpec.py
CyrilleB79: Agree with @lukaszgo1's suggestion. Unless there is a good reason to do otherwise, using `int_list` would be consistent with what is done for speech mode, thus easier to understand quickly.

source/gui/settingsDialogs.py
CyrilleB79: So let's take an example to check we understand well each other:
* default preferred mode = disabled
* second mode rarely needed, but switching to it has to be done fast = NVDA left and other apps right.

The user uses the second mode when receiving a call, so that switching can be done very fast.

If there are two modes available in the cycling command, I do not know why switching would be less fast:
* if the user is already in split mode when the call arrives, there is nothing to do, so it's as fast as possible
* if the user is in default mode when the call arrives, just pressing NVDA+alt+S is needed, which is as fast as with only one mode in the cycling command.

Have I missed something?
I maintain that a cycling command with only one possible value is a confusing user experience. So if there is no use case where a one-value cycling command is more efficient than a two-value cycling command, there is no point in allowing it.

lukaszgo1:
> * Sound split only works in wasapi mode. The reason for that is that when NVDA is using WinMM, adjusting its volume via core audio API doesn't work. In theory this can be worked around by still applying volume inside NVWave player; however since NVDA is migrating to wasapi, I thought this doesn't make much sense at this point.

Thanks for the additional explanation @mltony. I'm slightly worried that this may mean that sound split will not work for arbitrary applications using WinMM. @jcsteh Is this something you could comment on?


> * multiple lint errors are coming from pycaw interface definitions. I hope an exemption can be made here, since we don't want to make potential updating of these files to require manual reformatting every time just to make linter happy.

I guess Flake8 relies on Python's `glob`, which does not recurse to directories by default. I'd try changing the exclusion to `source/comInterfaces/**`.

> 
> * During shotdown we try to restore audio volume of all applications. However if an app is currently not playing any sounds, it doesn't have an active audio session and therefore there is no way to update its volume. As a result some applications might be playing sounds only in 1 channel long after NVDA is shut down. This is a known issue and there appears to be no work around.

This does not need to be done in this PR, however if users will complain we can probably store all sessions for which we changed 
channels and restore them to the original setting when exiting NVDA.

source/audio.py
```diff
@@ -0,0 +1,204 @@
...
+from typing import Callable
+import ui
+from utils.displayString import DisplayStringIntEnum
+import _ctypes
+
+VolumeTupleT = tuple[float, float]
+
+
+@unique
+class SoundSplitState(DisplayStringIntEnum):
```
LeonarddeR, line 31: I think I'd prefer having two options, one to select where NVDA's sound is played and one for the application sounds. That feels more intuitive to me.

source/audio.py
```diff
@@ -0,0 +1,204 @@
...
+			# Translators: Sound split state
+			SoundSplitState.NVDA_BOTH_APPS_LEFT: _("NVDA in both channels and applications on the left"),
+			# Translators: Sound split state
+			SoundSplitState.NVDA_BOTH_APPS_RIGHT: _("NVDA in both channels and applications on the right"),
+		}
+	
+	def getAppVolume(self) -> VolumeTupleT:
+		if self == SoundSplitState.OFF or 'APPS_BOTH' in self.name:
+			return (1.0, 1.0)
+		elif 'APPS_LEFT' in self.name:
```
LeonarddeR, line 62: Having two options no longer requires you to do this kind of slightly hacking things.

source/comInterfaces/coreAudio/audioclient/__init__.py
```diff
@@ -0,0 +1,155 @@
+# This file is a part of pycaw library (https://github.com/AndreMiras/pycaw).
```
LeonarddeR, line 1: Why don't we include pycaw as a dependency?

source/audio.py
LeonarddeR, line 1: Is there anything aagainst having this code as part of nvWave? If you insist on having an extra module, I'd personally call it soundSplitting or something similar.

source/config/configSpec.py
mltony: > Can't this use `int_list` from configobj?

I haven't found a way to assign a non-empty default value to intList - do you know how to do so?

mltony:
> Thanks for the additional explanation @mltony. I'm slightly worried that this may mean that sound split will not work for arbitrary applications using WinMM. @jcsteh Is this something you could comment on?

In my understanding that's correct, this won't work for winmm applications. However, sound split functionality has been around for many years and nobody complained about that which makes me think that winmm applications are exceptionally rare these days.

> I guess Flake8 relies on Python's `glob`, which does not recurse to directories by default. I'd try changing the exclusion to `source/comInterfaces/**`.

That doesn't work - at least with runlint command ran locally.

> This does not need to be done in this PR, however if users will complain we can probably store all sessions for which we changed channels and restore them to the original setting when exiting NVDA.

SG as in we can try to preserve COM object. But I suspect that won't work since session enumerator only enumerates active sessions and so I suspect all the calls on saved com objects might fail. But again happy to try if we receive complaints.

source/audio.py
mltony: The original and main use case of this PR is sound split, so I would like users to be able to turn it on with a single keystroke instead of having to press two separate keystrokes one for NVDA and one for other apps.

source/comInterfaces/coreAudio/audioclient/__init__.py
mltony: Pycaw has its own dependencies that we don't need. I don't want to pollute NVDA with subdependency libraries that we don't need.

source/audio.py
mltony: Semantically this doesn't belong to nvwave as nvwave deals with sound output. I mentioned in PR description that I called this audio.py because there are two more PRs I plan to send that would also fit in here and that are not related to sound splitting per se.

source/config/configSpec.py
lukaszgo1: This can be written as follows:
`includedSoundSplitModes =int_list(default=list(0, 1, 3))`

source/audio.py
CyrilleB79: I agree that the list of possible sound split modes is a bit long, and if we can find something clearer, why not.

But, as @mltony writes, @LeonarddeR's suggestion of one option for NVDA sound and one for the sound of other apps would completely defeat the main use case of this feature.

lukaszgo1:
> That doesn't work - at least with runlint command ran locally.

It turns out our exclusions weren't working for some time now. I have created #16087 to address this, so after that PR is merged and master is merged to this branch Linter should no longer complain.

source/config/configSpec.py
mltony: Thanks, that works!

source/audio.py
CyrilleB79: Re the concern with the sound split modes being too long:
If we really want 2 options, we may divide the size of the list by 2 and use a checkbox to switch left and right.

LeonarddeR: AH yes, I see. Having two combo boxes would also allow a user to set NVDA right, apps right, which makes no sense at all.

source/audio.py
LeonarddeR: Still, I think i prefer having this stated explicitly. Also, please consider using a match statement, which in a Python 3.10+ word is more readable then a bunch of f/elif statements.

CyrilleB79: I'd also like to see this "hack" replaced by something more maintainable, e.g. in case we change the names or add new members to the enum in the future:

```suggestion
		elif self in [SoundSplitState.NVDA_RIGHT_APPS_LEFT, SoundSplitState.NVDA_BOTH_APPS_LEFT]:
```

hwf1324:
Hi.

I am using an audio streaming software ([AudioRelay](https://audiorelay.net/)) to stream audio to other devices for use. When I apply this PR, all other modes work fine, but there is no sound return from NVDA in "NVDA on the left and applications on the right" and "NVDA on the right and applications on the left" modes.

This audio streaming software creates a virtual audio device through which it gets audio from the rest of the applications. In both modes NVDA's visual audio bar is jumping, but the virtual audio device backend has no audio output.

Sorry, I'm not sure if this is a known issue with this PR.

mltony:
@hwf1324, could you explain more?\
1. Which modes do work? Paste the whole list.
2. Which device you're streaming from and which device streaming to?
In general I have the feeling that sound split just won't be compatible with that application you're using. Target audience of sound split would be people who use either speakers or headphones connected directly.

source/audio.py
LeonarddeR: In that case, please consider making audio a package where soundSPlitting is a module, e.g. audio\soundSplitting.py

source/comInterfaces/coreAudio/audioclient/__init__.py
LeonarddeR: As far as I can see, the only extra dependency Pycaw introduces is psutil. It is very likelike that py2exe will strip that, so in the end, there's no extra dependency.

mltony: 1. IIRC psutil has a bunch of dependencies of its own.
2. Didn't know about py2exe is that smart.
3. pycaw is missing one of the com interfaces - I had to add it myself. Sure it can be argued that we can import pycaw and write that new interface in a separate file - but current solution is more concise.
4. pycaw is just a wrapper around com interfaces. It is marginally more useful when working with raw com interfaces. But NVDA is working with raw com interfaces all over the place. Plus the new com interface will have to be interacted with outside of pycaw object model. So in my mind there is not much utility in pycaw object model.
5. Another PR that I'm planning - microphone mute - required changes to pycaw object model as well. So again it would be more convenient to work with raw com interfaces.

source/audio.py
```diff
@@ -0,0 +1,205 @@
...
+	def getAppVolume(self) -> VolumeTupleT:
+		match self:
+			case SoundSplitState.OFF | SoundSplitState.NVDA_LEFT_APPS_BOTH | SoundSplitState.NVDA_RIGHT_APPS_BOTH:
+				return (1.0, 1.0)
+			case SoundSplitState.NVDA_RIGHT_APPS_LEFT | SoundSplitState.NVDA_BOTH_APPS_LEFT:
+				return (1.0, 0.0)
+			case SoundSplitState.NVDA_LEFT_APPS_RIGHT | SoundSplitState.NVDA_BOTH_APPS_RIGHT:
+				return (0.0, 1.0)
+			case _:
+				raise RuntimeError
```
CyrilleB79, line 67: This error may occur if we add new items in the enum in the future. So let's log the faulty member:

```suggestion
				raise RuntimeError(f"{self=}")
```

source/audio.py
```diff
@@ -0,0 +1,205 @@
...
+	def getNVDAVolume(self) -> VolumeTupleT:
+		match self:
+			case SoundSplitState.OFF | SoundSplitState.NVDA_BOTH_APPS_LEFT | SoundSplitState.NVDA_BOTH_APPS_RIGHT:
+				return (1.0, 1.0)
+			case SoundSplitState.NVDA_LEFT_APPS_RIGHT | SoundSplitState.NVDA_LEFT_APPS_BOTH:
+				return (1.0, 0.0)
+			case SoundSplitState.NVDA_RIGHT_APPS_LEFT | SoundSplitState.NVDA_RIGHT_APPS_BOTH:
+				return (0.0, 1.0)
+			case _:
+				raise RuntimeError
```
CyrilleB79, line 78: Idem:

```suggestion
				raise RuntimeError(f"{self=}")
```

CyrilleB79 (commented):
The new `match` syntax is very nice and clear. Here are just some requests to clarify an hypothetical future error.

Regarding `match`, I want also to take the opportunity to remind that it is not detected by our version of flake8 to evaluate the complexity. Thus we have to be careful not to write functions too complex when using it because the linter won't complain of it. @mltony, this is a general reminder, but it does not apply at all to the use done in this PR, which is very clear and short instead.

hwf1324:
>     1. Which modes do work? Paste the whole list.

- "Disabled sound split": normalcy
- "NVDA on the left and applications on the right": does not transmit NVDA sound
- "NVDA on the left and applications in both channels": supporting
- "NVDA on the right and applications on the left": does not transmit NVDA sound
- "NVDA on the right and applications in both channels": supporting
- "NVDA in both channels and applications on the left": supporting
- "NVDA in both channels and applications on the right": supporting

>     2. Which device you're streaming from and which device streaming to?

Transfer sound from your laptop to your cell phone and listen to multiple devices with one headset.

>        In general I have the feeling that sound split just won't be compatible with that application you're using. Target audience of sound split would be people who use either speakers or headphones connected directly.

I think it should only be partially incompatible, if you can set which audio streams can follow the NVDA sound, or set the mode of certain audio streams, or set to exclude certain audio streams, you can solve this problem. But I think it goes beyond the scope of this PR, and even beyond NVDA.

mltony:
@hwf1324 , based on your response it appears that your application doesn't work when the set of NVDA channels and the set of channels for other apps are disjoints - e.g. no common elements. We can thus deduce that the app takes NVDA sound, but also the sound goes through setting applied to all other apps and then it transmits sound to mobile phone, which ends up being zero in both channels. I don't see a way to fix that except by not using sound split in conjunction with that ap.

cary-rowen:
> global command NVDA+alt+s that toggles sound split between off, NVDA on the left and NVDA on the right (default behavior).

Although this is configurable, I was wondering, if there is a need to improve this default behavior:
If we confirm that we only need three options to cycle, Then one of the options can be removed to include NVDA on the left(or right) and applications in both channels".

To me, if there are too many options to cycle, then we might need a reverse cycle gesture.
The current default behavior "NVDA on the left and NVDA on the right" doesn't make much difference to me between the two cases.

source/audio/soundSplit.py
```diff
@@ -0,0 +1,205 @@
...
+			# Translators: Sound split state
+			SoundSplitState.NVDA_RIGHT_APPS_LEFT: _("NVDA on the right and applications on the left"),
+			# Translators: Sound split state
+			SoundSplitState.NVDA_RIGHT_APPS_BOTH: _("NVDA on the right and applications in both channels"),
+			# Translators: Sound split state
+			SoundSplitState.NVDA_BOTH_APPS_LEFT: _("NVDA in both channels and applications on the left"),
+			# Translators: Sound split state
+			SoundSplitState.NVDA_BOTH_APPS_RIGHT: _("NVDA in both channels and applications on the right"),
+		}
+
```
hwf1324, line 57: Remove redundant indentation

```suggestion

```

CyrilleB79:
> > global command NVDA+alt+s that toggles sound split between off, NVDA on the left and NVDA on the right (default behavior).
> 
> Although this is configurable, I was wondering, if there is a need to improve this default behavior: If we confirm that we only need three options to cycle, Then one of the options can be removed to include NVDA on the left(or right) and applications in both channels".

I fully agree. Given it is configurable, there is no point in including symmetrical modes in the default available modes. Either we include only 2 modes (off and NVDALeft+AppsRight) or 3 modes as described by @cary-rowen.
Having 2 modes would satisfy the needs of the majority of the people, so it would make sense to set it as a default. On the opposite, a third mode would help promote the movie/music use case (NVDALeft+AppsBoth). I have no strong preference among these 2 options.

> To me, if there are too many options to cycle, then we might need a reverse cycle gesture. The current default behavior "NVDA on the left and NVDA on the right" doesn't make much difference to me between the two cases.

I agree here too. There are actually too many options to cycle if all the modes are included in the cycling command. Though, that's not a common use case. Thus, I'd leave the backward cycling command unassigned.

mltony:
I updated default setting as suggested by #cary-rowen.
I think it is premature to introduce reverse cycle command, as I would think most of the users will only cycle between off and their favorite sound split mode.

hwf1324:
This sets the volume of the corresponding process by its PID.

Is it possible to add a way to ignore volume settings by process name?

First get the PID by process name, then skip the process with that PID.

mltony:
@hwf1324 ,
You can implement this in an add-on. You can call `applyToAllAudioSessions` function with a custom lambda function that performs any logic you can think of. I'm not sure there is a concept of process name, but you can lookup information given pid by calling `appModuleHandler.AppModule(16064)`

CyrilleB79 (commented):
Some more little changes to improve UI and documentation.

source/audio/__init__.py
```diff
@@ -0,0 +1,16 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2015-2021 NV Access Limited
```
seanbudd, line 2:
```suggestion
# Copyright (C) 2024 NV Access Limited
```

source/comInterfaces/coreAudio/audioclient/__init__.py
seanbudd: I think generally we should avoid copying code from external dependencies where possible.
I think it would be much more elegant to use pycaw as a pip dedependency and patch as needed from NVDA, and/or make upstream PRs to fix pycaw.

mltony:
@seanbudd, that would be a significant rewrite. I can try to send pr to pycaw maintainer.
Also if I switch to pycaw API, then we'll automatically bring a new useless dependency psUtil package that we don't really need. Most likely it won't be stripped away by py2exe as looking at pycaw code querying audioSessions also resolves process names through psUtil even though we don't need that. Are you going to be ok with that?

XLTechie:
@seanbudd @mltony for what it's worth, there are a few add-ons which would very 
much welcome the dependency of psutil. I believe Resource Monitor is one. Sound 
Splitter also used it, though that one will be decommissioned obviously.

There are others, though I can't think of them now.

cary-rowen:
UnMute and VolumeAdjustment add-on also depends on psutil

mltony:
I sent PR to pycaw author AndreMiras/pycaw#87 - let's see if it gets merged.

cary-rowen:
Congratulations to @mltony on the PR that was quickly merged into pycaw, which is great news.

mltony:
@seanbudd, I switched to pycaw as you suggested.

seanbudd:
Could you please update the PR description?

mltony:
@seanbudd , just updated description.

AndreMiras:
FYI I've made a pycaw release 20240210 that has the `IChannelAudioVolume` interface in.

mltony:
@AndreMiras, thanks, I just updated reference to pycaw production version 20240210 instead of github commit ID.

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2015-2021 NV Access Limited
```
seanbudd, line 2:
```suggestion
# Copyright (C) 2024 NV Access Limited
```

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+	def getAppVolume(self) -> VolumeTupleT:
+		match self:
+			case SoundSplitState.OFF | SoundSplitState.NVDA_LEFT_APPS_BOTH | SoundSplitState.NVDA_RIGHT_APPS_BOTH:
+				return (1.0, 1.0)
+			case SoundSplitState.NVDA_RIGHT_APPS_LEFT | SoundSplitState.NVDA_BOTH_APPS_LEFT:
+				return (1.0, 0.0)
+			case SoundSplitState.NVDA_LEFT_APPS_RIGHT | SoundSplitState.NVDA_BOTH_APPS_RIGHT:
+				return (0.0, 1.0)
+			case _:
+				raise RuntimeError(f"{self=}")
```
seanbudd, line 59: could you explain what message is logged with this run time error? Would it be better to make the message more explicit?

```suggestion
				raise RuntimeError(f"Unexpected or unknown state {self=}")
```

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+			case SoundSplitState.NVDA_LEFT_APPS_RIGHT | SoundSplitState.NVDA_LEFT_APPS_BOTH:
+				return (1.0, 0.0)
+			case SoundSplitState.NVDA_RIGHT_APPS_LEFT | SoundSplitState.NVDA_RIGHT_APPS_BOTH:
+				return (0.0, 1.0)
+			case _:
+				raise RuntimeError(f"{self=}")
+
+
+audioSessionManager: IAudioSessionManager2 | None = None
+activeCallback: AudioSessionNotification = None
```
seanbudd, line 74:
```suggestion
activeCallback: AudioSessionNotification | None = None
```

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def initialize() -> None:
+	if nvwave.usingWasapiWavePlayer():
+		global audioSessionManager
+		audioSessionManager = AudioUtilities.GetAudioSessionManager()
+		state = SoundSplitState(config.conf['audio']['soundSplitState'])
```
seanbudd, line 81:
```suggestion
		state = SoundSplitState(config.conf["audio"]["soundSplitState"])
```

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+			case _:
+				raise RuntimeError(f"{self=}")
+
+
+audioSessionManager: IAudioSessionManager2 | None = None
+activeCallback: AudioSessionNotification = None
+
+
+def initialize() -> None:
+	if nvwave.usingWasapiWavePlayer():
```
seanbudd, line 78: can we log a debug message on "else" here - "Cannot initialize sound split as WASAPI is disabled"

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+		global audioSessionManager
+		audioSessionManager = AudioUtilities.GetAudioSessionManager()
+		state = SoundSplitState(config.conf['audio']['soundSplitState'])
+		setSoundSplitState(state)
+
+
+@atexit.register
+def terminate():
+	if nvwave.usingWasapiWavePlayer():
```
seanbudd, line 87: same here for adding logging

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+		setSoundSplitState(SoundSplitState.OFF)
+		unregisterCallback()
+
+
+def applyToAllAudioSessions(
```
seanbudd, line 92: please add a docstring explaining the purpose/use of this function

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+def toggleSoundSplitState() -> None:
+	if not nvwave.usingWasapiWavePlayer():
+		message = _(
+			# Translators: error message when wasapi is turned off.
+			"Sound split cannot be used. "
+			"Please enable WASAPI in the Advanced category in NVDA Settings to use it."
+		)
+		ui.message(message)
+		return
+	state = SoundSplitState(config.conf['audio']['soundSplitState'])
```
seanbudd, line 147:
```suggestion
	state = SoundSplitState(config.conf["audio"]["soundSplitState"])
```

mltony:
@seanbudd,
> Please merge this section into SelectSoundSplitMode and CustomizeSoundSplitModes
Done. What do you want me to do with the original section? I copied this style from speech modes command, which has one entry for its shortcut and another for its counterpart in settings. So far I cut down the original section down to brief description and a link to sound split settign.

user_docs/en/userGuide.t2t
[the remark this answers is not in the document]
seanbudd: I think as cyrille outlined here, the paragraph is not necessary. The documentation in settings should be sufficient

source/audio/soundSplit.py
```diff
@@ -0,0 +1,157 @@
...
+				channelVolume.SetChannelVolume(0, leftNVDAVolume, None)
+				channelVolume.SetChannelVolume(1, rightNVDAVolume, None)
+
+	volumeSetter = VolumeSetter()
+	applyToAllAudioSessions(volumeSetter)
+
+
+def toggleSoundSplitState() -> None:
+	if not nvwave.usingWasapiWavePlayer():
+		message = _(
```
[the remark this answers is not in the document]
mltony: I added a spoken warning if we found at least one audio session has other than 2 channels.

source/gui/settingsDialogs.py
seanbudd: @CyrilleB79 - I mentioned here: https://github.com/nvaccess/nvda/pull/16071#discussion_r1498438921

I agree, but would like to hear more feedback on this. The warning certainly doesn't hurt much, but I don't imagine anyone wanting remove "Off" from the list.

source/audio/soundSplit.py
```diff
@@ -0,0 +1,188 @@
...
+	def getNVDAVolume(self) -> VolumeTupleT:
+		match self:
+			case SoundSplitState.OFF | SoundSplitState.NVDA_BOTH_APPS_LEFT | SoundSplitState.NVDA_BOTH_APPS_RIGHT:
+				return (1.0, 1.0)
+			case SoundSplitState.NVDA_LEFT_APPS_RIGHT | SoundSplitState.NVDA_LEFT_APPS_BOTH:
+				return (1.0, 0.0)
+			case SoundSplitState.NVDA_RIGHT_APPS_LEFT | SoundSplitState.NVDA_RIGHT_APPS_BOTH:
+				return (0.0, 1.0)
+			case _:
+				raise RuntimeError(f"{self=}")
```
seanbudd, line 71: please raise a similar message as `getAppVolume`

source/audio/soundSplit.py
```diff
@@ -0,0 +1,188 @@
...
+		log.debug("Cannot initialize sound split as WASAPI is disabled")
+
+
+@atexit.register
+def terminate():
+	if nvwave.usingWasapiWavePlayer():
+		setSoundSplitState(SoundSplitState.OFF)
+		unregisterCallback()
+	else:
+		log.debug("Skipping terminating sound split as wasapi is mode is not enabled.")
```
seanbudd, line 94:
```suggestion
		log.debug("Skipping terminating sound split as WASAPI is disabled.")
```

source/audio/soundSplit.py
seanbudd:
```suggestion
				raise RuntimeError(f"Unexpected or unknown state {self=}")
```

seanbudd (dismissed):
Thanks @mltony

source/gui/settingsDialogs.py
```diff
@@ -2736,6 +2750,48 @@ def makeSettings(self, settingsSizer: wx.BoxSizer) -> None:
 		self.bindHelpEvent("AudioAwakeTime", self.audioAwakeTimeEdit)
 		self.audioAwakeTimeEdit.Enable(nvwave.usingWasapiWavePlayer())
 
+	def _appendSoundSplitModesList(self, settingsSizerHelper: guiHelper.BoxSizerHelper) -> None:
+		self._allSoundSplitModes = list(audio.SoundSplitState)
+		self.soundSplitModesList: nvdaControls.CustomCheckListBox = settingsSizerHelper.addLabeledControl(
+			# Translators: Label of the list where user can select sound split modes that will be available.
+			_("&Modes available in the Cycle sound split mode command:"),
```
seanbudd, line 2757:
```suggestion
			_("&Modes available in the 'Cycle sound split mode' command:"),
```

user_docs/en/userGuide.t2t
```diff
@@ -1957,6 +1957,38 @@ Setting the value too high may cause the battery of the sound output device to d
...
+%kc:endInclude
+
+By default this command will cycle between the following modes:
+- Disabled sound split: both NVDA and other applications output sounds to both left and right channels.
+- NVDA on the left and applications on the right: NVDA will speak in the left channel, while other applications will play sounds in the right channel.
+- NVDA on the right and applications on the left: NVDA will speak in the right channel, while other applications will play sounds in the left channel.
+-
+
+There are more advanced sound split modes available in NVDA setting combo box.
+Please note, that sound split doesn't work as a mixer. For example, if an application is playing a stereo sound track while sound split is set to "NVDA on the left and applications on the right", then you will only hear the right channel of the sound track, while the left channel of the sound track will be muted.
```
seanbudd, line 1978:
```suggestion
Please note, that sound split doesn't work as a mixer.
For example, if an application is playing a stereo sound track while sound split is set to "NVDA on the left and applications on the right", then you will only hear the right channel of the sound track, while the left channel of the sound track will be muted.
```

user_docs/en/userGuide.t2t
```diff
@@ -1957,6 +1957,38 @@ Setting the value too high may cause the battery of the sound output device to d
 
 
 You can set the time to zero in order to disable this feature.
+
+==== Sound split====[SelectSoundSplitMode]
+
+The sound split feature allows users to make use of their stereo output devices, such as headphones and speakers. Sound split makes it possible to have NVDA speech in one channel (e.g. left) and have all other applications play their sounds in the other channel (e.g. right).
```
seanbudd, line 1963:
```suggestion
The sound split feature allows users to make use of their stereo output devices, such as headphones and speakers.
Sound split makes it possible to have NVDA speech in one channel (e.g. left) and have all other applications play their sounds in the other channel (e.g. right).
```

user_docs/en/userGuide.t2t
```diff
@@ -1957,6 +1957,38 @@ Setting the value too high may cause the battery of the sound output device to d
...
 You can set the time to zero in order to disable this feature.
+
+==== Sound split====[SelectSoundSplitMode]
+
+The sound split feature allows users to make use of their stereo output devices, such as headphones and speakers. Sound split makes it possible to have NVDA speech in one channel (e.g. left) and have all other applications play their sounds in the other channel (e.g. right).
+By default sound split is disabled, which means that all applications including NVDA will play sounds in both left and right channels.
+A gesture allows cycling through the various sound split modes:
+%kc:beginInclude
+|| Name | Key | Description |
+| Cycle Sound Split Mode | ``NVDA+alt+s`` | Cycles between speech modes. |
```
seanbudd, line 1968:
```suggestion
| Cycle Sound Split Mode | ``NVDA+alt+s`` | Cycles between sound split modes. |

```

source/gui/settingsDialogs.py
```diff
@@ -2707,8 +2708,63 @@ def makeSettings(self, settingsSizer: wx.BoxSizer) -> None:
 		self.bindHelpEvent("SoundVolume", self.soundVolSlider)
 		self.soundVolSlider.SetValue(config.conf["audio"]["soundVolume"])
 
+		# Translators: This is a label for the sound split combo box in the Audio Settings dialog.
+		soundSplitLabelText = _("Sound split mode:")
```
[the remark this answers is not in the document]
mltony: Uhm, what is accelerator key?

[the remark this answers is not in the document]
seanbudd: It's a keyboard shortcut key denoted by a preceding ampersand 

```suggestion
		soundSplitLabelText = _("&Sound split mode:")
```

user_docs/en/changes.t2t
```diff
@@ -25,6 +25,7 @@ What's New in NVDA
   -
 - Added support for the BrailleEdgeS2 braille device. (#16033)
 - NVDA will keep the audio device awake after speech stops, in order to prevent the start of the next speech being clipped with some audio devices such as Bluetooth headphones. (#14386, @jcsteh, @mltony)
+- Sound split: allows to split NVDA sounds in one channel (e.g. left) while sounds from all other applications in the other channel (e.g. right) - toggled by ``NVDA+alt+s`` (#12985, @mltony)
```
seanbudd, line 28:
```suggestion
- Sound split:  (#12985, @mltony)
  - Allows splitting NVDA sounds in one channel (e.g. left) while sounds from all other applications in the other channel (e.g. right).
  - Toggled by ``NVDA+alt+s``.
  -
```

user_docs/en/changes.t2t
```diff
@@ -25,6 +25,10 @@ What's New in NVDA
   -
 - Added support for the BrailleEdgeS2 braille device. (#16033)
 - NVDA will keep the audio device awake after speech stops, in order to prevent the start of the next speech being clipped with some audio devices such as Bluetooth headphones. (#14386, @jcsteh, @mltony)
+- Sound split:  (#12985, @mltony)
```
seanbudd, line 28:
```suggestion
- Sound split: (#12985, @mltony)
```

seanbudd (approved):
Thanks @mltony

paulber19:
What is the reason for soundSplit to be limited to only "Wasapi" mode?
THANKS.

CyrilleB79:
Because its technical implementation is not the same with and without WASAPI. Since WASAPI is the default mode and should remain the only one possible in the future, there was no point in spending time to also support sound split without WASAPI.

PR #16074, opened 2024-01-21
Make logging of exception which occurred on non main thread less verbose

### LINKED ISSUE #16070: Issue when logging error or printing to stderr in another thread
Opened in replacement of #16065 for clarity.

### Steps to reproduce:
Execute the following instructions in the console:
```
>>> import threading
>>> import core
>>> def f():
...  return 0/0  # raises an error
... 
>>> core.callLater(1000, lambda: threading.Thread(None, f).start())
<wx.core.CallLater object at 0x05166470>
```
### Actual behavior:
The following messages are logged
```
ERROR - stderr (14:08:47.089) - Thread-71 (f) (15936):
Exception in thread
ERROR - stderr (14:08:47.117) - Thread-71 (f) (15936):
Thread-71 (f)
ERROR - stderr (14:08:47.126) - Thread-71 (f) (15936):
:
ERROR - stderr (14:08:47.135) - Thread-71 (f) (15936):
ERROR - stderr (14:08:47.145) - Thread-71 (f) (15936):
  File "threading.pyc", line 1045, in _bootstrap_inner
ERROR - stderr (14:08:47.164) - Thread-71 (f) (15936):
  File "threading.pyc", line 982, in run
ERROR - stderr (14:08:47.200) - Thread-71 (f) (15936):
  File "<console>", line 2, in f
ERROR - stderr (14:08:47.210) - Thread-71 (f) (15936):
ZeroDivisionError
ERROR - stderr (14:08:47.262) - Thread-71 (f) (15936):
:
ERROR - stderr (14:08:47.287) - Thread-71 (f) (15936):
division by zero
```
### Expected behavior:
The following message should be logged instead.
```
Exception in thread Thread-71 (f):
  File "threading.pyc", line 1045, in _bootstrap_inner
  File "threading.pyc", line 982, in run
  File "<console>", line 2, in f
ZeroDivisionError: division by zero
```
None
Installed
#### NVDA version:
2024.1beta3
#### Windows version:
Windows 10 22H2 (AMD64) build 19045.3803
N/A
N/A
Not re-tested but probably yes.
* NVDA 2023.3 (run from installer): No issue
* nvda_snapshot_alpha-29722,b67e7f46 (run from installer): Issue reproduced

Thus the regression is between these two versions. But I have not been able to narrow down the range for now since running from source will require more time (only tested with installers for now).

Yes
Not tested

lukaszgo1:
### Link to issue number:
Fixes #16070
### Summary of the issue:
To log unhandled exceptions NVDA relies on a custom except hook installed when initializing logging. Until Python 3.8 custom except hooks were not called for exception which occurred outside of the main thread - they were simply written to stderr (see https://github.com/python/cpython/issues/42148). This worked well for NVDA, since it replaces stderr with a custom object which logs everything written to stderr, so the exceptions ended up in our log, just in a different way. In Python 3.8 it become possible to install a custom hook which gets called for exceptions which are raised in non main threads, but as part of this change the way in which they are handled by default has changed - the traceback is written to stderr line by line, rather than as a single call to `print` as it used to be. For NVDA it means that each line of the traceback results in a separate log entry making logging extremely noisy.
Exception from threads different than the main one are logged in the same way as they were in NVDA 2023.3.
A custom except hook for exception raised in threads was introduced. It logs them as an exception with additional information explaining in which thread they were raised.
### Testing strategy:
Performed STR from #16070 - ensured that logging looks similarly to the one from 2023.3 and that single exception results in a single log entry.
None known
### Code Review Checklist:

- [X] Documentation:
  - Change log entry
  - User Documentation
- [X] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/logHandler.py
+++ b/source/logHandler.py
@@ -15,11 +15,13 @@ import logging
 import inspect
 import winsound
 import traceback
-from types import FunctionType
+from types import FunctionType, TracebackType
 import globalVars
 import winKernel
 import buildVersion
 from typing import (
+	Literal,
+	NamedTuple,
 	Optional,
 	TYPE_CHECKING,
 )
@@ -192,6 +194,9 @@ else:
 		return text
 
 
+_excInfo_t = tuple[type[BaseException] | None, BaseException | None, TracebackType | None]
+
+
 class Logger(logging.Logger):
 	# Import standard levels for convenience.
 	from logging import DEBUG, INFO, WARNING, WARN, ERROR, CRITICAL
@@ -258,7 +263,7 @@ class Logger(logging.Logger):
 			return
 		self._log(log.IO, msg, args, **kwargs)
 
-	def exception(self, msg="", exc_info=True, **kwargs):
+	def exception(self, msg: str = "", exc_info: Literal[True] | _excInfo_t = True, **kwargs):
 		"""Log an exception at an appropriate level.
 		Normally, it will be logged at level "ERROR".
 		However, certain exceptions which aren't considered errors (or aren't errors that we can fix) are expected and will therefore be logged at a lower level.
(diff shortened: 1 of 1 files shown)


CyrilleB79 (approved):
Thanks for this fix @lukaszgo1. Working well!

seanbudd (approved):

PR #16098, opened 2024-01-26
Report the code of combined characters

### LINKED ISSUE #5011: NVDA doesn't speak the diacritics of some IPA symbols in applications other than MS Word
**Reported by Sukil on 2015-03-31 18:14**
Steps to reproduce:
1. Type for example the letter "s" in MS Word, followed by the symbol 0x33A. Both symbols should merge and eSpeak should only play the sound associated with the symbol when reading character by character.
2. Copy that symbol into some other application (tested with EspeakEdit and PowerPoint). When reading character by character, NVDA ignores the diacritics and says "s". In addition, when pressing NVDA+Numpad5 two and three tymes, the description of a regular "s" is given.

CyrilleB79:
### Link to issue number:
Closes #5011.

### Summary of the issue:
Some characters are built with two or more Unicode characters: the first is a normal character and the following ones are combining characters. For example the character "é". It is built from the character "e" and the character "combining acute accent".

Pressing numpad2 twice, the description of the two characters are reported. However pressing numpad2 3 times, no character code is reported and the compound character is just reported as at first press; thus, triple press brings no useful information in this case.

When pressing three times numpad2:
* if the character under the review cursor is a single character, its decimal and hexadecimal codes will be reported as in pevious NVDA versions
* if the character under the review cursor is compound of two or more characters, the codes for each character building it will be reported:
  * In braille, both decimal and hexadecimal codes will be reported for each character
  * With speech, only the decimal code is reported to avoid cluttering the speech output with too much numeric information that may be hard to parse.

Modified the dedicated function.

### Testing strategy:
Manual test with the character in #5011 and with "é".
In some applications (Firefox edit field, NVDA Python console output), navigating with the caret by character jumps over the whole compound character but navigating with the review cursor goes through each individual character building the compound character. In this case, each character is considered a separate character and handled as such. This PR does not change anything in this case.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -1922,13 +1922,17 @@ class GlobalCommands(ScriptableObject):
 			speech.spellTextInfo(info,useCharacterDescriptions=True)
 		else:
 			try:
-				c = ord(info.text)
+				cList = [ord(c) for c in info.text]
 			except TypeError:
 				c = None
-			if c is not None:
-				speech.speakMessage("%d," % c)
-				speech.speakSpelling(hex(c))
-				braille.handler.message(f"{c}, {hex(c)}")
+			if cList:
+				
+				for c in cList:
+					speech.speakMessage("%d," % c)
+					# Report hex along with decimal only when there is one character; else, it's confusing.
+					if len(cList) == 1:
+						speech.speakSpelling(hex(c))
+				braille.handler.message("; ".join(f"{c}, {hex(c)}" for c in cList))
 			else:
 				log.debugWarning("Couldn't calculate ordinal for character %r" % info.text)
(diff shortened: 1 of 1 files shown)


CyrilleB79:
Notes:
* The User Guide or the script's description in input help mode does not provide details information in the very specific case of compound characters, so I have not modified them
* I have not added anything in the change log, since this change seems quite minor, and I feel that a change log item would bring more confusion than clarification on such a detail.

Also cc @sukiletxe if you wish to comment or test the build once it is ready.

seanbudd (approved):

PR #16103, opened 2024-01-28
Update translations

michaelDCurran:


gerald-hartig (approved):
It looks like only a few of the translation files have changes to their contents. Just wanted to check that is expected.

PR #16147, opened 2024-02-08
Various fix-up of keep audio awake feature

### LINKED ISSUE #16148: Play silence to keep audio device awake
### Is your feature request related to an issue? Please define.
It will be difficult for those who want to change the silence duration (default = 30 seconds) setting in the advanced settings panel that comes with the play silence feature to keep the audio device awake.
### Describe your desired solution
Moving it to the sound panel in the settings would be better for users who want to change this setting.
### Describe the alternatives you considered
No.
### Additional context

CyrilleB79:
This PR contains various fix-ups of PR #16099, some less important than others. Cc @mltony 

### Link to issue number:
Fix-up of #16099

Fixes #16148

### Summary of the issue:
1. Error when disabling WASAPI and reopening advanced settings:
```
ERROR - unhandled exception (11:14:54.349) - MainThread (6028):
  File "gui\settingsDialogs.pyc", line 4517, in onCategoryChange
  File "gui\settingsDialogs.pyc", line 693, in onCategoryChange
  File "gui\settingsDialogs.pyc", line 675, in _doCategoryChange
  File "gui\settingsDialogs.pyc", line 603, in _getCategoryPanel
  File "gui\settingsDialogs.pyc", line 362, in __init__
  File "gui\settingsDialogs.pyc", line 372, in _buildGui
  File "gui\settingsDialogs.pyc", line 3514, in makeSettings
  File "gui\settingsDialogs.pyc", line 3259, in __init__
  File "gui\settingsDialogs.pyc", line 3352, in _onWasapiChange
  File "gui\guiHelper.pyc", line 232, in Enable
TypeError: Window.Enable(): argument 1 has unexpected type 'FeatureFlag'
ERROR - unhandled exception (11:14:54.361) - MainThread (6028):
  File "gui\guiHelper.pyc", line 218, in _onEnableChanged
TypeError: Window.Enable(): argument 1 has unexpected type 'FeatureFlag'
```

2. Oversight in `AdvancedPanelControls.haveConfigDefaultsBeenRestored` in `settingsDialogs.py`.

3. In `settingsDialogs.py`, rename variable to be consistent with the GUI ("silence" -> "keep audio awake")

4. Documentation (User Guide and Change log) updates

* The option is enabled or disabled depending on the real current WASAPI state, not the state defined in the configuration for the next restart of NVDA.
* The "keep audio awake" option is moved to audio settings
* The documentation does not mention "playing silence" at all anymore

For point 1: use real WASAPI state instead of what is in the config for the next restart, as it was already done in other audio settings depending on WASAPI.
### Testing strategy:
Checked that the error reported in point 1 does not occur anymore.
* Checked that the feature still works, using the debug config parameter `whiteNoiseVolume`.

Note: I am not able to directly test the feature to keep audio awake, since I have no impacted device (e.g. Bluetooth headset).
None
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


mltony:
Looks good to me!

seanbudd:
>should we use "speech sequence", "spoken phrase", "utterance" or something else be used in the documentation

Why not just: "This edit box specifies how long NVDA plays silence after speech ends"

seanbudd:
I think it would be good to move the setting to the audio panel in this PR

cary-rowen:
> I think it would be good to move the setting to the audio panel in this PR

The default value is sufficient for me and even most users, it is not a setting that needs to be changed frequently, so I think it is appropriate to put it in the advanced settings.

mltony:
I agree with @cary-rowen - most users won't need to change this value, so I'd think it belongs on the advanced panel.

XLTechie:
> The default value is sufficient for me and even most users, it is not a setting that needs to be changed frequently, so I think it is appropriate to put it
> in the advanced settings.

It is not a matter of whether the default is sufficient. It is a matter of where 
users would expect to find it. IMO, that's in audio. There is nothing overly 
advanced about this setting.

Plus, though I haven't tried the latest build, I think the only way to turn this 
feature on/off, is to change this setting, is it not? If that is so, even more 
reason to keep it out of Advanced and put it in Audio. It advertises the 
feature, and what can be done with it. Some users are afraid of anything in 
Advanced, as anyone inclined to be afraid of it likely should be.

XLTechie:
> Why not just: "This edit box specifies how long NVDA plays silence after speech ends"

Yes, exactly the kind of thing I was targeting. Explains it with no technical 
terminology at all. CC @Adriani90

CyrilleB79:
I have addressed review comments:
* improved the documentation
* moved the parameter to audio settings

To see the documentation improvements before the paragraph was moved, do:
`git diff master 1825f0bfeef3e0ecd7a163c4fd13f3057ae0c89b -- Copie""$ user_docs/en/userGuide.t2t`
or better, if master has moved in the meantime:
`git diff 49d5e581d499768d3858443ce25d2ccba4319630 1825f0bfeef3e0ecd7a163c4fd13f3057ae0c89b -- Copie""$ user_docs/en/userGuide.t2t`

The following changes have been done in documentation:
* replace "speech sequence" by "after speech end" (see https://github.com/nvaccess/nvda/pull/16147#issuecomment-1935073292).
* Removed completely the notion of playing silence, which is of no use for the end user. (in line with https://github.com/nvaccess/nvda/pull/16099#issuecomment-1931352113)
* For unavailability without WASAPI, mention the whole feature reather than only the control, which is not a slider actually. When there is WASAPI off, not only the control is unavailable, but the whole feature.

Regarding the move of the option to audio settings:
* Added a shortcut key. My opinion is having shortcut keys for any control can always be a plus for some people, at least in all panels except advanced settings.

Brian1Gaff:
If it can work at the very start of a spoken item after a long gap, then I'd agree. Luckily my main computer has no latency sound turn off as its a desktop, but have encountered these effects on Bluetooth devices and some laptops, and there that program Silencio seemed to be the best approach. 
Brian

-- 
***@***.*** 
in the display name field.
  ----- Original Message -----

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
 This setting only takes effect when "Volume of NVDA sounds follows voice volume" is disabled.
 This option is not available if you have started NVDA with [WASAPI disabled for audio output #WASAPI] in Advanced Settings.
 
+==== Duration of keeping audio device awake ====[KeepAudioAwakeDuration]
```
XLTechie, line 1937: Personally, in English, I find this title slightly awkward.
Here is a suggested alternative. Note that I didn't change the anchor name, as that is referenced else ware, and should be changed holistically if it is changed.
I'm not in love with this alternative either, but it may be somewhat better.

```suggestion
==== Keep audio device awake time ====[KeepAudioAwakeDuration]
```

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
 This setting only takes effect when "Volume of NVDA sounds follows voice volume" is disabled.
 This option is not available if you have started NVDA with [WASAPI disabled for audio output #WASAPI] in Advanced Settings.
 
+==== Duration of keeping audio device awake ====[KeepAudioAwakeDuration]
+
+This edit box specifies how long NVDA keeps the audio device awake after speech end.
+This allow to avoid certain glitches like dropped parts of speech sequences.
```
XLTechie, line 1940:
```suggestion
This allows NVDA to avoid certain speech glitches like dropped parts of words
```

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
 This setting only takes effect when "Volume of NVDA sounds follows voice volume" is disabled.
 This option is not available if you have started NVDA with [WASAPI disabled for audio output #WASAPI] in Advanced Settings.
 
+==== Duration of keeping audio device awake ====[KeepAudioAwakeDuration]
+
+This edit box specifies how long NVDA keeps the audio device awake after speech end.
+This allow to avoid certain glitches like dropped parts of speech sequences.
+This can happen due to audio devices (especially Bluetooth and wireless devices) entering stand-by mode.
+This might also be helpful in certain other use cases, such as running NVDA inside a virtual machine, such as Citrix Virtual Desktop, or on certain models of laptops.
```
XLTechie, line 1942:
```suggestion
This might also be helpful in other use cases, such as when running NVDA inside a virtual machine (e.g. Citrix Virtual Desktop), or on certain laptops.
```

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
...
 This option is not available if you have started NVDA with [WASAPI disabled for audio output #WASAPI] in Advanced Settings.
 
+==== Duration of keeping audio device awake ====[KeepAudioAwakeDuration]
+
+This edit box specifies how long NVDA keeps the audio device awake after speech end.
+This allow to avoid certain glitches like dropped parts of speech sequences.
+This can happen due to audio devices (especially Bluetooth and wireless devices) entering stand-by mode.
+This might also be helpful in certain other use cases, such as running NVDA inside a virtual machine, such as Citrix Virtual Desktop, or on certain models of laptops.
+
+Lower values may mean audio would be cut-off more often, as a device may enter stand-by mode, causing the next speech to be clipped.
```
XLTechie, line 1944:
```suggestion
Lower values may allow audio to be cut-off more often, as a device may enter stand-by mode too soon, causing the start of the following speech to be clipped.
```

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
...
+A too high value may mean the battery of the sound output device discharges faster, as it stays active for longer while no sound is being sent.
```
XLTechie, line 1945:
```suggestion
Setting the value too high, may cause the battery of the sound output device to discharge faster, as it stays active for longer while no sound is being sent.
```

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
...
+
+You can specify the duration to be 0 in order to disable this feature.
```
XLTechie, line 1947: This suggestion is applicable only if you change the word from "duration" to "time" as I commented above.
Note, that in networking and other compute contexts in my experience, the phrase is usually "keep-alive time", not "keep-alive duration". That may be why I mislike the word "duration" here.

```suggestion
You can set the time to zero in order to disable this feature.
```

XLTechie (changes requested):
The following suggestions are mainly to improve the reading of this in my opinion as an English speaker and editor. @Qchristensen might also have some thoughts.

CyrilleB79:
Thanks @XLTechie for your valuable feedback about English spelling / grammar, as well as other wording improvements.

Regarding the name of the option, I'd say that we need to agree on the label of the control in the GUI. Then I'll derive the following:
* name of the paragraph in the User Guide = same as the GUI option, just omitting the part between parenthesis, i.e. "(sec)"
* variable names: something shorter with the most significant words of the option's name
* anchor: idem

I'll wait for other feedback (@seanbudd, @Qchristensen and community).

I agree that having a suitable name in the GUI and in the user guide can participate to the success of a new feature.

*

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
...
 
+==== Duration of keeping audio device awake ====[KeepAudioAwakeDuration]
+
+This edit box specifies how long NVDA keeps the audio device awake after speech ends.
+This allows NVDA to avoid certain speech glitches like dropped parts of words
+This can happen due to audio devices (especially Bluetooth and wireless devices) entering stand-by mode.
+This might also be helpful in other use cases, such as when running NVDA inside a virtual machine (e.g. Citrix Virtual Desktop), or on certain laptops.
+
+Lower values may allow audio to be cut-off more often, as a device may enter stand-by mode too soon, causing the start of the following speech to be clipped.
+Setting the value too high, may cause the battery of the sound output device to discharge faster, as it stays active for longer while no sound is being sent.
```
SaschaCowley, line 1945: There's an extra comma here.
```suggestion
Setting the value too high may cause the battery of the sound output device to discharge faster, as it stays active for longer while no sound is being sent.

```

user_docs/en/userGuide.t2t
SaschaCowley: I think I would just go with `Time to keep audio device awake`

user_docs/en/userGuide.t2t
```diff
@@ -1934,6 +1934,19 @@ This slider allows you to set the volume of NVDA sounds and beeps.
...
 This option is not available if you have started NVDA with [WASAPI disabled for audio output #WASAPI] in Advanced Settings.
 
+==== Duration of keeping audio device awake ====[KeepAudioAwakeDuration]
+
+This edit box specifies how long NVDA keeps the audio device awake after speech ends.
+This allows NVDA to avoid certain speech glitches like dropped parts of words
+This can happen due to audio devices (especially Bluetooth and wireless devices) entering stand-by mode.
+This might also be helpful in other use cases, such as when running NVDA inside a virtual machine (e.g. Citrix Virtual Desktop), or on certain laptops.
+
+Lower values may allow audio to be cut-off more often, as a device may enter stand-by mode too soon, causing the start of the following speech to be clipped.
```
SaschaCowley, line 1944: The various instances of "stand-by" should be replaced with "standby".

source/gui/settingsDialogs.py
```diff
@@ -2708,6 +2708,23 @@ def makeSettings(self, settingsSizer: wx.BoxSizer) -> None:
 		self.soundVolSlider.SetValue(config.conf["audio"]["soundVolume"])
 
 		self._onSoundVolChange(None)
+		
+		audioAwakeDurationLabelText = _(
+			# Translators: The label for a setting in Audio settings panel
+			# to change the duration of keeping audio device awake
+			"Duration of &keeping audio device awake (sec)"
```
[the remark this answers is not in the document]
seanbudd: I think this suggestion is good

user_docs/en/userGuide.t2t
seanbudd: agreed with Sascha here

seanbudd (approved):

PR #16166, opened 2024-02-11
Update translations

michaelDCurran:


source/locale/af_ZA/LC_MESSAGES/nvda.po
```diff
@@ -2658,18 +2658,28 @@ msgid "Toggles native selection mode on and off"
...
+#: browseMode.py:2041
+msgid "Native selection mode unsupported in this browse mode document"
+msgstr ""
+
+#. Translators: the message when native selection mode cannot be turned off in this browse mode document.
+#: browseMode.py:2044
+msgid "Native selection mode cannot be turned off in this browse mode document"
+msgstr ""
+
+#. Translators: the message when native selection mode is not available in this browse mode document.
```
gerald-hartig, line 2670: This Translator note is the same as the browse mode msgid above.
```suggestion
#. Translators: the message when native selection mode is not available in this document.

```

gerald-hartig (approved):

michaelDCurran:
I'd advise against making edits to the po files once they come out of 
Crowdin.

They should really be opaque at this point.

If a change should be made to a translator comment, it is best made in 
the source code itself.

E.g. browseMode.py:2041

Issue #13517, opened 2022-03-21
Use custom extensions to move by sentence in Microsoft Word controls


codeofdusk:
### Is your feature request related to a problem? Please describe.
NVDA is moving toward UIA with custom extensions in Microsoft Office. In older versions of Word without support for custom extensions, it was necessary to use the legacy word `textInfo` implementation to move by sentence, even when accessing via UIA.

### Describe the solution you'd like
Now that there is [support via custom extensions to move by sentence](https://docs.microsoft.com/en-gb/office/uia/word/wordcustompatterns), take advantage of this support.

### Describe alternatives you've considered
Continue using the object model.

LeonidhaMara:
Hi,

I'm new to this project but i would like to work with this feature.

dkager:
We should pick this up, now that UIA in Word has been around for a few more years. I'd like to look into it.

Adriani90:
Cc: @mltony

Adriani90:
@dkager did you have a chance to look into this? See also the PR #16384.

PR #16214, opened 2024-02-22
Implement support for writing UI Automation Remote Operations natively in NVDA using Python

michaelDCurran:
### Link to issue number:
None

### Background
Several years ago Microsoft added a low-level API in Windows to allow the caller to execute multiple actions in a UI Automation remote provider via only one cross-process call.
Limiting the amount of cross-process calls is necessary for providing a responsive experience, especially for interacting with applications hosted in Windows Defender Application Guard (WDAG) or as a Remote Application Integrated Locally (RAIL). There are also scenarios such as searching for / listing headings in large documents, which can bennifit from limiting cross-process calls, even when the application is running on the local machine.
 the [Windows.UI.UIAutomation.Core.CoreAutomationRemoteOperation](https://learn.microsoft.com/en-us/uwp/api/windows.ui.uiautomation.core.coreautomationremoteoperation?view=winrt-22621) is a winRT interface that allows the caller to specify the actions as a string of byte code representing a specific set of instructions for a conceptual virtual machine which is executed within the remote provider. The instructions allow you to manipulate UI Automation elements, patterns and text ranges, as well as perform basic logic and collect information using ints, floats, strings, GUIDs and arrays created within the conceptual virtual machine.
Although these instructions are well-defined, public documentation is limited and programming with these raw instructions can be slow and complicated to read.
Microsoft released the [Microsoft-UI-UIAutomation Remote Operations Library](https://github.com/microsoft/microsoft-ui-uiautomation) which provides some higher level ways of building and executing these instructions, including intermediate winRT interfaces, and high-level pure c++/winRT interfaces.
NV Access has made many prototypes with this library, and has made significant contributions to that project over the years. 
Today, NVDA includes this library as a depedency, and currently uses it to access the new UI Automation Custom Extensions feature, which is required to access certain information in Microsoft Word. In theory, NVDA could make a much greater uuse of UI automation Remote Operations to improve performance across many  parts of Windows and other applications.
However, there are several issues with the Microsoft UI Automation Remote Operations Library which slow down or limit greater usage in NVDA, including:
* All remote operations logic must be written in C++ and not Python. This creates a large learning curve, if not a quite uncomfortable context switch at very least.
 * the Remote Operations Library seems to be no longer maintained by Microsoft.
* The Remote Operations Library currently contributes an extra massive 16 minutes to NvDA's build time, as it must compile the whole of the C++/winRT API set.

### Solution
This PR introduces a new remote operations library written in pure Python, which wraps the low-level CoreAutomationRemoteOperation winRT interfaces in Windows. This new library replaces the need for the older Microsoft Remote Operations Library, and therefore is no longer a dependency.

I had several initial goals in mind with this new library:
1. To support our current usage of remote operations (namely UI Automation custom extensions: isExtension and CallExtension). If I could not achieve our current requirements, there would be no point continuing.
2. To make a significant performance improvement to at least one extra action or scenario in NvDA (E.g. quick navigation by heading in Microsoft Word). This would prove that implementation of a fix / feature was easier with the new library, and would make it worth spending the initial time on it.

I was able to achieve both of these goals.

* In browse mode in Microsoft Word with UIA enabled, quick nav to headings, and listing headings in Elements list, is now up to 30 times faster. E.g. Jumping to a heading at the bottom of a 160 page document all the way from the top previously took ~4 seconds, now it takes ~0.13 seconds.

* A new `UIAHandler._remoteOps` package has been added, which interfaces with the low-level Windows Remote Operations API, providing higher-level APIs and types to aide in building and executing easy to read remote algorithms. See [The NVDA remote operations readme](https://github.com/nvaccess/nvda/blob/remoteOpsLowLevel/source/UIAHandler/_remoteOps/readme.md) for explanations and example code for all features.
* UIAHandler.remote's msWord_getCustomAttributeValue has been re-written to use the new NVDA remote operations library.
* Some extra functions were added to UIAHandler.remote, including findFirstHeadingInTextRange and collectAllHeadingsInTextRange, built on the NVDA remote ops library. UIA browse mode makes use of these functions on Windows 11, to majorly speed up jumping to / listing headings.

### Testing strategy:
* ~40 unit tests were added for the NVDA remote operations library, see [tests/unit/test_UIARemoteOps/test_highLevel.py](https://github.com/nvaccess/nvda/blob/remoteOpsLowLevel/tests/unit/test_UIARemoteOps/test_highLevel.py)
* In browse mode in Microsoft Word with UIA enabled, tested pressing h to jump to each heading in this document, also tested opening the Elements list and listing headings.

* Documentation could be improved.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

## Summary by CodeRabbit

- **Chores**
  - Removed the `microsoft-ui-uiautomation` submodule and its references from the project.
  
- **New Features**
  - Introduced functionality for creating, manipulating, and retrieving results from remote operations in UI Automation.
  - Added support for building and managing remote operations within NVDA.
  - Enhanced local execution of operations with new classes and functions.
  - Implemented a function for iterating over text units in a remote text range.
  
- **Documentation**
  - Updated readme files to reflect the removal of `microsoft-ui-uiautomation` and added guidelines for new remote operations features.


nvdaHelper/UIARemote/lowLevel.cpp
```diff
@@ -0,0 +1,191 @@
...
+	AutomationRemoteOperationResult results {nullptr};
+	winrt::copy_from_abi(results, arg_pResults);
+	if(!results) {
+		return E_INVALIDARG;
+	}
+	*pHasOperand = results.HasOperand(AutomationRemoteOperationOperandId{arg_registerID});
+	return S_OK;
+}
+
+extern "C" __declspec(dllexport) void __stdcall remoteOpResult_free(void* arg_pResults) {
```
dkager, line 123: Perhaps for consistency this function should also return `HRESULT` (always `S_OK`) like all the others in the API?

source/UIAHandler/_remoteOps/lowLevel.py
```diff
@@ -0,0 +1,376 @@
...
+			raise LookupError(f"Operand {operandId} not found in results")
+		return val.value
+
+	def __del__(self):
+		_dll.remoteOpResult_free(self._pResults)
+
+
+class RemoteOperation:
+	"""
+	Creats and wraps a Windows.UI.UIAutomation.Core.CoreAutomationRemoteOperation.
```
dkager, line 127: Small typo.

source/UIAHandler/_remoteOps/instructions.py
```diff
@@ -0,0 +1,758 @@
...
+
+
+@dataclass
+class Fork(_TypedInstruction):
+	opCode = lowLevel.InstructionType.Fork
+	jumpTo: lowLevel.RelativeOffset
+
+
+@dataclass
+class ForkIfFalse(_TypedInstruction):
```
dkager, line 41: Why is `ForkIfTrue` not implemented? Same for some other instructions. Would it be worth it to implement more instructions? (Good for new dev, or to spend a Sunday afternoon.)

source/UIAHandler/_remoteOps/readme.md
```diff
@@ -0,0 +1,483 @@
...
+textRange = ra.newTextRange(UIATextRange)
+textRange.move(TextUnit_Word, 1)
+```
+In this example, the values TextUnit_Word and 1 will be automatically remoted.
+
+### Basic remote types
+#### Equality checks
+All types support equality checks:
+```
+a = ra.newBool(True)
```
dkager, line 105: I think the variable `a` should be `i`.

source/UIAHandler/_remoteOps/readme.md
```diff
@@ -0,0 +1,483 @@
...
+
+### UI Automation text ranges
+#### Declaring a text range 
+To create a new remote text range, call `ra.newTextRange`, giving it an existing IUIAutomationTextRange comtypes pointer as its argument:
+```
+textRange = ra.newElement(UIATextRange)
+```
+Note that under the hood the text range is automatically cloned after it has been remoted, so that any manipulation of the remote text range (such as moving its ends) is not reflected in the original IUIAutomationTextRange you gave it.
+
+#### Retreaving text, comparison and manipulation
```
dkager, line 234: 'Retreaving' should be 'Retrieving'

source/UIAHandler/_remoteOps/readme.md
```diff
@@ -0,0 +1,483 @@
...
+tempRange.moveEndpointByRange(TextPatternRangeEndpoint_End, tempRange, TextPatternRangeEndpoint_Start)
+with ra.whileBlock(lambda: tempRange.move(TextUnit_word, 1) == 1):
+	with ra.ifblock(tempRange.compareEndpoints(textPatternRangeEndpoint_Start, textRange, TextPatternRangeEndpoint_End) >= 0):
+		ra.breakLoop()
+	wordCount += 1
+ra.Return(wordcount)
+```
+
+#### Text range logical adapter to improve logic and readability
+The verboseness of many of the compare and move text range methods can make it hard to quickly read the code and gain a good idea of what the algorithm is actually doing. It is also quite tricky to write an algorithm that is easily reversed. Therefore remote text ranges have a `getLogicalAdapter` method, taking a single boolean `reverse` argument which returns a special object which raps a remote text range, and provides friendly start and end properties, which take the reversal into account.
```
dkager, line 259: 'raps' should be 'wraps', although rapping a text range sounds like fun. :-)

ABuffEr:
CC @mltony, maybe this PR could interest you :)

Adriani90:
Wow what an amazing work, thank you very much Mick this is really great decision to make NVDA more independent.
I guess this will open up many possibilities to solve also other issues we have with UIA in MS Office products at least.

source/UIAHandler/_remoteOps/readme.md
```diff
@@ -0,0 +1,483 @@
...
+Again, in future the library could be extended to support this, but it would involve having to marshal out all items, and marshal them back in, appending them to the array one at a time, which would also be costly.
+
+#### Building iterable functions
+A commn use of remote operations is to walk a text range or element tree, and collect data which would be returned in an array. However, as arrays can not be marked as `static`, this would involve a lot of extra code to handle execution continuation after the instruciton limit is reached. Therefore the library supports a `Operation.buildIterableFunction` decorator, which can be used in place of `Operator.bildFunction`. Within a fuction that uses this decorator, rather than using ra.Return to return value and halt, instead you can use `ra.Yield` which will yield a value and continue to execute (until the instruction limit is reached of course). To actually execute an iterable function though, instead of using `Operation.execute`, you use `Operation.iterExecute` as the generator to a `for` loop, which will iterate over the yielded values.
```
dkager, line 378: Typo in `Operator.bildFunction`

source/UIAHandler/browseMode.py
```diff
@@ -160,6 +174,7 @@ def UIAHeadingQuicknavIterator(
 		direction: str = "next"
 ):
 	reverse = bool(direction == "previous")
+	wantedLevel = int(itemType[7:]) if len(itemType) > 7 else None
```
dkager, line 177: It would be a bit cleaner to use a constant instead of the index 7 (e.g. the length of the word 'heading').

source/UIAHandler/remote.py
```diff
@@ -4,36 +4,150 @@
...
-		raise RuntimeError("UIARemote not initialized")
-	customAttribValue = VARIANT()
-	if _dll.msWord_getCustomAttributeValue(docElement, textRange, customAttribID, byref(customAttribValue)):
-		return customAttribValue.value
-	return None
+	docElement: UIA.IUIAutomationElement,
+	textRange: UIA.IUIAutomationTextRange,
+	customAttribID: int
+) -> Optional[Any]:
+	guid_msWord_extendedTextRangePattern = GUID("{93514122-FF04-4B2C-A4AD-4AB04587C129}")
```
dkager, line 46: Likewise these could also should be constants in the final version.

source/UIAHandler/remote.py
```diff
@@ -4,36 +4,150 @@
...
+	for level, label, paragraphRange in op.iterExecute(maxTries=20):
+		yield level, label, paragraphRange
+
+
+def findFirstHeadingInTextRange(
+	textRange: UIA.IUIAutomationTextRange,
+	wantedLevel: int | None = None,
+	reverse: bool = False
+) -> tuple[int, str, UIA.IUIAutomationElement] | None:
+	op = operation.Operation(enableCompiletimeLogging=True)
```
dkager, line 120: Should compile-time logging be enabled for this once it's moved into a release build? Maybe this can be another logging category in the Advanced settings?

dkager:
So I've been reading and trying to understand this PR today. It must be the most awesome computer science project in years. Great incentive to further optimize the use of UIA in MS Office!

michaelDCurran:
ForkIfTrue has not been implemented as none of the higher-level APIs or 
types depend upon it currently. Thus, the unit tests as currently 
written would not exercise it either.

But of course if there is a clear need, then someone is welcome to add it.

michaelDCurran:
It has certainly been a very fun personal project that up to now has 
been outside of work time, as I was really not sure just how successful 
it would be.

lukaszgo1 (commented):
Note that I've just skimmed the code - I'll probably take another look later on, as this looks like a quite interesting piece of work
- Please remove mentions of the UIA Remote library from the developer documentation and include's readme
- Has it been considered to add the library documentation to the developer guide rather than into a separate read me?

source/UIAHandler/_remoteOps/builder.py
```diff
@@ -0,0 +1,299 @@
...
+	def getByteCode(self) -> bytes:
+		byteCode = struct.pack('l', self.opCode.value)
+		params = list(self.params.values())
+		if len(params) > 0 and isinstance(params[-1], list):
+			# If the last parameter is a list, it is a variable length parameter.
+			params[-1:] = params[-1]
+		for param in params:
+			if isinstance(param, enum.IntEnum):
+				param = c_long(param.value)
+			if isinstance(param, Operand):
```
dkager, line 126: Should this be an elif, or can `param` be both an enum and an Operand?

michaelDCurran:
> Has it been considered to add the library documentation to the developer guide rather than into a separate read me?
@lukaszgo1  Perhaps eventually, but as this is so new, I would like to keep it contained to just UIAHandler, and keep the `_remoteOps` package private, until it has been road tested a bit internally by NVDA.
it can definitely provide extraordinary speedups if used appropriately, but may actually be much slower for very small scenarios with only 2 or 3 cross-process calls without using the feature.

codeofdusk:
Wow, haven't looked at the code yet but this is fantastic!

Another good test candidate for this might be #13517.

dkager:
I did some testing in a document that was notoriously slow. I navigated from one h1 to the next h1. They are about four pages apart.
* Without this PR (release build 2023.3.3): 592 ms
* With this PR (running from source): 157 ms

Interestingly, the elements list (headings selected) did not show significant speed improvements. In any case, this should greatly speed up quicknav. I think it could also be very effective for navigating hyperlinks in Word.

nvdaHelper/UIARemote/lowLevel.cpp
aursulis, line 1: Hey @michaelDCurran, this is Antanas from HP.

A question about the wrappers in this lowLevel.cpp file: are they here because C++/WinRT is convenient? I am asking because Windows Runtime, at its deepest level, is just plain old COM with a few bells and whistles, so given the correct definitions for Python's comtypes, you wouldn't need to jump through a C++ DLL here. It would probably not be very elegant Python code because the C++/WinRT projection does a fair amount of heavy lifting, but it should be possible.

source/UIAHandler/_remoteOps/remoteFuncWrapper.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+_remoteFunc_self = TypeVar('_remoteFunc_self', bound=builder._RemoteBase)
+_remoteFunc_paramSpec = ParamSpec('_remoteFunc_paramSpec')
+_remoteFunc_return = TypeVar('_remoteFunc_return')
+
+
+class _BaseRemoteFuncWrapper:
+
+	def generateArgsKwargsString(self, *args, **kwargs):
```
seanbudd, line 28:
```suggestion
	def generateArgsKwargsString(self, *args, **kwargs) -> str:
```

source/UIAHandler/remote.py
```diff
@@ -1,39 +1,150 @@
...
+	StyleId
+)
 
 
-_dll = None
-
-
-def initialize(doRemote: bool, UIAClient: POINTER(UIA.IUIAutomation)):
+def initialize(doRemote: bool, UIAClient: UIA.IUIAutomation):
 	"""
```
seanbudd, line 30: please update the docstring to note the parameters are ignored now

source/UIAHandler/remote.py
```diff
@@ -1,39 +1,150 @@
...
+		textRange: UIA.IUIAutomationTextRange,
 		customAttribID: int
-) -> Optional[Union[int, str]]:
-	if _dll is None:
-		raise RuntimeError("UIARemote not initialized")
-	customAttribValue = VARIANT()
-	if _dll.msWord_getCustomAttributeValue(docElement, textRange, customAttribID, byref(customAttribValue)):
-		return customAttribValue.value
-	return None
+) -> Optional[Any]:
```
seanbudd, line 42: is this an API breaking change (change of return type)

source/UIAHandler/remote.py
```diff
@@ -1,39 +1,150 @@
...
+			ra.logRuntimeMessage("doing callExtension for extendedTextRangePattern")
+			remoteDocElement.callExtension(
+				guid_msWord_extendedTextRangePattern,
+				remoteResult
+			)
+			with ra.ifBlock(remoteResult.isNull()):
+				ra.logRuntimeMessage("extendedTextRangePattern is null")
+				ra.halt()
+			with ra.elseBlock():
+				ra.logRuntimeMessage("got extendedTextRangePattern ")
```
seanbudd, line 64:
```suggestion
				ra.logRuntimeMessage("got extendedTextRangePattern")
```

tests/unit/test_UIARemoteOps/test_highLevel.py
```diff
@@ -0,0 +1,700 @@
...
+from ctypes import POINTER
+from UIAHandler import UIA
+from UIAHandler._remoteOps import operation
+from UIAHandler._remoteOps import remoteAPI
+from UIAHandler._remoteOps.lowLevel import (
+	PropertyId,
+)
+
+
+class TestHighLevel(TestCase):
```
seanbudd, line 21: I think it would be valuable to split this suite up into operands e.g. test bool, test if, etc

seanbudd (commented):
I've reviewed everything except the following files:
- source/UIAHandler/_remoteOps/instructions.py
- source/UIAHandler/_remoteOps/remoteAPI.py
- tests/unit/test_UIARemoteOps/test_highLevel.py

source/UIAHandler/_remoteOps/readme.md
```diff
@@ -0,0 +1,504 @@
+# UI Automation Remote Operations for NVDA
+
+## Introduction
+If you have code that makes more than a few UI Automation calls all in a row, you may want to consider rewriting the code as a Remote Operation. This will allow Windows to execute the code all in one cross-process call, thereby significantly speeding up execution.
```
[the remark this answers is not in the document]
seanbudd: for the rest of the file too please

source/UIAHandler/_remoteOps/readme.md
```diff
@@ -0,0 +1,504 @@
...
+
+Following is a simple example of a remote operation.
+### Example 1: Fetching the names of all ancestors of a UI Automation element
+```
```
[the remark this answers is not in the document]
XLTechie: @seanbudd are you sure it should be "py"? That is not a recognized language according to [GitHub's list](https://raw.githubusercontent.com/github-linguist/linguist/master/lib/linguist/languages.yml), unless I missed it.
The convention is usually
````
```python
````

````suggestion
```python
````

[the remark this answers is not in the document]
Comment (seanbudd) [reply source/UIAHandler/_remoteOps/readme.md:8]: you can use any of the file extensions as well as the full name - see this stackoverflow answer for an explanation of using GitHub linguist's yml

https://stackoverflow.com/a/45786100

File: nvdaHelper/UIARemote/lowLevel.cpp
Comment (michaelDCurran) [reply nvdaHelper/UIARemote/lowLevel.cpp:1]: I agree it would be possible to replace the c++/winrt code here with pure Python code, but I don't really know what advantage that would provide. I personally would not be motivated to do this.
You'd also have to write out all the COM interfaces by hand which would take a while and just adds extra code that could introduce bugs.

File: source/UIAHandler/remote.py (hunk 14 above)
Comment (michaelDCurran) [reply source/UIAHandler/remote.py:42]: The function will still return either an int or a string or None. In deed union of just int or string (not None) was originally incorrect.
Also, as execute is marked as always returning Any, then it is a bit tricky to really enforce  the right return type. I could cast the result of execute, but is it really worth it?

File: source/UIAHandler/remote.py (hunk 9 above)
Comment (michaelDCurran) [reply source/UIAHandler/remote.py:43]: I would argue these are already constants. They are not used anywhere else, and already have descriptive names.

**Comment by seanbudd** [comment]
@coderabbitai review

Testing code rabbit on this large PR

**Comment by seanbudd** [comment]
@coderabbitai resolve

File: tests/unit/test_UIARemoteOps/test_highLevel/test_iterable.py [hunk 19]
```diff
@@ -0,0 +1,88 @@
...
+
+from unittest import TestCase
+from unittest.mock import Mock
+from ctypes import POINTER
+from UIAHandler import UIA
+from UIAHandler._remoteOps import operation
+from UIAHandler._remoteOps import remoteAPI
+from UIAHandler._remoteOps.lowLevel import (
+	PropertyId,
+)
```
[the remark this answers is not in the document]
michaelDCurran: Looks like this suggestion causes a syntax error as the previous line should also be removed: `from UIAHandler._remoteOps.lowLevel import (`

michaelDCurran:
@coderabbitai review

seanbudd:
Note this might also close out https://github.com/nvaccess/nvda/issues/13995

michaelDCurran:
@coderabbitai resolve

source/UIAHandler/remote.py
seanbudd: I don't think that's needed, particularly if the original documentation was wrong

source/UIAHandler/_remoteOps/instructions/general.py
```diff
@@ -0,0 +1,65 @@
...
+
+from __future__ import annotations
+from dataclasses import dataclass
+from .. import lowLevel
+from .. import builder
+from ._base import _TypedInstruction
+
+
+@dataclass
+class Set(_TypedInstruction):
```
seanbudd, line 20: perhaps this should be renamed Assign as it this could be confused with a `set(), {1, 2, 3}`

source/UIAHandler/_remoteOps/instructions/general.py
```diff
@@ -0,0 +1,65 @@
...
+	opCode = lowLevel.InstructionType.Compare
+	result: builder.Operand
+	left: builder.Operand
+	right: builder.Operand
+	comparisonType: lowLevel.ComparisonType
+
+	def localExecute(self, registers: dict[lowLevel.OperandId, object]):
+		localLeft = registers[self.left.operandId]
+		localRight = registers[self.right.operandId]
+		if self.comparisonType == lowLevel.ComparisonType.Equal:
```
seanbudd, line 41: this should use match/case syntax

source/UIAHandler/_remoteOps/instructions/null.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@dataclass
+class IsNull(_TypedInstruction):
+	opCode = lowLevel.InstructionType.IsNull
+	result: builder.Operand
+	target: builder.Operand
+
+	def localExecute(self, registers: dict[lowLevel.OperandId, object]):
+		registers[self.result.operandId] = (registers[self.target.operandId] is None)
```
seanbudd, line 35:
```suggestion
		registers[self.result.operandId] = registers[self.target.operandId] is None
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+	RemoteTextRange,
+)
+
+
+class RemoteAPI(builder._RemoteBase):
+	_op: operation.Operation
+	_rob: builder.RemoteOperationBuilder
+	_logObj: RemoteString | None = None
+
+	def __init__(self, op: operation.Operation, enableRemoteLogging=False):
```
seanbudd, line 49:
```suggestion
	def __init__(self, op: operation.Operation, enableRemoteLogging: bool = False):
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+		self.addCompiletimeComment(f"Yielding {remoteValue}")
+		self._op._yieldListOperand.append(remoteValue)
+
+	_newObject_RemoteType = TypeVar('_newObject_RemoteType', bound=RemoteBaseObject)
+
+	def _newObject(
+			self,
+			RemoteType: Type[_newObject_RemoteType],
+			value: Any,
+			static=False
```
seanbudd, line 98:
```suggestion
			static: bool = False
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+	def newFloat(self, value: float = 0.0, static=False) -> RemoteFloat:
+		return self._newObject(RemoteFloat, value, static=static)
+
+	def newString(self, value: str = "", static=False) -> RemoteString:
+		return self._newObject(RemoteString, value, static=static)
+
+	def newBool(self, value: bool = False, static=False) -> RemoteBool:
+		return self._newObject(RemoteBool, value, static=static)
+
+	def newGuid(self, value: GUID | str | None = None, static=False) -> RemoteGuid:
```
seanbudd, line 122:
```suggestion
	def newUint(self, value: int = 0, static: bool = False) -> RemoteUint:
		return self._newObject(RemoteUint, value, static=static)

	def newInt(self, value: int = 0, static: bool = False) -> RemoteInt:
		return self._newObject(RemoteInt, value, static=static)

	def newFloat(self, value: float = 0.0, static: bool = False) -> RemoteFloat:
		return self._newObject(RemoteFloat, value, static=static)

	def newString(self, value: str = "", static: bool = False) -> RemoteString:
		return self._newObject(RemoteString, value, static=static)

	def newBool(self, value: bool = False, static: bool = False) -> RemoteBool:
		return self._newObject(RemoteBool, value, static=static)

	def newGuid(self, value: GUID | str | None = None, static: bool = False) -> RemoteGuid:
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+			realValue = value
+		return self._newObject(RemoteGuid, realValue, static=static)
+
+	def newVariant(self) -> RemoteVariant:
+		return RemoteVariant.createNew(self.rob)
+
+	def newArray(self) -> RemoteArray:
+		return RemoteArray.createNew(self.rob)
+
+	def newElement(self, value: UIA.IUIAutomationElement | None = None, static=False) -> RemoteElement:
```
seanbudd, line 137:
```suggestion
	def newElement(
		self,
		value: UIA.IUIAutomationElement | None = None,
		static: bool = False
	) -> RemoteElement:
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+		with self.rob.overrideDefaultSection(section):
+			if value is not None:
+				obj = self._op.importElement(value)
+				if static:
+					self._op._registerStaticOperand(obj)
+				return obj
+			else:
+				return self._newObject(RemoteElement, value, static=static)
+
+	def newTextRange(self, value: UIA.IUIAutomationTextRange | None = None, static=False) -> RemoteTextRange:
```
seanbudd, line 148:
```suggestion
	def newTextRange(
		self,
		value: UIA.IUIAutomationTextRange | None = None,
		static: bool = False
	) -> RemoteTextRange:
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+		instructionList.addInstruction(
+			instructions.SetOperationStatus(
+				status=RemoteInt.ensureRemote(self.rob, status)
+			)
+		)
+
+	_scopeInstructionJustExited: instructions.InstructionBase | None = None
+
+	@contextlib.contextmanager
+	def ifBlock(self, condition: RemoteBool, silent=False):
```
seanbudd, line 181:
```suggestion
	def ifBlock(self, condition: RemoteBool, silent: bool = False):
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+			instructionList.addComment("If block body")
+		yield
+		if not silent:
+			instructionList.addComment("End of if block body")
+		nextInstructionIndex = instructionList.getInstructionCount()
+		conditionInstruction.branch = RelativeOffset(nextInstructionIndex - conditionInstructionIndex)
+		self._scopeInstructionJustExited = conditionInstruction
+
+	@contextlib.contextmanager
+	def elseBlock(self, silent=False):
```
seanbudd, line 198:
```suggestion
	def elseBlock(self, silent: bool = False):
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+	def continueLoop(self):
+		instructionList = self.rob.getDefaultInstructionList()
+		instructionList.addInstruction(instructions.ContinueLoop())
+
+	def breakLoop(self):
+		instructionList = self.rob.getDefaultInstructionList()
+		instructionList.addInstruction(instructions.BreakLoop())
+
+	@contextlib.contextmanager
+	def whileBlock(self, conditionBuilderFunc: Callable[[], RemoteBool], silent=False):
```
seanbudd, line 230:
```suggestion
	def whileBlock(self, conditionBuilderFunc: Callable[[], RemoteBool], silent: bool = False):
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+	@remoteContextManager
+	def forEachItemInArray(
+			self,
+			array: RemoteArray
+	) -> Generator[RemoteVariant, None, None]:
+		with self.forEachNumInRange(0, array.size()) as index:
+			yield array[index]
+
+	@contextlib.contextmanager
+	def tryBlock(self, silent=False):
```
seanbudd, line 289:
```suggestion
	def tryBlock(self, silent: bool = False):
```

source/UIAHandler/_remoteOps/remoteAPI.py
```diff
@@ -0,0 +1,371 @@
...
+		if not silent:
+			instructionList.addComment("End of try block body")
+		instructionList.addInstruction(instructions.EndTryBlock())
+		# update the try block catch offset to jump to the end of the try block body
+		nextInstructionIndex = instructionList.getInstructionCount()
+		tryBlockInstruction.catchBranch = RelativeOffset(nextInstructionIndex - tryBlockInstructionIndex)
+		self._scopeInstructionJustExited = tryBlockInstruction
+
+	@contextlib.contextmanager
+	def catchBlock(self, silent=False):
```
seanbudd, line 309:
```suggestion
	def catchBlock(self, silent: bool = False):
```

source/UIAHandler/_remoteOps/remoteTypes/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+
+	def _generateDefaultInitialValue(self) -> LocalTypeVar:
+		if self.LocalType is None:
+			raise TypeError("LocalType not set")
+		return self.LocalType()
+
+	def _generateInitInstructions(self) -> Iterable[instructions.InstructionBase]:
+		raise NotImplementedError()
+
+	def _initOperand(self, initialValue: LocalTypeVar | None = None, const=False):
```
seanbudd, line 71:
```suggestion
	def _initOperand(self, initialValue: LocalTypeVar | None = None, const: bool =False):
```

source/UIAHandler/_remoteOps/remoteTypes/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+		return remoteObj
+
+	@property
+	def initialValue(self) -> LocalTypeVar:
+		if self._initialValue is not None:
+			return self._initialValue
+		return self._generateDefaultInitialValue()
+
+	@property
+	def isLocalValueAvailable(self):
```
seanbudd, line 141:
```suggestion
	def isLocalValueAvailable(self) -> bool:
```

source/UIAHandler/_remoteOps/remoteTypes/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+
+	@property
+	def localValue(self) -> LocalTypeVar:
+		if self._executionResult is None:
+			raise RuntimeError("Operation not executed")
+		value = self._executionResult.getOperand(self.operandId)
+		return cast(LocalTypeVar, value)
+
+	@remoteMethod_mutable
+	def set(self, other: Self | LocalTypeVar):
```
seanbudd, line 152: consider renaming this to assign

source/UIAHandler/_remoteOps/instructions/arithmetic.py
```diff
@@ -0,0 +1,110 @@
...
+
+	def localExecute(self, registers: dict[lowLevel.OperandId, object]):
+		registers[self.target.operandId] -= registers[self.value.operandId]
+
+
+@dataclass
+class InplaceMultiply(_TypedInstruction):
+	opCode = lowLevel.InstructionType.Multiply
+	target: builder.Operand
+	value: builder.Operand
```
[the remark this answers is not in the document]
seanbudd: or would it be

```suggestion
	target: builder.Operand[RemoteInt | RemoteFloat]
	value: builder.Operand[RemoteInt | RemoteFloat]
```

source/UIAHandler/_remoteOps/remoteTypes/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+
+
+class RemoteString(RemoteBaseObject[str]):
+	_IsTypeInstruction = instructions.IsString
+	LocalType = str
+	_defaultInitialValue = ""
+
+	def _generateInitInstructions(self) -> Iterable[instructions.InstructionBase]:
+		initialValue = self.initialValue
+		stringLen = (len(initialValue) + 1)
```
seanbudd, line 557:
```suggestion
		stringLen = len(initialValue) + 1
```

source/UIAHandler/_remoteOps/remoteTypes/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+			instructions.StringConcat(
+				result=self,
+				left=self,
+				right=type(self).ensureRemote(self.rob, other)
+			)
+		)
+		return self
+
+	@remoteMethod_mutable
+	def set(self, other: Self | str):
```
seanbudd, line 596: assign?

source/UIAHandler/_remoteOps/remoteTypes/element.py
```diff
@@ -0,0 +1,95 @@
...
+class RemoteElement(RemoteExtensionTarget[POINTER(UIA.IUIAutomationElement)]):
+	"""
+	Represents a remote UI Automation element.
+	Allows for navigation and property retrieval.
+	"""
+
+	_IsTypeInstruction = instructions.IsElement
+	LocalType = POINTER(UIA.IUIAutomationElement)
+
+	def _initOperand(self, initialValue: None = None, const=False):
```
seanbudd, line 37:
```suggestion
	def _initOperand(self, initialValue: None = None, const: bool = False):
```

source/UIAHandler/_remoteOps/remoteTypes/intEnum.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+_RemoteIntEnum_LocalTypeVar = TypeVar('_RemoteIntEnum_LocalTypeVar', bound=enum.IntEnum)
+
+
+class RemoteIntEnum(RemoteInt, Generic[_RemoteIntEnum_LocalTypeVar]):
+	localType = enum.IntEnum
+	_enumType: _RemoteIntEnum_LocalTypeVar
+
+	def _initOperand(self, initialValue: _RemoteIntEnum_LocalTypeVar, const=False):
```
seanbudd, line 77:
```suggestion
	def _initOperand(self, initialValue: _RemoteIntEnum_LocalTypeVar, const: bool = False):
```

source/UIAHandler/_remoteOps/remoteTypes/textRange.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+class RemoteTextRange(RemoteExtensionTarget[POINTER(UIA.IUIAutomationTextRange)]):
+	"""
+	Represents a remote UI Automation text range.
+	"""
+
+	LocalType = POINTER(UIA.IUIAutomationTextRange)
+
+	def _initOperand(self, initialValue: None = None, const=False):
```
seanbudd, line 40:
```suggestion
	def _initOperand(self, initialValue: None = None, const: bool = False):
```

source/UIAHandler/_remoteOps/remoteTypes/textRange.py
```diff
@@ -0,0 +1,271 @@
...
+			)
+		)
+
+	@remoteMethod_mutable
+	def moveEndpointByUnit(
+			self,
+			endpoint: RemoteIntEnum[lowLevel.TextPatternRangeEndpoint] | lowLevel.TextPatternRangeEndpoint,
+			unit: RemoteIntEnum[lowLevel.TextUnit] | lowLevel.TextUnit,
+			count: RemoteInt | int
+	):
```
seanbudd, line 101: please add a return type

source/UIAHandler/_remoteOps/remoteTypes/textRange.py
```diff
@@ -0,0 +1,271 @@
...
+				result=result,
+				target=self,
+				thisEndpoint=RemoteIntEnum.ensureRemote(self.rob, thisEndpoint),
+				otherRange=otherRange,
+				otherEndpoint=RemoteIntEnum.ensureRemote(self.rob, otherEndpoint)
+			)
+		)
+		return result
+
+	def getLogicalAdapter(self, reverse=False) -> RemoteTextRangeLogicalAdapter:
```
seanbudd, line 164:
```suggestion
	def getLogicalAdapter(self, reverse: bool = False) -> RemoteTextRangeLogicalAdapter:
```

source/UIAHandler/_remoteOps/remoteTypes/textRange.py
```diff
@@ -0,0 +1,271 @@
...
+		obj = RemoteTextRangeLogicalAdapter(self.rob, self, reverse=reverse)
+		return obj
+
+
+class _RemoteTextRangeEndpoint(builder._RemoteBase):
+
+	def __init__(
+			self,
+			rob: builder.RemoteOperationBuilder,
+			textRangeLA: RemoteTextRangeLogicalAdapter, isStart: bool
```
seanbudd, line 174:
```suggestion
			textRangeLA: RemoteTextRangeLogicalAdapter,
			isStart: bool
```

source/UIAHandler/_remoteOps/remoteTypes/textRange.py
```diff
@@ -0,0 +1,271 @@
...
+	def __ne__(self, other: _RemoteTextRangeEndpoint) -> RemoteBool:
+		return (self == other).inverse()
+
+
+class RemoteTextRangeLogicalAdapter(builder._RemoteBase):
+
+	def __init__(
+			self,
+			rob: builder.RemoteOperationBuilder,
+			textRange: RemoteTextRange, reverse=False
```
seanbudd, line 238:
```suggestion
			textRange: RemoteTextRange,
			reverse: bool = False
```

source/UIAHandler/_remoteOps/remoteTypes/textRange.py
```diff
@@ -0,0 +1,271 @@
...
+	@property
+	def end(self) -> _RemoteTextRangeEndpoint:
+		obj = _RemoteTextRangeEndpoint(self.rob, self, isStart=False)
+		return obj
+
+	@end.setter
+	def end(self, value: _RemoteTextRangeEndpoint):
+		self.end.moveTo(value)
+
+	def clone(self):
```
seanbudd, line 270: please add a return type

seanbudd (commented):
I only have the unit tests left to review now

tests/unit/test_UIARemoteOps/test_highLevel/test_bool.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+
+from unittest import TestCase
+from UIAHandler._remoteOps import operation
+from UIAHandler._remoteOps import remoteAPI
+
+
+class Test_bool(TestCase):
+
+	def test_bool_false(self):
```
seanbudd, line 17: Since "Test_bool" is in the test suite name these could probably be simplified. Same in other files

```suggestion
	def test_false(self):
```

tests/unit/test_UIARemoteOps/test_highLevel/test_bool.py
```diff
@@ -0,0 +1,91 @@
+# A part of NonVisual Desktop Access (NVDA)
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+# Copyright (C) 2023 NV Access Limited
```
seanbudd, line 4: Same in other files

```suggestion
# Copyright (C) 2024 NV Access Limited
```

source/UIAHandler/_remoteOps/readme.md
```diff
@@ -0,0 +1,556 @@
...
+	# Increment the counter by 1.
+	counter += 1
+# Return the words array.
+ra.Return(words)
+```
+
+By simply changing the False to True, the algorithm is automatically reversed, as the start and end properties reverse, and the `numUnits` argument on the methods have their sign flipped.
+As shown above, `start` and `end` properties can be assigned to which moves the endpoint, and they can be moved by a unit with `moveByUnit`.
+They can also be compared with `<`, `<=`, `==`, `>=`, and `>`.
+If you still want the comparison delta (like with `compareEndpoints` the properties also have a `compareWith` method which takes another endpoint and gives back a number less than 0, equal to 0 or greater than 0.
```
XLTechie, line 320: Missing punctuation.
```suggestion
If you still want the comparison delta (like with `compareEndpoints`), the properties also have a `compareWith` method, which takes another endpoint and gives back a number less than 0, equal to 0 or greater than 0.
```

tests/unit/test_UIARemoteOps/test_highLevel/test_bool.py
```diff
@@ -0,0 +1,91 @@
...
+
+"""
+High-level UIA remote ops Unit tests for setting and comparing booleans.
```
seanbudd, line 7:
```suggestion
High-level UIA remote ops unit tests for setting and comparing booleans.
```

same in other files

tests/unit/test_UIARemoteOps/test_highLevel/test_errorHandling.py
```diff
@@ -0,0 +1,60 @@
...
+
+		@op.buildFunction
+		def code(ra: remoteAPI.RemoteAPI):
+			i = ra.newInt(3)
+			was_in_catch = ra.newBool(False)
+			with ra.tryBlock():
+				i / 0
+			with ra.catchBlock():
+				was_in_catch.set(True)
+			ra.Return(was_in_catch)
```
seanbudd, line 57: could you return `i` and confirm it is unchanged and equal to 3

tests/unit/test_UIARemoteOps/test_highLevel/test_instructionLimit.py
```diff
@@ -0,0 +1,44 @@
...
+			executionCount = ra.newInt(0, static=True)
+			executionCount += 1
+			i = ra.newInt(0, static=True)
+			with ra.whileBlock(lambda: i < 20000):
+				i += 1
+			ra.Return(executionCount, i)
+
+		executionCount, i = op.execute(maxTries=20)
+		self.assertEqual(i, 20000)
+		self.assertEqual(executionCount, 9)
```
seanbudd, line 44: could you explain this test case more in comments? I'm not sure I understand the results here

tests/unit/test_UIARemoteOps/test_highLevel/test_iterable.py
```diff
@@ -0,0 +1,82 @@
...
+			with ra.whileBlock(lambda: i < 4):
+				ra.Yield(i)
+				i += 1
+
+		results = []
+		for i in op.iterExecute():
+			results.append(i)
+		self.assertEqual(results, [0, 1, 2, 3])
+
+	def test_long_iterableFunction(self):
```
seanbudd, line 32: could you explain this test case more in comments? I'm not sure I understand the results here

seanbudd (commented):
Tests generally look good

source/UIAHandler/browseMode.py
```diff
@@ -170,6 +186,19 @@ def UIAHeadingQuicknavIterator(
 			searchArea.start = entireDocument.start
 		else:
 			searchArea.end = entireDocument.end
+	if winVersion.getWinVer() >= winVersion.WIN11:
```
LeonarddeR, line 189: This winVersion check has to be replicated for every case where UIA Remote is to be used. Rather than having an explicit winVersion check, could you consider having a function or variable on UIAHandler.remote, e.g. `UIAHandler.remote.isSupported`? This also makes it easier to test the pre remote code by changing that variable to False temporarily.

source/UIAHandler/_remoteOps/instructions/__init__.py
```diff
@@ -0,0 +1,111 @@
...
+"""
+This package contains all the instructions that can be executed by the remote ops framework.
+Each instruction contains the appropriate op code and parameter types.
+Most instructions also contain a `localExecute` method,
+which provides an implementation of the instruction that can be executed locally.
+"""
+
+
+# Import all instructions so that they can be accessed as attributes of this module.
+# flake8: noqa: F401
```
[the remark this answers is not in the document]
michaelDCurran: I'm not a fan of this personally as it means duplicating all the names which adds further maintenance  and risk of something being forgotten. If the over all goal was to specifically limit the packages `*` symbols then I'd  support `__all__` but using it purely to stop a linting check, I'd prefer the explicit noqa comment myself.

source/UIAHandler/_remoteOps/instructions/arithmetic.py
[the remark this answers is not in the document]
michaelDCurran: I think it could actually be
```suggestion
	target: RemoteNumber
	value: RemoteNumber
```
Or yes, in theory we could use a TypeVar limited to RemoteInt, RemoteUint or RemoteFlaot, and use the Typevar for both, thus then the target and value would have to also be the same type. However, I think that Dataclass may actually use type annotations internally at runtime for docstring generation or validation, thus I don't think it would support using generics.  I could play with it, but at a minimum I could use RemoteNumber.  
I had avoided using higher-level remote types in these instructions as I wanted to keep them lower level and more true to how they are specified in the remote operations documentation. I.e. they are just operands. But I could try using remote types for all of them.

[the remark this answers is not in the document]
seanbudd: More specific typing is nice if possible, but I'd consider this a nice to have

source/UIAHandler/_remoteOps/instructions/controlFlow.py
```diff
@@ -0,0 +1,67 @@
...
+
+@dataclass
+class NewLoopBlock(_TypedInstruction):
+	opCode = lowLevel.InstructionType.NewLoopBlock
+	breakBranch: lowLevel.RelativeOffset
+	continueBranch: lowLevel.RelativeOffset
+
+
+@dataclass
+class EndLoopBlock(_TypedInstruction):
```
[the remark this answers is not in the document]
michaelDCurran: This reflects Microsofts naming of the low-level instructions. I'd prefer to keep them matching.

source/UIAHandler/_remoteOps/instructions/general.py
michaelDCurran: Again, this reflects Microsoft's low-level instruction name though I see your point. However, these instructions are only to be used internally by the _remoteOps package, and not by code that uses remote operations through out NVDA.

source/UIAHandler/remote.py
```diff
@@ -26,20 +27,32 @@
...
+
 def initialize(doRemote: bool, UIAClient: UIA.IUIAutomation):
 	"""
 	Initializes UI Automation remote operations.
-	This function is deprecated and will be removed in the future.
-	The following parameters are ignored:
+	The following parameters are deprecated and ignored:
 	@param doRemote: true if code should be executed remotely, or false for locally.
 	@param UIAClient: the current instance of the UI Automation client library running in NVDA.
 	"""
```
seanbudd, line 46: I think it would be worth logging a warning if the parameters are set

tests/unit/test_UIARemoteOps/test_highLevel/test_iterable.py
```diff
@@ -66,6 +74,13 @@ def code(ra: remoteAPI.RemoteAPI):
 		self.assertEqual(results, [10, 11, 12, 13, 14])
 
 	def test_forEachItemInArray(self):
+		"""
+		Tests that a long operation is automatically re-executed multiple times
+		until it is successfully run without exceeding the instruction limit.
+		Using a static variable which is not reset between executions,
+		and then incremented once each execution,
+		We can track how many times the operation was executed.
```
seanbudd, line 82:
```suggestion
		we can track how many times the operation was executed.
```

tests/unit/test_UIARemoteOps/test_highLevel/test_iterable.py
```diff
@@ -30,27 +30,35 @@ def code(ra: remoteAPI.RemoteAPI):
 		self.assertEqual(results, [0, 1, 2, 3])
 
 	def test_long_iterableFunction(self):
+		"""
+		Tests that a long operation is automatically re-executed multiple times
+		until it is successfully run without exceeding the instruction limit.
+		Using a static variable which is not reset between executions,
+		and then incremented once each execution,
+		We can track how many times the operation was executed.
+		The operation itself  involves incrementing a counter variable 'i'
```
seanbudd, line 39:
```suggestion
		we can track how many times the operation was executed.
		The operation itself involves incrementing a counter variable 'i'
```

tests/unit/test_UIARemoteOps/test_highLevel/test_instructionLimit.py
```diff
@@ -28,6 +28,16 @@ def code(ra: remoteAPI.RemoteAPI):
 			op.execute()
 
 	def test_instructionLimitExceeded_with_static(self):
+		"""
+		Tests that a long operation is automatically re-executed multiple times
+		until it is successfully run without exceeding the instruction limit.
+		Using a static variable which is not reset between executions,
+		and then incremented once each execution,
+		We can track how many times the operation was executed.
+		The operation itself  involves incrementing a counter variable 'i'
```
seanbudd, line 37:
```suggestion
		we can track how many times the operation was executed.
		The operation itself involves incrementing a counter variable 'i'
```

source/UIAHandler/_remoteOps/instructions/_base.py
```diff
@@ -0,0 +1,17 @@
...
+from __future__ import annotations
+from ..builder import (
+	InstructionBase,
+)
+
+
+class _TypedInstruction(InstructionBase):
+
+	@property
+	def params(self):
```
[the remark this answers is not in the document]
seanbudd: can you add a return type here

source/UIAHandler/_remoteOps/instructions/arithmetic.py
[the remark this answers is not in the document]
seanbudd: Is it possible to add better typing instead of `builder.Operand` to instructions e.g. for target, value, etc?

source/UIAHandler/remote.py
michaelDCurran: At the moment, they are always set, when called by `UIAHandler`. They were never optional arguments.

source/UIAHandler/_remoteOps/instructions/arithmetic.py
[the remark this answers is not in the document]
michaelDCurran: Unfortunately it causes circular imports to try and import and use remoteTypes such as `remoteTypes.RemoteString` on instructions such as `instructions.string.IsString` because RemoteString would be used on an annotation of the IsString instruction class at class creation time, and  the string instruction is referenced on RemoteString at class creation time. 
There may be a nice way to fix this - I agree it would be really nice if we could. But I fear I may need to refactor things quite a bit to make that work. Thus I would prefer not to block this pr on that as things are functioning well currently, and perhaps do this as future work.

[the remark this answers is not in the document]
LeonarddeR: Have you tried/considered using `typing.TYPE_CHECKING` to only import the types when checking, or is that not possible in this case?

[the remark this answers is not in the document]
michaelDCurran: Because these classes are DataClasses, I think that the type annotations are required at runtime.

seanbudd (approved):

cary-rowen:
cc @mltony 
This PR is merged, will it have a positive impact on style navigation in MSWord?

dkager:
It's great to see this coming to master, thank you very much!
Since on older versions of Windows the remote ops will not be available, am I correct to assume that the DLL will still be required? If so, it would be nice to also work on #16217. If not, that issue can be closed.

mltony:
@cary-rowen - that's actually not up to me - the problem is that textInfo implementation is too slow in MSWord in current implementation. If this PR would allow faster textInfo then yes. But I don't understand low level UIA enough to tell whetehr this will allow to speed up TextInfo enough.

michaelDCurran:
this pr does not speedup any textInfos yet. However it does lay down 
some of the groundwork to do this in the future.

seanbudd:
The docs here are pretty thorough, do you think anything is missing? 
https://github.com/nvaccess/nvda/blob/master/source/UIAHandler/_remoteOps/readme.md

That wiki page is pretty out of date, we have goals of moving anything relevant into projectDocs/dev/ like the current up-to-date technical design overview

Adriani90:
I think it is not obvious for a new dev where to look for the thorough documentation for it, so I think it would need at least some introductory notes with relevant links in the internal design overview or in the developers guide or so. Von meinem iPhone gesendetAm 26.06.2024 um 02:32 schrieb Sean Budd ***@***.***>:﻿
The docs here are pretty thorough, do you think anything is missing?
https://github.com/nvaccess/nvda/blob/master/source/UIAHandler/_remoteOps/readme.md
That wiki page is pretty out of date, we have goals of moving anything relevant into projectDocs/dev/ like the current up-to-date technical design overview

seanbudd:
Would you open an issue to track any remaining work you think there is?

PR #16223, opened 2024-02-25
A gui.nvdaControls.MessageDialog with default type of standard, should no longer throw a None conversion exception

XLTechie:
### Link to issue number:

None

### Summary of the issue:

Using `gui.nvdaControls.MessageDialog` with standard type (the default), generates this exception:

```
ERROR - unhandled exception (03:21:22.395) - MainThread (16784):
  File "gui\nvdaControls.pyc", line 335, in _onShowEvt
  File "gui\nvdaControls.pyc", line 294, in _playSound
TypeError: 'NoneType' object cannot be interpreted as an integer
```

This is happening because there is no sound assigned for standard type dialogs, but the method which plays sound does not check for this. Which, ironically, caused a sound, at least in test builds.

Beta/alpha users will no longer hear an error sound when one of these is used.

Check whether the int is actually a None, before trying to play the sound. If so, don't play it.

### Testing strategy:

TBD

none

### Code Review Checklist:
- [X] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/nvdaControls.py
+++ b/source/gui/nvdaControls.py
@@ -1,6 +1,6 @@
 # -*- coding: UTF-8 -*-
 # A part of NonVisual Desktop Access (NVDA)
-# Copyright (C) 2016-2022 NV Access Limited, Derek Riemer, Cyrille Bougot
+# Copyright (C) 2016-2024 NV Access Limited, Derek Riemer, Cyrille Bougot, Luke Davis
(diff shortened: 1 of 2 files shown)


seanbudd (approved):
Thanks

PR #16239, opened 2024-02-29
removed unusable shortcuts from the (Synthesizer) and (Braille display) groupings

HanaDadfar:
### Link to issue number:
non found
### Summary of the issue:
in the nvda settings dialog, under Speech and Braille Categories, we have Synthesizer, and Braille display  groupings, which these groupings had shortcuts. (Synthesizer alt+s), and (Braille display alt+d). But, when we were navigating these groupings with tab key, their shortcuts were not announced, and it just announces (name) grouping. Only when we were navigating these groupings with object navigation, their shortcuts were announced, but again, when pressing these shortcuts, there were not reachable with these shortcuts. For example, when we pressed alt+s in the speech Categorie so many times, we would only reach (Use spelling functionality if supported  check box  checked  Alt+s). Or when pressed alt+d in Braille Categorie, we'd just here the windows sound and not reach any option. So, the shortcuts for these groupings were completely unusable, and also, the shortcut of the change button in these groupings can take you to these groupings, so all these reasons can be the result of removing the shortcuts of these groupings.
When navigating the (Synthesizer) and the (Braille display) groupings with nvda's object navigation, their unusable shortcuts (alt+s) and (alt+d) will no longer be announced, they have been removed because they didn't have any usage.
removed & symbol from the related code lines, to remove the (S) and (D) shortcuts from them.
### Testing strategy:
tested that the shortcuts of these groupings are no longer announced when navigating them with object navigation.
non
### Code Review Checklist:

- [ ] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -3584,7 +3584,7 @@ class BrailleSettingsPanel(SettingsPanel):
 			evt.Skip()
 
 	def onChangeDisplay(self, evt):
-		changeDisplay = BrailleDisplaySelectionDialog(None, multiInstanceAllowed=True)
+		changeDisplay = BrailleDisplaySelectionDialog(self, multiInstanceAllowed=True)
 		ret = changeDisplay.ShowModal()
 		if ret == wx.ID_OK:
 			self.Freeze()
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -1073,7 +1073,7 @@ class SpeechSettingsPanel(SettingsPanel):
 			evt.Skip()
 
 	def onChangeSynth(self, evt):
-		changeSynth = SynthesizerSelectionDialog(self, multiInstanceAllowed=True)
+		changeSynth = SynthesizerSelectionDialog(None, multiInstanceAllowed=True)
 		ret = changeSynth.ShowModal()
 		if ret == wx.ID_OK:
 			self.Freeze()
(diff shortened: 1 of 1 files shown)


HanaDadfar:
@seanbudd hi.
i'm trying to make it so that the Select Synthesizer  dialog is not a children of the settings dialog, but i don't know what to do. I've tryed to change self to none in that change sinth function, but it doesn't work. Can you give me some ideas? Thanks.

XLTechie:
> i'm trying to make it so that the Select Synthesizer dialog is not a children of the settings dialog, but i don't know what to do.

How is that relevant to what this PR says it is trying to do? That seems like 
trying to achieve a separate goal.

HanaDadfar:
@XLTechie my point is to do 2 things in this pr, but i do not know how to go forward with the last one in standard way. Once i'm able to, i'll update the discription and title of this pr completely.

HanaDadfar:
i want to do this with out breaking the ability of updating the sinth name in that read only edit box.
i could just do wx.call_after for that change button function, but it'll break this. I'll do my last trys, and if i can't go, this pr will just do one point, which i already mentioned.

XLTechie:
@HanaDadfar may I suggest that you stick to the thing you opened your PR to do?

I don't think the main objective of removing the useless key assignments is 
controversial. But the second thing you are trying to do may have wider 
implications, and need a larger discussion.
I am aware of that issue and what is being discussed there.

Trying to do two unrelated things in the same PR may cause this simple PR to be 
rejected when it otherwise wouldn't be.

As to your other comment, sometimes core.callLater works better than 
wx.CallAfter. Though I don't know that either will help in this situation.

HanaDadfar:
this pr is now ready

HanaDadfar:
this pr shouldn't be merged because I have to open an Issue, which is related.

HanaDadfar:
this pr is not blocked and it can be merged, because its point stil stands, as even in 2020.4 where the shortcuts of these options were announced, these options were not reachable when their shortcut keys were pressed, and also, the shortcut keys of the change buttons are stil stands and can take you in to these groupings.

seanbudd (approved):
Thanks @HanaDadfar

PR #16286, opened 2024-03-09
Mute microphone command

### LINKED ISSUE #16056: Feature request: mute microphone command
I would like to be able to mute microphone easily during online calls in Skype, Teams, Zoom and browser-based VC apps.
NVDA keystroke that mutes/unmutes microphone. I propose to bind it by default to `NVDA+Delete`.
1. Windows 11 claims to have `Win+Alt+K` global keyboard command to mute/unmute microphone. However that's not true. Quick investigation shows that it only works in Microsoft Teams. Ironically, it doesn't work in Skype, a Microsoft application too. There is evidence all over Internet that it doesn't work in a bunch of other VC apps. Even if it works in Teams, there is no spoken notification, so NVDA users won't know whetehr the mic is muted or unmuted, rendering this keystroke useless.
2. Some Laptop manufacturers  have  a keystroke on the laptop, such as `Fn+F4` for Lenovo. However this is not a good solution either, because again it doesn't trigger any spoken  notification, and it doesn't work with external keyboards.
### Additional context
Prototype is implemented in Tony's enhancements add-on.
Prior discussion: #16037

mltony:
### Link to issue number:
Closes #16056
### Summary of the issue:
A command to mute/unmute microphone.
Added a global command to mute/unmute microphone. No default gesture assigned.
Added wasapi notification listener; when it detects microphone mute/unmute event - it speaks it. This would speak notification even if microphone is muted by other means, e.g. laptop-specific keystroke or windows settings.
Made use of pycaw for both mute command and notification listener.
### Testing strategy:
Tested manually on Windows 11.
Wasapi notification comes in with a noticeable delay of about 1 second.
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/audio/notifications.py
+++ b/source/audio/notifications.py
@@ -43,7 +43,7 @@ def terminate():
 			try:
 				deviceEnumerator.UnregisterEndpointNotificationCallback(mmClient)
 			except COMError:
-				log.exception("Could not terminate audio notifications module"):
+				log.exception("Could not terminate audio notifications module")
 				return
 
 
--- a/source/audio/notifications.py
+++ b/source/audio/notifications.py
@@ -35,8 +35,11 @@ def initialize() -> None:
 
 @atexit.register
 def terminate():
-	if nvwave.usingWasapiWavePlayer():
-		if mmClient is not None and deviceEnumerator is not None:
+	if (
+		nvwave.usingWasapiWavePlayer():
+		and mmClient is not None
+		and deviceEnumerator is not None
+	):
 			try:
 				deviceEnumerator.UnregisterEndpointNotificationCallback(mmClient)
 			except COMError:
--- a/source/audio/notifications.py
+++ b/source/audio/notifications.py
@@ -1,0 +1,74 @@
+# A part of NonVisual Desktop Access (NVDA)
+# Copyright (C) 2024 NV Access Limited
+# This file is covered by the GNU General Public License.
+# See the file COPYING for more details.
+
+from comtypes import GUID, COMError
+from pycaw.callbacks import MMNotificationClient
+from pycaw.api.mmdeviceapi.depend.structures import PROPERTYKEY
+from pycaw.utils import AudioUtilities
+from pycaw.constants import EDataFlow
+from typing import Any
+from logHandler import log
+import ui
+import nvwave
+import atexit
+
+deviceEnumerator: Any | None = None
+mmClient: MMNotificationClient | None = None
+
+
+def initialize() -> None:
+	if nvwave.usingWasapiWavePlayer():
+		global mmClient, deviceEnumerator
+		try:
+			mmClient = MMNotificationClientImpl()
+			deviceEnumerator = AudioUtilities.GetDeviceEnumerator()
+			deviceEnumerator.RegisterEndpointNotificationCallback(mmClient)
+		except COMError as e:
+			log.exception("Could not initialize audio notifications module", e)
+			return
+
+	else:
+		log.debug("Cannot initialize audio notifications as WASAPI is disabled")
+
+
+@atexit.register
+def terminate():
+	if nvwave.usingWasapiWavePlayer():
+		if mmClient is not None and deviceEnumerator is not None:
+			try:
+				deviceEnumerator.UnregisterEndpointNotificationCallback(mmClient)
+			except COMError as e:
+				log.exception("Could not terminate audio notifications module", e)
+				return
+
+
+GUID_DEVICE_VOLUME_OR_MUTE = GUID("{9855C4CD-DF8C-449C-A181-8191B68BD06C}")
+EVENT_MUTE = (GUID_DEVICE_VOLUME_OR_MUTE, 1)
+
+
+class MMNotificationClientImpl(MMNotificationClient):
+	def on_property_value_changed(self, device_id: str, property_struct: PROPERTYKEY, fmtid: GUID, pid: int):
+		global EVENT_MUTE
+		match (fmtid, pid):
+			case EVENT_MUTE:
+				defaultMicrophone = AudioUtilities.GetMicrophone()
+				defaultMicrophoneId = defaultMicrophone .GetId() if defaultMicrophone is not None else None
+				device = deviceEnumerator.GetDevice(device_id)
+				deviceObject = AudioUtilities.CreateDevice(device)
+				if device_id == defaultMicrophoneId:
+					microphoneName = ""
+					isMicrophone = True
+				else:
+					microphoneName = deviceObject.FriendlyName
+					isMicrophone = AudioUtilities.GetEndpointDataFlow(device_id, 1) == EDataFlow.eCapture.value
+				if isMicrophone:
+					isMuted = deviceObject.EndpointVolume.GetMute()
+					if isMuted:
+						# Translators: notification spoken when microphone muted/unmuted
+						msg = _("Muted microphone %s") % microphoneName
+					else:
+						# Translators: notification spoken when microphone muted/unmuted
+						msg = _("Unmuted microphone %s") % microphoneName
+					ui.message(msg)
--- a/source/audio/notifications.py
+++ b/source/audio/notifications.py
@@ -36,7 +36,7 @@ def initialize() -> None:
 @atexit.register
 def terminate():
 	if (
-		nvwave.usingWasapiWavePlayer():
+		nvwave.usingWasapiWavePlayer()
 		and mmClient is not None
 		and deviceEnumerator is not None
 	):
--- a/source/audio/notifications.py
+++ b/source/audio/notifications.py
@@ -39,8 +39,8 @@ def terminate():
 		if mmClient is not None and deviceEnumerator is not None:
 			try:
 				deviceEnumerator.UnregisterEndpointNotificationCallback(mmClient)
-			except COMError as e:
-				log.exception("Could not terminate audio notifications module", e)
+			except COMError:
+				log.exception("Could not terminate audio notifications module"):
 				return
 
 
--- a/source/audio/notifications.py
+++ b/source/audio/notifications.py
@@ -25,8 +25,8 @@ def initialize() -> None:
 			mmClient = MMNotificationClientImpl()
 			deviceEnumerator = AudioUtilities.GetDeviceEnumerator()
 			deviceEnumerator.RegisterEndpointNotificationCallback(mmClient)
-		except COMError as e:
-			log.exception("Could not initialize audio notifications module", e)
+		except COMError:
+			log.exception("Could not initialize audio notifications module")
 			return
 
 	else:
--- a/source/audio/notifications.py
+++ b/source/audio/notifications.py
@@ -57,7 +57,7 @@ class MMNotificationClientImpl(MMNotificationClient):
 		match (fmtid, pid):
 			case EVENT_MUTE:
 				defaultMicrophone = AudioUtilities.GetMicrophone()
-				defaultMicrophoneId = defaultMicrophone .GetId() if defaultMicrophone is not None else None
+				defaultMicrophoneId = defaultMicrophone.GetId() if defaultMicrophone is not None else None
 				device = deviceEnumerator.GetDevice(device_id)
 				deviceObject = AudioUtilities.CreateDevice(device)
 				if device_id == defaultMicrophoneId:
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -4481,7 +4481,7 @@ class GlobalCommands(ScriptableObject):
 	@script(
 		description=_(
 			# Translators: Describes a command.
-			"Toggles microphone mute",
+			"Toggle microphone mute",
 		),
 		category=SCRCAT_AUDIO,
 		gesture=None,
--- a/source/globalCommands.py
+++ b/source/globalCommands.py
@@ -4489,7 +4489,7 @@ class GlobalCommands(ScriptableObject):
 	def script_toggleMicrophoneMute(self, gesture: "inputCore.InputGesture") -> None:
 		if not nvwave.usingWasapiWavePlayer():
 			# Translators: error message for toggle microphone mute command
-			ui.message(_("Please enable wasapi in order to mute microphone"))
+			ui.message(_("Please enable WASAPI in order to mute microphone"))
 			return
 		microphoneInterface = AudioUtilities.GetMicrophone()
 		if microphoneInterface is None:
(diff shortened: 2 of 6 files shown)


codeofdusk:
> Added wasapi notification listener; when it detects microphone mute/unmute event - it speaks it.

Makes sense – I'm guessing this is indicated visually, and a screen reader report here gives us equivalent access.

> Added a global command to mute/unmute microphone. No default gesture assigned.

This feels... out of place in a screen reader. Doesn't Windows have a hotkey to do this now?

mltony:
@codeofdusk ,

> Doesn't Windows have a hotkey to do this now?

No, see explanation in the original feature request.

XLTechie:
How does the mute control work if you have more than one microphone, or more 
than one sound device with microphone inputs?

I, for example, have an on-board microphone in one of my laptops. Of course it's 
junk, so I never use it. However, on that same laptop, I have a six-channel 
audio interface attached, with two XLR microphone inputs (one of which I use).

What is muted if I use this control?

(And yes, I know I could just try the build. But I would like to know the 
intended outcome first.)

mltony:
@XLTechie, the global command would mute the default microphone. Notification will be spoken if any microphone is muted. If not default microphone is muted, it would also speak its name; while for default one it'll just say "muted/unmuted microphone".

Brian1Gaff:
Many individual programs do, like Zoom etc. It all depends on what you term microphone here. I use line in and out as a direct loop through the sound card.
 However a program can still mute any sound going to zoom etc in software.
 Brian

-- 
***@***.*** 
in the display name field.
  ----- Original Message -----

bramd:
The notifications may indeed be useful, but I wonder if we are getting more information than visual users here. As far as I know, Windows shows a notification icon in the notification area/system tray when an app uses the microphone to alert the user to this, however I don't think there is a clear global visual indicator for the mic status. Would it be a better translation of visual information if we give a notification as soon as the microphone notification icon is present?

Some programs, like Zoom or Teams send their own notifications when the microphone is muted/unmuted, so with the global notification on top that might be a bit much? In other programs, an active notification might not be send and there we could enable this feature to have information that is equal to what is visible on the screen. Also, a short sound may be considered instead to not disturb an ongoing voice communication so much.

To recap, I'm not against these notifications and think they might be very useful in some circumstances, but it raises the question if we should provide more information than what is given to visual users.

beqabeqa473:
I think it is upto software what kind of notification it subscribes
to. I don't see a point in this last comment.
Lots of wireless headsets have a possibility to identify in their
software if microphone is muted or unmuted, and why nvda should not
have such possibility? I don't think there is something bad with it.

On 3/11/24, Bram Duvigneau ***@***.***> wrote:
> The notifications may indeed be useful, but I wonder if we are getting more
> information than visual users here. As far as I know, Windows shows a
> notification icon in the notification area/system tray when an app uses the
> microphone to alert the user to this, however I don't think there is a clear
> global visual indicator for the mic status. Would it be a better translation
> of visual information if we give a notification as soon as the microphone
> notification icon is present?
>
> Some programs, like Zoom or Teams send their own notifications when the
> microphone is muted/unmuted, so with the global notification on top that
> might be a bit much? In other programs, an active notification might not be
> send and there we could enable this feature to have information that is
> equal to what is visible on the screen. Also, a short sound may be
> considered instead to not disturb an ongoing voice communication so much.
>
> To recap, I'm not against these notifications and think they might be very
> useful in some circumstances, but it raises the question if we should
> provide more information than what is given to visual users.
>
> --
> https://github.com/nvaccess/nvda/pull/16286#issuecomment-1987863694
>
> Message ID: ***@***.***>


-- 
Tell: +995593454005
Email: ***@***.***
Web: https://gozaltech.org
Skype: beqabeqa473

XLTechie:
@bramd wrote:
> The notifications may indeed be useful, but I wonder if we are getting more information than visual users here. As far as I know, Windows shows a
> notification icon in the notification area/system tray when an app uses the microphone to alert the user to this, however I don't think there is a clear
> global visual indicator for the mic status. Would it be a better translation of visual information if we give a notification as soon as the microphone
> notification icon is present?

It is not always best to get precisely analogous information to what the sighted 
get. We have to get information that gives us comparable understanding. There is 
a difference.

A sighted user can see that the microphone icon is visible. Seeing this does not 
impede the use of the microphone.

But if NVDA were to indicate as soon as that icon lights up, it would probably 
be picked up by the microphone. That means the user will have to take steps to 
avoid hearing or picking up that notification.
This is a case where precisely analogous information hurts accessibility, not 
promotes it.

Whereas an indication when the microphone is muted/unmuted, will probably come 
before the microphone needs to be used. The user gains access to the same 
information, but in a way that is far more useful to a speech user, than the 
more accurate sight-compared notification would be.

Adriani90:
Sighted people are also looking for a solution in this regard, see different discussions on Redit. I found a tool, that solves exactly this use case:
https://github.com/bralexc/mute-unmute-meetings

@mltony given this tool is also something sighted people use, is it something you could do too for solving your use case? Otherwise I don't see a strong argument why this feature should be part of NVDA core if there is already a tool for it. That tool even allows you to define audio indication when the Microfone is muted or unmuted.

seanbudd:
@bramd - many keyboards (including my laptop) have microphone mute function key with an associated light to indicate if it is toggled. This issue was initially triaged as there is usually a visual notification on the screen and sometimes keyboard lights to indicate microphone muting. I don't think we necessarily need the keyboard command to perform this, as there are other (e.g. keyboard, app specific) solutions and it is slightly out of scope, but the feature is pretty minor and very handy.

seanbudd (commented):
Generally looks good to me, thanks @mltony

source/globalCommands.py
```diff
@@ -4477,6 +4478,32 @@ def script_cycleParagraphStyle(self, gesture: "inputCore.InputGesture") -> None:
...
+			# Translators: Describes a command.
+			"Toggles microphone mute",
+		),
+		category=SCRCAT_AUDIO,
+		gesture=None,
+	)
+	def script_toggleMicrophoneMute(self, gesture: "inputCore.InputGesture") -> None:
+		if not nvwave.usingWasapiWavePlayer():
+			# Translators: error message for toggle microphone mute command
+			ui.message(_("Please enable wasapi in order to mute microphone"))
```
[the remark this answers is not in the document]
codeofdusk: Or, even simpler:

```suggestion
			ui.message(_("Please enable WASAPI to mute microphone"))
```

XLTechie:
> toggle microphone mute command allows you to mute or unmute default microphone 
with a single keystroke.

"the" default microphone

Adriani90:
@mltony can you make sure the original state of the microfone is returned when closing NVDA? Otherwise we run into the same problem as in #16292.

And we don't really want to make sighted users working on the same machine confused.

I still think implementing an announcement of Microfone state when changed globally might be rather something that belongs to NVDA, and not the command itself. However, the native notification of mic state should be preserved when changing with the native scripts of the coresponding app.

source/audio/notifications.py
```diff
@@ -0,0 +1,77 @@
...
+			return
+
+	else:
+		log.debug("Cannot initialize audio notifications as WASAPI is disabled")
+
+
+@atexit.register
+def terminate():
+	if (
+		nvwave.usingWasapiWavePlayer():
```
CyrilleB79, line 39:
```suggestion
		nvwave.usingWasapiWavePlayer()
```

source/audio/notifications.py
```diff
@@ -0,0 +1,77 @@
...
+		and mmClient is not None
+		and deviceEnumerator is not None
+	):
+			try:
+				deviceEnumerator.UnregisterEndpointNotificationCallback(mmClient)
+			except COMError:
+				log.exception("Could not terminate audio notifications module"):
```
CyrilleB79, line 46:
```suggestion
				log.exception("Could not terminate audio notifications module")
```

CyrilleB79 (commented):
@mltony, I have tried to test this build, fixing the two syntax errors below.

Just wanted to report the following issue:
I have assigned a key (`NVDA+shift+delete`) to the mute / unmute command. Trying to use the command, it seems to change the mute state of the mic.
However, I cannot hear any mute / unmute notification. Is this expected?

I am running from source, under Windows 10 22H2 (AMD64) build 19045.4046.

mltony:
@CyrilleB79 , I can reproduce this on Windows 10. Will try to investigate.

mltony:
Ok, here are more results of my testing:
1. I have a Windows 11 laptop. Mute notifications work as expected.
2. I have a Windows 10 laptop. Mute notifications work as expected.
3. I have a virtual machine with Windows 10. It's got a microphone device. Mute notifications are not coming at all.

@CyrilleB79 , do you experience your issue on a real installation or VM?
I couldn't figure out the difference between two windows 10 installations. I verified that notification are not coming even when I use pycaw example, so the issue is not with my code, but most likely with wasapi itself.
Googling doesn't return any relevant results at all - I guess not many people use this type of audio notifications.
The conclusion I would like to make is this: wasapi notifications appear to be unreliable. Given that, do we want to still include them in the coreand risk having bug reports from people for whom they don't work? Or better not include them?
I am thinking that the best course would be to make mute mic command to speak its own update, but keep audio notifications, but make a note in user_guide saying that these notifications are not reliable.

CyrilleB79:
My test was done on a real installation, but with no recording or application using the mic.

Same result with a second Windows 10 machine running Windows 10 2004 (AMD64) build 19041.388.
On this second machine, I have tested the mute/unmute command while audacity is recording and the notification is not heard; the mic is correctly muted/unmuted though.

In addition, I have 3 or 4 notifications heard when I open audacity, before I record anything. May it be a matter of sound card disabling the mic when not needed.

mltony:
At this point I'm biased to drop mute notifications all together. While it works perfectly on my Windows 11 laptop, it has problems on both my Windows 10 installations and also Cyrille reported similar problems on Windows 10 as well.
So the symptoms  I'm observing are:
1. Sometimes mute notification doesn't fire at all.
2. Sometimes we receive multiple mute notification. On my Windows 10 laptop manually unmuting it in settings (well unchecking "disable microphone" checkbox) triggers 37 notifications in a row.
3. Sometimes we receive mute notification on Windows  login - e.g. without even explicitly muting/unmuting microphone.
4. (reported by Cyrille) Mute notifications are fired when opening Audacity.

I did some debugging, and checked notifications with pycaw example - I see the same symptoms, so likely this is not caused by a bug in my code. I don't think pycaw is at fault either - it's just a thin wrapper over com interfaces. Most likely something is wrong with Windows 10. It's possible that Microsoft fixed that only in Windows 11. Would be great if anyone else could test if this works on Windows 11. But other than that - if people don't have any better ideas - I might get rid of notifications and have mute microphone command speak microphone status.
Another potential idea is to have a thread check whetehr microphone is muted say every 1 second and announce updates there. But I'm not excited about creating a new thread.

Adriani90:
But if it works on Win 11 it should be good enough to go.  for Win 10 it will need an external fix which means we have to take it up with Microsoft. This is not the only case which works better in Win 11 and needs an external fix on Win 10. If you are sure that the root of this problem is not in your code, then this should be fine for this case. But ofcourse NV Access has the last decision hereVon meinem iPhone gesendetAm 15.03.2024 um 22:55 schrieb mltony ***@***.***>:﻿
At this point I'm biased to drop mute notifications all together. While it works perfectly on my Windows 11 laptop, it has problems on both my Windows 10 installations and also Cyrille reported similar problems on Windows 10 as well.
So the symptoms  I'm observing are:

Sometimes mute notification doesn't fire at all.
Sometimes we receive multiple mute notification. On my Windows 10 laptop manually unmuting it in settings (well unchecking "disable microphone" checkbox) triggers 37 notifications in a row.
Sometimes we receive mute notification on Windows  login - e.g. without even explicitly muting/unmuting microphone.
(reported by Cyrille) Mute notifications are fired when opening Audacity.

I did some debugging, and checked notifications with pycaw example - I see the same symptoms, so likely this is not caused by a bug in my code. I don't think pycaw is at fault either - it's just a thin wrapper over com interfaces. Most likely something is wrong with Windows 10. It's possible that Microsoft fixed that only in Windows 11. Would be great if anyone else could test if this works on Windows 11. But other than that - if people don't have any better ideas - I might get rid of notifications and have mute microphone command speak microphone status.
Another potential idea is to have a thread check whetehr microphone is muted say every 1 second and announce updates there. But I'm not excited about creating a new thread.

mltony:
Makes sense; then it would be good if people can test if this works correctly on Win11.

mltony:
More testing shows notifications are not reliable on my Windows 11 as well. My configuration is laptop with windows 11; it has two microphones: builtin microphone array, and microphone in my bluetooth headphones.
1. When Bluetooth microphone is set as default, I get all notifications correctly. From both microphones. I get no repeating notifications.
2. When builtin microphone array is set as default, then I get no notifications at all. Neither from builtin nor from bluetooth. Pycaw example listener also doesn't get any notifications in this case.

I am biased to conclude that notification API (specifically `IMMNotificationClient::OnPropertyValueChanged`) is broken on Windows side (also seems like in Windows 11 Microsoft tried to fix it and mostly fixed it, but still not all the way) and we shouldn't rely on it.

seanbudd:
@mltony - you say "we shouldn't rely on it", does that mean this PR should be closed as abandoned due to the approach not being feasible?

mltony:
Yes, sorry, forgot about this PR. TLDR: I found wasapi notifications regarding microphone mute status to be quite unreliable - likely  a problem of wasapi implementation. In theory microphone mute command can still be implemented in NVDA, but many NVDA contributors expressed negative sentiments regarding this feature so I kind of abandoned it - for those who need it - it is still available in Tony's enhancements add-on. Closing this as abandoned.

XLTechie:
Unless I misread what was posted, these appear reliable in real machines, but unreliable in VMs, according to @mltony's testing. If that's true, is that really a problem? Most people are not using VMs, and even if they are, a note in the guide that such notifications may fail in virtual machines appears sufficient.

mltony:
@XLTechie, wasapi notifications are unreliable on both  virtual and real machines according to my testing.

XLTechie:
Sorry @mltony, I missed your [comment about two microphones on Windows 11](https://github.com/nvaccess/nvda/pull/16286#issuecomment-2000751863).

Issue #16287, opened 2024-03-10
when sound split is disabled, the feature as it is in NVDA core now interferes with everything else that changes channel volumes, whether an add-on or not.


paulber19:
### Describe the solution you'd like
Some add-ons allow you to modify application channel volumes individually.
Since the introduction of the "sound split" feature, NVDA systematically returns sound to both channels of applications to hear sound equally on both channels when "Disabled" mode is selected.
In this case, the modifications made by the add-ons are canceled

### Describe alternatives you've considered
The "disabled" mode should mean that NVDA does nothing in this case and works as before the feature was introduced.
### Additional context
This concerns alpha versions of NVDA.

Adriani90:
Cc:  @mltony

paulber19:
Correction: it is not the applications that are centered, but only the 
system sounds.

CyrilleB79:
@paulber19 could you indicate the names of these add-ons as examples? Thanks.

mltony:
I implemented sound split in a generic enough way, so that add-on authors can reuse it to customize functionality - use `audio.soundSplit.applyToAllAudioSessions` function with a custom lambda. I think add-ons should go through this way, because pycaw is now included in NVDA and it's not a good idea to bring another version of pycaw in an add-on.

XLTechie:
@mltony
> I implemented sound split in a generic enough way, so that add-on authors can reuse it to customize functionality - use
> audio.soundSplit.applyToAllAudioSessions function with a custom lambda. I think add-ons should go through this way, because pycaw is now included in NVDA

Would this be worth documenting in the dev guide?

paulber19:
Totally agree. However, to continue providing functionality in previous 
versions of NVDA, this is necessary.

paulber19:
NVDAExtensionGlobalPlugin
With this add-on, it is possible to place nvda on one side and all 
applications on the other side, including system sounds.
But, although NVDA's "sound split" feature mode is disabled, NVDA puts 
the system sounds in the center.

CyrilleB79:
@paulber19
 what you are describing just seems another sound split feature; am I correct?
If yes, there is no point in keeping it enabled in the add-on for NVDA 2024.2 since the feature is in core. You may though want to keep it in the add-on for older versions of NVDA though.

However, if @paulber19's add-on only uses functions in the API (i.e. public) for this feature, the introduction in NVDA core of Sound Split feature would be an API breaking change, what should not be allowed for 2024.2.
[...]

PR #16312, opened 2024-03-15
Restore channel volume of applications upon exit

### LINKED ISSUE #16292: Sound split: closing NVDA doesn't restore panning for closed applications
### Steps to reproduce:
1. Start Foobar2000 and play some audio
2. Set sound split in NVDA, e.g. to NVDA left, apps right
3. Close Foobar2000
4. Disable sound split
5. Close NVDA
6. Start Foobar2000 and start playing again.

### Actual behavior:
Foobar is still panned to the right.

### Expected behavior:
Foobar should play through both speakers.

### Additional context
Note that NVDA seems to restore Foobars panning as soon as you start it. This basically means that sound split, even when it is disabled, changes the panning for applications.

Installed

#### NVDA version:
alpha-31398

#### Windows version:
Windows 11 23H2

Foobar2000 2.1.2

Not tested

No

Yes

N/a

mltony:
### Link to issue number:
Closes #16292
### Summary of the issue:
When sound split is enabled, if an application is closed while NVDA is still running, its channel volume would be preserved even after the app is restarted later without NVDA.
N/A
Catching `on_state_changed` notification and when `new_state == "Expired"` - this means that the applciation just exited - we restore both left and right channel to full volume.
### Testing strategy:
Manual.
N/A
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/audio/soundSplit.py
+++ b/source/audio/soundSplit.py
@@ -10,12 +10,14 @@ import globalVars
 from logHandler import log
 import nvwave
 from pycaw.api.audiopolicy import IAudioSessionManager2
-from pycaw.callbacks import AudioSessionNotification
+from pycaw.callbacks import AudioSessionNotification, AudioSessionEvents
 from pycaw.utils import AudioSession, AudioUtilities
 import ui
 from utils.displayString import DisplayStringIntEnum
 from dataclasses import dataclass
 from comtypes import COMError
+from threading import Lock
+import core
 
 VolumeTupleT = tuple[float, float]
 
@@ -99,6 +101,20 @@ def terminate():
 		log.debug("Skipping terminating sound split as WASAPI is disabled.")
 
 
+@dataclass(unsafe_hash=True)
+class AudioSessionNotificationWrapper(AudioSessionNotification):
+	listener: AudioSessionNotification
+
+	def on_session_created(self, new_session: AudioSession):
+		pid = new_session.ProcessId
+		with applicationExitCallbacksLock:
+			if pid not in applicationExitCallbacks:
+				volumeRestorer = VolumeRestorer(pid, new_session)
+				new_session.register_notification(volumeRestorer)
+				applicationExitCallbacks[pid] = volumeRestorer
+		self.listener.on_session_created(new_session)
+
+
 def applyToAllAudioSessions(
 		callback: AudioSessionNotification,
 		applyToFuture: bool = True,
@@ -111,6 +127,7 @@ def applyToAllAudioSessions(
 		or until unregisterCallback() is called.
 	"""
 	unregisterCallback()
+	callback = AudioSessionNotificationWrapper(callback)
 	if applyToFuture:
 		audioSessionManager.RegisterSessionNotification(callback)
 		# The following call is required to make callback to work:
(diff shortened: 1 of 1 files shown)


source/audio/soundSplit.py
```diff
@@ -192,3 +209,45 @@ def toggleSoundSplitState() -> None:
...
+		# Application connected to this audio session is terminating. Restore its volume.
+		try:
+			channelVolume = self.audioSession.channelAudioVolume()
+			channelCount = channelVolume.GetChannelCount()
+			if channelCount != 2:
+				log.warning(
+					f"Audio session for pid {self.pid} has {channelCount} channels instead of 2 - cannot set volume!"
+				)
+				return
+			channelVolume.SetChannelVolume(0, 1.0, None)
```
LeonarddeR, line 234: Why should the volume be 1.0? What if the application itself lowered the volume of its channels?
I think the volume of both channels should be restored to the volume of the active channel. SO if soundsplit was active with NVDA left, application right and the application lowered its volume to 0.5, both channels should be set to 0.5 instead of 1.0.

LeonarddeR (changes requested):
I have reopened #16287. I'd prefer this pr to fix that as well, i.e. ensure that soundsplit disabled really means that it is disabled.

Looking at the code of sound split in general, I have another concern. It looks like that Sound SPlit ignores the volume of every channel, simply resetting it to 1.0. What if an app prefers to change channel volumes to a lower level? I think sound split should respect that by using the average channel volume rather than a hardcoded value of 1.0

mltony:
In practice I've never seen applications to lower down their volume using IChannelAudioVolume interface. So restoring to 1.0 seems to be good enough. Moreover, the volume set via ISimpleAudioVolume is different from IChanelAudioVolume. If we assume that apps adjust their channel volume using simple volume, then their volume will be preserved. If apps adjust their volume via channel volume, then it is going to interfere with sound split no matter what, but again, I haven't seen this.

paulber19:
Apart from the add-on that I already mentioned, there is also the 
"soundVolumeView" application from NirSoft which allows you to adjust 
the channel balance.
Under NVDA 2024beta, it is enough to adjust the channel balance of 
foobar2000 once with this software and foobar2000 can be stopped and 
restarted without losing this setting.
Under nvda lpha and when the sound separation mode is "Disabled", the 
setting is lost each time foobar2000 is restarted and the channels are 
systematically rebalanced.

LeonarddeR:
> In practice I've never seen applications to lower down their volume using IChannelAudioVolume interface. So restoring to 1.0 seems to be good enough. 

Honestly, I don't care much about practice. I think it is common sense that NVDA should only change/touch things that are absolutely necessary. In other words, NVDA should leave my system as it was before after stopping it, especially when running as a portable copy. This is a fundamental principle for me, and the current implementation violates that.

mltony:
Sound split already restores all volumes to 1 upon exit. I don't see why in this feature the behavior should be different.
In theory it would be possible to remember previous volume and try to restore to this volume when application exits. But:
1. I haven't observed any applications adjusting channel volume, so testing is going to be impossible.
2. This would make code more complicated - and I hesitate to add unnecessary complexity just for an imaginary and untestable scenario.
3. The only scenario I can imagine is when another application manages channel volumes. In this case that application would conflict with sound split no matter what. I can imagine corner cases where NVDA and said application adjust volumes at the same time. Race condition would be unavoidable. So Given utility of sound split I would say we should claim that sound split is not compatible with said application if one is ever to be discovered, because making it compatible is going to be hard if ever possible.

LeonarddeR (changes requested):
> Sound split already restores all volumes to 1 upon exit. I don't see why in this feature the behavior should be different. 

We are talking about Sound SPlit here. So if in this change the volume shouldn't be restored to 1.0. it should also be applied to sound split in general.

>     3. The only scenario I can imagine is when another application manages channel volumes. In this case that application would conflict with sound split no matter what. I can imagine corner cases where NVDA and said application adjust volumes at the same time. Race condition would be unavoidable. So Given utility of sound split I would say we should claim that sound split is not compatible with said application if one is ever to be discovered, because making it compatible is going to be hard if ever possible.

This actually emphasizes my concern that Sound Split touches channel volumes in the off state. I'm happy to accept that sound split is incompatible with some applications and even understand that these situations could be imaginary. But don't unnecessarily fiddle with my system parameters please.

michaelDCurran (approved):

PR #16331, opened 2024-03-23
Add sound split disabled mode

### LINKED ISSUE #16287: when sound split is disabled, the feature as it is in NVDA core now interferes with everything else that changes channel volumes, whether an add-on or not.

mltony:
### Link to issue number:
Closes #16287
### Summary of the issue:
Adding disabled sound split mode
Adding disabled sound split mode
1. If sound split mode is disabled at startup, we're not touching any volumes.
2. If user toggles to disabled mode, then first we set mode to NVDA_BOTH_APPS_BOTH to effectively restore the volumes, then set sound pslit state to off.
### Testing strategy:
Manual
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/audio/soundSplit.py
+++ b/source/audio/soundSplit.py
@@ -108,7 +108,9 @@ def initialize() -> None:
 @atexit.register
 def terminate():
 	if nvwave.usingWasapiWavePlayer():
-		setSoundSplitState(SoundSplitState.NVDA_BOTH_APPS_BOTH)
+		state = SoundSplitState(config.conf["audio"]["soundSplitState"])
+		if state != SoundSplitState.OFF:
+			setSoundSplitState(SoundSplitState.OFF)
 		unregisterCallback()
 	else:
 		log.debug("Skipping terminating sound split as WASAPI is disabled.")
--- a/source/audio/soundSplit.py
+++ b/source/audio/soundSplit.py
@@ -37,9 +37,9 @@ class SoundSplitState(DisplayStringIntEnum):
 		return {
 			# Translators: Sound split state
 			SoundSplitState.OFF: pgettext("SoundSplit", "Sound split disabled"),
-			# Translators: Sound split state
 			SoundSplitState.NVDA_BOTH_APPS_BOTH: pgettext(
 				"SoundSplit",
+				# Translators: Sound split state
 				"NVDA in both channels and applications in both channels",
 			),
 			# Translators: Sound split state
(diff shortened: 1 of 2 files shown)


cary-rowen:
Hi @mltony 

Thank you for your work.

I tested this snapshot, 
I first selected "NVDA on the left and applications on the right" in the "Audio" settings panel, clicked OK, and NVDA performed the expected.

Then, I slected "Sound split disabled" in combo box in the "audio" settings panel and click "OK" did not disable sound splitt.

However, when I use the gesture nvda+alt+S it can be disabled successfully.

Thanks

mltony:
@cary-rowen, just fixed that, please try again.

source/audio/soundSplit.py
```diff
@@ -85,15 +100,15 @@ def initialize() -> None:
...
+		setSoundSplitState(state, initial=True)
 	else:
 		log.debug("Cannot initialize sound split as WASAPI is disabled")
 
 
 @atexit.register
 def terminate():
 	if nvwave.usingWasapiWavePlayer():
-		setSoundSplitState(SoundSplitState.OFF)
+		setSoundSplitState(SoundSplitState.NVDA_BOTH_APPS_BOTH)
```
LeonarddeR, line 111: Why this change? This reads to me as sound split being set to both channels on every terminate, effectively meaning that parameters are still being changed when terminating.

LeonarddeR (commented):
Thanks for taking this

LeonarddeR:
Personally, I think the difference between `off` and `NVDA in both channels and applications in both channels` is a bit vague at first sight. The documentation definitely needs updating, but may be debate about the UX first? @Qchristensen What do you think?

source/audio/soundSplit.py
mltony: Fixed that.

michaelDCurran (approved):

PR #16340, opened 2024-03-28
Add com port option in Braille Display Select Dialog and fix Serial Connection error

EdKweon:
### Summary of the issue:
Add com port in Hims Braille Display and fix Serial Connection error
- Users can always perform serial connections normally.
- User can choose com port in Braille Display Selection Dialog(Hims).
- In the HimsDisplayDriver, due to a chip issue, there were instances where packet data was intermittently split arbitrarily during the serial connection verification process, necessitating modifications to the _onReceive function. 
### Testing strategy:
Lint, System, Unit
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

## Summary by CodeRabbit

- **New Features**
  - Introduced fallback device management for USB device drivers to improve reliability when primary drivers are unavailable.

- **Enhancements**
  - Improved handling of multiple device types and identifiers in automatic detection.
  - Enhanced the serial data management to handle data fragments more effectively.

- **Bug Fixes**
  - Enhanced logic to ensure fallback devices are correctly utilized and cleared when necessary.


source/bdDetect.py
```diff
@@ -462,14 +462,14 @@ def getConnectedUsbDevicesForDriver(driver: str) -> Iterator[DeviceMatch]:
 			DeviceMatch(DeviceType.CUSTOM, port["usbID"], port["devicePath"], port)
 			for port in deviceInfoFetcher.usbDevices
 		),
-		(
```
LeonarddeR, line 465: Could you please explain why you changed this?

source/brailleDisplayDrivers/hims.py
```diff
@@ -272,26 +272,23 @@ def registerAutomaticDetection(cls, driverRegistrar: bdDetect.DriverRegistrar):
...
 		super(BrailleDisplayDriver, self).__init__()
 		self.numCells = 0
 		self._model = None
+		self._serData = b''
 
 		for match in self._getTryPorts(port):
 			portType, portId, port, portInfo = match
 			self.isBulk = portType == bdDetect.DeviceType.CUSTOM
 			self.isHID = portType == bdDetect.DeviceType.HID
+			self.isSerial = portType == bdDetect.DeviceType.SERIAL
```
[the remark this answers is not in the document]
EdKweon: Thanks. I will delete that line.

source/bdDetect.py
EdKweon: Currently, BrailleEdge products may simultaneously connect through Serial and HID connections. In cases where connection requests for both occur simultaneously, we aim to automatically prioritize and establish a Serial connection. Merely altering the sequence appeared to be without issues. If there are any objections or potential problems, your input would be greatly appreciated.

LeonarddeR: Why would you prefer a USB serial connection over USB HID?

EdKweon: @LeonarddeR 
The previous himsDriver did not use HID, so hims users are more familiar with Serial. This is why there were many opinions suggesting it's better to first explore Serial. Please let me know if there's any part you don't understand, and I would be grateful.

LeonarddeR: I find it a bit tricky to adjust the logic in NVDA just for this. In ordering, HID has always been preferred over Serial. People might rely on this behavior and might get confused when we swap them.

Personally, I would say HID is the modern standard. Moreover, I assume that HID behaves exactly the same as serial? Or is HID a different choice in the device?

EdKweon: Yes, that's correct. The device supports Serial in all countries, but HID is only supported in specific countries. There are countries where people who have been using Hims products traditionally choose the Serial method. If you know other solution, please tell me. I will follow.

lukaszgo1: Perhaps we should discuss this in terms of user impact i.e. what issues your users are experiencing due to current ordering. I agree with @LeonarddeR that changing ordering for entire auto detection to accommodate a single Braille display driver seems a wrong choice.

EdKweon: Thank you for your explanation.
And then is that impossible only change the ordering of detection in hims display?

lukaszgo1: Could you please answer the first part of my question i.e. what is the user visible impact of the current ordering?

EdKweon: There is a product that supports both HID and Serial connections. However, in certain countries, HID usage is not desired. In such cases, if the device automatically connects via HID first, users in countries where only Serial is used will be unable to use screen readers.

lukaszgo1: > However, in certain countries, HID usage is not desired.

Are you saying that the Braille display technically supports HID, but users don't want to use it? If so I still have trouble understanding why they just cannot have it auto-detected using HID, and if desired switch to serial afterwards. Ideally if device exposes HID it should be as functional as serial is.

source/brailleDisplayDrivers/hims.py
```diff
@@ -497,15 +498,35 @@ def _onReceive(self, data: bytes):
 			firstByte = data
 			# data only contained the first byte. Read the rest from the device.
 			stream = self._dev
+
+		# sometimes serial data received splited data. so accmulate data until it reaches 10 bytes.
```
seanbudd, line 502:
```suggestion
		# sometimes serial data is received in fragments.
		# so accumulate data until it reaches 10 bytes.
```

source/brailleDisplayDrivers/hims.py
```diff
@@ -497,15 +498,35 @@ def _onReceive(self, data: bytes):
...
 		elif firstByte == b"\xfa":
-			# Command packets are ten bytes long
-			packet = firstByte + stream.read(9)
+			# serial data first received
+			if not self._serData:
+				try:
+					# Command packets are ten bytes long
+					packet = firstByte + stream.read(9)
+				except Exception:
+					# remained data will be received next onReceive
```
seanbudd, line 523:
```suggestion
					# remaining data will be received next onReceive
```

source/brailleDisplayDrivers/hims.py
```diff
@@ -497,15 +498,35 @@ def _onReceive(self, data: bytes):
...
+					self._serData = firstByte
+					return
+			else:
+				packet = self._serData
+				self._serData = b''
```
seanbudd, line 528:
```suggestion
				self._serData = b""
```

source/brailleDisplayDrivers/hims.py
```diff
@@ -497,15 +498,35 @@ def _onReceive(self, data: bytes):
 			firstByte = data
 			# data only contained the first byte. Read the rest from the device.
 			stream = self._dev
+
+		# sometimes serial data received splited data. so accmulate data until it reaches 10 bytes.
+		if self._serData:
+			self._serData += data
+			if len(self._serData) == 10:
+				firstByte = b'\xfa'
```
seanbudd, line 506:
```suggestion
				firstByte = b"\xfa"
```

seanbudd (dismissed):

source/bdDetect.py
seanbudd: Could you please answer this question from @lukaszgo1, @EdKweon ?

EdKweon: @lukaszgo1 
In the case of serial connections, most functionalities can be operated without the NVDA screen reader. However, with HID connections, there is currently a total dependency on the NVDA screen reader. Due to this situation, users in certain countries have decided not to use HID. If implementing this code proves difficult, please let me know. I will roll back the code and explore other alternatives.

LeonarddeR:
1. As far as I understand the product, you can choose two connection modes, either HID or Serial. Furthermore, for countries that don't want to use HID, you hide the HID method. Is that correct?
2. If the desired operating mode is Serial, shouldn't the device reject requests to connect using HID?
3. Should people who don't want to use HID have the possibility to disable HID discoverability of the device altogether?

EdKweon:
@LeonarddeR 
Your points accurately. However, the current issue is that this device cannot hide the HID mode. 
1. Yes, that is correct. 
2. Currently, the system team has stated that it is challenging to implement a method to reject HID connection requests.
3. As it stands, when the device is connected via USB, the HID connection is inevitably established. It would be ideal to disable it, but currently, it seems difficult. 

Given these issues, there is a need for a software solution. Consequently, my current plan is to modify the code to check via Serial if HID is connectable in software terms. Would this be acceptable?

EdKweon:
@LeonarddeR 
Yes . But I think that's impossible leaving out. I will commit code tomorrow. Thank you for your help

lukaszgo1:
@EdKweon I've looked briefly at the code and what you're trying to do seems doable but difficult. Here is how I'd go about de-prioritizing HID for a particular driver:
- Add an additional argument (something like `useAsAFallBack` which should default to `False`) on `DriverRegistrar.addUsbDevices` (for consistency the same can be done for `addBluetoothDevices`, though this is secondary)
- No longer store union of callables and sets of IDs in the `_driverDevices` for each device - abstract it into a class like `DeviceMatcher` - it should remember if the given match is a fallback and obviously allow to check if the device in question has a matching ID / Bluetooth name
- Then, in `getConnectedUsbDevicesForDriver` when iterating through found devices, for each one check if it should be used as a fallback, if so don't yield it immediately, but only at the end when there is no non-fallback match
- Finally, when registering HID ID for Hims, set it as a fallback device, which ensures it will be picked only when serial is not found.


Comments are appreciated.

EdKweon:
@lukaszgo1
Sorry for reply late. My work was busy. I will commit code this week. Thanks for your feedback

EdKweon:
@lukaszgo1
I'm really sorry for commit late. I think it is better than before that only check specific devices as you recommend. I really appreciate it if you check my code is right.

source/bdDetect.py
```diff
@@ -655,3 +666,8 @@ def addDeviceScanner(
 		scanForDevices.register(scanFunc)
 		if moveToStart:
 			scanForDevices.moveToEnd(scanFunc, last=False)
+
+
+class DeviceMathcer:
+    fallBackDevices = list()
+    matches = list()
```
seanbudd, line 673: Can you add the type to the lists here to replace `?`

```suggestion
    fallBackDevices: List[?] = []
    matches: List[?] = []
```

source/bdDetect.py
```diff
@@ -605,7 +612,7 @@ def _getDriverDict(self) -> DriverDictT:
 			ret = _driverDevices[self._driver] = DriverDictT(set)
 			return ret
 
-	def addUsbDevices(self, type: DeviceType, ids: Set[str]):
+	def addUsbDevices(self, type: DeviceType, ids: Set[str], useAsFallBack: bool = False):
```
seanbudd, line 615: Just modernizing this with the change

```suggestion
	def addUsbDevices(self, type: DeviceType, ids: set[str], useAsFallBack: bool = False):
```

source/brailleDisplayDrivers/hims.py
```diff
@@ -250,7 +250,7 @@ def registerAutomaticDetection(cls, driverRegistrar: bdDetect.DriverRegistrar):
 		# Hid device
 		driverRegistrar.addUsbDevices(bdDetect.DeviceType.HID, {
 			"VID_045E&PID_940A",  # Braille Edge3S 40
-		})
+		}, True)
```
seanbudd, line 253: Can you split each parameter up into new lines e.g.

```py
driverRegistrar.addUsbDevices(
	bdDetect.DeviceType.HID,
	{
		"VID_045E&PID_940A",  # Braille Edge3S 40
	},
	True
)

File: source/brailleDisplayDrivers/hims.py [hunk 10]
```diff
@@ -272,12 +272,13 @@ def registerAutomaticDetection(cls, driverRegistrar: bdDetect.DriverRegistrar):
 
 	@classmethod
 	def getManualPorts(cls):
```
seanbudd, line 274: can you add a return type here?

source/brailleDisplayDrivers/hims.py
```diff
@@ -272,12 +272,13 @@ def registerAutomaticDetection(cls, driverRegistrar: bdDetect.DriverRegistrar):
...
 	@classmethod
 	def getManualPorts(cls):
-		return braille.getSerialPorts(filterFunc=lambda info: "bluetoothName" in info)
+		return braille.getSerialPorts()
 
 	def __init__(self, port="auto"):
 		super(BrailleDisplayDriver, self).__init__()
 		self.numCells = 0
 		self._model = None
+		self._serData = b''
```
seanbudd, line 281: We generally prefer full words for names just to reduce ambiguity (e.g. numCells is fine), however we user `serial` in full elsewhere

```suggestion
		self._serialData = b''
```

seanbudd (commented):
Thanks, just minor suggestions

source/bdDetect.py
```diff
@@ -605,7 +612,7 @@ def _getDriverDict(self) -> DriverDictT:
 			ret = _driverDevices[self._driver] = DriverDictT(set)
 			return ret
 
-	def addUsbDevices(self, type: DeviceType, ids: Set[str]):
+	def addUsbDevices(self, type: DeviceType, ids: Set[str], useAsFallBack: bool = False):
 		"""Associate USB devices with the driver on this instance.
 		:param type: The type of the driver.
 		:param ids: A set of USB IDs in the form C{"VID_xxxx&PID_XXXX"}.
```
LeonarddeR, line 618: Please at a description of the new `useAsFallBack` parameter to the doc string

source/bdDetect.py
```diff
@@ -618,9 +625,13 @@ def addUsbDevices(self, type: DeviceType, ids: Set[str]):
...
-		devs = self._getDriverDict()
-		driverUsb = devs[type]
-		driverUsb.update(ids)
+		
+		if not useAsFallBack:
+			devs = self._getDriverDict()
+			driverUsb = devs[type]
+			driverUsb.update(ids)
+		else:
+			DeviceMathcer.fallBackDevices.extend([(type, id) for id in ids])
```
LeonarddeR, line 634: This looks like a typo to me
```suggestion
			DeviceMatcher.fallBackDevices.extend([(type, id) for id in ids])
```

source/bdDetect.py
```diff
@@ -655,3 +666,8 @@ def addDeviceScanner(
 		scanForDevices.register(scanFunc)
 		if moveToStart:
 			scanForDevices.moveToEnd(scanFunc, last=False)
+
+
+class DeviceMathcer:
```
LeonarddeR, line 671: Please explain the intention of this class. May be the name is too general:

```suggestion
class FallbackDevicesStore:
```

source/bdDetect.py
```diff
@@ -472,6 +472,10 @@ def getConnectedUsbDevicesForDriver(driver: str) -> Iterator[DeviceMatch]:
 		)
 	)
 	for match in usbDevs:
+		if (match.type, match.id) in DeviceMathcer.fallBackDevices:
+			DeviceMathcer.matches.append(match)
```
LeonarddeR, line 476: This looks like the `matches` list is ever growing.

LeonarddeR (changes requested):

source/brailleDisplayDrivers/hims.py
EdKweon: I fixed it. Thanks.

source/bdDetect.py
EdKweon: Thanks! I added description of useAsFallBack

source/brailleDisplayDrivers/hims.py
EdKweon: I fixed this problem. Thanks.

source/bdDetect.py
EdKweon: Okay. I fixed it with clear the mathces list before iteration.

LeonarddeR: Where are you doing that exactly, or wasn't that yet committed?

EdKweon: Sorry.  I forgot to push that code.

LeonarddeR (changes requested):
It was a bit difficult for me to see the changes in perspective, so I checked out the branch locally. I'm afraid I still have several concerns:

1. I actually think that storing the fallback matches on `FallbackDevicesStore` is not necessary, as the list of fallback matches is never leaving `getConnectedUsbDevicesForDriver`. So you can use a matches list that is local to the `getConnectedUsbDevicesForDriver` function.
2. You don't associate the driver to the fallback devices. This can become pretty problematic if one decides to register a device identifier as fallback that is shared by many braille displays, such as the FTDI controller with VID_0403&PID_6001. 
3.  Aare you sure that the current code actually works as advertised? The `getDriversForConnectedUsbDevices` function doesn't seem to care about fallback devices at all.

Personally I'd do something like the following:

1. Change `DeviceMatch` class to become a data class, add an extra fallback property to it. Ensure that it implements the `collections.abc.Sequence` and that the sequence methods exclude the new fallback property. The latter is to keep backwards compatibility.
2. Now, just treat device matches that have the fallback property as fallback devices (i.e. yield them last)

Every alternative I can think of has different disadvantages. Actually, I think Hims should just solve this in the hardware themselves...

EdKweon:
@LeonarddeR 
Okay. I rollback fallback code. Check code and PR again Please. I think above problem needs more time, If necessary, it might be better to create another PR and discuss it there. Thanks

EdKweon:
@LeonarddeR 
I have a question.
 - Now, just treat device matches that have the fallback property as fallback devices 
 I can't understand where to set fallback propery true. Can you tell me in my code?

Thnkas

EdKweon:
@LeonarddeR 
I fixed code as you recommend. It works well. Thanks.

LeonarddeR (changes requested):
Thanks for these changes. Unfortunately the points i mentioned earlier still stand:
1. Storing the fallback matches on FallbackDevicesStore is not necessary, as the list of fallback matches is never leaving getConnectedUsbDevicesForDriver.
2. You don't associate the driver to the fallback devices.
3. The getDriversForConnectedUsbDevices function doesn't seem to care about fallback devices. This means that the fallback doesn't work when disconnecting and reconnecting the device.

source/bdDetect.py
```diff
@@ -88,7 +90,8 @@ def __getattr__(attrName: str) -> Any:
 	raise AttributeError(f"module {repr(__name__)} has no attribute {repr(attrName)}")
 
 
-class DeviceMatch(NamedTuple):
+@dataclass
+class DeviceMatch(Sequence):
```
LeonarddeR, line 94: On reflection, I wonder whether this change contributes to a solution. Sorry about that. The entire detection system is not made for this kind of stuff.

EdKweon: @LeonarddeR 
Thank you for comment.
Do you think that the previous code is not suitable? Should we roll it back?

EdKweon: I rollback DeviceMatch code. And make not using that.

EdKweon:
@LeonarddeR 
I can't understand what you said.
- The getDriversForConnectedUsbDevices function doesn't seem to care about fallback devices. This means that the fallback doesn't work when disconnecting and reconnecting the device.
Can you explain with code?

Thanks

LeonarddeR:
> @LeonarddeR I can't understand what you said.
> 
>     * The getDriversForConnectedUsbDevices function doesn't seem to care about fallback devices. This means that the fallback doesn't work when disconnecting and reconnecting the device.
> 
> 
> Can you explain with code?

You need to apply the changes you performed to `getConnectedUsbDevicesForDriver` to `getDriversForConnectedUsbDevices` as well.

EdKweon:
@LeonarddeR 
Thanks. I also applied fallback in getDriversForConnecttedUsbDevices function. I'm really sorry but I have a last question.
- Storing the fallback matches on FallbackDevicesStore is not necessary, as the list of fallback matches is never leaving getConnectedUsbDevicesForDriver.
We have rolled back the parent class of the DeviceMatch class. Here is my understanding of the proposed solution:

- Remove the code that stores FallbackDevicesStore in the addUsbDevices function.
- Remove the FallBackDevicesStore class.
- Create a local matches list in the getDriversForConnectedUsbDevices and getConnectedUsbDevicesForDriver functions. For ----- example, matches = [("Hims", DeviceType.HID, "VID_045E&PID_940A")].
I find this a bit confusing and want to ensure accuracy before modifying the code. Given the rollback of the DeviceMatch class, I would like to confirm this approach once more.

Thank you.

EdKweon:
@LeonarddeR, @seanbudd 
When can I expect feedback? I'm concerned that the PR is taking too long to be merged.

source/bdDetect.py
```diff
@@ -655,3 +686,7 @@ def addDeviceScanner(
 		scanForDevices.register(scanFunc)
 		if moveToStart:
 			scanForDevices.moveToEnd(scanFunc, last=False)
+
+
+class FallbackDevicesStore:
+    fallBackDevices: List[Tuple[str, DeviceType, str]] = []
```
LeonarddeR, line 692: I think I prefer this to become a set:
```suggestion
    fallBackDevices: set[Tuple[str, DeviceType, str]] = set()
```

LeonarddeR (approved):
Looks good to me, just two small things. Could you please adapt the `terminate` function to clear `FallbackDevicesStore.fallBackDevices`

EdKweon:
@LeonarddeR 
> `terminate`
I clear fallback devcies in terminate function.

seanbudd:
@coderabbitai review

seanbudd:
Note @EdKweon - this is an AI review, so it may give unhelpful or irrelevant comments

source/brailleDisplayDrivers/hims.py
```diff
@@ -248,10 +249,13 @@ class BrailleDisplayDriver(braille.BrailleDisplayDriver):
...
-			"VID_045E&PID_940A",  # Braille Edge3S 40
-		})
-
+		driverRegistrar.addUsbDevices(
+			bdDetect.DeviceType.HID,
+			{
+				"VID_045E&PID_940A",  # Braille Edge3S 40
+			},
+			True
+		)
```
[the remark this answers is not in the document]
seanbudd: This is true, but this refactor is out of scope of this PR. If it's something that would be good to tackle separately

source/brailleDisplayDrivers/hims.py
```diff
@@ -497,15 +502,36 @@ def _onReceive(self, data: bytes):
...
+					# Command packets are ten bytes long
+					packet = firstByte + stream.read(9)
+				except Exception:
+					# remaining data will be received next onReceive
+					self._serialData = firstByte
+					return
+			else:
+				packet = self._serialData
+				self._serialData = b""
+
```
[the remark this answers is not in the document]
seanbudd: I agree that this could be refactored and split up into separate methods, @EdKweon what do you think?

source/bdDetect.py
```diff
@@ -153,13 +152,18 @@ def getDriversForConnectedUsbDevices(
...
+	fallbackDriversAndMatches = []
 	for match in itertools.chain(usbCustomDeviceMatches, usbHidDeviceMatchesForCustom, usbComDeviceMatches):
 		for driver, devs in _driverDevices.items():
 			if limitToDevices and driver not in limitToDevices:
 				continue
 			for type, ids in devs.items():
 				if match.type == type and match.id in ids:
-					yield driver, match
+					if (driver, match.type, match.id) in FallbackDevicesStore.fallBackDevices:
+						fallbackDriversAndMatches.append(Set(driver, match))
```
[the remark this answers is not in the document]
seanbudd: @EdKweon - please fix this up

source/brailleDisplayDrivers/hims.py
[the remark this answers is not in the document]
seanbudd: yes please do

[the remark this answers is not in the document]
seanbudd: actually can this be addressed in this PR?

source/bdDetect.py
```diff
@@ -618,6 +648,9 @@ def addUsbDevices(self, type: DeviceType, ids: Set[str]):
 				f"Invalid IDs provided for driver {self._driver!r}, type {type!r}: "
 				f"{', '.join(malformedIds)}"
 			)
+		if useAsFallBack:
+			FallbackDevicesStore.fallBackDevices.update([(self._driver, type, id) for id in ids])
```
LeonarddeR, line 652: @coderabbitai I have the feelling that this can be. It shouldn't have to be a list. Is that correct?
```suggestion
			FallbackDevicesStore.fallBackDevices.update(((self._driver, type, id) for id in ids))

```@coderabbitai Could you suggest how to improve this line?

File: source/brailleDisplayDrivers/hims.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py:534]: @seanbudd 
I agree with spliting it up into spearate methods.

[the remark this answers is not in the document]
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py]: @coderabbitai
I commit the code. Can you check this?

File: source/brailleDisplayDrivers/hims.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py:258]: @seanbudd 
Do I just need to change the HIMs display driver according to the method recommended by the bot?

File: source/brailleDisplayDrivers/hims.py
Comment (EdKweon) [inline source/brailleDisplayDrivers/hims.py:1]: @coderabbitai
I splited methods as you recoomended, but centralizing is quite difficult. Can you check this code?

File: source/brailleDisplayDrivers/hims.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (seanbudd) [reply source/brailleDisplayDrivers/hims.py:258]: yes, I think that would be good thanks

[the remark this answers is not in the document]
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py:258]: @coderabbitai
How can apply default parameter in addUsbDevices method? default value is False.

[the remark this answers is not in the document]
Comment (seanbudd) [reply source/brailleDisplayDrivers/hims.py:258]: Please consider the original suggestion of a refactor, how can this refactor be changed to include the default parameter (note a new Boolean value could be added to the tuples in the sets)

File: source/brailleDisplayDrivers/hims.py
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py:1]: Is this code need to merge _accumulateSerialData method and _processSerialData method? For centrailizing?

File: source/brailleDisplayDrivers/hims.py (hunk 19 above)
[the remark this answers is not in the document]
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py]: @seanbudd
I think this review is done. What do you think?

File: source/brailleDisplayDrivers/hims.py (hunk 18 above)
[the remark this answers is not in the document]
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py:258]: @coderabbitai
Please consider the original suggestion of a refactor, how can this refactor be changed to include the default parameter (note a new Boolean value could be added to the dictionaries)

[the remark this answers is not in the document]
Comment (EdKweon) [reply source/brailleDisplayDrivers/hims.py:258]: @seanbudd 
I think this one is the best.

File: source/bdDetect.py [hunk 22]
```diff
@@ -100,7 +100,6 @@ class DeviceMatch(NamedTuple):
 	deviceInfo: Dict[str, str]
 	"""All known information about a device."""
 
-
```
seanbudd, line 103: please retain this white space

source/bdDetect.py
```diff
@@ -153,13 +152,18 @@ def getDriversForConnectedUsbDevices(
...
+	fallbackDriversAndMatches = []
 	for match in itertools.chain(usbCustomDeviceMatches, usbHidDeviceMatchesForCustom, usbComDeviceMatches):
 		for driver, devs in _driverDevices.items():
 			if limitToDevices and driver not in limitToDevices:
 				continue
 			for type, ids in devs.items():
 				if match.type == type and match.id in ids:
-					yield driver, match
+					if (driver, match.type, match.id) in FallbackDevicesStore.fallBackDevices:
+						fallbackDriversAndMatches.append(set((driver, match)))
```
seanbudd, line 164:
```suggestion
						fallbackDriversAndMatches.append({driver, match})
```

source/bdDetect.py
```diff
@@ -153,13 +152,18 @@ def getDriversForConnectedUsbDevices(
 		for port in deviceInfoFetcher.hidDevices
 		if port["provider"] == "usb"
 	))
+
+	fallbackDriversAndMatches = []
```
seanbudd, line 156: Can you fill out this type hint please

```suggestion
	fallbackDriversAndMatches: list[set[str, DeviceMatch]] = []
```

source/bdDetect.py
```diff
@@ -169,7 +173,13 @@ def getDriversForConnectedUsbDevices(
 		# This ensures that a vendor specific driver is preferred over the braille HID protocol.
 		# This preference may change in the future.
 		if _isHIDBrailleMatch(match):
-			yield (hidName, match)
+			if (driver, match.type, match.id) in FallbackDevicesStore.fallBackDevices:
+				fallbackDriversAndMatches.append(Set(hidName, match))
```
seanbudd, line 177:
```suggestion
				fallbackDriversAndMatches.append({hidName, match})
```

source/bdDetect.py
```diff
@@ -169,7 +173,13 @@ def getDriversForConnectedUsbDevices(
 		# This ensures that a vendor specific driver is preferred over the braille HID protocol.
 		# This preference may change in the future.
 		if _isHIDBrailleMatch(match):
-			yield (hidName, match)
+			if (driver, match.type, match.id) in FallbackDevicesStore.fallBackDevices:
+				fallbackDriversAndMatches.append(Set(hidName, match))
+			else:
+				yield (hidName, match)
```
seanbudd, line 179:
```suggestion
				yield hidName, match
```

source/bdDetect.py
```diff
@@ -471,15 +483,27 @@ def getConnectedUsbDevicesForDriver(driver: str) -> Iterator[DeviceMatch]:
 			for port in deviceInfoFetcher.usbComPorts
 		)
 	)
+
+	fallbackMatches = []
```
seanbudd, line 487:
```suggestion
	fallbackMatches: list[DeviceMatch] = []
```

source/bdDetect.py
```diff
@@ -605,11 +629,17 @@ def _getDriverDict(self) -> DriverDictT:
 			ret = _driverDevices[self._driver] = DriverDictT(set)
 			return ret
 
-	def addUsbDevices(self, type: DeviceType, ids: Set[str]):
+	def addUsbDevices(self, type: DeviceType, ids: Set[str], useAsFallBack: bool = False):
 		"""Associate USB devices with the driver on this instance.
 		:param type: The type of the driver.
 		:param ids: A set of USB IDs in the form C{"VID_xxxx&PID_XXXX"}.
 			Note that alphabetical characters in hexadecimal numbers should be uppercase.
+   		:param useAsFallBack: A boolean flag to determine how USB devices are associated with the driver.
```
seanbudd, line 637: Use tabs not spaces
```suggestion
		:param useAsFallBack: A boolean flag to determine how USB devices are associated with the driver.

```

source/bdDetect.py
```diff
@@ -655,3 +688,7 @@ def addDeviceScanner(
 		scanForDevices.register(scanFunc)
 		if moveToStart:
 			scanForDevices.moveToEnd(scanFunc, last=False)
+
+
+class FallbackDevicesStore:
+    fallBackDevices: set[Tuple[str, DeviceType, str]] = set()
```
seanbudd, line 694: Use tabs not spaces

```suggestion
	fallBackDevices: set[Tuple[str, DeviceType, str]] = set()
```

source/bdDetect.py
```diff
@@ -655,3 +688,7 @@ def addDeviceScanner(
 		scanForDevices.register(scanFunc)
 		if moveToStart:
 			scanForDevices.moveToEnd(scanFunc, last=False)
+
+
+class FallbackDevicesStore:
```
seanbudd, line 693: Is there a reason why `fallBackDevices` is not a module level variable, do you expect more code to be added to `FallbackDevicesStore`?

source/bdDetect.py
seanbudd: can you add a docstring to this please?

source/bdDetect.py
```diff
@@ -605,11 +629,17 @@ def _getDriverDict(self) -> DriverDictT:
 			ret = _driverDevices[self._driver] = DriverDictT(set)
 			return ret
 
-	def addUsbDevices(self, type: DeviceType, ids: Set[str]):
+	def addUsbDevices(self, type: DeviceType, ids: Set[str], useAsFallBack: bool = False):
```
seanbudd, line 632:
```suggestion
	def addUsbDevices(self, type: DeviceType, ids: set[str], useAsFallBack: bool = False):
```

source/brailleDisplayDrivers/hims.py
```diff
@@ -4,6 +4,7 @@
 # Copyright (C) 2010-2023 Gianluca Casalino, NV Access Limited, Babbage B.V., Leonard de Ruijter,
 # Bram Duvigneau
 
+import typing
```
seanbudd, line 7: please import from typing rather than using typing directly

```suggestion
```

source/brailleDisplayDrivers/hims.py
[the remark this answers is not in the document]
seanbudd: can you please implement this suggestion @EdKweon ?

source/brailleDisplayDrivers/hims.py
```diff
@@ -271,13 +275,14 @@ def registerAutomaticDetection(cls, driverRegistrar: bdDetect.DriverRegistrar):
 		)))
 
 	@classmethod
-	def getManualPorts(cls):
-		return braille.getSerialPorts(filterFunc=lambda info: "bluetoothName" in info)
+	def getManualPorts(cls) -> typing.Iterator[typing.Tuple[str, str]]:
```
seanbudd, line 278:
```suggestion
	def getManualPorts(cls) -> Iterator[tuple[str, str]]:
```

source/brailleDisplayDrivers/hims.py
```diff
@@ -291,7 +296,7 @@ def __init__(self, port="auto"):
 					case bdDetect.DeviceType.CUSTOM:
 						# onReceiveSize based on max packet size according to USB endpoint information.
 						self._dev = hwIo.Bulk(port, 0, 1, self._onReceive, onReceiveSize=64)
-					case _:
+					case bdDetect.DeviceType.SERIAL:
```
seanbudd, line 299: please add a new case for `case _:` where an error is raised or logged

source/brailleDisplayDrivers/hims.py
```diff
@@ -489,37 +494,63 @@ def _hidOnReceive(self, data: bytes):
...
-			assert packet[9] == 0xfb # Command End
-			calcCheckSum: int = 0xff & sum(
-				c for index, c in enumerate(packet) if(
-					index != CHECKSUM_INDEX)
-			)
-			assert(calcCheckSum == checksum)
-			self._handlePacket(packet)
 		else:
-			log.debug("Unknown first byte received: 0x%x"%ord(firstByte))
+			log.debug("Unknown first byte received: 0x%x" % ord(firstByte))
```
seanbudd, line 518: Please make this an f-string

source/brailleDisplayDrivers/hims.py
```diff
@@ -489,37 +494,63 @@ def _hidOnReceive(self, data: bytes):
...
+				return False
+		return True
+
+	def _processSerialData(self, firstByte: bytes, stream):
+		# serial data first received
+		if not self._serialData:
+			try:
+				# Command packets are ten bytes long
+				packet = firstByte + stream.read(9)
+			except Exception:
```
seanbudd, line 536: what Exception is expected here? can you be more specific?

seanbudd (changes requested):

source/bdDetect.py
EdKweon: I added white space. Thanks.

source/bdDetect.py
EdKweon: I deleted redundant brace.

source/bdDetect.py
EdKweon: I removed redundant brace. Thanks.

source/bdDetect.py
EdKweon: I changed Set to brace

source/brailleDisplayDrivers/hims.py
EdKweon: I changed as you suggested. Thanks.

source/brailleDisplayDrivers/hims.py
EdKweon: I specified exception. Thanks

source/brailleDisplayDrivers/hims.py
EdKweon: I add log.error for case_ . Thanks.

source/brailleDisplayDrivers/hims.py
```diff
@@ -247,22 +247,14 @@ class BrailleDisplayDriver(braille.BrailleDisplayDriver):
...
-		driverRegistrar.addUsbDevices(bdDetect.DeviceType.HID, {
-			"VID_045E&PID_940A",  # Braille Edge3S 40
-		})
-
-		# Bulk devices
-		driverRegistrar.addUsbDevices(bdDetect.DeviceType.CUSTOM, {
-			"VID_045E&PID_930A",  # Braille Sense & Smart Beetle
-			"VID_045E&PID_930B",  # Braille EDGE 40
-		})
+		device_types = {
```
seanbudd, line 250:
```suggestion
		deviceTypes = {
```

source/brailleDisplayDrivers/hims.py
```diff
@@ -247,22 +247,14 @@ class BrailleDisplayDriver(braille.BrailleDisplayDriver):
...
+			bdDetect.DeviceType.HID: ({"VID_045E&PID_940A"}, True),
+			bdDetect.DeviceType.CUSTOM: ({"VID_045E&PID_930A", "VID_045E&PID_930B"}, False),
+			bdDetect.DeviceType.SERIAL: ({"VID_0403&PID_6001", "VID_1A86&PID_55D3"}, False)
```
seanbudd, line 253: can you please add comments indicating the devices like there were before

source/brailleDisplayDrivers/hims.py
```diff
@@ -247,22 +247,14 @@ class BrailleDisplayDriver(braille.BrailleDisplayDriver):
...
+		}
 
-		# Sync Braille, serial device
-		driverRegistrar.addUsbDevices(bdDetect.DeviceType.SERIAL, {
-			"VID_0403&PID_6001",
-			"VID_1A86&PID_55D3",  # Braille Edge2S 40
-		})
+		for device_type, (ids, useAsFallback) in device_types.items():
```
seanbudd, line 256:
```suggestion
		for deviceType, (ids, useAsFallback) in deviceTypes.items():
```

source/bdDetect.py
seanbudd: please add a docstring here

source/bdDetect.py
seanbudd: can you answer this please?

seanbudd: see auto hidden comments from the previous review

EdKweon:
I added docstring of FallbackDevicesStore class.

source/bdDetect.py
seanbudd: can you answer this please?

seanbudd:
@EdKweon - you may need to uncollapse hidden comments. search the page for "6 hidden comments"

seanbudd:
Alternatively, open the PR files page, select conversations, and check through the unresolved comments

seanbudd:
@EdKweon - this is looking almost ready other than the last comment. can you do a final round of testing?

@coderabbitai full review

source/brailleDisplayDrivers/hims.py
```diff
@@ -489,37 +501,63 @@ def _hidOnReceive(self, data: bytes):
...
+				return False
+		return True
+
+	def _processSerialData(self, firstByte: bytes, stream):
+		# serial data first received
+		if not self._serialData:
+			try:
+				# Command packets are ten bytes long
+				packet = firstByte + stream.read(9)
+			except IOError:
```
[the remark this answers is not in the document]
seanbudd: ignore this @EdKweon

source/brailleDisplayDrivers/hims.py
```diff
@@ -247,22 +247,31 @@ class BrailleDisplayDriver(braille.BrailleDisplayDriver):
...
+				False
+			),
+			bdDetect.DeviceType.SERIAL: (
+				{
+					"VID_0403&PID_6001",
+					"VID_1A86&PID_55D3"  # Braille Edge2S 40
+				},
+				False
+			)
+		}
```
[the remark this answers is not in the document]
seanbudd: ignore this @EdKweon

source/bdDetect.py
```diff
@@ -106,6 +106,11 @@ class DeviceMatch(NamedTuple):
 
 _driverDevices = OrderedDict[str, DriverDictT]()
 
+fallBackDevices: set[Tuple[str, DeviceType, str]] = set()
```
seanbudd, line 109:
```suggestion
fallBackDevices: set[tuple[str, DeviceType, str]] = set()
```

source/bdDetect.py
```diff
@@ -106,6 +106,11 @@ class DeviceMatch(NamedTuple):
 
 _driverDevices = OrderedDict[str, DriverDictT]()
 
+fallBackDevices: set[Tuple[str, DeviceType, str]] = set()
+"""This class is used to store fallback devices.
```
seanbudd, line 110:
```suggestion
"""
Used to store fallback devices.
```

seanbudd (approved):

seanbudd (approved):

seanbudd:
Thanks @EdKweon

PR #16352, opened 2024-04-02
Indicate category and option name in error message when validating settings dialog

CyrilleB79:
### Link to issue number:

None

### Summary of the issue:
The settings dialog contains always more error message boxes appearing when the validation of a parameter fails: modes available in cycle speech mode command or in cycle sound split mode command and regex validation in advanced settings have been added recently.

These messages do not formally mention the panel, nor the option which they are associated with. This may be difficult for the user to know which option is causing the error when the user has been modifying various settings in various setting panels (categories) before validating with "OK" or "Apply".

The validation error messages now indicate the category and the name of the option which is causing the error so that the user can go to this option and modify it.

Create a common dedicated error validation message with appropriate placeholders to be filled.

### Testing strategy:
* [ ]: Manual test of the 4 validation error messages (no NVDA key, customize cycle speech mode, customize sound split mode and paragraph text regexp)
* [ ]: Visual test from sighted people (see below)

#### Visual test
Sighted people collaboration needed please. Cc @Qchristensen, @gerald-hartig

The current aspect of the dialog is as follows:
image

The initial error message is written bigger (12 pt) and in aqua-blue color. The additional information (category and options) is written smaller (9 pt) and in black color.
This difference was not done on purpose and seems due to the empty line that I have added between the initial error text and the additional information.

If I remove this empty line, the dialog would becomes as follows, with all the text written in black, size 9pt:
image

What is the best appearance according to you?
Or should I explore a third solution, e.g. all the text in 9pt black, but with a blank line between first error text and additional information? I do not know if it is easily doable...

None

### Change log
Not needed IMO.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


gerald-hartig:
FWIW I quite like the 12pt blue and 9pt black created by the empty line. It has good visual semantics, better than the other two options.

CyrilleB79:
Thanks @gerald-hartig for your feedback.

source/gui/settingsDialogs.py
```diff
@@ -1719,13 +1745,13 @@ def isValid(self) -> bool:
...
 				# Translators: Message shown when not enough speech modes are enabled.
-				_("At least two speech modes have to be checked."),
-				# Translators: The title of the message box
-				_("Error"),
-				wx.OK | wx.ICON_ERROR,
-				self,
+				message=_("At least two speech modes have to be checked."),
+				# Translators: Same as the label for the list of checkboxes where user can select speech modes that will
+				# be available in Speech Settings, but without keyboard accelerator (& character) nor final colon.
+				option=_("Modes available in the Cycle speech mode command"),
```
lukaszgo1, line 1753: Can't you just retrieve the label from the control and remove the unnecessary characters yourself? These two strings can easily get out of sync otherwise. I've looked at this briefly - it is likely you may need to expand `guiHelper`, so that each labeled control stores its label.

CyrilleB79: > Can't you just retrieve the label from the control and remove the unnecessary characters yourself? These two strings can easily get out of sync otherwise. I've looked at this briefly - it is likely you may need to expand `guiHelper`, so that each labeled control stores its label.

It is not enough to just remove the "&" character from the label. E.g. in Chinese translation; see #15266 and #15258.
Then, the implementation recommended by @seanbudd in ,https://github.com/nvaccess/nvda/issues/15258#issuecomment-1667002771 was to duplicate the strings.
That's why I have followed the same approach in this PR.

seanbudd (approved):
Generally looks good, just thought that we could make the message title more descriptive potentially

source/gui/settingsDialogs.py
```diff
@@ -413,6 +413,32 @@ def isValid(self) -> bool:
...
+				"\n"
+				'Category: "{category}"\n'
+				'Option: "{option}"'
+				).format(
+			message=message,
+			category=category,
+			option=option,
+			),
+			# Translators: The title of the message box
+			caption=_("Error"),
```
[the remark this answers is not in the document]
CyrilleB79: Shouldn't it rather be "Invalid setting" with singular form. Only one invalid setting is reported in each dialog.

Or is "settings" always with a final "s" in English?

@seanbudd, feel free to commit yourself the more suitable proposal.

```suggestion
			caption=_("Invalid setting"),
```

XLTechie:
@CyrilleB79 I was going to suggest "invalid configuration", but "invalid 
setting" is also fine. "settings" is not correct for a single setting.

PR #16364, opened 2024-04-05
Revert "Ignore padding dots at low symbol reporting levels (#16141)"

### LINKED ISSUE #16345: Collapse of more than three dots into "padding dots" symbol cannot be disabled
### Steps to reproduce:
1. Set symbol level to "all".
2. Read the following string with NVDA: ....

### Actual behavior:

NVDA says "padding dots".

### Expected behavior:

NVDA says "dot dot dot dot". This behaviour changed in #16141, and the string "padding dots" is very confusing without the PR context. This collapsing should be disabled at the "all" symbol level (as all symbols aren't reported) or at very least user configurable even if enabled by default.

N/A

### System configuration
N/A

### Other questions
N/A

### Context
See https://github.com/nvaccess/nvda/pull/16141#issuecomment-2015747315

codeofdusk:
### Link to issue number:
Closes #16345.

### Summary of the issue:
In some situations, such as Chinese text (which often uses more than three dots to form an ellipsis) and Windows notifications read with a symbol level of "all", the collapsing of more than three dots into "padding dots" is confusing and unwanted.

Reverts the offending PR, pending a refined, more flexible approach.

Reopens #15845.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/characterProcessing.py
+++ b/source/characterProcessing.py
@@ -1,5 +1,5 @@
 # A part of NonVisual Desktop Access (NVDA)
-# Copyright (C) 2010-2024 NV Access Limited, World Light Information Limited,
+# Copyright (C) 2010-2023 NV Access Limited, World Light Information Limited,
 # Hong Kong Blind Union, Babbage B.V., Julien Cochuyt, Cyrille Bougot
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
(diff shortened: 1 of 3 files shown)


Adriani90:
See https://github.com/nvaccess/nvda/issues/16345#issuecomment-2039283991.

While I understand the annoyance behind hearing "padding dots", this solves a much more annoying use case where people have to hear lots of dots in table of contents. The chinese ellipsis problem is not alone. If you want to solve that problem, see my suggestion in the referenced comment. Afterwards reverting this could be discussed but at this stage I would not advice to revert this.
We could rename "padding dots" to "multiple dots" if it feels better for you.

hwf1324:
I don't think the Chinese use case needs to revert the whole PR, maybe we just need to make adjustments to the number of judgment points. Of course there is no perfect solution, we can only make trade-offs.

XLTechie:
For what it's worth, I'd vote for just saying the number of periods whenever 
there's more than six.

codeofdusk:
> For what it's worth, I'd vote for just saying the number of periods whenever
there's more than six.

Agreed.

Adriani90:
@XLTechie, @codeofdusk in https://github.com/nvaccess/nvda/issues/16345#issuecomment-2039278377 I have tried to bring up the complexity of elipsis. It is not about hearing dot dot dot dot dot dot, actually also not about hearing dot dot dot. You would understand it is an elipsis when you hear "6 dots" or "3 dots". What is the use case for hearing so many dots being reported each?
If there is indeed a valid use case for this, I vote to be optional whether the user hears every punctuation character or the number of characters when more than 2 periods are next to each other.
Even hearing each of 6 dots reported in table of contents is annoying.
This affects also dashes or underlines in email signatures. Voice over or Jaws count them as well as far as I know.

XLTechie:
Counting them has been standard since at least the 90s. For example, the old IBM 
Screen Reader I had in 1993, would say "All 16 dots" in a case of 16 periods. 
(The word "all" was thought to be a good trigger word for some strange reason)

So I prefer counting and announcing the number.

For three or six dots, I might suggest "Ellipsis" for three, and "double ellipsis" 
or "ellipsis ellipsis" or "2 ellipsis" for six in English.

But this must be translation aware, because in China the six dots is just an 
ellipsis I guess.

So yes, I agree that even hearing six dots in a row would be annoying if they 
were all spoken individually.

Adriani90:
There is not an obvious community support for this revert. Hearing each of the dots would introduce back a far more annoying issue.
I am closing this pr. Let‘s continue the discussion in the referenced underlying issue on  an alternative solution.

PR #16369, opened 2024-04-06
Give ui.browseableMessage the ability to show copy and close buttons after the message, and add buttons to some instances

### LINKED ISSUE #14641: Report "Press escape to close" on NVDA dialogs which the user needs to press escape to close

XLTechie:
### Link to issue number:

Fixes #14641 

### Summary of the issue:

@Qchristensen reported that some users are confused by browseableMessages, and their lack of definite closure mechanisms.
During the conversation, it was pointed out that in some cases, a user might also desire a copy button.

Added copy and close buttons to some browseableMessages.

* Thanks to @michaelDCurran's recent `browseableMessage` change--namely adding a `Scripting.Dictionary` to carry arbitrary query-string-equivalent  style parameters to the `mshtml` instance behind `browseableMessage`, it is now possible to pass in values for two new buttons, and various translatable messages, without any contortions.
* Tried different methods in the JS and CSS of `message.html`, and eventually settled on one which displayed the two buttons side-by-side, under a separator. If neither button's label is supplied, the div containing the HR and buttons remains hidden.
* The "Copy" button can be activated with `Alt+C`, and the "Close" button by `Escape`. `Alt+C` is indicated to the user via an accessibility label.
* While the description of the key has been made translatable, the key itself can not be changed currently.
* Fixed up some docstring parameter listings to match modern format.
* Per the issue, modified the Report Link Destination message, to have both a copy button and a close button.
* Per the issue, added a close button to the Report Character Information message.
* Per request in PR comments, used a live region to present a "Copy complete" or "copy failed" message to the user. This message remains on screen for five seconds.
* The AI suggested covering for more failures in the initialization and configuration of MSHTML. I used its suggested exception checking, but added a private function to display an error message to the user when one of the components fail, so at least some kind of user notice can be given. I based it on the already extant private function for messaging the user if a `browseableMessage` is called on a secure screen.
* Added an immediate return if the MSHTML window is opened without providing the dialog arguments (i.e. the message and title, at least). That situation should never actually occur, but in the unlikely event that a failure passes the error checking we have in `browseableMessage`, this will return immediately, instead of stranding the user in a blank window with no obvious close mechanism.
* Tightened the handling of text vs. HTML, to possibly prevent some hypothetical cases of raw HTML being given to the browser instance when we don't expect that to be the case. In particular, messages are now always assumed to be text by the javascript processing, unless a specific flag is passed via the `scriptable.dictionary`, instructing the JS to allow the message to be treated as HTML.

### Testing strategy:

Tested using mainly the report link destination window, and a few other core windows.

* In addition to the others, it was requested that the OCR results window provide a close button. This is more difficult, and I'd rather leave it to a separate PR in case it is not desired after all.
* I have not made any docs changes for this. I will look through existing docs and figure out if anything needs to be mentioned at appropriate places, in a follow-up PR, unless @seanbudd wants it here.
* The `Alt+C` key can not currently be changed to another key by keymap translators, because of the esoteric way javascript maps keys to known character sets. I continue to research this but didn't want to delay the PR for it.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

## Summary by CodeRabbit

- **New Features**
  - Added "Copy" and customizable "Close" buttons to message dialogs for enhanced usability.
  - Users can now copy text directly from message dialogs.
  
- **Documentation**
  - Updated user documentation to reflect new "Copy" and "Close" button features in message dialogs.


XLTechie:
@Qchristensen would you mind trying this build, and commenting on the UX?

Specifically, try NVDA+k NVDA+k, and NVDA+f NVDA+f, and observe the buttons which appear in each message.
Please consider them in the light of #14641.

I haven't written dev or user documentation yet, but I want to make sure this is going in a reasonable direction.

wmhn1872265132:
I prefer to change the behavior of copying to copy and exit, generally speaking once the information is copied the window loses its necessity to exist

CyrilleB79:
I am not convinced of the usefulness of this PR. As an experimented user, the change in this PR is a degradation of my personal experience, due to the "select all" issue, i.e. when selecting all, you select both the content and the control buttons.

But I understand that this PR rather targets beginners. Though, it seems to me that there has not been so many reporting from beginners or teachers with the current browseable message. Escape or Alt+F4 are classical commands in Windows, so I guess many beginners learn it quite quickly in any case. Also aren't touch device users accustomated to go to previous object in fow to reach the close button?

In any case, I guess that teachers (@XLTechie, @britechguy, maybe @cary-rowen) are the best persons to indicate if this PR is useful or not; and I will not go against their choice.

I have various remarks though on this PR:
1. The button containing "Press escape to exit" is a strange UX. Either you put a button (or a link)  with only "Exit" or just write "Press escape to exit" in static text, but not both. Did you ever see dialog boxes with a button labeled "Press escape to cancel"?
2. IMO, the user experience should be the same in all browseable messages, at least in NVDA core's ones. I do not know in which case the "Copy" and "Exit" button text should be configurable. Thus, the strings of the buttons should be defined in `ui.browseableMessage` (at least by default), not in the calling function. If we really want these buttons still to be configurable,maybe a simple True/False parameter would be enough?
3. When you press the "Copy" button, nothing is heard. If we keep this "Copy" button, a speech feedback is at least needed.
4. More specifically, I am not convinced of the "Copy" button. In NVDA 2024.1, we can easily copy the text with control+A and control+C. With this PR we add an additional footer with an "Exit" button which degrade the copy experience. Thus we need to add one more button which loads more this windows. I have the feeling that we have not found the good solution. Is there an alternative solution? E.g. a description of the document labelled "Press escape to exit".

Sorry Luke for this quite negative feedback, but I thought it was worth sharing.

XLTechie:
> I am not convinced of the usefulness of this PR. As an experimented user, the change in this PR is a degradation of my personal experience, due to the "select all" issue, i.e. when selecting all, you select both the content and the control buttons.

I believe that is why @nvdaes proposed the "copy" button in the original issue. The copy button specifically avoids copying the footer section.
For clarification: what is the purpose of selecting all and copying, if you can use the copy button provided? I think we could put a `Ctrl+c` keyboard shortcut on it if needed.

> But I understand that this PR rather targets beginners. Though, it seems to me that there has not been so many reporting from beginners or teachers with the current browseable message.

@Qchristensen originally raised this. That was my only context; I am assuming he, as NV Access, had sufficient complaints to open an issue.

I think the debate is supposed to be done in the original issue. Certainly the context is there.

>  1. The button containing "Press escape to exit" is a strange UX. Either you put a button (or a link) with only "Exit" or just write "Press escape to exit" in static text, but not both. Did you ever see dialog boxes with a button labeled "Press escape to cancel"?

This is a rip-off of Jaws, and in part a response to the request in the original 
issue. It would not bother me for this to be just "Close", leaving the escape 
part out entirely. This was in part why I requested a pre-review from 
@Qchristensen. I also do not care for this text.

The Jaws UI here, I believe, is just a text line saying "Press escape to close", that for some reason acts as a close button if you happen to interact with it. I could have done that here (with a clickable), but I thought people would find that as objectionable as you find this.
Jaws doesn't provide a copy button, and I often see their closing line show up in people's copied text from these Windows, which Jaws uses for more purposes I believe.

>     2. IMO, the user experience should be the same in all browseable messages, at least in NVDA core's ones.

I agree in general, and can easily do that. At this stage of the draft PR, I wished to demonstrate both possibilities.

However, add-ons may have other requirements, and I didn't want to insist that they have a copy button.

> I do not know in which case the "Copy" and "Exit" button text should be configurable. Thus, the strings of the buttons should be defined in `ui.browseableMessage` (at least by default), not in the calling function.

My original reason for that, along with hesitating to insist that add-ons adopt the core phrasing, was because I originally planned to implement this PR just to add the functionality to `browseableMessage`, without including any uses for it in core. Then to add the implementation for various dialogs in a later PR. Again though, I would be fine if these were "copy" and "close".

>     3. When you press the "Copy" button, nothing is heard. If we keep this "Copy" button, a speech feedback is at least needed.

Agreed.

>     4. More specifically, I am not convinced of the "Copy" button. In NVDA 2024.1, we can easily copy the text with control+A and control+C. With this PR we add an additional footer with an "Exit" button which degrade the copy experience. Thus we need to add one more button which loads more this windows.

"Copy" exists to enable copying without inclusion of the footer. That may be what you're saying there, the translation is unclear of "which loads more this windows", although I think it means that it adds more content to the window. If that is what you meant, then the reason is to provide a copy function that doesn't capture the footer, which is what I got from https://github.com/nvaccess/nvda/issues/14641#issuecomment-1443490395

> I have the feeling that we have not found the good solution. Is there an alternative solution? E.g. a description of the document labelled "Press escape to exit".

"Press escape to exit" could be added to the title bar of every message.
* I'm not sure how useful that is given that some title bars may already be quite long (if they contain the name of a link, for example), and visually I am not sure the title bar is even shown.
* That also still leaves us with the problem of touch users, which @josephsl raised originally.
* Lastly, @michaelDCurran preferred the footer close button in the original issue.

I am honestly not sure there is a "good" solution to this.
Actually, there is: replace everything currently shown in a `UI.browseableMessage`, as a real dialog in an appropriate `WX` control. Then provide the close button only. That should solve the select all problem, if it's 
a read-only editable text field.
`ui.browseableMessage` is a convenient, but ultimately lazy, way to get a non-modal dialog with HTML side-benefits.

XLTechie:
@CyrilleB79 did you notice my comment above? https://github.com/nvaccess/nvda/pull/16369#issuecomment-2042895733 I would be interested in your further comments.

CyrilleB79:
Re all the comments in https://github.com/nvaccess/nvda/pull/16369#issuecomment-2042895733:

I have the feeling that we did not investigate enough #14641, i.e. why some people may have difficulty with the browseable message. Are there many such people? What is the profile of these people? Have these people switched recently from Jaws and do they expect the same from NVDA?
@Qchristensen, can you provide answers to these questions?

It seems to me that almost every blind windows user know that pressing Alt+F4 or Escape in any window or dialog allows to close it; a user that would not be aware of this is lacking important information and thus would need extra teaching.

My personal need for select all is only to copy the content. So I agree that the "Copy" button does the same thing.

Why I am a bit reluctant to the change in this PR is that it seems to me a UX degradation. The browseable message becomes heavier (in term of UX) with the Close button, and moreover adding the Close button causes a new issue, which is resolved by adding the Copy button, which makes this browseable message more heavy and complex. But people have not expressed difficulty in copying the content of current version of the browseable message.

Also, I understand why the buttons (presence or not and label) may need to be configurable. However, this configurability will make the experience of browseable message less unified, which is a UX degradation.

The wx dialog is not an option due to the `isHtml` option, except if you remove this possibility which is not used in NVDA's core in any case.

Bu if I am the only one feeling that it's a UX degradation, go ahead anyway. Especially if @Qchristensen confirms the "go" for this PR.

Thanks for having changed the label of the "Close" button. The speech feedback of the "Copy" button is still needed. And also, what is copied when `isHtml` is True?

XLTechie:
@CyrilleB79 I just want to say that:

1. Most dialogs in NVDA that are real dialogs (actually all, that I can think of), have some sort of active close feature, be it an OK button, a Close button, or such. So this is not so abnormal. The average user doesn't know that this is not a dialog. I think that may be where some of the confusion comes from.

2. The close text does not have to be configurable. I did that for add-ons, but I don't need to keep it. I have already removed it for the Copy button.

3. I didn't yet implement the message when copy is performed, awaiting Quentin's pre-review before spending any more time on this.

Adriani90:
In general this is not a dialog, and in fact it is even not visible on the screen as far as I know. NVDA reports "document" when opening them, and not "dialog" as it does when opening settings dialogs. I agree with @CyrilleB79's doubts on the usefullness of this PR.
I would rather vote to add some lines to the user guide explaining the difference between a dialog and a virtual browseable document.

@XLTechie wrote:
> The average user doesn't know that this is not a dialog. I think that may be where some of the confusion comes from.

I think it would make more sense to tell users that there is a difference instead of changing the interface so the users don't notice a difference.

In NVDA terms virtual browseable documents are commonly used in many discussoins, so it is ok in my view not to treat them as if they were dialogs.

XLTechie:
@Adriani90 wrote:

> In general this is not a dialog, and in fact it is even not visible on the screen as far as I know.

It is absolutely visible on the screen.

Adriani90:
Even the OCR window?Von meinem iPhone gesendetAm 15.04.2024 um 12:46 schrieb Luke Davis ***@***.***>:﻿
@Adriani90 wrote:

> In general this is not a dialog, and in fact it is even not visible on the screen as far as I know.

It is absolutely visible on the screen.

XLTechie:
@Adriani90 wrote:

> Even the OCR window?

no, but that is a custom dialog, not an mshtml document. Not applicable to this 
conversation.

seanbudd:
Is this ready for review?

XLTechie:
@seanbudd back in April, I asked for comments from @Qchristensen prior to code review, as I believe he requested the original feature, which went through a few evolutions in style requested before implementation.

Not sure if you should review it before Quentin's comments on UX, as I might need to make changes.

Especially if he prefers the "copy then close" alteration someone proposed above.

Qchristensen:
Ok firstly @XLTechie, @Adriani90 & everyone, my apologies for the delay replying.  I got behind on issues over the last few months, and I am working on catching up!

Looking at the try build, and having just caught up on all the comments.  I like the NVDA+f functionality.  Just to save anyone needing to go back and test the try build, the behaviour is:
NVDA+F twice: Dialog has text in a read-only edit, with 'copy' and 'close' buttons.  If you press escape, the dialog closes, if you move to the close button and activate it, the dialog closes.  There is also an 'X' close button up the top-right for touch users (they could also use the close button).

NVDA+k twice: dialog is essentially the same, but the close button reads "Press escape to close".  I think just close is fine here.

Re the question of copying the information, as someone noted, if you press control+a, then control+c, you DO get the button text as well as the formatting (or link) information.  Whether that will annoy more users who expect to be able to press control+a then control+c, vs those who instead tab to the copy button and use that is a good question it's probably hard to know without having people test it in the wild.

The original idea behind "Press escape to close" was that it was read out automatically, and from when the dialog did not have any buttons.  From memory, I only had one user ask about this, but the confusion seemed like it could feasibly more widespread, so I created the issue - essentially when you press NVDA+f once NVDA reads the information.  When you press it twice, NVDA reads it and you can now navigate it, but it may not be 100% clear that you are now in a dialogue.  I can say since I created the issue, that I haven't had anyone else come to me confused.  Having created the buttons may make it easier to add additional functionality down the track (I can't think of anything else which might be useful to add here).  But whether the lack of additionally confused users pushes towards this not being such an important issue to solve at all?

Two additional mostly unrelated visual observations:
1) the dialog which is presented is about three times higher than needed.  That is nothing to do with this PR, it's how the dialogs have previously been presented (I created an issue for that - #16727)

2) When you do select all, only the text is highlighted with NVDA's focus highlight.  A user with enough sight to see that may not expect the button text to be copied as well.

source/message.html
```diff
@@ -9,19 +9,51 @@
...
+function onCopyButtonPress() {
+	// Copy code came from http://www.codestore.net/store.nsf/unid/DOMM-4QHQE8/
+	var rng = document.body.createTextRange();
+	rng.moveToElementText(messageID);
+	rng.scrollIntoView();
+	rng.select();
+	rng.execCommand("Copy");
+	rng.collapse(false);
+	rng.select();
+}
```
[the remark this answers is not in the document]
XLTechie: @coderabbitai can you suggest an implementation for this, specifically including alerts for success ("Copied to clipboard"), and failure ("Couldn't copy to clipboard")?

source/message.html
```diff
@@ -9,19 +9,51 @@
...
+
+function onCloseButtonPress() {
+	//window.open('', '_self', '');  // Only enable if window.close() stops working in a future version, may fix.
+	window.close();
+}
```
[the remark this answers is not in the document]
XLTechie: This shouldn't be able to fail in the mshtml context. Therefore errorhandling here isn't needed.

source/message.html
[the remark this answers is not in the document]
XLTechie: can you suggest an implementation for this, specifically including alerts for success ("Copied to clipboard"), and failure ("Couldn't copy to clipboard")?

source/globalCommands.py
```diff
@@ -2391,7 +2390,8 @@ def _reportFormattingHelper(self, info, browseable=False):
 			ui.browseableMessage(
 				message,
 				# Translators: title for formatting information dialog.
-				_("Formatting")
+				_("Formatting"),
+				copyButton=True, closeButton=True
```
seanbudd, line 2394:
```suggestion
				copyButton=True,
				closeButton=True,
```

source/globalCommands.py
```diff
@@ -4089,7 +4089,8 @@ def script_reportLinkDestination(
 					linkDestination,
 					# Translators: Informs the user that the window contains the destination of the
 					# link with given title
-					title=_("Destination of: {name}").format(name=obj.name)
+					title=_("Destination of: {name}").format(name=obj.name),
+					closeButton=True, copyButton=True
```
seanbudd, line 4093:
```suggestion
					closeButton=True,
					copyButton=True,
```

user_docs/en/changes.md
```diff
@@ -43,6 +43,7 @@ Unicode CLDR has also been updated.
 * In the Python console, the last unexecuted command will no longer be lost when moving in the input history. (#16653, @CyrilleB79)
 * A unique anonymous ID is now sent as part of optional NVDA usage statistics gathering. (#16266)
 * By default, a new folder will be created when making a portable copy. Warnings have been added when writing to a non-empty directory. (#16684)
+* The Report link destination,  Character formatting information, and Speak selection windows, now include "Close" and "Copy" buttons for user convenience. (#16369, @XLTechie)
```
seanbudd, line 46:
```suggestion
* The Report link destination, Character formatting information, and Speak selection dialogs, now include "Close" and "Copy" buttons for user convenience. (#16369, @XLTechie)
```

source/ui.py
```diff
@@ -89,12 +89,20 @@ def _warnBrowsableMessageNotAvailableOnSecureScreens(title: Optional[str]) -> No
 	)
 
 
-def browseableMessage(message: str, title: Optional[str] = None, isHtml: bool = False) -> None:
+def browseableMessage(
+		message: str,
+		title: Optional[str] = None,
```
seanbudd, line 94:
```suggestion
		title: str | None = None,
```

source/message.html
```diff
@@ -4,24 +4,66 @@
...
+			copyButtonSpan.style.display = 'inline';  // Display the button
+			buttonDiv.style.display = 'block';  // Display the div
+		}
 	}
 } 
 //-->
 </SCRIPT>
 </HEAD>
-<BODY tabindex='0' id='main_body' style="margin:1em" LANGUAGE=javascript onload="return windowOnLoad()" onkeypress="return escapeKeyPress()" >
+<BODY tabindex='0' id='main_body' style="margin:1em" LANGUAGE=javascript onload="return windowOnLoad()" onkeypress="return handleKeyPress()" >
```
seanbudd, line 62:
```suggestion
<body tabindex='0' id='main_body' style="margin:1em" LANGUAGE=javascript onload="return windowOnLoad()" onkeypress="return handleKeyPress()" >
```

source/message.html
```diff
@@ -4,24 +4,66 @@
 <TITLE></TITLE>
 <SCRIPT LANGUAGE=javascript>
 <!--
-function escapeKeyPress(e) {
+var args = window.dialogArguments;
```
seanbudd, line 7:
```suggestion
const args = window.dialogArguments;
```

source/message.html
```diff
@@ -4,24 +4,66 @@
...
 	if(e.keyCode == 27){ // code for escape
 		window.close();
+	} else if (e.altKey && e.keyCode === 67) {  // Code for Alt+c for copy button
+		onCopyButtonPress();
+	}
+}
+
+function onCopyButtonPress() {
+	// Copy code came from http://www.codestore.net/store.nsf/unid/DOMM-4QHQE8/
+	var rng = document.body.createTextRange();
```
seanbudd, line 20:
```suggestion
	const rng = document.body.createTextRange();
```

source/message.html
```diff
@@ -4,24 +4,66 @@
...
+	rng.moveToElementText(messageID);
+	rng.scrollIntoView();
+	rng.select();
+	var success = rng.execCommand("Copy");
```
seanbudd, line 24:
```suggestion
	const success = rng.execCommand("Copy");
```

source/message.html
```diff
@@ -4,24 +4,66 @@
...
+			alert(args.item('copySuccessfulAlertText'));
+		} else {
+			alert(args.item('copyFailedAlertText'));
+		}
 	}
-};
+}
+
+function onCloseButtonPress() {
+	//window.open('', '_self', '');  // Only enable if window.close() stops working in a future version, may fix.
```
seanbudd, line 37: can you explain why this may stop working?

source/message.html
```diff
@@ -4,24 +4,66 @@
...
+	rng.scrollIntoView();
+	rng.select();
+	var success = rng.execCommand("Copy");
+	rng.collapse(false);
+	rng.select();
+	if (args) {  // Notify the user about the copy result
+		if (success) {
+			alert(args.item('copySuccessfulAlertText'));
+		} else {
+			alert(args.item('copyFailedAlertText'));
```
seanbudd, line 31: where is this text coming from? I don't see any other references

seanbudd (changes requested):

source/ui.py
```diff
@@ -123,6 +131,12 @@ def browseableMessage(message: str, title: Optional[str] = None, isHtml: bool =
 		return
 	d.add("title", title)
 	d.add("message", message)
```
XLTechie, line 133: Adding this, although I can not think of circumstances where MSHTML would be unable to copy to clipboard, in the contexts where we allow it to run.

```suggestion
	d.add("message", message)
	# Translators: A notice to the user that the copy operation succeeded.
	d.add("copySuccessfulAlertText", _("Text copied."))
	# Translators: Notice to the user that the copy operation failed.
	d.add("copyFailedAlertText", _("Couldn't copy to clipboard."))
```

source/message.html
XLTechie: Fixing in suggestion https://github.com/nvaccess/nvda/pull/16369/files#r1654532903

Meant to add that over the weekend, but I got called away unexpectedly and am only now back.

source/message.html
XLTechie: I would have to dig back into some old research, but I remember coming across something on StackOverflow, that suggested a potential failure mode could exist. I suspect it would only be in the case of MS changing strictness or some other aspect of the Javascript engine of MSHTML, which is unlikely at this point. But it seemed like a harmless note to leave as an improbable just in case.
Happy to take it out if you prefer.

XLTechie:
@Qchristensen, a question:

>     2. When you do select all, only the text is highlighted with NVDA's focus highlight.  A user with enough sight to see that may not expect the button text to be copied as well.

This should be a normal browse mode select. If you do a select all in browse mode on a web page in FF or Edge, is the same problem extant?

It might be possible to detect a select all via javascript, and change the styling of the buttons to make them look selected. But given that this is browse mode, I'm not sure the browser instance even knows about the selection state of various elements.

XLTechie:
This is not working currently; I'll figure out why when I get back tonight. Problem in the JS.

source/message.html
seanbudd: It would be great to add the StackOverlow link if you can find it.

XLTechie:
@coderabbitai full review

source/message.html
```diff
@@ -4,24 +4,72 @@
...
 <SCRIPT LANGUAGE=javascript>
 <!--
-function escapeKeyPress(e) {
+// Though these should be constants, MSHTML does not support const as of 2024
+var args = window.dialogArguments;
+var copyStatusTimeout = 5000;  // milliseconds before copy status message goes away
+
+function handleKeyPress(e) {
 	e = e || window.event;
 	if(e.keyCode == 27){ // code for escape
```
XLTechie, line 13: @coderabbitai could you suggest an additional condition to this if statement, which will detect if the user presses `Alt+C`, and run the `onCopyButtonPress()` function if so? It must work in the context of MSHTML (Internet Explorer 10).

XLTechie: @coderabbitai that is the identical code I used in commit 6ee880cd224ca13eed671a8ff0c63549afcb93a1. It does not work to detect Alt+C in MSHTML; a Windows error sound is played if the key sequence is pressed, and it is not passed to the javascript. Can you suggest an alternative?

source/message.html
```diff
@@ -4,24 +4,72 @@
...
+			copyStatusDiv.innerHTML = '<p id="copyStatusText"><strong>' + args.item('copySuccessfulAlertText') + '</strong></p>';
+		} else {
+			copyStatusDiv.innerHTML = '<p id="copyStatusText"><strong>' + args.item('copyFailedAlertText') + '</strong></p>';
+		}
+		// Time out the user alert message
+		setTimeout(function() {
+			copyStatusDiv.innerHTML = '';
+		}, copyStatusTimeout);
+	}
+}
```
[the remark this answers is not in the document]
XLTechie: @coderabbitai can you provide commitable suggestions for the error handling code you proposed?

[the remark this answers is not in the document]
XLTechie: @coderabbitai the user feedback is already being handled by using a live region. That is a superior user experience to that provided by the javascript alert function in this situation.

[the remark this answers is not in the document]
XLTechie: > Learning: When suggesting improvements to JavaScript functions in the NVDA codebase, include error handling and immediate user feedback using `alert` for better user experience.

Wow. Not only is that exactly contrary to what I said, it's directly contra to what the bot itself said earlier in the same comment. It's really started going off the rails in this conversation.

source/message.html
```diff
@@ -4,24 +4,72 @@
...
+			closeButton.style.display = 'inline';  // Display the button
+			buttonDiv.style.display = 'block';  // Display the button section
+		}
+		// If caller wants a copy to clip button
+		if (args.item('copyButtonText') != null) {
+			copyButton.innerHTML = args.item('copyButtonText');  // Assign the (translated) label
+			copyButtonSpan.style.display = 'inline';  // Display the button
+			copyStatusDiv.style.display = 'block';  // Display the copy status
+			buttonDiv.style.display = 'block';  // Display the section
+		}
```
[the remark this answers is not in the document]
XLTechie: @coderabbitai these variables are either already checked for by the code in the browseableMessage function, or the likelihood is that they will never be missing. I do not think error checking is necessary for these conditions.

XLTechie:
@Qchristensen, @seanbudd, The AI suggested that we add Close and Copy buttons for some additional uses of browseableMessage.
I was aware of those, but had initially planned to do these in a separate PR, so we could get some feedback on whether they should have the buttons, and so rolling them back wouldn't un-do this one.
If you would prefer, however, I can do them here.

##### Effected:

* Apparently the Poedit AppMod uses one. (@LeonarddeR?)
* Word and Excel use a few of them.

###### Detailed file listings:

* source/appModules/poedit.py: line 30
* source/NVDAObjects/window/winword.py: line 5
* source/NVDAObjects/IAccessible/winword.py: line 20
* source/NVDAObjects/UIA/wordDocument.py: line 5
* source/NVDAObjects/UIA/excel.py: line 5
* source/NVDAObjects/window/excel.py: line 10

XLTechie:
@Qchristensen, @CyrilleB79, others: from the PR comment:
> The "Copy" button can be activated with `Alt+C`, and the "Close" button by `Escape`. However these keys are not indicated to the user in any way.

Any ideas what user experience might be desirable here? We could underline the "C" of "Copy", which I believe is the form used for other Alt plus keys, but even if I'm right about that, there's nothing I can do for `Escape` in that regard.

britechguy:
Luke Davis wrote, in part: "Any ideas what user experience might be
desirable here? We could underline the "C" of "Copy", which I believe is
the form used for other Alt plus keys, but even if I'm right about that,
there's nothing I can do for Escape in that regard."

The underlining of the C for Copy on a button is indeed an age-old Windows
convention for button controls, and the "Alt +" part goes without saying
for most experienced Windows users, particularly those who have a
keyboard-centric focus.

Escape has been a "often, but not always" method to close almost any
dialog/pop-up you can name, and I have never seen that "officially
documented" anywhere.  The key name itself explains what it is supposed to
trigger, an escape, if such can be triggered by it.  I've actually used
Escape to close lots of things for decades even when working in "point and
click" mode.

It is only reasonable to presume that most users will not be complete
Windows neophytes and that certain conventions that have been in use for a
very, very long time are already known and understood, including the
shorthands to indicate "Alt +" codes for buttons via underlining the
character to be struck.  It should, of course, also result in an
announcement from the screen reader, e.g., "Close button, Alt + C".

XLTechie:
> It should, of course, also result in an announcement from the screen reader,

Which, in this case, it won't. Hence my question.

I think I may try putting an aria-label on it. Not sure if MSHTML respects that, but we shall see. Or not see, hopefully.

Also, I have just realized that once the word "Copy" gets translated, Alt+C is probably going to stop making sense.
I have an idea on how to deal with that, but it's annoying.

seanbudd:
@XLTechie - I'm happy either way with adding more usages of the buttons. Ones that are likely controversial may be best kept to a subsequent PR.

XLTechie:
@Qchristensen Can you please take a look at https://github.com/nvaccess/nvda/pull/16369#issuecomment-2191355897, and also comment upon, using the latest build:

* The alignment of the buttons.
* The look of the message which appears when the copy button is pressed.

XLTechie:
@britechguy, @Qchristensen  I can not find a convenient way to underline the "c" in "copy", unfortunately, since it's a translatable string without the ability to change the key when translated.
I may work on this in the future, but javascript makes it non-trivial.

I do now have an accessibility label set up to announce the translated version of "Alt+C", along with the button, without actually displaying it. I'll see how well that works when the current build finishes, but it is likely as far as I can take this for now.

source/message.html
XLTechie: I have come to question the currentness of this comment after further research. Removed.

LeonarddeR:
I like this pr, but there is a lot of nasty stuff going on because we insist on supporting MSHTML.
I wonder, are you aware of the fact that wx python seems to support the Edge Webview2 control? May be we should consider supporting that, using the old urlmon based control as a fallback, eventually facing it out when dropping support for Windows 8.1?

XLTechie:
@LeonarddeR

> I like this pr

Thanks. :)

> but there is a lot of nasty stuff going on because we insist on supporting MSHTML.

Tell me about it! This was going to be a quick solve. About 80% of the time I spent on it, was researching workarounds for ridiculous MSHTML edge cases.

> I wonder, are you aware of the fact that wx python seems to support the Edge Webview2 control? May be we should consider supporting that, using the old urlmon based control as a fallback, eventually facing it out when dropping support for Windows 8.1?

I thought it might. Investigating switching us over to that has been on my to-do 
list since I realized 24 hours ago that mshtml only partially supports aria-labelledby. That was the last straw.

But this was pretty much done, so I wanted to ship it before delving into a complete re-engineering.
It already took way way longer than I expected because of my schedule, and all the IE compatibility crap.

XLTechie:
@LeonarddeR The other thing I would like to do long term, is move many of these use cases out of webviews entirely.

Most of the common ones (link destination viewer, character information, the Office usages) don't need HTML support.

They would be fine with modeless dialogs. But we don't have a standard in core for modeless dialogs, as we do with, for example, gui.message.messageBox.

There is an issue somewhere for that, but once it finally happens I would like to convert most of these to dialogs. We would still have to keep browseableMessage around, of course, for the few cases that don't fit, and add-ons, but most of the time I don't think we need to be using it.

XLTechie:
@seanbudd What do I need to do to get the remaining two checks to either finish, or start over? It's been two days since the last commit, and it seems that if they were going to finish running, they would have done so.

seanbudd:
I think you may need to resolve merge conflicts, or close re-open the PR

XLTechie:
@seanbudd Okay, I've fixed the merge conflict caused by the CLI fixes PR.

CodeQL is still "expected", which doesn't usually take that long (I think), and Analyze (python) is still "expected". I don't even know where that one is coming from, so no idea how long it is supposed to take.

I guess if those two still aren't done by the time AppVeyor is, I'll close-open.

It's ironic that the external process AppVeyor can give a "pending" status, but the GitHub internal options can't report work in progress and just indicate "expected".

XLTechie:
@seanbudd At this point: CodeQL and Analise are still indefinitely expecting, even after a close-open. Assign milestone has failed outright, although I suppose that can be investigated. No idea how to progress this though.

source/message.html
```diff
@@ -4,24 +4,82 @@
...
+}
 
 function windowOnLoad() {
-	var args = window.dialogArguments;
 	if (args) {
 		document.title = args.item('title');
 		messageID.innerHTML = args.item('message');
+		// If caller wants a close button
+		if (args.item('closeButtonText') != null) {
+			closeButton.innerHTML = args.item('closeButtonText');  // Assign the (translated) label
```
seanbudd, line 52: can this use `innerText` instead? Otherwise RCE injection is possible through translations

Can you check other usages of `innerHTML` to ensure translatable strings cannot inject code?

source/message.html
```diff
@@ -4,24 +4,82 @@
...
+
+function onCloseButtonPress() {
+	window.close();
+}
 
 function windowOnLoad() {
-	var args = window.dialogArguments;
 	if (args) {
 		document.title = args.item('title');
 		messageID.innerHTML = args.item('message');
```
seanbudd, line 49: this may also allow RCE, can this be fixed?

source/message.html
```diff
@@ -4,24 +4,82 @@
...
+	rng.scrollIntoView();
+	rng.select();
+	var success = rng.execCommand("Copy");
+	rng.collapse(false);
+	rng.select();
+	if (args) {
+		if (success) {  // Notify the user about the copy result
+			copyStatusDiv.innerHTML = '<p id="copyStatusText"><strong>' + args.item('copySuccessfulAlertText') + '</strong></p>';
+		} else {
+			copyStatusDiv.innerHTML = '<p id="copyStatusText"><strong>' + args.item('copyFailedAlertText') + '</strong></p>';
```
seanbudd, line 33: can these args be escaped somehow before adding to html?

source/message.html
```diff
@@ -4,24 +4,82 @@
...
+				copyButtonAcceleratorAccessibilityLabel.innerHTML = args.item('copyButtonAcceleratorAccessibilityLabel');
+			}
+		}
 	}
 } 
 //-->
 </SCRIPT>
 </HEAD>
-<BODY tabindex='0' id='main_body' style="margin:1em" LANGUAGE=javascript onload="return windowOnLoad()" onkeypress="return escapeKeyPress()" >
+<body tabindex='0' id='main_body' style="margin:1em" LANGUAGE=javascript onload="return windowOnLoad()" onkeydown="return handleKeyPress()" >
```
seanbudd, line 71: I don't think this language attribute is valid, if anything it should use the NVDA language code

```suggestion
<body tabindex='0' id='main_body' style="margin:1em" onload="return windowOnLoad()" onkeydown="return handleKeyPress()" >
```

source/ui.py
```diff
@@ -103,37 +158,59 @@ def browseableMessage(message: str, title: Optional[str] = None, isHtml: bool =
...
+	# Translators: A notice to the user that a copy operation failed.
+	d.add("copyFailedAlertText", _("Couldn't copy to clipboard."))
+	if closeButton:
+		# Translators: The text of a button which closes the window.
+		d.add("closeButtonText", _("Close"))
+	if copyButton:
+		# Translators: The label of a button to copy the text of the window to the clipboard.
+		d.add("copyButtonText", _("Copy"))
+		# Translators: A portion of an accessibility label for the "Copy" button,
+		# describing the key to press to activate the button. Currently, this key may only be Alt+C.
```
seanbudd, line 194: If this is only the case, should it be translatable?
what can translators change it to?

seanbudd:
I'm starting to think CodeQL checks only run on PRs where the fork base has security checking enabled

source/ui.py
XLTechie: Is the key known as "Alt" the same in all languages? Can all keyboards generate an `Alt+C`, or will some translations need to set it to the empty string because that key can not be used there?

These were my thoughts. I am only familiar with English language and keyboards.
If they are invalid, then I will remove the translatability of that string.
It was the best I could come up with short of making the key remappable, which is impractical as far as I can tell.

seanbudd: it's probably worth noting that it should be set to an empty string if alt+c is not supported in the local keyboard

source/message.html
XLTechie: That's been there for a very long time. No idea when it was valid, if ever.

source/message.html
XLTechie: I will work around their unescapedness by adding them as text, which is hopefully sufficient?

seanbudd: No, you'd need to escape any HTML characters if adding the text using `innerHTML`.
You may need to find a legacy solution for MSHTML JS
https://stackoverflow.com/questions/6234773/can-i-escape-html-special-chars-in-javascript

It might be easier to do it from python with the module `nh3`, if you need to use `innerHTML` for adding translated strings

source/message.html
XLTechie: Not without breaking API.
`ui.browseableMessage` specifically supports the `isHtml` flag, which, if set, passes the message as unaltered HTML by intent.

I have two suggestions to clean this up a little.

1. If `isHtml` is `False`, we can assign this as text only. Not much help as it can be easily gotten around, but it's maybe something.
2. In milestone 2025.1, we can cause `ui.browseableMessage`'s `isHtml` flag's truth to be ignored if running in secure mode, or with add-ons disabled. That way, even if some code tries to pass a script through here, it will render as text once I implement my point 1, which I can do in this PR.

Is that an acceptable solution, or do you have other ideas?

@coderabbitai, can you open an issue covering point 2 above?

source/message.html
XLTechie: Yes. Will handle in upcoming commits.

source/message.html
seanbudd: If isHtml is false, it should already be getting escaped on the python side of things.
Using `nh3` will preserve the HTML syntax that is not a valid XSS threat.
I think we should implement nh3 sanitization now and log a warning if it's set. We can add exceptions to nh3 as necessary if cases arise from add-ons.

source/message.html
XLTechie: @seanbudd I have been able to switch to `.innerText` for everything but the case of the consumer passing in a message it intends to be HTML.
For that case, at this stage, I'm logging a warning on the NVDA side, and otherwise passing it unchanged.

I don't have any objection to using NH3, but I think:
* That would need some implementation discussion as it's got some configurability;
* I would consider it API breaking, since we can't say what behavior add-ons are depending on in raw HTML that they have been passing, that we would now sanitize out; and
* I would prefer it was its own PR.

Particularly in relation to my last point: I think this one leaves us no worse off than we were before, if not better since this has better error handling if nothing else.

XLTechie:
@seanbudd I have updated this and marked it as ready again, pursuant to my comment https://github.com/nvaccess/nvda/pull/16369#discussion_r1708847560
If you disagree about NH3 being a separate action/API breaking, then we can put it back to draft again and I'll work on that.

seanbudd:
@XLTechie - I'll need to discuss internally on whether we want to perform the sanitization in this release.

The vector we're concerned about is NVDA translations, as these are fairly unregulated, translation strings are the only "code" included in NVDA without a direct review from NV Access. If NVDA translations can perform RCE, we have a problem.
Considering no NVDA source code uses `isHTML` currently, this isn't an active vector.
However, if we ever start using `isHTML`, it becomes an active vector, which is something we want to avoid and prevent from becoming a possibility. 

I think we can tackle this in 2025.1 as the sanitization rules might break add-on functionality, and there's no active risk right now.

source/message.html
seanbudd: Thanks, I agree that it is probably out of scope of this PR

XLTechie:
> The vector we're concerned about is NVDA translations, as these are fairly unregulated, translation strings are the only "code" included in NVDA without a direct review from NV Access. If NVDA translations can perform RCE, we have a problem.

Well, I would imagine they could, if we had any HTML based messages in play, which, as you say, we don't. I don't think there are any active proposals that would introduce one, either.

I can even think of a policy based mitigation that would work for the time being, in case some weird need for an HTML message comes up before 2025.1 release.

Specifically, require that all internal HTML messages have only their non-HTML strings translatable, and run those that are through one of the escaping functions.

seanbudd:
@XLTechie I'll write up an issue to do this in 2025.1. I think we could make the sanitization rules for this message as part of our public API so add-ons can adjust them as necessary

seanbudd:
I've opened https://github.com/nvaccess/nvda/issues/16985

source/message.html
```diff
@@ -46,30 +45,39 @@
...
+		window.close();
 	}
 } 
 //-->
 </SCRIPT>
 </HEAD>
-<body tabindex='0' id='main_body' style="margin:1em" LANGUAGE=javascript onload="return windowOnLoad()" onkeydown="return handleKeyPress()" >
-<div id=messageID></div>
+<body tabindex='0' id='main_body' style="margin:1em" onload="return windowOnLoad()" onkeydown="return handleKeyPress()" >
+<div id=messageDiv></div>
```
seanbudd, line 80:
```suggestion
<div id="messageDiv"></div>
```

seanbudd (approved):

SaschaCowley:
Re the "ding" issue, did you try `e.stopPropagation` or `e.preventDefault` in the key handler?

XLTechie:
@SaschaCowley Yes, with no change in outcome. I also tried other ways of picking up the event than the body onKeyDown handler, but was unsuccessful in even receiving the key then, unless I was doing it wrong.
(E.g. an addEventListener on the button.)

PR #16381, opened 2024-04-11
Improve speech output for UIA notifications, honoring all NotificationProcessing_* constants

michaelDCurran:
this is a prototype to see how NVDA's UIA notification support can be improved, both for the bennofit of UIA notifications in general, but also laying groundwork for supporting  the proposed ariaNotify specification allowing web authors to instruct screen readers to speak messages without resorting to live regions: https://github.com/WICG/aom/blob/gh-pages/notification-api.md
 
### Link to issue number:
None

### Summary of the issue:
Currently, NVDA's speech output for UIA notifications is very basic in that it just cancels speech and then speaks the display string. It does not honor the importants, or whether only the most recent notification should be spoken etc.
NVDA should appropriately support all NotificationProcessing_* constants:
* ImportantAll
* ImportantMostRecent
* ImportantCurrentThenMostRecent
* MostRecent
* CurrentThenMostRecent

NVDA now honors the importants of UI Automation notifications sent by applications.

ui.message: add optional arguments:
* messageID: a unique string identify for this message. This could be used to filter or customize the message, and is used with the following  arguments to configure how multiple messages with the same message ID interrupt each other.
    * speakOnlyMostRecent: If a message with the same message ID has already been queued, the already queued message/s are dropped and only the most recent one is spoken.
    * interuptCurrentIfSameID: if speakOnlyMostRecentIfSameID is also true, but this argument is false, the first existing message in the queue is not dropped, thus a full message is always spoken. E.g. a progress bar quickly going from 0 to 100 in jumps of 1, might say "0, 25, 55, 100" rather than just stuttering until it gets to 100.
                                                                                                                                                            
UIA NvDAObject's event_UIA_notify: suitably map the UIA notificationProgress argument to appropriate values for speakOnlyMostRecentIfSameID and interuptCurrentIfSameID. Also map them to an appropriate speech priority. E.g.:
* NotificationProcessing_Important*: priority=NOW
* NotificationProcessing_*ThenMostRecent: speakOnlyMostRecentIfSameID=True
* NotificationProcessing_*CurrentThenMostRecent: interuptcurrentIfSameID=False
* NotificationProcessing_*: priority=NEXT
 
### Testing strategy:
Private testing done, but will be outlined when browser prototypes become public.

this is a prototype draft. It is functional, but the code could be better abstracted.

### Code Review Checklist:

- [ ] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -2463,7 +2463,7 @@ class UIA(Window):
 			speechPriority=speechPriority,
 			messageID=messageID,
 			speakOnlyMostRecentWithSameID=speakOnlyMostRecentWithSameID,
-			interuptCurrentWithSameID=interuptCurrentWithSameID
+			interuptCurrentWithSameID=interuptCurrentWithSameID,
 		)
 
 	def event_UIA_dragDropEffect(self):
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -2440,13 +2440,31 @@ class UIA(Window):
 			messageID = None
 		# Add a new UIA notification processing constant.
 		NotificationProcessing_ImportantCurrentThenMostRecent = 5
-		speakOnlyMostRecentWithSameID = notificationProcessing in (UIAHandler.NotificationProcessing_MostRecent, UIAHandler.NotificationProcessing_CurrentThenMostRecent, UIAHandler.NotificationProcessing_ImportantMostRecent, NotificationProcessing_ImportantCurrentThenMostRecent)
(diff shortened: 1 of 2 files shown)


codeofdusk:
@michaelDCurran Any updates on this PR?

michaelDCurran:
I was very unhappy with the speed of which interupted speech would be truly interupted. There was a great deal of stuttering when being flooded with notifications such as a progress bar increasing. 
I'm going to close this for now, and hopefully I, or someone, might be able to come up with something a little better. I think more limiting of UIA events is necessary, in particular limiting how many can be in NVDA's main event queue at any given time.

PR #16383, opened 2024-04-11
Changed keyboard command for find previous and next search string after searching with NVDA find dialog

### LINKED ISSUE #6499: In Browse mode, an easier key should be bound to find, find next, and find previous.
As discussed on IRC:
Currently, to bring up find, the user has to do finger gymnastics. Several NVDA power users have switched their find commands to / in the input gestures dialog, because it is faster. While it's not perfect because the user can sometimes (if they are an extremely fast typer) type faster than the dialog comes up, it's much easier to use. We also discussed something like [ and ] or ; and ' being next/previous find by default. however we are in a shortage for keys, so maybe nvda+slash for forward and shift+slash or shift+nvda+slash for backward.
cc @tspivey @jcsteh @camlorn

Adriani90:
### Link to issue number:
Fixes #6499 
### Summary of the issue:
The current commands (nvda+f3 and nvda+shift+f3 are too heavy to perform for many users, finger gymnastics is needed.
The users can use now nvda+x or nvda+shift+x.
Advantages:
* The x letter is on most keyboards on the same place, and thus very near to capslock key
* this command does not depend on whether the f keys are assigned to other functions or not, i.e. on Lenovo laptops f keys are used for other features by default such as volume change, brightness change etc.

Changed the kb commands in cursor manager and adjusted docs

### Testing strategy:
Tested running from source in a browser that the command works as expected.

None

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
x  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/cursorManager.py
+++ b/source/cursorManager.py
@@ -215,7 +215,7 @@ class CursorManager(documentBase.TextContainerObject,baseObject.ScriptableObject
 			# Translators: Input help message for find next command.
 			"find the next occurrence of the previously entered text string from the current cursor's position"
 		),
-		gesture="kb:NVDA+f3",
+		gesture="kb:NVDA+x",
 		resumeSayAllMode=sayAll.CURSOR.CARET,
 	)
 	def script_findNext(self,gesture):
@@ -233,7 +233,7 @@ class CursorManager(documentBase.TextContainerObject,baseObject.ScriptableObject
 			# Translators: Input help message for find previous command.
 			"find the previous occurrence of the previously entered text string from the current cursor's position"
 		),
-		gesture="kb:NVDA+shift+f3",
+		gesture="kb:NVDA+shift+x",
 		resumeSayAllMode=sayAll.CURSOR.CARET,
 	)
 	def script_findPrevious(self,gesture):
--- a/user_docs/en/changes.t2t
+++ b/user_docs/en/changes.t2t
@@ -52,6 +52,7 @@ What's New in NVDA
   -
 - Padding dots commonly used in tables of contents are not reported anymore at low punctuation levels. (#15845, @CyrilleB79)
 - Shortcut for finding next and previous search string in a document changed from nvda+f3 and nvda+shift+f3 to nvda+x and nvda+shift+x (#6499, @adriani90)
+-
 
 == Bug Fixes ==
 - Backspace now works correctly when using Nudi 6.1 with NVDA's "Handle keys from other applications" setting enabled. (#15822, @jcsteh)
--- a/user_docs/en/changes.t2t
+++ b/user_docs/en/changes.t2t
@@ -51,7 +51,7 @@ What's New in NVDA
     -
   -
 - Padding dots commonly used in tables of contents are not reported anymore at low punctuation levels. (#15845, @CyrilleB79)
--
+- Shortcut for finding next and previous search string in a document changed from nvda+f3 and nvda+shift+f3 to nvda+x and nvda+shift+x (#6499, @adriani90)
 
 == Bug Fixes ==
 - Backspace now works correctly when using Nudi 6.1 with NVDA's "Handle keys from other applications" setting enabled. (#15822, @jcsteh)
--- a/user_docs/en/userGuide.t2t
+++ b/user_docs/en/userGuide.t2t
@@ -804,8 +804,8 @@ In addition, you can manually force focus mode, after which it will remain in ef
 | Exit focus mode | escape | Switches back to browse mode if focus mode was previously switched to automatically |
 | Refresh browse mode document | NVDA+f5 | Reloads the current document content (useful if certain content seems to be missing from the document. Not available in Microsoft Word and Outlook.) |
 | Find | NVDA+control+f | Pops up a dialog in which you can type some text to find in the current document. See [searching for text #SearchingForText] for more information. |
-| Find next | NVDA+f3 | Finds the next occurrence of the text in the document that you previously searched for |
-| Find previous | NVDA+shift+f3 | Finds the previous occurrence of the text in the document you previously searched for |
+| Find next | NVDA+x | Finds the next occurrence of the text in the document that you previously searched for |
+| Find previous | NVDA+shift+x | Finds the previous occurrence of the text in the document you previously searched for |
 %kc:endInclude
 
 ++ Single Letter Navigation ++[SingleLetterNavigation]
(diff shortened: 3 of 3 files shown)


cary-rowen:
I don't think this would be better, I think F3/Shift+F3 is very intuitive, and adding the NVDA key on top of that means it's an NVDA gesture.
So I think consistency is more important, especially for novices. They already have to memorize many commands. Let's not make it worse, Why not let novices find some memory rules?

You wrote:
> this command does not depend on whether the f keys are assigned to other functions or not, i.e. on Lenovo laptops f keys are used for other features by default such as volume change, brightness change etc.

If so, then other gestures involving the F keys should also be modified, right? It's obviously impossible for us to do this. So far, I have not seen a laptop that cannot switch the function of the F keys. Many laptops I know can restore the function of the F keys by pressing FN+ESC.


I'm not going to change the find previous/next gesture assignments because I've never found that to be an issue.

#6499 was raised because NVDA+Shift+F3 might require two hands to complete, If you want to change this feature because it is used frequently in a specific case, there are too many NVDA gestures that require two hands that will be changed for this reason.

Many of NVDA's gestures take into account the user's memory base. If you only consider the convenience of key presses and give up uniformity, there are also many gestures that need to be changed.

Finally, this gesture has been around since version 0.6p2 and we were definitely breaking user muscle memory by changing it

Adriani90:
I think we need a decision by NV Access here.
We start having two types of users:
1. Users who always argue with muscle memory which is completely out of date in modern software management and UX design, and which is a serious impediment for improvements, especially when we talk about accessibility, and
2. Users who are really impacted by this key stroke because they cannot move their fingers in that position, especially when pressing nvda+shift+f3, so the key stroke is not accessible.

We have enough evidence, also in #6499, that users prefer to change this key stroke. However, especially for beginners it is a process that would not have to be done if the default key is more attractive.

We see more often that users have to change a lot of things in NVDA until they can really work efficiently with it. It might be worth it to do a community research on which default behaviors are changed most of the times and maybe adjust those default values to values that are convenient to most users and not only to those that work with NVDA since 0.6p2.
Before changing further relevant values as requested by @cary-rowen, this one is a concrete case that has been requested and for which there is evidence in the community.
All other behaviors can be requested specifically as separate issues on Github, or via a community survey.
So that argument brought by @cary-rowen might be valid but is not a reason for blocking this PR.

@gerald-hartig maybe you can have a look, in case NV Access decides that muscle memory is more important than accessibility, then this should be stated in a policy, otherwise we will always have to argue against such discussions when improving things, especially old behaviors in NVDA which will definitely need some redesign at some point.
Even Windows changed lots of its key strokes between Windows 8.1 and Windows 11, so muscle memory is not really the point here but I would like to have NV Access opinion on this before closing this PR.

burmancomp:
Well if I understand correctly: nvda+f3/shift+nvda+f3 may be difficult to use. Is there reason not to assume that this applies to all gestures which contain function keys. Then this would mean that there could not be any gestures which would use function keys.

Maybe quite many users like to use alt+f4 to close applications, and maybe some also use shift+f10, or f1. Because function keys are not f1 - f12 by default in many laptops, many users have to find a way to change default behavior to function keys behavior.

If there are given gestures which currently use function keys, and there is rationale to have them some other gestures, is it not possible to add extra key strokes for these gestures by default. For example nvda+x for find next and nvda+shift+x for find previous but keep also nvda+f3 and nvda+shift+f3.

CyrilleB79:
First of all, it's worth noting that #6499 has not been discussed in the last 5 years. During this time, NV Access' position may have evolved, NVDA's user base may have changed and the computer's world has changed.

For example, we may re-discuss using `control+F`, `f3` and `shift+f3` in browse mode. NV Access was not for this solution in the past. But if we consider that this solution has always been present in Jaws without any complain from the users, we may reconsider this choice. Also have we an idea on how many users need to use the standard browser's find commands (i.e. requiring to press `NVDA+f2` in this case)? And I imagine that among them, the majority are advanced who can easily remap their keyboard gestures.

Personally, I appreciate that NVDA has defined for its find function shortcuts that remind standard Windows commands, i.e. using F key or F3 key. I think that it's quite easy to memorize while X key has no special meaning.
If we really want to dissociate NVDA's find keys from standard windows find key, why not just use `j`, `shift+j` for NVDA's find next/previous and even `NVDA+j` for NVDA's initial find command?

XLTechie:
@Adriani90 wrote:

> The current commands (nvda+f3 and nvda+shift+f3 are too heavy to perform for many users, finger gymnastics is needed.

Then they can re-map them. Some, for example, have overridden those gestures to 
remove the NVDA key entirely, like Jaws apparently does. Since the browser find 
is not really useful to screen reader users much of the time, as long as the 
user doesn't use Word regularly, this is low impact.
Users who do use Word regularly, can map them somewhere else.

I also agree with the observations about function keys made by others.

> The users can use now nvda+x or nvda+shift+x.

Which now would divorce NVDA search, from search in most other programs. 
Notepad, Word, browsers (if partially sighted users do use the browser search), 
all of which use F3/Shift+F3 for next/prev search.

I am not in favor of this change.
Though I would be less opposed if we did it as @burmancomp suggested.

Adriani90:
Ok, it seems community support seems limited for this. I am closing. Though I hope NV Access will give their opinion on whether it is worthy to change or not. I don't want to invest more energy in this discussion.

I understand though the argument behind the consistency of this, but users in some countries request the change of this behavior specifically because this is the most used feature for them among all other function key commands.

Maybe someone will invest more time in providing a better solution.

XLTechie:
> I understand though the argument behind the consistency of this, but users in some countries request the change of this behavior specifically because this
> is the most used feature for them among all other function key commands.

Can you encourage them to comment on the original issue for this, so we can 
directly hear what they have to say instead of by proxy?

This is one of those "me and my silent army" situations, that I think really 
needs the silent army to be less silent.

Adriani90:
That‘s just utopic. I can ofcourse motivate some people to do it, but what you just bring up is utopic. Many people are not on github, and a lot of people don‘t speak english at all or not good enough. I am really concerned now that the trust in local community representants is put in question on this level. I see serious impediments in NVDA development comming up through people who really don‘t trust. I think this will cost a lot of energy for community representants.If this is the proposed way to go, I am seriously thinking to step down from this role. It is just utopic to bring all people to Github.Von meinem iPhone gesendetAm 12.04.2024 um 03:47 schrieb Luke Davis ***@***.***>:﻿
> I understand though the argument behind the consistency of this, but users in some countries request the change of this behavior specifically because this
> is the most used feature for them among all other function key commands.

Can you encourage them to comment on the original issue for this, so we can 
directly hear what they have to say instead of by proxy?

This is one of those "me and my silent army" situations, that I think really 
needs the silent army to be less silent.

beqabeqa473:
I am also against this change, because f3 shift+f3 are standard next/previous search result navigation keys and i don't think these should be changed in core.

Users can always remap shortcuts if they want.

Issue #13612, opened 2022-04-15
Adding an audio cue if nvda cursor has reached the boundary


LittleStar-VIP:
### Is your feature request related to a problem? Please describe.
1. user press up arrow
2. if current nvda cursor position has already reached the boundary, ie.g. cursor is already at the first item, e.g. the top edge of a word file, the top edge of the folderview listview, etc
3. word
4. if it is inside a editor, like notepad, nvda will repeat what is on the first line, but if it is in a file explore, nvda will silent

the above problem also apply to situations when cursor position is at the lowest edge
### Describe the solution you'd like
a special audio cue can be played to signify that it is already at the first item or last item
this is like what is currently handled by the IOS Voiceover

### Describe alternatives you've considered
if no audio cue, can nvda add a speech like 'this is the end' or 'this is the first item'

### Additional context
if it is inside a text editor and pressing up arrow repeating what is inside the first line, user may get confused of whether there are multiple lines that contain the same line or text

XLTechie:
+1

I find this particularly annoying on web pages which have multiple repeated 
items before the end of the page, or in text editors with multiple repeated 
lines.
In neither case can you know with any reliability that you're on the last one, 
unless you can check the number of lines, or move up and count the number of 
same items until you reach the top.

The Voiceover drum sound would be ideal for this.

cary-rowen:
yes i need this

mazen428:
+2

Brian1Gaff:
Not sure aboutthis one myself. If you had a window that made the line wrap 
artificially so to speak, where would you put the boundary. The I-phone 
needs this sound as its mostly touch screen driven, but why would oone need 
it in a keyboard driven system? I mean I guess you could have something like 
that for those modes where you were moving about inside a window with no 
tabable buttons etc, ie not really accessible, but I myself find that it can 
be annoying if you use a bluetooth keyboard on a phone.
 Brian

***@***.***
Sent via blueyonder.
Please address personal E-mail to:-
***@***.***, putting 'Brian Gaff'
in the display name field.
Newsgroup monitored: alt.comp.blind-users
----- Original Message ----- 
From: "LITTLE STAR☆VIP" ***@***.***>
To: "nvaccess/nvda" ***@***.***>
Cc: "Subscribed" ***@***.***>
Sent: Friday, April 15, 2022 11:54 AM
Subject: [nvaccess/nvda] Adding an audio cue if nvda cursor has reached the 
boundary (Issue #13612)

[...]

PR #16384, opened 2024-04-11
Sentence navigation

### LINKED ISSUE #8518: Feature request: Include Sentence Nav as part of NVDA

mltony:
### Link to issue number:
Closes #8518.
### Summary of the issue:
Sentence navigation.
1. `Alt+Down/UpArrow` now navigates by sentence in editables and in browse mode.
2. Sentence reconstruction option added to document navigation panel in settings. This defines to what extent sentence navigation can look into adjacent paragraphs when trying to reconstruct a sentence.
3. Alsoadded an option to specify non-breaking prefixes for current language.
1. Sentence splitting algorithm is implemented in `documentNavigation\sentenceNavigation.py`. It is generic enough and works with different types of textInfos. I left some comments throughout to explain inner workings, but LMK if more clarification/ explanation is needed. I ended up rewriting the algorithm (compared to SentenceNav add-on) from scratch - now the code should be cleaner and more elegant.
2. In `editableText.py` I added `speakCurrentSentence` script - not assigned any default keystroke. Hope that's ok - since SentenceNav users requested that in the past.
4. In `OffsetsTextInfo` updated `_getUnitOffsets()` function to deal with `UNIT_SENTENCE`.
5. In `UIATextInfo` updated `expand()` and `move()` methods to work with `UNIT_SENTENCE`.
6. In `browseMode.py` updated `script_collapseOrExpandControl`: if current item is not collapsable, then treat alt+Up/Down as sentence navigation commands.
7. Existing sentence navigation logic for MSWord is not affected.
### Testing strategy:
* Manual testing in:
  * Notepad
  * Notepad++
  * Visual Studio
  * VSCode
  * Chrome - browse mode and textArea
  * Firefox browse mode and TextArea
* Unit tests
Ready for review.
~~This is a draft, but fully working. Feel free to test or even start reviewing. Things I plan to address in the coming few days:~~
* ~~Figure out how to deal with language-specific non-breaking prefixes.~~
* ~~Add docuemntation~~
* ~~Add unit tests~~
### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/source/config/configDefaults.py
+++ b/source/config/configDefaults.py
@@ -6,7 +6,7 @@ import json
 
 
 DEFAULT_TEXT_PARAGRAPH_REGEX = (
-	r"^|({plb}{nlb}{optQuote}{dot}{optQuote}|{punc2}{optQuote}|{cjk}){optWiki}{spaces}|{n2}|\Z".format(
+	r"^|({plb}{nlb}{optQuote}{dot}{optQuote}|{punc2}{optQuote}){optWiki}{spaces}|{cjk}|{n2}|\Z".format(
 		# Look behind clause ensures that we have a text character before text punctuation mark.
 		# We have a positive lookBehind \w that resolves to a text character in any language,
 		# coupled with negative lookBehind \d that excludes digits.
--- a/source/config/configDefaults.py
+++ b/source/config/configDefaults.py
@@ -34,7 +34,7 @@ DEFAULT_TEXT_PARAGRAPH_REGEX = (
 		# since they don't trigger as many false positives.
 		punc2=r"[?!…]",
 		# We also check for CJK full-width punctuation marks without any extra rules.
-		cjk=r"[．！？：；]",
+		cjk=r"[。．！？]",
 		# Double newline means end of sentence too.
 		n2=r"([  \t]*\n){{2,}}",
 	)
--- a/source/config/configDefaults.py
+++ b/source/config/configDefaults.py
@@ -2,13 +2,19 @@
 # Copyright (C) 2024 NV Access Limited
 # This file is covered by the GNU General Public License.
 # See the file COPYING for more details.
+import json
+
 
 DEFAULT_TEXT_PARAGRAPH_REGEX = (
-	r"({lookBehind}{optQuote}{punc}{optQuote}{optWiki}{lookAhead}|{punc2}|{cjk})".format(
+	r"^|({plb}{nlb}{optQuote}{dot}{optQuote}|{punc2}{optQuote}|{cjk}){optWiki}{spaces}|{n2}|\Z".format(
 		# Look behind clause ensures that we have a text character before text punctuation mark.
 		# We have a positive lookBehind \w that resolves to a text character in any language,
 		# coupled with negative lookBehind \d that excludes digits.
-		lookBehind=r'(?<=\w)(?<!\d)',
+		plb=r"(?<=\w)(?<!\d)",
+		# Language-specific exceptions: characters suggesting that the following dot is not indicative
+		# of a sentence stop.
+		# This is a negative look-behind and will be inserted later when language is specified.
+		nlb="{nonBreakingRegex}",
 		# In some cases quote or closing parenthesis might appear right before or right after text punctuation.
 		# For example:
 		# > He replied, "That's wonderful."
@@ -16,18 +22,26 @@ DEFAULT_TEXT_PARAGRAPH_REGEX = (
 		# Actual punctuation marks that suggest end of sentence.
 		# We don't include symbols like comma and colon, because of too many false positives.
 		# We include question mark and exclamation mark below in punc2.
-		punc=r'[.…]{1,3}',
+		dot=r"[.]{{1,3}}",
 		# On Wikipedia references appear right after period in sentences, the following clause takes this
 		# into account. For example:
 		# > On his father's side, he was a direct descendant of John Churchill.[3]
-		optWiki=r'(\[\d+\])*',
-		# LookAhead clause checks that punctuation mark must be followed by either space,
+		optWiki=r"(\[\d+\])*",
+		# spaces clause checks that punctuation mark must be followed by either space,
 		# or newLine symbol or end of string.
-		lookAhead=r'(?=[\r\n  ]|$)',
+		spaces=r"([  ]+|([  \t]*\n)+|$)",
 		# Include question mark and exclamation mark with no extra conditions,
 		# since they don't trigger as many false positives.
-		punc2=r'[?!]',
+		punc2=r"[?!…]",
 		# We also check for CJK full-width punctuation marks without any extra rules.
-		cjk=r'[．！？：；]',
+		cjk=r"[．！？：；]",
+		# Double newline means end of sentence too.
+		n2=r"([  \t]*\n){{2,}}",
 	)
 )
+
+
+nonBreakingPrefix = json.dumps({
+	"en": r"\b[A-Z]|\bMr|\bMs|\bMrs|\bDr|\bProf|\bSt|\be.g|\bi.e",
+	"ru": r"\b[A-ZА-Я]|\b[Тт]ов|\b[Уу]л|\bт.[ке]|\bт. [ке]",
+})
--- a/source/gui/settingsDialogs.py
+++ b/source/gui/settingsDialogs.py
@@ -78,6 +78,8 @@ import weakref
 import time
 import keyLabels
 from .dpiScalingHelper import DpiScalingHelperMixinWithoutInit
+import json
+from documentNavigation import sentenceNavigation
 
 #: The size that settings panel text descriptions should be wrapped at.
 # Ensure self.scaleSize is used to adjust for OS scaling adjustments.
@@ -2645,9 +2647,60 @@ class DocumentNavigationPanel(SettingsPanel):
 		)
 		self.bindHelpEvent("ParagraphStyle", self.paragraphStyleCombo)
 
+		# Translators: This is a label for sentence reconstruction in the document navigation dialog
+		label = _("&Sentencew Sentence reconstruction:")
+		self.sentenceReconstructionCombo: nvdaControls.FeatureFlagCombo = sHelper.addLabeledControl(
+			labelText=label,
+			wxCtrlClass=nvdaControls.FeatureFlagCombo,
+			keyPath=["documentNavigation", "sentenceReconstruction"],
+			conf=config.conf
+		)
+		self.bindHelpEvent("SentenceReconstruction", self.sentenceReconstructionCombo)
+
+		language = speech.getCurrentLanguage()
+		prettyLanguage = languageHandler.getLanguageDescription(language)
+		# Translators: This is placeholder for the situation when we couldn't determine current language.
+		# It is used in document navigation settings panel to describe non-breaking prefixes edit box.
+		prettyLanguage = prettyLanguage or _("Unknown language")
+		# Translators: This is the label for a textfield in the
+		# Document Navigation settings panel.
+		label = _("Non-breaking prefixes [%s]")
+		label = label % prettyLanguage
+		self.nonBreakingPrefixesEdit = sHelper.addLabeledControl(
+			label,
+			wxCtrlClass=wx.TextCtrl,
+		)
+		j = json.loads(config.conf["documentNavigation"]["nonBreakingPrefixRegex"])
+		value = j.get(language, "")
+		self.nonBreakingPrefixesEdit .SetValue(value)
+		self.bindHelpEvent("NonBreakingPrefixesEdit ", self.nonBreakingPrefixesEdit)
+
+	def isValid(self) -> bool:
+		nbp = self.nonBreakingPrefixesEdit .GetValue()
+		try:
+			sentenceNavigation.getSentenceStopRegex(nonBreakingPrefixes=nbp)
+		except re.error as e:
+			log.debugWarning("Failed to compile text paragraph regex", exc_info=True)
+			gui.messageBox(
+				# Translators: Message shown when invalid text paragraph regex entered
+				_("Failed to compile text paragraph regular expression: %s") % str(e),
+				# Translators: The title of the message box
+				_("Error"),
+				wx.OK | wx.ICON_ERROR,
+				self,
+			)
+			return False
+		return True
+
+
 	def onSave(self):
 		self.paragraphStyleCombo.saveCurrentValueToConf()
-
+		self.sentenceReconstructionCombo.saveCurrentValueToConf()
+		language = speech.getCurrentLanguage()
+		j = json.loads(config.conf["documentNavigation"]["nonBreakingPrefixRegex"])
+		nbp = self.nonBreakingPrefixesEdit .GetValue()
+		j[language] = nbp
+		config.conf["documentNavigation"]["nonBreakingPrefixRegex"] = json.dumps(j)
 
 def _synthWarningDialog(newSynth: str):
 	gui.messageBox(
--- a/user_docs/en/userGuide.t2t
+++ b/user_docs/en/userGuide.t2t
@@ -594,8 +594,8 @@ NVDA provides the following key commands in relation to the system caret:
 | Report text formatting | NVDA+f | NVDA+f | Reports the formatting of the text where the caret is currently situated. Pressing twice shows the information in browse mode |
 | Report link destination | ``NVDA+k`` | ``NVDA+k`` | Pressing once speaks the destination URL of the link at the current caret or focus position. Pressing twice shows it in a window for more careful review |
 | Report caret location | NVDA+numpadDelete | NVDA+delete | Reports information about the location of the text or object at the position of system caret. For example, this might include the percentage through the document, the distance from the edge of the page or the exact screen position. Pressing twice may provide further detail. |
-| Next sentence | alt+downArrow | alt+downArrow | Moves the caret to the next sentence and announces it. (only supported in Microsoft Word and Outlook) |
-| Previous sentence | alt+upArrow | alt+upArrow | Moves the caret to the previous sentence and announces it. (only supported in Microsoft Word and Outlook) |
+| Next sentence | alt+downArrow | alt+downArrow | Moves the caret to the next sentence and announces it. |
+| Previous sentence | alt+upArrow | alt+upArrow | Moves the caret to the previous sentence and announces it. |
 
 When within a table, the following key commands are also available:
 || Name | Key | Description |
@@ -2501,6 +2501,26 @@ Note that this paragraph style cannot be used in Microsoft Word or Microsoft Out
 -
 
 You may toggle through the available paragraph styles from anywhere by assigning a key in the [Input Gestures dialog #InputGestures].
+
+==== Sentence reconstruction = ====[SentenceReconstruction]
+
+|| . {.hideHeaderRow} | . |
+| Options | Default (Reconstruct across same style paragraphs), Reconstruct across same style paragraphs, Never reconstruct sentences across paragraphs, Reconstruct sentences across any paragraphs|
+| Default | Reconstruct across same style paragraphs |
+
+
+This combo box allows you to select whether NVDA will try to identify and reconstruct sentences spanning across multiple paragraphs when navigating by sentence via``alt+upArrow`` and ``alt+downArrow``.
+Sentence reconstruction comes in handy in two typical use cases:
+- In most Most PDF files text is incorrectly split into paragraphs. Every line typically denotes a paragraph and thus most sentences span many adjacent paragraphs and in order to identify complete sentences NVDA must look into adjacent paragraphs.
+- Many email clients send emails broken up into lines, where each line also appears as a paragraph.
+-
+
+The available sentence reconstruction modes are:
+- Reconstruct sentences across same style paragraphs: NVDA will analyze preceding and following paragraphs only if they look visually similar.
+This style works best in most use cases.
+- Never reconstruct sentences across paragraphs: NVDA will never reconstruct sentences that span into adjacent paragraphs.
+- Reconstruct sentences across any paragraphs: NVDA will always attempt to reconstruct sentences spanning across paragraphs, even if paragraphs are of different style.
+-
 
 +++ Windows OCR Settings +++[Win10OcrSettings]
 The settings in this category allow you to configure [Windows OCR #Win10Ocr].
--- a/source/NVDAObjects/UIA/__init__.py
+++ b/source/NVDAObjects/UIA/__init__.py
@@ -65,6 +65,7 @@ import locationHelper
 import ui
 import winVersion
 import NVDAObjects
+from documentNavigation import sentenceNavigation
 
 
 paragraphIndentIDs = {
@@ -899,8 +900,13 @@ class UIATextInfo(textInfos.TextInfo):
 		return self._getBoundingRectsFromUIARange(self._rangeObj)
 
 	def expand(self, unit: str) -> None:
-		UIAUnit=UIAHandler.NVDAUnitsToUIAUnits[unit]
-		self._rangeObj.ExpandToEnclosingUnit(UIAUnit)
+		if unit == textInfos.UNIT_SENTENCE:
+			context = sentenceNavigation.SentenceContext(self)
+			sentence = context.moveSentence(0)
+			self._rangeObj = sentence._rangeObj
+		else:
+			UIAUnit = UIAHandler.NVDAUnitsToUIAUnits[unit]
+			self._rangeObj.ExpandToEnclosingUnit(UIAUnit)
 
 	def move(
 			self,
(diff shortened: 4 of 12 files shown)


mltony:
I have a technical question regarding how to deal with non-breaking prefixes.
Define non-breaking prefix is a word that goes right before period, but that suggests that the period doesn't serve as a sentence breaker. Here are some examples in English:
* Titles like Dr., Mr, Ms, Prof.
* Some abbreviations like i.e., s.t.
* Single capital letters. E.g. "George H. W. Bush"
* As a counterexample, not all abbreviations should be included. For example, "etc." should not be included, since a sentence can end with "etc.". In this case we should prefer to have a false positive rather than a false negative.

Non-breaking prefixes are obviously language-dependent. So far I just hardcoded English non-breaking prefixes, but that obviously needs to be changed before merging. I think we should make non-breaking prefixes to go through translation system to have it translated to all NVDA languages. But I see a few issues here to call out:
1. This is not a typical string to translate and it would require some research of the language. I will write a long #translators message explaining what is non-breaking prefixes and what's the expected format.
2. That string must be `|`-separated list of valid fixed-length regular expressions. Fixed length requirement is coming from the fact that each regex in the list is packed into a negative look-behind clause, which can only accept fixed length regex. So for example `S*` would be invalid for NLB. I expect some translators will not be familiar either with regexes in general or this specific requirement, so mistakes will be made. I haven't dealt with NVDA translation system before, so wondering, is there a way to set up a custom validator for a translateable string? If not, then can it be arranged to manually check all incoming translations of the string and invalidate those that will turn out to be invalid?
3. Another issue is identifying current language. Often times people use NVDA to read context in multiple languages, so perhaps a good idea would be to take the current language of synthesizer and retrieve non-breaking prefixes in that language, instead of default NVDA language. There are two problems here:
    * How to map current synth language onto NVDA languages? In the past I remember for Russian some synthesizers return "ru_ru", whereas in NVDA I only see "ru". An obvious solution would be to try to identify language by the first two letters, but then I see some country-specific languages in `user_docs` directory, such as `de_CH`. How big deal would it be to ignore such country-specific languages?
    * What is API call to retrieve a string in another language? I bet it must be possible, but I looked through NVDA source code and didn't find any examples - would be great if someone knows.
4. Do we need to create a setting to specify non-breaking prefixes? In the past I've been given feedback that I tend to expose too many settings and it's confusing to the users - so asking here to get community's opinion.

cary-rowen:
Thanks @mltony,

Based on Sentence-Nav I want to know:
1. Will phrase navigation be included here?
2. Will there also be a user-defined regular expression for sentence splitting?
3. How do we deal with MSWord? There is an option in the Sentence-Nav add-on to override the default sentence navigation of MSWord. It is very useful not only for MSWord but also for Bookworm.

mltony:
> 1. Will phrase navigation be included here?

No. But it should be very straightforward to add it later provided NVAccess is aligned.

> 2. Will there also be a user-defined regular expression for sentence splitting?

It currently uses the same regex as for text paragraph navigation, which is configurable. But its only a single regex.

> 3. How do we deal with MSWord? There is an option in the Sentence-Nav add-on to override the default sentence navigation of MSWord. It is very useful not only for MSWord but also for Bookworm.

I don't override MSWord sentence navigation here. It should be easy to override later - again as long as NVAccess is aligned.

In general I think getting this PR approved and merged is going to be somewhat challenging because there are many unknowns regarding non-breaking prefixes. So at this point I would like to focus on core sentence navigation, and postpone bells and whistels to later PRs.

XLTechie:
@mltony wrote:

>    Also, not sure how to fix lint error in configSpec.py - line is indeed to long, but trying to break it up leads to invalid config.

At end of line, after two spaces:

# noqa: E501 Breaking this line makes invalid config

XLTechie:
@mltony wrote:
> Also, not sure how to fix lint error in configSpec.py - line is indeed to long, but trying to break it up leads to invalid config.

At end of line, after two spaces:

```
# noqa: E501 Breaking this line makes invalid config
```

XLTechie:
@mltony wrote:
>     2. That string must be `|`-separated list of valid fixed-length regular expressions. [.] I expect some translators will not be familiar either with regexes in general or this specific requirement, so mistakes will be made.

Can you use a tuple of strings, then join them with or bars at runtime?

```python
sentenceSeparators = "|".join((
        "string1",
        "string2",
        "etc.",
))
```

I don't know how translation for this would work--I imagine you couldn't use normal translatable strings, but I'm not sure. Regardless you must solve that question either way, and this is probably better than trying to make translators juggle a large regex.

mltony:
> Can you use a tuple of strings

How can we send that tuple to translators? If translation system supports tuples of variable length then should be doable, but please show me an example.

XLTechie:
@mltony I have no idea. Nor do I know how you'll move a long and fragile regular 
expression string through the translation process.

Especially with the translation system the way it is now, I know even 
less about it than I did before.

bad
Apologies for the inapplicable idea.

mltony:
This is now ready for review.

cary-rowen:
Hi @mltony 
I'm getting the following error when trying to do sentence navigation using the latest build.
```
IO - speech.speech.speak (07:12:27.736) - MainThread (11956):
Speaking ['日志片段开始点已标记，再按一次复制到剪贴板']
IO - inputCore.InputManager.executeGesture (07:12:28.564) - winInputHook (3860):
Input: kb(laptop):alt+downArrow
ERROR - scriptHandler.executeScript (07:12:28.582) - MainThread (11956):
error executing script: <bound method BrowseModeDocumentTreeInterceptor.script_collapseOrExpandControl of <NVDAObjects.IAccessible.chromium.ChromeVBuf object at 0x0B435150>> with gesture 'Alt+下光标'
  File "scriptHandler.pyc", line 295, in executeScript
  File "browseMode.pyc", line 1689, in script_collapseOrExpandControl
  File "cursorManager.pyc", line 157, in _caretMovementScriptHelper
  File "textInfos\offsets.pyc", line 538, in expand
  File "virtualBuffers\__init__.pyc", line 405, in _getUnitOffsets
  File "textInfos\offsets.pyc", line 512, in _getUnitOffsets
  File "documentNavigation\sentenceNavigation.pyc", line 210, in __init__
  File "documentNavigation\sentenceNavigation.pyc", line 145, in getSentenceStopRegex
KeyError: '1,3'
IO - inputCore.InputManager.executeGesture (07:12:29.599) - winInputHook (3860):
Input: kb(laptop):control+shift+NVDA+f1
```
There seems to be no language difference, I used the English and Chinese versions of the NVDA User Guide for testing.

source/config/configDefaults.py
```diff
@@ -2,32 +2,46 @@
...
 		# or newLine symbol or end of string.
-		lookAhead=r'(?=[\r\n  ]|$)',
+		spaces=r"([  ]+|([  \t]*\n)+|$)",
 		# Include question mark and exclamation mark with no extra conditions,
 		# since they don't trigger as many false positives.
-		punc2=r'[?!]',
+		punc2=r"[?!…]",
 		# We also check for CJK full-width punctuation marks without any extra rules.
-		cjk=r'[．！？：；]',
+		cjk=r"[．！？：；]",
```
cary-rowen, line 37: In Chinese, the period symbol is "。"

mltony:
@cary-rowen, try resetting your value for text paragraph regex. It should work with the default value.

Adriani90:
@mltony did you test this in browse mode in pdf documents, with Adobe reader, outlook messages in MS outlook and browse mode in MS Word? Does it give an error in browse mode in MS Excel? Or did you implement a message that this is not supported there?

Adriani90:
@mltony Re non breaking prefixes, it might be worth to add all so called "period words" in a database first and import them from there. Maybe there could be a file similar to symbols.dic called periodwords.dic. At least with such a file you could pass it to the translation system and we would get all of the international prefixes translated while translators can also add their language specific period words to their own language's periodwords.dic file.
Ofcourse the correct sentence navigation would depend on translation work.

mltony:
@Adriani90 

* Browse mode in browsers: tested, works fine
* PDF reader browse mode: tested, works fine
* MSWord and Outlook: legacy behavior overrides functionality oof my PR, so no change here
* In MS Excel alt+Down arrow seems to trigger expand action, so seems no regressions here.
Re symbols file - that's a reasonable idea. I guess I'll just wait for a few days before working on that just in case anyone else has objections or better ideas.

cary-rowen:
Hi @mltony 

Thank you for your efforts on this. I will continue to pay attention to this feature and provide the following test work for your reference.

# Test in Windows 10 Notepad:
## Steps
1. Write the following Chinese content : "今天天气很好。我想去爬山。" These are two sentences containing two periods.
2. Press Alt+Up/DownArrow.
3. Add a space after the first period.
4. Place the caret at the beginning of the text and press Alt+Downarrow repeatedly.
5. Press Alt+UpArrow once.

## Actual behavior:
* Unable to navigate by sentence when performing step (2).
* NVDA reports the second sentence when performing step (4).

## expected behavior
* Perform step 2 to navigate by sentence. No matter whether there is a space after the period, because in Chinese writing, there is generally no space after the period.
* When performing step (4), "blank" should be reported, consistent with the behavior in MS Word.

# Test in NVDA Speech Viewer or Bookworm:
Sentence navigation is not working. If you use the Sentence-nav add-on, you need to enable overriding MS Word's sentence navigation behavior in settings.
So, as far as this PR is concerned, I think we should have a solution for similar situations like the Bookworm rich text control.

Thanks

mltony:
@cary-rowen

* Re step 2: fixed regexp, try again.
* Also did I get Chinese question mark and exclamation mark correctly in the regexp?
* Re step 4: This one is complicated. So I implement in this PR functions like `UIATextInfo.expand(UNIT_SENTENCE)`. If you want NVDA to report blank in this case, that means that when the cursor is at the last position of the document, it should expand into empty sentence. I actually designed sentences to be non-empty. Trying to redesign just to account for this case is going to be difficult and any attempt to redesign is likely to cause bugs in many other edge cases.
Trying to analyze behavior with different textInfo units, I played with `UNIT_WORD`. If you type: "Hello world" without quotes into Notepad on Windows 11 and try to press control+rightArrow repeatedly, the cursor would stay at the word "world" and not find a blank word in the end. Although in notepad++ it does find the blank word. So behavior is not consistent here.
So given the fact that there is already some inconsistency present with other textInfo units and given how hard and hacky it's going to be to add fake empty sentence at the end, I would ask whether it's even worth it to tackle this issue. WDYT?
* Re NVDA Speech Viewer : acctually it works on my side. It doesn't appear to be using MSWord control inside.
* Re Bookworm: I prefer to postpone this until after this PR is merged. I'm worried that the more features we add here, the more likely we'll get bogged down with discussions on implementation details.

cary-rowen:
> Re step 2: fixed regexp, try again.
Yes, sentence navigation now works as expected.

---

> Also did I get Chinese question mark and exclamation mark correctly in the regexp?

Yes.

---

> So given the fact that there is already some inconsistency present with other textInfo units and given how hard and hacky it's going to be to add fake empty sentence at the end, I would ask whether it's even worth it to tackle this issue. WDYT?

OK, regarding this behavior, if it is very difficult then I don't think there is a need to put too much effort into it.

---

> If you type: "Hello world" without quotes into Notepad on Windows 11 and try to press control+rightArrow repeatedly, the cursor would stay at the word "world" and not find a blank word in the end. 

Yes, I reproduce this on Notepad on Windows 10 as well.

---

As an improvement on step 4, I think another behavior might be better: 
NVDA will report the last sentence repeatedly when we press Alt+DownArrow repeatedly, And the caret has been moved to the end of the last sentence. This means that if the user presses Alt+UpArrow once you will hear NVDA continue reporting the last sentence.

Personally, I would like it to report the first sentence, ie: when there are no more sentences our caret should be placed at the beginning of the last sentence, no matter how many times I press Alt+DownArrow.

This is consistent with the paragraph navigation pattern in document navigation.
This is also consistent with the behavior of the sentence-nav add-on, and users of the sentence navigation add-on will have a consistent experience as before.

---

> Re NVDA Speech Viewer : acctually it works on my side. It doesn't appear to be using MSWord control inside.

That's weird, can you confirm this again, it seems to me that Speech Viewer behaves the same as Bookworm, i.e. sentence navigation doesn't work at all.

The test I performed was with Speech Viewer open: I asked NVDA to report the Chinese sentences given in the comments above. Then try sentence navigation in Speech Viewer.
NVDA seems to treat the entire Speech Viewer as one sentence.

cary-rowen:
Hi @mltony 

While reminding you not to forget the above comment, I thought of another question.

I'm wondering, is it necessary that we share the regex option with paragraph navigation?
The current regular expression options are very complex and may be completely uncustomizable by novice users. Even someone like me who can read regular expressions may need to spend a lot of time understanding it.

Is there a way we can make the regex options more understandable?
If not, that’s no problem. Our goal is to provide users with an experience that they can use right out of the box, rather than requiring users to modify it before they can use it.

Thanks

CyrilleB79:
I have not reviewed this PR.

But, on contrary to what I wrote earlier, I understand now why the regexp for text paragraph navigation and the one for sentence navigation may be separately configurable.
For now, the criterion to identify a text paragraph is if it contains a sentence. However, we may imagine in the future that we use a different criterion (even if I do not have it in mind now).

Of course, since we do not have another criterion now, we may also wait for a real need before duplicating these fields.
@cary-rowen, have you a real use case / interest to configure differently text paragraph navigation and sentence navigation?

Re the possibility to better understand the regexp, we may use a verbose regexp in the field, rather than a normal one. Verbose regexps ignore comments, spaces and newline characters (unless escaped). We should however take into account that verbose regexps are less known than normal ones; I have discovered this option only some weeks ago...

Adriani90:
I agree in case of sentence navigation, the carret should always stay at the beginning of the sentence as it does with paragraph navigation when the paragraph is handled by NVDA in document navigation settings, no matter if it is the first or the last sentence or any senence in the middle.
Instead of repeating the first or the last sentence, there could be a message reporting "no next sentence" or "no previous sentence". This is also consistent with NVDA paragraph navigation.

cary-rowen:
@CyrilleB79 
> Have you a real use case / interest to configure differently text paragraph navigation and sentence navigation?

No, I haven't yet. I just have the above thoughts from the perspective of ease of understanding.
However, we can focus on the core implementation of this PR, and put other things into separate PRs to discuss or implement.

cary-rowen:
> Instead of repeating the first or the last sentence, there could be a message reporting "no next sentence" or "no previous sentence". This is also consistent with NVDA paragraph navigation.

To be honest, I don’t really like using speech to indicate boundaries.
At this point, NVDA's paragraph navigation is inconsistent with other behaviors, such as paragraph navigation in Word.

I prefer to use beeps or play sound to indicate boundaries.

A similar request has already been made by #13612.

cary-rowen:
By the way, sentence-nav add-on is better and there are also beep when navigating from one paragraph to another.

Adriani90:
There is no consistency in paragraph navigation between apps. Libreoffice behaves differently compared to ms Word for example. Nvda paragraph navigation is the most consistent paragraph handling we have so far.Von meinem iPhone gesendetAm 24.04.2024 um 17:36 schrieb Rowen ***@***.***>:﻿

Instead of repeating the first or the last sentence, there could be a message reporting "no next sentence" or "no previous sentence". This is also consistent with NVDA paragraph navigation.

To be honest, I don’t really like using speech to indicate boundaries.
At this point, NVDA's paragraph navigation is inconsistent with other behaviors, such as paragraph navigation in Word.
I prefer to use beeps or play sound to indicate boundaries.
A similar request has already been made by #13612.

Adriani90:
Indicating boundaries by sounds is a valid issue but out of scope for this PR I guess.Von meinem iPhone gesendetAm 24.04.2024 um 17:36 schrieb Rowen ***@***.***>:﻿

Instead of repeating the first or the last sentence, there could be a message reporting "no next sentence" or "no previous sentence". This is also consistent with NVDA paragraph navigation.

To be honest, I don’t really like using speech to indicate boundaries.
At this point, NVDA's paragraph navigation is inconsistent with other behaviors, such as paragraph navigation in Word.
I prefer to use beeps or play sound to indicate boundaries.
A similar request has already been made by #13612.

Adriani90:
What I wanted to say is that when paragraphs are handled by applications, in some applications NVDA repeats the first or last paragraph, in some applications NVDA is silent, in some applications a windows bing sound is heared, etc.
NVDA has a good basis for reporting the boundaries, and this can ofcourse be improved with sounds or what ever, but this does not affect only paragraphs and is in my view at least still out of scope for this PR.

For now The boundary reporting for paragraphs in NVDA is consistent all over the place when paragraphs are handled by NVDA.
Tony can make use of the internal paragraph handling to design the boundaries for documents and if we improve the boundary reporting, then this should be done for both in a separate PR.
But this affects also other navigation paterns handled by NVDA such as line by line navigation, every kind of review cursor or object navigation that reaches the beginning or the end of an element etc.

user_docs/en/userGuide.t2t
```diff
@@ -2502,6 +2502,26 @@ Note that this paragraph style cannot be used in Microsoft Word or Microsoft Out
...
+==== Sentence reconstruction = ====[SentenceReconstruction]
+
+|| . {.hideHeaderRow} | . |
+| Options | Default (Reconstruct across same style paragraphs), Reconstruct across same style paragraphs, Never reconstruct sentences across paragraphs, Reconstruct sentences across any paragraphs|
+| Default | Reconstruct across same style paragraphs |
+
+
+This combo box allows you to select whether NVDA will try to identify and reconstruct sentences spanning across multiple paragraphs when navigating by sentence via``alt+upArrow`` and ``alt+downArrow``.
+Sentence reconstruction comes in handy in two typical use cases:
+- In most Most PDF files text is incorrectly split into paragraphs. Every line typically denotes a paragraph and thus most sentences span many adjacent paragraphs and in order to identify complete sentences NVDA must look into adjacent paragraphs.
```
Qchristensen, line 2514: Duplicate word: most (In most Most PDF files...).  Also "Every line typically denotes a...." should be on a new line.  I would also split into two sentences as:

Every line typically denotes a paragraph and thus most sentences span many adjacent paragraphs.
In order to identify complete sentences NVDA must look into adjacent paragraphs.

Qchristensen (changes requested):
Minor recommended change to user guide text, otherwise docs read well.

CyrilleB79 (commented):
I have done a first review of a part of the code and listed comments/suggestions here.

However, I'll also provide more important remarks in my next comment.

CyrilleB79:
Here are some more general remarks.
Note that the feature does not seem to work on my end.

### Test in Chrome's browse mode
Pressing `Alt+downArrow` in this GitHub's issue:

```
IO - inputCore.InputManager.executeGesture (16:05:38.597) - winInputHook (9664):
Input: kb(desktop):alt+downArrow
DEBUG - config.featureFlag._validateConfig_featureFlag (16:05:38.643) - MainThread (1980):
Validating feature flag: DEFAULT, optionsEnum: SentenceReconstructionFlag, behaviorOfDefault: SAME_STYLE_PARAGRAPHS
ERROR - scriptHandler.executeScript (16:05:38.657) - MainThread (1980):
error executing script: <bound method BrowseModeDocumentTreeInterceptor.script_collapseOrExpandControl of <NVDAObjects.IAccessible.chromium.ChromeVBuf object at 0x0B7F3FF0>> with gesture 'alt+flèche bas'
  File "documentNavigation\sentenceNavigation.py", line 146, in getSentenceStopRegex
    regex = regex.format(nonBreakingRegex=nonBreakingNLBs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: '1,3'

During handling of the above exception, another exception occurred:

    script(gesture)
  File "browseMode.py", line 1689, in script_collapseOrExpandControl
    return self._caretMovementScriptHelper(gesture, textInfos.UNIT_SENTENCE, direction)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "cursorManager.py", line 157, in _caretMovementScriptHelper
    info.expand(unit)
  File "textInfos\offsets.py", line 538, in expand
    self._startOffset,self._endOffset=self._getUnitOffsets(unit,self._startOffset)
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "virtualBuffers\__init__.py", line 405, in _getUnitOffsets
    return super(VirtualBufferTextInfo, self)._getUnitOffsets(unit, offset)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "textInfos\offsets.py", line 512, in _getUnitOffsets
    context = SentenceContext(textInfo)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "documentNavigation\sentenceNavigation.py", line 213, in __init__
    regex = getSentenceStopRegex()
            ^^^^^^^^^^^^^^^^^^^^^^
  File "documentNavigation\sentenceNavigation.py", line 148, in getSentenceStopRegex
    raise RuntimeError("Failed to substitute non-breaking prefixes into sentence regular expression. Please make sure your sentence regular expression is valid.", e)
RuntimeError: ('Failed to substitute non-breaking prefixes into sentence regular expression. Please make sure your sentence regular expression is valid.', KeyError('1,3'))
```

### Documentation navigation panel
* Open this panel
* Validate without changing anything

Same style of error as above.

### In Notepad

The feature is not working with nothing logged in Notepad (Windows 10). Is it expected to work there?

### Exception listing
The exceptions for sentence stop are listed for all expressions in all languages.

I am not totally convinced of this way to do. However, if we do not find a better solution and if it is validated by NV Access:
- If this field remains in this panel, I'd prefer a comma separated list rather than a regex; it's probably easier for all users. In advanced settings panel, a regex is OK though.
- May we consider a default exception list for all languages, e.g. a single letter followed by a dot?
- We should ask translators to provide exceptions for their language as much as possible. Maybe after having merged this PR, translators can test sentence nav in NVDA alpha; it will be easy to provide one or more subsequent PRs to add language specific exceptions.
- I did not check it. But how have you handled language variants such as "pt" vs "pt_PT" / "pt_BR" or "en" vs "en_AU"?

### Unknown languages

The following languages are recognized as unknown in the settings panel:
* many eSpeak languages such as (list not complete):
  * English (Landcaster)
  * Esperanto
  * Vietnamese (Central)
  * Chinese (Mandarin, latin as Pinyin)
  * Chinese (Mandarin, latin as English)
  * Chinese (Cantonese)
  * Armenian (West Armenia)
  * Haitian Creole
* IBMTTS: English UK

IMO, something should be done to have less unrecognized languages. At least English variants should fallback to English (neutral) and main Chinese languages should be recognized.

More generally, if possible, we may log a warning and display the language code in case the language name is not recognized by Windows/NVDA, but not default to Unknown language.

source/gui/settingsDialogs.py
```diff
@@ -2645,9 +2647,60 @@ def makeSettings(self, settingsSizer: wx.BoxSizer) -> None:
...
+		self.bindHelpEvent("SentenceReconstruction", self.sentenceReconstructionCombo)
+
+		language = speech.getCurrentLanguage()
+		prettyLanguage = languageHandler.getLanguageDescription(language)
+		# Translators: This is placeholder for the situation when we couldn't determine current language.
+		# It is used in document navigation settings panel to describe non-breaking prefixes edit box.
+		prettyLanguage = prettyLanguage or _("Unknown language")
+		# Translators: This is the label for a textfield in the
+		# Document Navigation settings panel.
+		label = _("Non-breaking prefixes [%s]")
```
CyrilleB79, line 2667: Why is it called prefix?

```suggestion
		label = _("Sentence stop exceptions [%s]")
```

or something else similar or even more suitable.

mltony:
@CyrilleB79, thanks for comments, I'll address them soon;
If you are getting exception with message:
```
KeyError: '1,3'
```
You would need to delete `textParagraphRegex` item in `[virtualBuffers]` section from your `nvda.ini` file to revert to the default value.
Although since we missed the window to merge this into v2024.2, most likely I'd need to update this PR to use a separate option for sentence regexp.
As for notepad being silent - there is legacy code that swallows any when trying to navigate by unit - most likely it's just swallowing the same `KeyError` in your case.

seanbudd (commented):
Hi @mltony - this is just a partial review, as I've already accumulated a large amount of feedback

CyrilleB79:
> @CyrilleB79, thanks for comments, I'll address them soon; If you are getting exception with message:
> 
> ```
> KeyError: '1,3'
> ```
> 
> You would need to delete `textParagraphRegex` item in `[virtualBuffers]` section from your `nvda.ini` file to revert to the default value.

OK. It is working now.
Maybe it was not a good idea to put the text paragraph nav regexp in the advanced settings panel; since when you want to update it, you may create API-breaking changes...

> As for notepad being silent - there is legacy code that swallows any when trying to navigate by unit - most likely it's just swallowing the same `KeyError` in your case.

Would it be the opportunity to fix this? Or do you think it's better in a separate PR? In the latter case, it would be worth to open an issue describing your finding.

Adriani90:
> Would it be the opportunity to fix this? Or do you think it's better in a separate PR? In the latter case, it would be worth to open an issue describing your finding.

I would advise to fix this as part of this PR, otherwise users will be confused because sentence navigation will be promoted as a global feature.

Also @michaelDCurran, @seanbudd I really hope you have a careful review on this before merging.
As I said in #16271, this Regexes can really have bad effects when users change them accidentally. I hope you reconsider giving #16271 a higher priority.

Also when reviewing following should be taken into account:
* Does the caret / virtual cursor  always land on the beginning of a sentence? This is what we expect.
* Is the beginning and the end of a document indicated via "no previous sentence" or "no next sentence"? This is what we expect and is in line with paragraph navigation in document navigation settings when paragraph is handled by NVDA.

seanbudd:
@mltony - do you plan to do any further work here?

mltony:
@seanbudd, I plan to take some time off work around January 2025 and finish this PR. Would that work for you?

seanbudd:
Yep no problem - I'm going to close this for now, as merge conflicts will continue to accumulate and discussion on this PR is already long. This PR will serve as good history to inform any future re-attempt.

PR #16396, opened 2024-04-15
Added Windows tips to the NVDA user guide

Adriani90:
### Link to issue number:
Addresses #6499.

### Summary of the issue:
During discussions in #6499 and #16383, changing default NVDA key commands has turned out to be undesirable due to several reasons pointed out by some users. Though there are still a lot of users having problems with current NVDA key stroke environment as reported in different mailing lists and local communities. During the process it turned out that it would be very useful to have an overview of Windows built in features and settings in the NVDA user guide which have an effect on NVDA experience.

Users will have a quick overview of Windows built in accessibility features and settings which have an effect on NVDA experience in the NVDA user guide.

Added those to the user guide.

### Testing strategy:
Tested that the lists show up properly after conversion.

None

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


CyrilleB79:
As indicated by the label added by @seanbudd, NV Access will have to make a decision if such information needs to be added in the User Guide.

If they agree with such addition, I have the following remarks:
* We cannot say that this issue addresses #6499, at least not completely, as per https://github.com/nvaccess/nvda/issues/6499#issuecomment-2053756432.
* If we begin to accept Windows native feature description, what was not done before, NV Access should clearly state where the limit lies.
* The accessibility features of Windows may evolve in time; the risk is that the User Guide becomes outdated. Why not rather put a link to Microsoft Accessibility help page? It's worth noting that this page is usually translated.
* Use a more descriptive title than "Tips"; else we will end up with many tips of all sorts.

Adriani90:
Thanks @CyrilleB79 for your comment,
You wrote:
> If we begin to accept Windows native feature description, what was not done before, NV Access should clearly state where the limit lies.

I agree. these tips are at least not meant to be a full list or a full description, it is just a hint for users to get a quick clue of some of the settings and how they are related to NVDA experience.

> The accessibility features of Windows may evolve in time; the risk is that the User Guide becomes outdated. Why not rather put a link to Microsoft Accessibility help page? It's worth noting that this page is usually translated.

There are many things that need to be maintained in the user guide from time to time, so this should not block improving the user guide in my view.
The Microsoft help page depends on internet connection, having these tips in a quick way via the user guide is more convenient for many people who do not have internet connectoin but still want to improve their NVDA experience.
Also the Microsoft page does not describe any relations to NVDA, so the users might even not try to search for such features at all.

> Use a more descriptive title than "Tips"; else we will end up with many tips of all sorts.

Would this be a problem? Actually giving users some tips out of our experience via the user guide would be a good thing. In fact this could be organized in a structured way so that a tips reference could be generated from the user guide.

It is important that these tips do not have to be a complete description of a feature or setting, but their effects or relation to NVDA should become obvious so that users can decide whether they need that for their NVDA experience or not.

Qchristensen:
Thanks @Adriani90 for raising this.  After discussion among NV Access, we have decided not to proceed with adding Windows features to the user guide.  What this issue has raised as worth doing however, is:

1) A link to Microsoft's Windows / accessibility documentation in the User Guide introduction.  I agree that users want the information, but it should be documented by Microsoft - that way users can also get the correct information for their version of Windows.  We've just dropped support for Windows 7, but looking at Magnifier for instance, it was quite different in Windows 7 to how it is in Windows 10 (I can't recall its behaviour in Windows 8.1 offhand but even so it was just an example, other accessibility features will work differently between Windows 8.1 and 11).

2) One Windows feature we should mention is sticky keys because that does directly affect the behaviour of the NVDA modifier key.

(Expanding on "2", the original suggestions in this PR are all arguably useful Windows features, which may be helpful for some users, but don't specifically change how NVDA behaves.  E.G. NVDA shouldn't behave any differently with Windows magnifier running as without it - if there is an instance I've forgotten about, that would be worth an issue - either to correct if it shouldn't, or to note in the User Guide if it is deliberate).

I've got actions for myself for both of those unless anyone else is keen to jump on it first.

Adriani90:
Thanks Quentin for following up on this. Actually the link to Microsoft documentation will be out of date at a certain time in future anyway, so it will be a matter of maintaining the links.
Re: sticky keys, also filter keys are worth mentioning, these allow to adjust the sensitivity of key presses when a key is pressed multiple times, this will also affect how NVDA will work.

PR #16477, opened 2024-05-02
Use moveToCodepointOffset for cursor routing

### LINKED ISSUE #10960: Braille: Variation Selectors break cursor positions
### Steps to reproduce:
1. Consider the following line:
  > ⚠️ test
2. Move the cursor on the word (test) using routing cursors.

### Actual behavior:
The cursor is moved on — the desired position + 1 —. Also ⚠️ takes 2 braille cells.

### Expected behavior:
The cursor should be moved on the desired position. Also ⚠️ symbol should takes one cell only.
With wordpad and in form mode in Firefox, cursor is properly moved.

Installed and portable

#### NVDA version:
2019.3, 2020.1

#### Windows version:
10 Insider (64-bit) build 19592.1001

notepad, Microsoft Word, Firefox (in browse mode), Run dialog (windows+r), etc..

Yes

No

Yes

No

CC @leonardder

LeonarddeR:
### Link to issue number:
Fixes #10960

### Summary of the issue:
There are cases where moving one character on a textInfo instance actually moves more than one unicode point offset. This is described by @mltony in the doc string for `textInfos.TextInfo.moveToCodepointOffset`.
This causes of by one errors when cursor routing, since we're asking the textInfo to move by 1 characters, that might be presented by two or even more characters within the liblouis mapping.

Cursor routing should be more reliable.

@mltony's creation of `moveToCodepointOffset` allows us to move x code points from the start of the reading unit. As we're using 32 bit for liblouis, every character as presented by liblouis is equal to one code point. Therefore we can safely assume that this method to move is much more reliable than the previous method.

### Testing strategy:
- Test str from #10960
- - Unit test that basically mimics the str from  #10960

None known

### Code Review Checklist:

- [ ] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


LeonarddeR:
This is ready for review. Only the changelog entry is missing, because there's no 2024.3 section yet.

seanbudd (approved):

LeonarddeR:
I added a changelog entry. Note that i also credited @mltony, since this pr is basically a very small one that completely builds on his work for unicode point offset navigation.

PR #16490, opened 2024-05-07
Standardized settings in user guide, + some wording corrections

Adriani90:
### Link to issue number:
None

### Summary of the issue:
In #15902 it became obvious that we need structured data in the user guide, so I prepared a PR.

Settings and their options along with the effect on NVDA behavior are now much clearer documented.

Adjusted the user guide in line with user standards, among others introduced in #15950.

### Testing strategy:
Tested that tables are shown correctly in browser when opening the user guide.

Not sure if there is something messed up visually though, someone needs to check.

### Code Review Checklist:

- [ ] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


XLTechie:
@Adriani90 I am hesitant to review this, though I'm sure I'll have comments. You have conflicts with master, because you submitted this right before #16492, which removes/renames-and-changes the t2t files.
Please synchronize.

seanbudd:
Unfortunately you'll need to redraft this due to the timing of #16419 and #16492
Good news is that the [markdown build artfiact](https://ci.appveyor.com/api/buildjobs/4ng5wde6bjg9ooqe/artifacts/output%2FuserGuide.md) from this PR will make it much easier

Adriani90:
Yeah It seems I chose the best timing to submit this. :))
Joke aside, I will draft it again soon. Actually it should not be a big deal given the md file is aalready converted. I there are any comments, you can also put them here so I can already consider them while looking into the md directly.

user_docs/en/userGuide.md
```diff
@@ -434,8 +434,8 @@ However, you can enable or disable the data gathering process manually in NVDA's
 Most NVDA-specific keyboard commands consist of pressing a particular key called the NVDA modifier key in conjunction with one or more other keys.
 Notable exceptions to this are the text review commands for the desktop keyboard layout which just use the numpad keys by themselves, but there are some other exceptions as well.
 
-NVDA can be configured so that the Insert, numpad Insert, and/or Caps Lock key can be used as the NVDA modifier key.
-By default, both the Insert and numpad Insert keys are set as NVDA modifier keys.
+NVDA can be configured so that the numpad Insert, Extended Insert and/or Caps Lock key can be used as the NVDA modifier key.
+By default, both the numpad Insert and Extended Insert keys are set as NVDA modifier keys.
```
seanbudd, line 438: please undo these changes

user_docs/en/userGuide.md
```diff
@@ -567,17 +567,22 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
...
+|---|---|
+|Options |off, beeps, talk, on-demand|
+|Default |talk|
+
+| Option |Behaviour|
+|---|---|
+|off |NVDA will not speak anything, however it will silently react to commands.|
+|beeps |NVDA will replace normal speech with short beeps.|
+|talk |NVDA will speak normally in reaction to screen changes, notifications, and actions such as moving the focus, or issuing commands.|
+|on-demand |NVDA will only speak when you use commands with a reporting function (e.g. report the title of the window); but it will not speak in reaction to actions such as moving the focus or the cursor.|
```
seanbudd, line 580: can you please copy the same order as before? it matches the default order of the commands when switching

user_docs/en/userGuide.md
```diff
@@ -567,17 +567,22 @@ When the menu comes up, You can use the arrow keys to navigate the menu, and the
 
 ### Speech modes {#SpeechModes}
 
+| . {.hideHeaderRow} |.|
+|---|---|
+|Options |off, beeps, talk, on-demand|
```
seanbudd, line 572: can you please copy the same order as before?

user_docs/en/userGuide.md
```diff
@@ -1653,53 +1658,107 @@ When the confirmation dialog appears, select "restart now" or "restart later" if
 
 ##### Save configuration on exit {#GeneralSettingsSaveConfig}
 
-This option is a checkbox that, when checked, tells NVDA to automatically save the current configuration when you exit NVDA.
```
seanbudd, line 1656: can you please keep the original sentence as a summary?

user_docs/en/userGuide.md
```diff
@@ -1653,53 +1658,107 @@ When the confirmation dialog appears, select "restart now" or "restart later" if
 
 ##### Save configuration on exit {#GeneralSettingsSaveConfig}
 
-This option is a checkbox that, when checked, tells NVDA to automatically save the current configuration when you exit NVDA.
+| . {.hideHeaderRow} |.|
```
seanbudd, line 1661: In our original discussion we raised the potential of moving these tables to the bottom of each user guide section.
I think this is a good idea and more common in reference materials - most people want a basic description first, then tables for further reference.

https://github.com/nvaccess/nvda/pull/15902#issuecomment-1867093375

Could  you also update userGuideStandards.md to note this change of policy?

user_docs/en/userGuide.md
```diff
@@ -1653,53 +1658,107 @@ When the confirmation dialog appears, select "restart now" or "restart later" if
...
+|---|---|
+|enabled |NVDA will automatically save the current configuration when you exit.|
+|disabled |NVDA will exit directly without saving the configuration. The changes to the settings of the user configuration will be lost.|
 
 ##### Show exit options when exiting NVDA {#GeneralSettingsShowExitOptions}
 
-This option is a checkbox that allows you to choose whether or not a dialog appears when you exit NVDA that asks what action you want to perform.
-When checked, a dialog will appear when you attempt to exit NVDA asking whether you want to exit, restart, restart with add-ons disabled or install pending updates (if any).
-When unchecked, NVDA will exit immediately.
+This setting is a checkbox that allows you to choose whether or not a dialog appears when you exit NVDA that asks what action you want to perform.
```
seanbudd, line 1673: I think it would be better to retain longer descriptions like this.

Thoughts @Qchristensen ?

seanbudd:
It seems like this PR is trying to achieve a large number of different things.
This makes it quite challenging to review.
I would encourage splitting out independent changes (i.e. options/settings terminology, settings tables, etc) and same with writing the user guide standards.
Alternatively, splitting this up into the sections of the guide being rewritten.
It also seems like some changes that have occurred on master have been reverted in this PR.

Qchristensen (changes requested):
There are a lot of different changes through this and it is very hard to review all at once.  Could I please ask to break it up into several PRs?  I suggest:

One for changing all the "options" to "settings"
One for fixing the formatting of tables
One (or more) for various specific changes after that (eg the numpad insert one which is a distinct rewording of that paragraph - although in that case both Sean and I were not in favour of the change without further explanation, as we had just approved a PR to make that change (perhaps this was inadvertent if this branch was created before that issue was merged?)

Adriani90:
@Qchristensen 
> I am in favour of keeping the longer description here. I don't think it is overly verbose and may aid understanding for some users.

However, long descriptions contain lot of duplicate information that we already explain in a focused way within the tables.
Keeping the long descriptions makes the user guide really unvonfortable to read in my view.
Which long descriptions do you think are not well covered in the tables? Is there any information missing?
I am really reluctant to keep double descriptions within the same sections. The user guide is already blown up so we should try to structure as much as possible in my view. But if you guys insist on this I can do ofcourse. There is the risk that people will be confused though. I think nowadays every blind user knows how to deal with simple tables like this.


@seanbudd 
Re moving tables at the bottom of sections: I designed the content in the tables in a way that makes the settings behavior really readable, and we have to decide how the reader flow should be in a case by case basis.
For example
1. We explaing in a short sentence what the setting is about ("This checkbox decides whether the voice should automatically change languages when navigating through a text"
2. We explain in two tables what options are available and what each of them does
3. Suppose you read the content in the tables, you now might better understand impacts of the options on NVDA behavior, so we may add some more information such as recommendations, definitions, explanations etc.

That older discussion actually refered only to the first table, where we just show up which options are available and which one is the default.
Now when connecting that table to the next table focusing on the settings behavior, it makes sense to keep them together at the top, in the middle or at the end of a section, what ever is better for contextual reading. Do you agree? Then I would adjust the standards accordingly.
Otherwise I can bring them all to the bottom ofcourse.

Adriani90:
@Qchristensen 
> One (or more) for various specific changes after that (eg the numpad insert one which is a distinct rewording of that paragraph - although in that case both Sean and I were not in favour of the change without further explanation, as we had just approved a PR to make that change (perhaps this was inadvertent if this branch was created before that issue was merged?)

Yes that seems to have conflicted, it is reverted now.
Re separate PR for various changes, there are not that much of them, actually the major changes are in the tables because I had to reword stuff to avoid lots of duplicate descriptions.
So if I create a separate PR with the tables, these are I would say 90% of this PR so it would not make things much easier.

I think we can solve your suggestions via single commits, it is in my view better administrated in this case.
Note that you don't need to review this PR all at once, I am ready to fix conflicts on the fly when ever they appear, so if possible, just take as much time as needed to review.

Adriani90:
@seanbudd 
> It also seems like some changes that have occurred on master have been reverted in this PR.

I think that was because of some conflicts which should be solved now.

CyrilleB79:
> @Qchristensen
> 
> > One (or more) for various specific changes after that (eg the numpad insert one which is a distinct rewording of that paragraph - although in that case both Sean and I were not in favour of the change without further explanation, as we had just approved a PR to make that change (perhaps this was inadvertent if this branch was created before that issue was merged?)
> 
> Yes that seems to have conflicted, it is reverted now. Re separate PR for various changes, there are not that much of them, actually the major changes are in the tables because I had to reword stuff to avoid lots of duplicate descriptions. So if I create a separate PR with the tables, these are I would say 90% of this PR so it would not make things much easier.

In this case, keep this PR for the tables and remove other changes from this PR to put it in separate PRs please.
From a reviewer point of view, it's really difficult to know what has been modified for the tables and what has been modified for other reasons.
Also separating in different PRs will prevent the biggest change (tables) to be blocked due to unrelated comments.
Thanks.

@Adriani90, I know that you have done a big job in this PR.
Though, I am sorry to say that I am a bit concerned by the amount of changes in the User Guide for translators. I fear that not all translators will be able to translate this big content on time. These big changes will also not be very popular among translators, so it's important that NV Access announces it with the correct justification to translators IMO (e.g. as it was done for the quick startt guide).
@seanbudd, @Qchristensen and @gerald-hartig, what is your opinion?

Adriani90:
@CyrilleB79 
> In this case, keep this PR for the tables and remove other changes from this PR to put it in separate PRs please.

However, changes made to the text which have been incorporated into the tables will be part of this PR though. We cannot merge the tables only and let double statements and descriptions in the same sections, this will definitely create confusions. The only part that I can logically separate is the lines with the word "option" becomming "setting" in appropriate place. This will be a small part of this PR.

I strongly suggest you take the html file generated by this pr, look at the line numbers and compared side to side what changed in case you are not confortable with how github displays diffs.
I don't think splitting the small part from this PR will make things easier. The major part with the tables will still have to be reviewed though.
Can you tell exactly where you have difficulties to review? The line numbers are indicated correctly in the diff.

> From a reviewer point of view, it's really difficult to know what has been modified for the tables and what has been modified for other reasons.

Can you explain more exactly what you mean? When you compare the both html generated user guides you can exactly see what changed.

All the settings part of the user guide has been converted into table structures, and the original text descriptions have been removed and incorporated into tables instead, in order to avoid duplications.

> Also separating in different PRs will prevent the biggest change (tables) to be blocked due to unrelated comments.

What do you mean by that? All changes I made in this PR are related to the user guide standard, which is directly related to the tables, even changing the "option" to "setting" is required to be in line with the table structure.

It is not feasible to split this work into lots of PRs for each section of the user guide.

> Though, I am sorry to say that I am a bit concerned by the amount of changes in the User Guide for translators. I fear that not all translators will be able to translate this big content on time.

This is a change in user guide standard compared to what we had before, I think translators will have to go through this work unfortunately, in case NV Access wants these standards to be pushed.

> so it's important that NV Access announces it with the correct justification to translators IMO (e.g. as it was done for the quick startt guide).

i agree.

Qchristensen:
Re the longer description comment - [Here](https://github.com/nvaccess/nvda/pull/16490#issuecomment-2104073154) - I was thinking of the names of things like "Restart with add-ons disabled", or install pending updates (I haven't got any to check the exact wording) which actually aren't explicitly mentioned anywhere else - I was thinking if a user was looking at the exit screen options and found that and wasn't sure what it does, it is good to be able to search the user guide for the wording they are presented with

seanbudd:
It is useful to have a plain english description as well as a table. There's no need to remove this content when creating tables.
Tables should be moved to the bottom of sections. Having a general description first and then structured data is more common in reference materials. For example:
- https://learn.microsoft.com/en-us/windows/win32/api/audiopolicy/nn-audiopolicy-iaudiosessionevents
- https://git-scm.com/docs/git-revert
- https://linux.die.net/man/8/apt-get

Adriani90:
Might be true for sighted people, but having to read lots of double descriptions is confusing when you use a screen reader. I can redesign that but I am quite sure we will come back to this conclusion. Did you look at the html file and compared it with what we have in master?Von meinem iPhone gesendetAm 22.05.2024 um 07:25 schrieb Sean Budd ***@***.***>:﻿
It is useful to have a plain english description as well as a table. There's no need to remove this content when creating tables.
Tables should be moved to the bottom of sections. Having a general description first and then structured data is more common in reference materials. For example:

https://learn.microsoft.com/en-us/windows/win32/api/audiopolicy/nn-audiopolicy-iaudiosessionevents
https://git-scm.com/docs/git-revert
https://linux.die.net/man/8/apt-get

CyrilleB79:
> @CyrilleB79
> 
> > In this case, keep this PR for the tables and remove other changes from this PR to put it in separate PRs please.
> 
> However, changes made to the text which have been incorporated into the tables will be part of this PR though. We cannot merge the tables only and let double statements and descriptions in the same sections, this will definitely create confusions. The only part that I can logically separate is the lines with the word "option" becomming "setting" in appropriate place. This will be a small part of this PR.
> 
> I strongly suggest you take the html file generated by this pr, look at the line numbers and compared side to side what changed in case you are not confortable with how github displays diffs. I don't think splitting the small part from this PR will make things easier. The major part with the tables will still have to be reviewed though. Can you tell exactly where you have difficulties to review? The line numbers are indicated correctly in the diff.

This PR is very large. And I will not be able to review it carefully. If changes unrelated to structure had been isolated in another PR, I could have reviewed them carefully. Just my opinion though.
Anyway, NV Access has not insisted on this separation, so you can continue as you started.

Adriani90:
@Qchristensen 
> I was thinking if a user was looking at the exit screen options and found that and wasn't sure what it does, it is good to be able to search the user guide for the wording they are presented with

Looking at the HTML file generated through this PR, the section for exit options looks as follows:

##### Show exit options when exiting NVDA {#GeneralSettingsShowExitOptions}

This setting is a checkbox that allows you to choose whether or not a dialog appears when you exit NVDA that asks what action you want to perform.

| . {.hideHeaderRow} |.|
|---|---|
|Options |enabled, disabled|
|Default |enabled|

| Option |Behaviour|
|---|---|
|enabled |NVDA displays a dialog when you attempt to exit. You will be asked whether you want to exit, restart, restart with add-ons disabled, restart with logging level set to debug or install pending updates (if any).|
|disabled |NVDA will exit directly without showing any confirmation dialog.|

So all the information that was in the long plain text description before is still here, it is just described in the behavior table. So in my view it doesn't make sense to keep the same description also above the tables, because then the user guide becomes really blown up and quite unreadable.
That's why I suggested to compare the html files to see exactly what happens on the screen.

@CyrilleB79 everything in this PR is related to the tables and the user guide standards.

CyrilleB79:
> @CyrilleB79 everything in this PR is related to the tables and the user guide standards.

In this case, you may drop "+ some wording corrections" from the title of this issue. What was it referring to?

In any case, there is at least the "option" vs "parameter" changes which are not linked to the tables. Why this specific rewording?

Adriani90:
Ofcourse they are related to the tables. Tables describe the options for specific settings, so this difference needs to be consistent throughout the user guide. To be honnest I am a bit shaking my head when reading about what is questioned here. You should really check the html file.Von meinem iPhone gesendetAm 22.05.2024 um 08:48 schrieb Cyrille Bougot ***@***.***>:﻿

@CyrilleB79 everything in this PR is related to the tables and the user guide standards.

In this case, you may drop "+ some wording corrections" from the title of this issue. What was it referring to?
In any case, there is at least the "option" vs "parameter" changes which are not linked to the tables. Why this specific rewording?

seanbudd:
>Might be true for sighted people, but having to read lots of double descriptions is confusing when you use a screen reader. I can redesign that but I am quite sure we will come back to this conclusion. Did you look at the html file and compared it with what we have in master?

@Adriani90 - we discussed this internally and Sascha and Mick don't see this issue you describe. This structure is extremely common in reference materials, and they don't understand the confusion you describe. Surely table content can be skipped over when the reader is not interested in it?
Please change this PR to only adding tables.

Adriani90:
@seanbudd 
> This structure is extremely common in reference materials, and they don't understand the confusion you describe. Surely table content can be skipped over when the reader is not interested in it?

I don't have anything against tables, I was only refering to the fact that the long descriptions of settings behavior have been moved into tables, so no need to display them again above tables.

What you linked above is the style of Microsoft to display reference material, which is article style. But the user guide as it is now does not contain that much description in the settings areas apart from what we describe in tables. I moved now all tables to the bottom of each section and kept the old descriptions above. If you don't like this, I can revert the last commit. In my view the user guide as it is generated now per last commit is not really well readable, contains many unclear settings and contains lot of duplicate descriptions which makes it difficult to read for people who need easy english language.

I wished you at NV Access would have spent some time to provide some constructive arguments, but it is how it is. Now you can decide the next steps, I will stop the work on the user guide at this point because for me it is too much work to adjust everything in new pull requests. I understand this for a coding perspective, but not really for the user guide.

Now we have sections, where the setting is called option, and the option contains options. Confusing... but yeah.

CyrilleB79:
I fully agree with @Adriani90 that duplicating the content makes things less easy to read.
Of course, one can skip the table. But readers won't be sure that exactly the same information is duplicated in the table.
And from a maintainability point of view, duplicating so much content is bad practice and will likely lead to inconsistencies.
Either we put more information in table or we put more information in the text paragraph, not both.

Except for the revert of last commit, I'd say that next changes should be done on one or two paragraphs illustrating the wanted change. Once an agreement has been found on the structure and content of these paragraphs, the work can be duplicated to the rest of the User Guide.
@seanbudd sorry, it's a pity IMO [to have encouraged](https://github.com/nvaccess/nvda/pull/15902#issuecomment-1869800898) @Adriani90 to propose changes directly in a 1000+ line PR rather than first opening an issue describing the intended changes and the motivation behind.

seanbudd:
@CyrilleB79 - to be clear, what was proposed in https://github.com/nvaccess/nvda/pull/15902#issuecomment-1867321541 was a PR that just added tables to the bottom of each section. My previous comment requested opening a discussion or issue first for any other scenario.

I would highly encourage a separate issue or discussion before proposing wider changes than this.
I'd also encourage tackling it as a separate PR if the issue is accepted. Large refactors in general are hard to review, particularly if they are trying to achieve multiple things. 
As it is, we just want to add structured content, without losing the plain text accessible descriptions.

Adriani90:
But a table contains already accessible plain text, there is no need for duplicates in this regard.
Sean the changes to the user guide have to be this huge if you guys propose such new standards. Putting e.g. every table in a single PR is not efficient at all. And me at least speaking for myself out of my voluntary perspective, I will not do that work. I don‘t think reviewing such complex changes to the user guide are more challenging than reviewing e.g. 1000+ lines of code. And we have several PRs with that amount of code.
At least with the user guide you can put the html files side to side and compare the real changes much easier. I am not really following the argument you bring on the table.
It is certain that if we merge the tables only as per current stage of this PR, the user guide will become a chaos and less accessible.

CyrilleB79:
@seanbudd, I have the feeling that NV Access expectations are not clear.

Let's take an example to be extra clear: the paragraph "Save configuration on exit"

In the current version of this PR,this paragraph consists in 3 parts in this order:

1. A plain text description of the option
`This option is a checkbox that, when checked, tells NVDA to automatically save the current configuration when you exit NVDA.`

2. A table listing the possible choices of the option and indicating the default one
```
| . {.hideHeaderRow} |.|
|---|---|
|Options |enabled, disabled|
|Default |enabled|
```

3. A table describing the effect (behaviour) of each possible choice
```
| Option |Behaviour|
|---|---|
|enabled |NVDA will automatically save the current configuration when you exit.|
|disabled |NVDA will exit directly without saving the configuration. The changes to the settings of 
```

Before this PR, the description of an option was done as follows:
1. Either only plain text (part 1)
2. Or, for newer ones and as per the new guidelines, choice listing table (part 2) followed by plain text (part 1)

Is NV Access expecting to do:
* Choice 1: only move the option listing table (or create it) below the plain text description, i.e.  part 1 followed by part 2.
* Choice 2: move the option listing table (or create it) below the plain text description and below this table, create a new table explaining the behaviour of each choice, i.e.  part 1 followed by part 2 and then part 3.
* Choice 3: another choice (please explain)

As explained by @Adriani90 and myself, choice 2 is probably not desirable due to readability and maintainability issues.

Thanks for future clarifications!

Adriani90:
Choice 2 was what I did first, I bassically created the behavior tables and incorporated all the old descriptions there. That version of this PR would have been easily maintainable. But that was not what NV Access seemed to expect, so now we have all the descriptions double, both in the behavior tables and above the tables.Von meinem iPhone gesendetAm 24.05.2024 um 10:21 schrieb Cyrille Bougot ***@***.***>:﻿
@seanbudd, I have the feeling that NV Access expectations are not clear.
Let's take an example to be extra clear: the paragraph "Save configuration on exit"
In the current version of this PR,this paragraph consists in 3 parts in this order:


A plain text description of the option
This option is a checkbox that, when checked, tells NVDA to automatically save the current configuration when you exit NVDA.


A table listing the possible choices of the option and indicating the default one


| . {.hideHeaderRow} |.|
|---|---|
|Options |enabled, disabled|
|Default |enabled|


A table describing the effect (behaviour) of each possible choice

| Option |Behaviour|
|---|---|
|enabled |NVDA will automatically save the current configuration when you exit.|
|disabled |NVDA will exit directly without saving the configuration. The changes to the settings of 

Before this PR, the description of an option was done as follows:

Either only plain text (part 1)
Or, for newer ones and as per the new guidelines, choice listing table (part 2) followed by plain text (part 1)

Is NV Access expecting to do:

Choice 1: only move the option listing table (or create it) below the plain text description, i.e.  part 1 followed by part 2.
Choice 2: move the option listing table (or create it) below the plain text description and below this table, create a new table explaining the behaviour of each choice, i.e.  part 1 followed by part 2 and then part 3.
Choice 3: another choice (please explain)

As explained by @Adriani90 and myself, choice 2 is probably not desirable due to readability and maintainability issues.
Thanks for future clarifications!

CyrilleB79:
I have edited my previous comment (https://github.com/nvaccess/nvda/pull/16490#issuecomment-2128891935).

Choice 2 is what has been done initially as well as the current state of this PR. But it was initially choice 2 + choice B. Now it is choice 2 + choice A, which is not desirable.

gerald-hartig:
There have been several requests in the PR for NV Access to clarify their position. 

NV Access's stance is to retain both the text and table formats, while minimising redundancy. The text should come first, offering a concise summary and providing necessary context and clarification without detailing specific behaviours. The table should follow, containing all the detailed information, including the behaviour of each choice.

The question is, how do we get to there from where we are now? We also need to consider if right here is a place that we want to be. And if not, how do we avoid finding ourselves in this situation again?

To move forward, we have to recognise that any effort to change the user guide is magnified 50x by the need for translations. Therefore, we need to be considerate of our translators and be sure that the cost of any changes is justified by the nature of the changes. We also need to consider how they work, for example translators must translate the entire diff each commit. 

We have a few options: 
1. copy the content on options & behaviours from the text to the table and have duplicated content between text and tables, then at a later date simplify and remove redundancy from the text, 
1. all in one go, move all the content on options and behaviours from the text to tables and simplify the text, 
1. move and simplify at the same time, but only do one section of the user guide at a time. 

Due to the magnitude of the change I suggest we need feedback from our translators on what method works best for them. 

Additionally, we must consider if our current starting point is optimal. Large diffs with numerous changes can overwhelm translators. More worryingly there appear to be changes to the content as well. Consider the changes in "Move system caret when routing review cursor" which changes the listed options. Refactoring must be done methodically and carefully to avoid error slipping in, particularly with large refactors. Keeping PRs separate for different goals can help with this.

I suggest the way forward is to get confirmation from the translators on their preferred approach, and for the developers to make a call on whether the current PR can be salvaged. I hate to be so blunt, but if github is struggling to show rendered diffs, this might be too large to work with. Adriani should maintain a backup to verify the intended changes are all in the final product. 

Looking forward to your insights and suggestions.

Adriani90:
@gerald-hartig thanks for your contribution here, following are my thoughts.
> NV Access's stance is to retain both the text and table formats, while minimising redundancy.

That was the initial state of this PR and still NV Access was not keen with it. Redundant text was moved into tables, and the rest of the description, if any, was still outside of the tables. For translators this would have been basically a copy and paste task, + some rewordings for more clarity and coresponding additions for opsions behaviors that are not documented yet in the master.

> The table should follow, containing all the detailed information, including the behaviour of each choice.

How does this relate to the previous point of minimizing redundancy? In my view the tables should contain only the behavior of each option, and the available options out there. Outside of the tables is anything else related to description of specific use cases for a specific setting or option, keystroke documentation etc.

> The text should come first, offering a concise summary and providing necessary context and clarification without detailing specific behaviours.

That's not possible without redundancy. Our settings subsections are very short as such. Apart from the behavior description, we have a clear title for a setting which in many cases does not need further introductory description, and in some cases we have recommendations and descriptions of use cases. But if you want to understand the setting from a readers point of view, in some cases you have to first understand the behavior of the setting's options, and then read the recommendations, descriptions of use cases, warnings etc. That's why initially the tables were not always at the bottom of each section.
Again, I think people compare apples with pairs when referencing to Microsoft's article style documentation. If we want the user guide to look like that in the end, we will need much more work than what I've done so far.

> We also need to consider if right here is a place that we want to be. And if not, how do we avoid finding ourselves in this situation again?

The current state of this PR as of now is definitely not the place we want to be. I think it would have been better to have a more in depth look at the initial state of this PR and add introductory text if really needed, and only for the sections where it makes sense, but at a later stage.

> 1. copy the content on options & behaviours from the text to the table and have duplicated content between text and tables, then at a later date simplify and remove redundancy from the text,

This option is not viable in my view, theredundant text was moved into tables in the initial state. Translators would have seen that as well while translating, because the coresponding lines would have been preceded by a dash and the lines in the table would have been shown with a plus sign.
Moreover, users would not get confused about why there are so many duplicate descriptions.

> 2. all in one go, move all the content on options and behaviours from the text to tables and simplify the text,

That was my initial state, while the word "option" was changed to "settings" in appropriate places to make the difference clear throughout the user guide.

> 3. move and simplify at the same time, but only do one section of the user guide at a time.
Due to the magnitude of the change I suggest we need feedback from our translators on what method works best for them.

I am a translator myself, and I noticed that almost all countries have more than one translator. So the work would be splitted to more than one person.
It is more efficient to split the translation work by the translators themselves instead of letting one person to create so many pull requests and manage so many discussions. I would have to create more than 60 pull requests for each sub section of the user guide, or more than 20 pull requests if we take a section with all its sub sections and the consistent change of the word options to settings in the appripriate places to keep the difference clear.

> Consider the changes in "Move system caret when routing review cursor" which changes the listed options.

The current description of that setting in the current master is quite difficult to read, I just splited the parts accordingly and put them into the behavior table which makes it much more easy to follow.

> Refactoring must be done methodically and carefully to avoid error slipping in, particularly with large refactors. Keeping PRs separate for different goals can help with this.

Making the user guide compliant with the user guide standards definitely will need rewording of some text because it needs to fit into behavior tables. From my voluntary perspective this would be too much work to split it into multiple PRs, but I am happy to keep this work documented and someone else can take it into separate pull requests.
I could try to split the large initial commit into smaller commits if this helps.

> I suggest the way forward is to get confirmation from the translators on their preferred approach, and for the developers to make a call on whether the current PR can be salvaged. I hate to be so blunt, but if github is struggling to show rendered diffs, this might be too large to work with.

I agree with this approach as long as there is full clarity on how the user guide should really look like in the end. To me it is still not clear enough at this stage.

Maybe it would make sense that you at NV Access change some settings sections in a separate pull request so that we can see a clear practical example of what you expect. Then I could easily adapt this work to your expectactions for the rest of the settings sections.

CyrilleB79:
Currently (master branch), we have a mix of option tables at the beginning of the paragraph before the description text (legacy) and option table at the end of the paragraph after the text (new user guide standard).

This is the worse situation. because the user would expect something consistent etween paragraphs, no matter if the table are at the beginning or at the end.

@Adriani90 do you plan to continue with this PR? Merge conflicts may accumulate, leading to more difficult review.

Alternatively, if this PR is not progressed, we may consider a smaller work, that is moving only the already existing tables at the end of the paragraphs without creating new ones.

seanbudd:
We think it's best to wait until we move user guide translations to Crowdin, before resolving merge conflicts.

seanbudd:
Thanks for your work on this. Due to the size of the diff, it is very hard for us to see how this PR can be move forward without continually fighting merge conflicts.
We would accept smaller PRs moving these tables in smaller groupings, with a max diff of 300 lines.

PR #16499, opened 2024-05-07
Revert "Remove redundant braille.BrailleHandler.update executions"

LeonarddeR:
### Reverts PR
Reverts #16463

### Issues fixed
None

### Issues reopened
Reopens #16456

### Reason for revert
The reverted pr causes unexpected side effects for message buffers while the scope of the issue it fixes is a bit unclear.

### Can this PR be reimplemented? If so, what is required for the next attempt
Braille updates should still properly dismiss message buffers.

### DIFF
--- a/source/braille.py
+++ b/source/braille.py
@@ -2535,9 +2535,8 @@ class BrailleHandler(baseObject.AutoPropertyObject):
 		self.mainBuffer.update()
 		# Last region should receive focus.
 		self.mainBuffer.focus(region)
-		if isinstance(region, TextInfoRegion):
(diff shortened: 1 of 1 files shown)


LeonarddeR:
@burmancomp While I"m convinced that you will certainly try to fix this, I think it is better to revert this first and then try to fix it, afterwards rather than providing a fixup pr for #16463.

burmancomp:
I am quite optimistic that I would have fix for this problem.

If I open new pr, can you then test if it works for you?

LeonarddeR:
Sure, go ahead if you wish.

PR #16504, opened 2024-05-08
Report position and set size for gridcells in treegrids

### LINKED ISSUE #15818: Position and set size not narrated for gridcell in treegrid
### Steps to reproduce:

1. Turn on NVDA.
2. Visit this [W3C tree grid cell navigation example #1][cell example].
3. Using the Tab key, navigate to the inbox example tree grid element located under the "Example" heading.
4. Using the Down or Up arrow key, navigate by rows in the tree grid, and observe the narration while navigating from one row to another.
5. Visit this [W3C tree grid row navigation example #2][row example].
6. Using the Tab key, navigate to the inbox example tree grid element located under the "Example" heading.
7. Using the Down or Up arrow key, navigate by rows in the tree grid, and observe the narration while navigating from one row to another.

### Actual behavior:

In the example #1, when nnavigating in the tree grid with the Down or Up arrow key, NVDA narrates the row and colun coordinates, but not the position of the row and the set size relative to the current tree level. Compare this with the example #2, where NVDA narrates the position and set size correctly, e.g. "2 of 3". The narration is missing regardless of whether the aria-posinset and aria-setsize are present on the row or not.

### Expected behavior:

The position and set size should be narrated even when role="gridcell" element is focused, inheriting the data from the parent role="row" element, in the same manner how JAWS does.

Installed.

#### NVDA version:

NVDA 2023.3.0.29780.

#### Windows version:

Windows 11, version 21H2 (OS Build 22621.2715)

Google Chrome, version 119.0.6045.160.

Yes.

No.

Yes.

Yes.

[cell example]: https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/?cell=force
[row example]: https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/?cell

SaschaCowley:
### Link to issue number:

Fixes #15818 

### Summary of the issue:

When navigating an ARIA treegrid's cells (as opposed to its rows), the position in set and set size are not supported.

Position information are now reported.

Modified `Ia2Web._get_positionInfo` to return the closest ancestor treeitem's position information if the current focus is a gridcell.

### Testing strategy:

Manually tested with the [APG Treegrid Email Inbox Example](https://www.w3.org/WAI/ARIA/apg/patterns/treegrid/examples/treegrid-1/?cell=force).

Level information is reported at the end of speech output when navigating by cell, unlike when navigating by row, where it is reported first if the level has changed.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [ ] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers


source/NVDAObjects/IAccessible/ia2Web.py
```diff
@@ -121,6 +121,10 @@ def isDescendantOf(self, obj: "NVDAObjects.NVDAObject") -> bool:
 		return bool(res)
 
 	def _get_positionInfo(self):
+		if self.role == controlTypes.Role.TABLECELL:
+			for anc in reversed(api.getFocusAncestors()):
```
LeonarddeR, line 125: This is definitely going to return wrong information when the table cell doesn't have focus, such as when using object navigation

I think you want to traverse the parents of the objects instead.

SaschaCowley:
@LeonarddeR  Thanks for a good catch; I could have really had egg on my face there.

SaschaCowley:
Closing as more input is needed on the UX expected.

PR #16505, opened 2024-05-08
Run the uninstaller with administrator privileges

wmhn1872265132:
### Link to issue number:
none

### Summary of the issue:
NVDA cannot be uninstalled correctly by directly running `"%ProgramFiles(x86)%\NVDA\uninstall.exe"`. You must run it as an administrator to uninstall it correctly.

You can directly run `"%ProgramFiles(x86)%\NVDA\uninstall.exe"` to uninstall NVDA

Change `RequestExecutionLevel` to `admin`

### Testing strategy:
Manually run `"%ProgramFiles(x86)%\NVDA\uninstall.exe"`, the uninstaller should correctly apply for administrator rights and uninstall NVDA

The portable version of NVDA's `uninstall.exe` will also apply for administrator rights when running.

### Code Review Checklist:

- [x] Documentation:
  - Change log entry
  - User Documentation
- [x] Testing:
  - Unit tests
  - System (end to end) tests
  - Manual testing
  - Speech 
  - Braille
  - Low Vision
  - Different web browsers

### DIFF
--- a/uninstaller/uninst.nsi
+++ b/uninstaller/uninst.nsi
@@ -9,7 +9,7 @@ SetCompressor /SOLID LZMA
 CRCCheck On
 XPStyle on
 InstProgressFlags Smooth
-RequestExecutionLevel user
+RequestExecutionLevel admin
 
 Name "${appName}"
 VIProductVersion "${VERSION_YEAR}.${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_BUILD}" ;Needs to be here so other version info shows up


seanbudd (approved):

seanbudd:
>The portable version of NVDA's uninstall.exe will also apply for administrator rights when running.

I think that's okay. I assume that tool is used to uninstall NVDA from a computer, not delete the portable copy. A portable copy cannot be installed or uninstalled

seanbudd:
I'm reverting this PR as it makes admin permissions required to generate the uninstaller. I don't believe this is setup correctly.
The installer doesn't require this and uses `RequestExecutionLevel user`.

Please file an issue first describing the problem you are facing, as generally the uninstaller works for me and requests admin level.

michaelDCurran:
It seems that When running the uninstaller from control panel / programs and features, or Windows Settings -> Apps -> installed apps: Windows automatically elevates the uninstaller. I.e. I get a UAC screen each time.
However, when running uninstall.exe directly from c:\program files (x86)\nvda, for instance using File Explorer, the uninstaller ls launched with no UAC elevation.
I would not think it is a very likely / supported pattern to run it directly. Generally people should use the installed apps / programs and features UI.

michaelDCurran:
Good point.

Yes, we should either remove that shortcut, or change it to use the 
shell to request elevation.

wmhn1872265132:
In addition, I really don’t think that running the uninstall program directly should not be supported, unless the program is invisible to the user, otherwise there should be no position to prevent the user from doing this.

seanbudd:
Please open a new issue for the shortcut bug

